PR #4, opened 2021-03-22
use rustc_index instead of custom structures.

christianpoveda:
I think this is a substantial change so there might be a lot of stuff to check here.


liquid-rust-driver/src/lower/basic_block.rs
```diff
@@ -9,7 +8,7 @@ impl<'tcx> Lower<'tcx> for mir::BasicBlock {
     type Output = BasicBlock;
 
     fn lower(&self, _lcx: LowerCtx<'tcx>) -> LowerResult<Self::Output> {
-        Ok(BasicBlock::new(self.index()))
+        Ok(*self)
```
nilehmann, line 11: I think we can just skip the `impl Lower for BasicBlock` altogether, no?

liquid-rust-driver/src/lower/local.rs
```diff
@@ -9,7 +8,7 @@ impl<'tcx> Lower<'tcx> for mir::Local {
     type Output = Local;
 
     fn lower(&self, _lcx: LowerCtx<'tcx>) -> LowerResult<Self::Output> {
-        Ok(Local::new(self.index()))
+        Ok(*self)
```
nilehmann, line 11: Same here, I think we can just drop the `impl`.

liquid-rust-lrir/src/mir/place.rs
```diff
@@ -58,7 +58,7 @@ impl From<Local> for PlaceRef<'_> {
 
 impl<'a> fmt::Display for PlaceRef<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut s = format!("{}", self.local);
+        let mut s = format!("{:?}", self.local);
```
nilehmann, line 61:
```suggestion
        let mut s = format!("_{}", self.local.as_usize());
```
Otherwise, places would look weird when printing them?
 `Local(0).0.1.2` vs `_0.0.1.2`

liquid-rust-typeck/src/bblock_env.rs
```diff
@@ -11,20 +11,21 @@ use liquid_rust_lrir::{
 };
 
 pub struct BBlockEnv {
-    pub ghost_vars: OrderedHashMap<GhostVar, Ty>,
+    pub ghost_vars: BTreeMap<GhostVar, Ty>,
```
nilehmann, line 14: We need to preserve insertion order and inserting them on a btree would reorder them. In practice, I think we always use the "alphabetical" order would match the insertion order but maybe that's too much to ask from context.

liquid-rust-typeck/src/bblock_env.rs
```diff
@@ -11,20 +11,21 @@ use liquid_rust_lrir::{
...
         mut refiner: Refiner<'_, 'tcx>,
         vars_in_scope: &mut Vec<Var>,
+        ghost_gen: &IndexGen<GhostVar>,
     ) -> Self {
-        let mut ghost_vars = OrderedHashMap::new();
+        let mut ghost_vars = BTreeMap::new();
         let mut locals = vec![];
         for (local, local_decl) in local_decls.iter_enumerated() {
-            let fresh_gv = refiner.var_gen.fresh();
+            let fresh_gv = ghost_gen.fresh();
```
nilehmann, line 28: I wasn't sure if it was necessary to use the atomic in the index generator, but this convinced me :+1:

liquid-rust-typeck/src/binding_tree.rs
```diff
@@ -1,30 +1,33 @@
...
+        const FIRST_NODE = 0
+    }
 }
 
 pub struct BindingTree {
-    nodes: IndexMap<NodeId, Node>,
+    nodes: IndexVec<NodeId, Node>,
     curr_path: Vec<NodeId>,
-    curr_bindings: OrderedHashMap<Var, Ty>,
+    curr_bindings: BTreeMap<Var, Ty>,
```
nilehmann, line 17: Thinking over this, now I believe we can even use hash map directly. Originally, I was using an "ordered hash map" to preserve the order of binding as they were inserted, but since we keep track of that in `curr_path` I think it's not necessary to do anything fancy anymore.

liquid-rust-typeck/src/bblock_env.rs
```diff
@@ -63,7 +64,7 @@ impl fmt::Display for BBlockEnv {
         let locals = self
             .locals
             .iter()
-            .map(|(local, gv)| format!("{}: {}", local, gv))
+            .map(|(local, gv)| format!("{:?}: {}", local, gv))
```
nilehmann, line 67: I think this would make it more readable.
```suggestion
            .map(|(local, gv)| format!("_{}: {}", local.as_usize(), gv))
```

liquid-rust-typeck/src/local_env.rs
```diff
@@ -236,7 +234,7 @@ impl fmt::Display for LocalEnv<'_> {
             .iter()
             .copied()
             .flatten()
-            .map(|(local, gv)| format!("{}: {}", local, gv))
+            .map(|(local, gv)| format!("{:?}: {}", local, gv))
```
nilehmann, line 237: Same here
```suggestion
            .map(|(local, gv)| format!("_{}: {}", local.as_usize(), gv))
```

liquid-rust-lrir/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(rustc_private)]
+#![feature(const_panic)]
```
nilehmann, line 2: Why do we need this?

nilehmann (commented):
My main concern is `BTreeMap` vs `OrderedHashMap` in `BBlockEnv`. A block env should morally be an ordered pair of ghost vars and types. This is important in theory because one can only talk about the well-formedness of the environment if it is ordered (one can only mention variables in scope). But since we also need to lookup by ghost var I created the concept of an OrderedHashMap to preserve insertion order. In practice, though, I think the only place where the order matters is when we are printing it.

christianpoveda:
> My main concern is `BTreeMap` vs `OrderedHashMap` in `BBlockEnv`. A block env should morally be an ordered pair of ghost vars and types. This is important in theory because one can only talk about the well-formedness of the environment if it is ordered (one can only mention variables in scope). But since we also need to lookup by ghost var I created the concept of an OrderedHashMap to preserve insertion order. In practice, though, I think the only place where the order matters is when we are printing it.

hmmm... interesting. I'm going to leave the `OrderedHashMap` API then. Do you think we have to back it with this IndexMap or could we use a `Vec<(K, V)>` instead given that we're not expected to have a lot of keys in the env anyway?

liquid-rust-lrir/src/lib.rs
christianpoveda: Because the `newtype_index` macro panics if you try to create an index that doesn't fit inside an `u32`. I don't love it either.

liquid-rust-typeck/src/bblock_env.rs
christianpoveda: Me neither! I just did it beforehand because I didn't want to risk requiring a mutable reference somewhere where it wasn't available hahaha.

liquid-rust-lrir/src/mir/place.rs
christianpoveda: I think this is fine because https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/mir/mod.rs.html#709-715 implements Debug as `_0`.

liquid-rust-driver/src/lower/basic_block.rs
christianpoveda: yes totally.

liquid-rust-driver/src/lower/local.rs
christianpoveda: yes

nilehmann:
> Do you think we have to back it with this IndexMap or could we use a `Vec<(K, V)>` instead given that we're not expected to have a lot of keys in the env anyway?

A `Vec<(K, V)>` may be faster for a few keys. One can even do two vecs, but we'd never know without benchmarking. At this point I'm fine with any option. For reference, the only place where we do lookups is in `LocalEnv::infer_subst`

christianpoveda:
I went for the `Vec` impl, maybe it can be slower in some cases but is one crate less in compilation :P

nilehmann (approved):

PR #12, opened 2021-05-13
Inline checking bbs with one predecessor

dcao:
Implements for switchint and goto terminators

### DIFF
--- a/liquid-rust-typeck/src/lib.rs
+++ b/liquid-rust-typeck/src/lib.rs
@@ -20,8 +20,8 @@ use liquid_rust_common::index::{Idx, IndexGen, IndexVec};
 use liquid_rust_fixpoint::{Fixpoint, Safeness};
 use liquid_rust_lrir::{
     mir::{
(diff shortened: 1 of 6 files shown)


liquid-rust-typeck/src/lib.rs
```diff
@@ -96,17 +99,20 @@ impl<'tcx, 'a> Checker<'tcx, 'a> {
             tcx,
             global_env: task.global_env,
             bb_envs,
+            body: &task.body,
             ret_env,
             local_gen,
         };
 
+        let mut seen = HashSet::new();
```
nilehmann, line 107: I'd add `seen` as a field on `Checker` and pass `self` mutably.

dcao: Adding `seen` as a field on `Checker` messes with some borrow checking stuff - in particular, in `check_basic_block`, we can't borrow the `bb_env` while also doing the stuff in `enter_basic_block` (the closure needs "exclusive access to `self`").

PR #17, opened 2021-12-22
Test using Executable

ranjitjhala:
Tweak the test code to use the `liquid-rust` executable. Gets @nilehmann 's `heapsort.rs` test working.

### DIFF
--- a/tests/pos/rvec.rs
+++ b/tests/pos/rvec.rs
@@ -1,5 +1,3 @@
-#![feature(register_tool)]
-#![register_tool(lr)]
 #![allow(dead_code)]
 
 pub struct RVec<T> {
--- a/tests/pos/rvec.rs
+++ b/tests/pos/rvec.rs
@@ -1,3 +1,5 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
 #![allow(dead_code)]
 
 pub struct RVec<T> {
--- a/tests/pos/rvec.rs
+++ b/tests/pos/rvec.rs
@@ -1,0 +1,58 @@
+#![allow(dead_code)]
+
+pub struct RVec<T> {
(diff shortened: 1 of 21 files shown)


tests/pos/rvec.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(register_tool)]
```
nilehmann, line 1: Do we really need this here? I think we only need to register the tool in the crate root.

nilehmann (approved):

tests/pos/rvec.rs
ranjitjhala: Good point I removed it.

ranjitjhala:
Any idea why the `neg` tests are failing in the CI ? It seems to complain about the paths, but AFAICT those files `tests/neg/test0{0,1}.rs` are already checked in? Can you try a `cargo test` locally?

nilehmann:
@ranjitjhala is failing for me locally as well with the same error.

ranjitjhala:
puzzling. let me make a fresh clone and try to replicate.

nilehmann:
@ranjitjhala ok, it makes sense the `liquid-rust` binary won't be in `target/debug` unless you run `cargo build` (or `cargo run`). `cargo test` won't create the binary.

ranjitjhala:
aha. So the issue is that you have to run a `cargo build` first and then
the `cargo test`.

(Odd that only the `neg` tests complain about it...)

ranjitjhala:
@nilehmann -- any idea how to fix? Is there a way to tell the `cargo test` to make sure it does a `build` first? OR should we do it in the CI script?

nilehmann:
@ranjitjhala changing the CI script to run `cargo build` first is the easiest.

PR #19, opened 2022-01-18
Mod support

atgeller:
Adds support for modulo in specifications and expressions. As a result, also adds support for assert terminators and BinAnd (&& in Rust rather than in liquid-rust).


nilehmann (changes requested):
@atgeller can you split this into three PRs (assert, new ops, todo). I don't have many comments for the todo section and would like to merge asap. Assert also seems fine. I have more subtle comments on the new operations, particularly about the distinction between reminder and modulo which differs for negative numbers. It'll be easier if we can focus on those problems independently.

PR #24, opened 2022-01-19
Support for reminder and bitwise and

atgeller:
Adds support for modulo % in liquid-rust specifications as well as in rust operands (note: rust treats % as remainder while liquid-fixpoint treats it as modulo,). Add support for BitAnd && on boolean values in rust operands.


nilehmann:
@atgeller can you confirm what's the semantics of `mod` in liquid Haskell/Z3 for negative numbers and check if it differs from that of `%` in rust. If they do, we can't really give `%` the type:

```
(%) : fn<a: int, b : int>(i32@a, i32@b) -> i32@{a mod b}
```

nilehmann:
@atgeller as discussed in the meeting then we should give (%) type

nilehmann:
@atgeller as discussed we should give `%` type for unsigned integers

```rust
(%) : fn<a: int, b: int{b != 0}>(i32 @ a, i32 @ b) -> i32{v: a > 0 && b > 0 => v = a mod b}
```

liquid-rust-typeck/src/checker.rs
```diff
@@ -459,12 +461,74 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
...
+    }
+
+    fn check_bitwise_op(&self, op: BinOp, ty1: Ty, ty2: Ty) -> Ty {
+        match (ty1.kind(), ty2.kind()) {
+            (TyKind::Refine(BaseTy::Bool, e1), TyKind::Refine(BaseTy::Bool, e2)) => TyKind::Refine(
+                BaseTy::Bool,
+                ExprKind::BinaryOp(op, e1.clone(), e2.clone()).intern(),
+            )
+            .intern(),
+            _ => unreachable!(
```
nilehmann, line 479: I assume you can also have bitwise operation between integers? In that case, this is a `todo!` not an `unreachable!`. But since we are at it, I think it's not too hard to implement it for integers as well. We can just return `i32{true}` in that case for now.

liquid-rust-core/src/ir.rs
```diff
@@ -88,6 +88,8 @@ pub enum BinOp {
     Sub,
     Mul,
     Div,
+    Mod,
```
nilehmann, line 91: This should be called `Rem` for consistency with `mir`. I think this also helps to differentiate it with `Mod` in types which has different semantics.

```suggestion
    Rem,
```

liquid-rust-syntax/src/grammar.lalrpop
```diff
@@ -138,7 +138,8 @@ BinOp5: ast::BinOp = {
 
 BinOp6: ast::BinOp = {
     "+" => ast::BinOp::Add,
-    "-" => ast::BinOp::Sub
+    "-" => ast::BinOp::Sub,
+    "%" => ast::BinOp::Mod,
```
nilehmann, line 142: This is more like a mental note, no need to address it now as we can change it at any point:

Maybe the syntax in the logic should be `mod` instead of `%` to stress the semantics difference between both.

tests/pos/new_ops.rs
```diff
@@ -0,0 +1,30 @@
...
+    input_must_be_odd(5);
+    input_must_be_odd(27);
+}
+
+#[lr::ty(fn(bool{x: x == true}) -> bool{y: y == true})]
+pub fn must_be_true(a: bool) -> bool {
+    a && true
+}
+
+pub fn test_and() {
```
nilehmann, line 22: I don't know if you did this intentionally, but we don't yet check the body of a function if it doesn't have a refinement annotation so this will be skiped.

nilehmann (changes requested):
@atgeller sorry for the delay. A couple of small comments, but overall looking very good!

tests/pos/new_ops.rs
atgeller: Whoops, not intentional, just forgot.

liquid-rust-typeck/src/checker.rs
atgeller: Done.

liquid-rust-syntax/src/grammar.lalrpop
atgeller: Agreed. Eventually perhaps we can even generate a configurable compiler warning if used on i32s and not usizes.

liquid-rust-core/src/ir.rs
atgeller: Done.

liquid-rust-typeck/src/checker.rs
```diff
@@ -471,6 +471,28 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
...
+            (
+                TyKind::Refine(BaseTy::Int(int_ty1), _e1),
+                TyKind::Refine(BaseTy::Int(int_ty2), _e2),
+            ) => {
+                debug_assert_eq!(int_ty1, int_ty2);
+                TyKind::Refine(
+                    BaseTy::Int(*int_ty1),
+                    ExprKind::Constant(liquid_rust_fixpoint::Constant::Bool(true)).intern(),
+                )
+                .intern()
```
nilehmann, line 483: This is not well-formed, it should be:

```suggestion
TyKind::Exists(BaseTy::Int(*int_ty1), Expr::tt().into()).intern()
```

nilehmann (changes requested):
@atgeller one more comment

liquid-rust-typeck/src/checker.rs
atgeller: Ah, that looks much better, thanks.

nilehmann (approved):

PR #25, opened 2022-01-19
Assert terminator support

atgeller:
Assert, a terminator in rust's mir, seems to be generated for under/overflow checking, as well as checking for div-by-zero. This includes supporting it essentially as a guarded goto, but also demonstrates the code needed to validate the absence of possible assert failures.

### DIFF
--- a/liquid-rust-core/src/ir.rs
+++ b/liquid-rust-core/src/ir.rs
@@ -53,6 +53,11 @@ pub enum TerminatorKind {
         place: Place,
         target: BasicBlock,
     },
+    Assert {
+        cond: Operand,
+        expected: bool,
+        target: BasicBlock,
+    },
 }
 
 pub struct Statement {
@@ -191,6 +196,9 @@ impl fmt::Debug for Terminator {
             }
             TerminatorKind::Drop { place, target } => {
                 write!(f, "drop({:?}) -> {:?}", place, target)
+            }
+            TerminatorKind::Assert { cond, target, expected } => {
+                write!(f, "Assert({:?} is expected to be {:?}) -> {:?}", cond, expected, target)
             }
         }
     }
--- a/liquid-rust-core/src/ir.rs
+++ b/liquid-rust-core/src/ir.rs
@@ -198,7 +198,7 @@ impl fmt::Debug for Terminator {
                 write!(f, "drop({:?}) -> {:?}", place, target)
             }
             TerminatorKind::Assert { cond, target, expected } => {
-                write!(f, "Assert({:?} is expected to be {:?}) -> {:?}", cond, expected, target)
+                write!(f, "assert({:?} is expected to be {:?}) -> {:?}", cond, expected, target)
             }
         }
     }
--- a/liquid-rust-typeck/src/checker.rs
+++ b/liquid-rust-typeck/src/checker.rs
@@ -229,7 +229,7 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
                 )?;
             }
             TerminatorKind::Assert { cond, expected, target } => {
-                self.check_assert(env, cursor, cond, expected, *target)?;
+                self.check_assert(env, cursor, cond, *expected, *target)?;
             }
             TerminatorKind::Drop { place, target } => {
                 let _ = env.move_place(place);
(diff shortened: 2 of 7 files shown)


liquid-rust-typeck/src/checker.rs
```diff
@@ -302,6 +305,33 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
...
+        let cond_ty = self.check_operand(env, cond);
+
+        let pred = match cond_ty.kind() {
+            TyKind::Refine(BaseTy::Bool, e) => {
+                e.clone()
+            }
+            _ => unreachable!("unexpected cond_ty {:?}", cond_ty),
+        };
+
+        let assert = ExprKind::BinaryOp(BinOp::Eq, pred, ExprKind::Constant(liquid_rust_fixpoint::Constant::Bool(*expected)).intern()).intern();
```
nilehmann, line 325: I would go with the following as I think it'll generate a slightly simpler constraint:

```suggestion
        let assert = if cond { pred } else { pred.not() }
```

liquid-rust-typeck/src/checker.rs
```diff
@@ -302,6 +305,33 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
         Ok(())
     }
 
+    fn check_assert(
+        &mut self,
+        env: &mut TypeEnv<'tcx>,
+        cursor: &mut Cursor,
+        cond: &Operand,
+        expected: &bool,
```
nilehmann, line 313:
```suggestion
        expected: bool,
```
https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

tests/neg/assert_terminator.rs
```diff
@@ -0,0 +1,9 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
+
+// Note: only fails if line 329 of liquid-rust-typeck/src/checker.rs is uncommented
+#[lr::ty(fn(usize, usize) -> usize)]
+pub fn assert_terminator_test(a: usize, b: usize) -> usize {
```
nilehmann, line 6: You need to add test explicitly to `liquid-rust-driver/tests/pos.rs`

tests/pos/assert_terminator.rs
```diff
@@ -0,0 +1,8 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
+
+#[lr::ty(fn(usize, usize{x: x > 0}) -> usize)]
+pub fn assert_terminator_test(a: usize, b: usize) -> usize {
```
nilehmann, line 5: You need to add test explicitly to `liquid-rust-driver/tests/neg.rs`

liquid-rust-typeck/src/checker.rs
```diff
@@ -302,6 +305,33 @@ impl<'a, 'tcx> Checker<'a, 'tcx> {
...
+            TyKind::Refine(BaseTy::Bool, e) => {
+                e.clone()
+            }
+            _ => unreachable!("unexpected cond_ty {:?}", cond_ty),
+        };
+
+        let assert = ExprKind::BinaryOp(BinOp::Eq, pred, ExprKind::Constant(liquid_rust_fixpoint::Constant::Bool(*expected)).intern()).intern();
+
+        // Uncomment the below line to allow pre-catching of possible divide-by-zero, underflow, and overflow
+        // WARNING: rust is very eager about inserting under/overflow checks, so be warned that uncommenting this will likely break everything
```
nilehmann, line 328: More than a `warning` I would add a `TODO` to make this user-configurable. We have some support for config flags here https://github.com/liquid-rust/liquid-rust/blob/main/liquid-rust-common/src/config.rs so we could make a new config to choose between both. It doesn't have to go in this PR though.

liquid-rust-core/src/ir.rs
```diff
@@ -192,6 +197,9 @@ impl fmt::Debug for Terminator {
             TerminatorKind::Drop { place, target } => {
                 write!(f, "drop({:?}) -> {:?}", place, target)
             }
+            TerminatorKind::Assert { cond, target, expected } => {
+                write!(f, "Assert({:?} is expected to be {:?}) -> {:?}", cond, expected, target)
```
nilehmann, line 201:
```suggestion
                write!(f, "assert({:?} is expected to be {:?}) -> {:?}", cond, expected, target)
```

nilehmann (changes requested):
@atgeller I just realized that no matter which optimization flags are passed to `rustc` it will always generate an assert for integer division and reminder, so we should land this before adding support for `%`.

Some comments below

atgeller:
I think the latest commits have resolved everything. I'll take a look at adding an option that just generates a warning/todo if set to true for now, and we can worry about threading it through later.

nilehmann (approved):
Lgtm

PR #48, opened 2022-02-21
Port tests to `surface` syntax

ranjitjhala:
Porting a bunch more of the top-level `pos` tests.


liquid-rust-tests/tests/pos/surface/kmp.rs
```diff
@@ -0,0 +1,66 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
+
+#[path = "../../lib/rvec.rs"]
+mod rvec;
+use rvec::RVec;
+
+// TODO: original lacks `&mut` but leaving it out triggers a "cannot instantiate generic" error
```
nilehmann, line 8: I think the reason this fails without the `mut` is because the desugaring incorrectly creates a `loc` and `ast::TyKind::Ref` to that `loc` for shared references, it should instead create an `ast::TyKind::ShrRef` and no `loc`.

ranjitjhala: This is fixed now.

PR #61, opened 2022-02-25
Explicit singletons in Surface syntax

ranjitjhala:
Add support for explicit singleton refinements e.g.

```rust
fn(x: &mut n@i32) -> i32; x: i32[n+1]
```


liquid-rust-driver/src/collector.rs
```diff
@@ -133,10 +133,12 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
         let kind = match (segment.ident.as_str(), &attr_item.args) {
             ("sig", MacArgs::Delimited(span, _, tokens)) => {
                 let fn_sig = self.parse(tokens.clone(), span.entire(), parse_fn_surface_sig)?;
+                // print!("LR::SIG {:#?} \n", fn_sig);
```
nilehmann, line 136: I recently added basic support for https://docs.rs/tracing/0.1.31/tracing/ we should gradually start building infrastructure around it. Just a mental note.

nilehmann (approved):
@ranjitjhala lgtm

Issue #73, opened 2022-03-08
Missing index triggers failures


ranjitjhala:
See 

1. In the example below, `lr` produces an error about expecting a pair `(int, int)` but only getting an `int`.

   https://github.com/liquid-rust/examples/blob/main/src/bug/mat-bug-00.rs
   
2. In the second example, `lr` just triggers some `fixpoint` "elaboration" crash perhaps due to some malformed sorts in the generated constraints.

   https://github.com/liquid-rust/examples/blob/main/src/bug/mat-bug-01.rs

nilehmann:
@ranjitjhala both errors are caused by the desugaring creating bindings with the wrong sort. There was a bug in the wf checker that let the second one go through, but I fixed that.

ranjitjhala:
Sweet I’ll take a look!

PR #75, opened 2022-03-09
Fix for #73

ranjitjhala:
Add a param in desugar `from_ty` to _not_ generate singleton/generics for references.


nilehmann (commented):
@ranjitjhala Can you also add a test with a mutable reference? I think that may go through a different path.

ranjitjhala:
Good idea, I added a test for that. To my great surprise, I had already accounted for that case :-)

PR #84, opened 2022-04-28
Default unrefined AdtDef

### LINKED ISSUE #83: Default specs with Polymorphism
Code like the below should work, using just the plain rust type signatures and nothing else

```rust
#![feature(register_tool)]
#![register_tool(lr)]

#[lr::sig(fn() -> i32[1])]
pub fn test2() -> i32 {
   let opt = Option::Some(1);
   opt.unwrap()
}
```

ranjitjhala:
Fixes #83


liquid-rust-tests/tests/neg/structs/default-option.rs
```diff
@@ -0,0 +1,16 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
+
+// We need this right now to not deal with the enum `Some`
+
+#[lr::assume]
+#[lr::sig(fn(T) -> Option<T>)]
```
nilehmann, line 7: You should be able to drop the `lr::sig` as this is the default signature, no?

liquid-rust-typeck/src/global_env.rs
```diff
@@ -24,6 +24,10 @@ pub struct GlobalEnv<'tcx> {
     adt_defs: FxHashMap<DefId, ty::AdtDef>,
 }
 
+fn default_adt_def() -> ty::AdtDef {
+    ty::AdtDef::Opaque { refined_by: crate::intern::Interned::from(vec![]) }
```
nilehmann, line 28:
```suggestion
    ty::AdtDef::Opaque { refined_by: crate::intern::List::from(vec![]) }
```

nilehmann (approved):
@ranjitjhala Nice! Some small comments, but it looks good to me

liquid-rust-tests/tests/neg/structs/default-option.rs
ranjitjhala: Yup, good catch!

PR #87, opened 2022-05-03
Add support for configuration of assert terminator behavior

atgeller:
Add support for configuring assert terminator behavior, either through a default environment variable, or a per-crate annotation. There are three options for the behavior:

1. "Ignore" (`default_assert_terminator_behavior=0/#![lr::assert_behavior(ignore)]`) -- Essentially treats an assert terminator as a goto, neither checking nor assuming the assert condition.
2. "Assume" (default option, `default_assert_terminator_behavior=1/#![lr::assert_behavior(assume)]`) -- Assume the asserted property but does not check it. Sound because rust execution will halt if the asserted property does not hold.
3. "Check" (`default_assert_terminator_behavior=2/#![lr::assert_behavior(check)]`) -- Checks the asserted condition and generates an error if the condition does not hold. The condition is also then "assumed" since it is known to hold.


liquid-rust-tests/tests/neg/assert_terminators/test-check.rs
```diff
@@ -0,0 +1,9 @@
+#![feature(register_tool)]
+#![register_tool(lr)]
+#![feature(custom_inner_attributes)]
+#![lr::assert_behavior(check)]
```
nilehmann, line 4: @atgeller I was thinking that it may be better to have a different syntax for these configurations, I was thinking something like:

```
#![lr::cfg(assert="check")]
```

I'm not quite sure how this attribute syntax works though.

nilehmann (commented):
@atgeller looks good but I have a comment about the syntax to use for the configuration.

liquid-rust-tests/tests/neg/assert_terminators/test-check.rs
atgeller: Sounds good, are you thinking a comma'ed list of `key=value`, and then validation would be performed when processing the attributes, or a comma'ed list of configuration settings, each with their own grammar object? Either way you would have to check for duplicates. In the first case error-handling would have to be done manually, but you only have to maintain it in one place (`collector.rs`, or some sub-file). In the second case, error-handling would be done automatically, but you have to maintain the syntax as well as the collector.

nilehmann: Rust have attributes which are a comma separated list of `key=value` so I'd use that. My hunch is that the values here are typically simple (something you'd expect as a cli argument) so no need for complex parsing. Consider also that the current implementation also needs to check for duplicates, you can specify the same crate attribute multiple times.

atgeller: I've figured out how to reuse the rustc tools for this, but it'll be a bit messy at first. I'm trying to make sure the code I write is mostly reusable for a generic cfg, but it's slow going. May not finish today.

atgeller:
We now support a general `#![lr::cfg(keyword = "value")]` attribute at the crate level. However, I don't think it's quite ready to merge yet, as I refactored the error handling and now it's a mess, so I should fix that and thoroughly test all the possible error cases.

atgeller:
Finally, at the moment, everything is expected to be a string that can then be parsed into the appropriate type. That's because it's very easy to get the value of a string from an `MetaItem` using `item.value_str()`. However, the Rust attribute language is much richer, and we could theoretically support `ints`, `bools`, and even nested lists using it. See the below links for more detail.

https://doc.rust-lang.org/stable/nightly-rustc/rustc_ast/ast/struct.MetaItem.html
https://doc.rust-lang.org/stable/nightly-rustc/rustc_ast/ast/enum.MetaItemKind.html
https://doc.rust-lang.org/stable/nightly-rustc/rustc_ast/ast/enum.LitKind.html

liquid-rust-common/src/config.rs
```diff
@@ -16,6 +45,7 @@ pub static CONFIG: SyncLazy<Config> = SyncLazy::new(|| {
             .set_default("log_dir", "./log/")?
             .set_default("dump_constraint", false)?
             .set_default("dump_checker_trace", false)?
+            .set_default("assert_terminator_behavior", "Assume")?
```
nilehmann, line 48: shouldn't this be "assume"?

liquid-rust-common/src/config.rs
```diff
@@ -3,11 +3,40 @@ use serde::Deserialize;
...
+    pub dump_checker_trace: bool,
+    pub assert_terminator_behavior: AssertBehaviorOptions,
+}
+
 #[derive(Deserialize)]
 pub struct Config {
     pub log_dir: PathBuf,
     pub dump_constraint: bool,
     pub dump_checker_trace: bool,
+    pub assert_terminator_behavior: AssertBehaviorOptions,
```
nilehmann, line 39: Let's call this `check_asserts`

liquid-rust-common/src/config.rs
```diff
@@ -3,11 +3,40 @@ use serde::Deserialize;
 use std::{io::Read, lazy::SyncLazy, path::PathBuf};
 pub use toml::Value;
 
+#[derive(Debug, Deserialize, Copy, Clone)]
+pub enum AssertBehaviorOptions {
```
nilehmann, line 7: Let's call this `AssertBehavior`

liquid-rust-common/src/config.rs
atgeller: I thought so too, but serde doesn't seem to use FromStr when deserializing, so it has to be uppercase to match the enum variant.

PR #89, opened 2022-05-10
Add hole to typeck::Pred

nilehmann:


ranjitjhala:
What is this doing? Adding an explicit "hole" form to simplify/clarify
downstream typing?

nilehmann:
Exactly, to simplify/clarify typing. More specifically, before we only had holes in `core::ty::Pred` but no in `typeck::ty:Pred` (stuff in `core::ty` are fundamentally source code annotations and stuff in `typeck::ty` are only generated during type checking). This worked well enough because we mostly instantiated refinements in function calls, which have a corresponding `core::ty` annotation, but now I want to implement "polymorphic folding" which requires creating types with holes on-the-fly when type checking. I think this also makes the inference phase cleaner which now runs on holes instead of dummy kvars.

PR #90, opened 2022-05-11
Preliminary support for type aliases

ranjitjhala:
See 

* [pos](liquid-rust-tests/tests/pos/surface/alias00.rs)
* [neg](liquid-rust-tests/tests/neg/surface/alias00.rs)

Now also works across files, e.g.

* [library](liquid-rust-tests/tests/lib/nat.rs) 
* [client](liquid-rust-tests/tests/neg/surface/alias01.rs)


liquid-rust-syntax/src/lexer.rs
```diff
@@ -11,6 +11,7 @@ use rustc_span::{symbol::kw, BytePos, Symbol};
 pub enum Token {
     Caret,
     EqEq,
+    Eq,
```
nilehmann, line 14: I think the problem is that you have to add a case to `Cursor::map_token` to add new tokens.

liquid-rust-syntax/src/lexer.rs
```diff
@@ -27,6 +28,7 @@ pub enum Token {
     Ge,
     At,
     Fn,
+    Type,
```
nilehmann, line 31: This is probably a failed experiment?

liquid-rust-syntax/src/resolve.rs
```diff
@@ -106,6 +106,7 @@ impl<'tcx> Resolver<'tcx> {
             }
             surface::Arg::StrgRef(loc, ty) => Ok(surface::Arg::StrgRef(loc, self.resolve_ty(ty)?)),
             surface::Arg::Ty(ty) => Ok(surface::Arg::Ty(self.resolve_ty(ty)?)),
+            _ => panic!("unexpected: resolve_arg"),
```
nilehmann, line 109: I prefer to be explicit about the missing case to catch errors if we add more variants.

nilehmann (approved):
Looks reasonable

liquid-rust-syntax/src/lexer.rs
ranjitjhala: Ah yes, that fixes it!

liquid-rust-syntax/src/lexer.rs
ranjitjhala: oops deleted!

liquid-rust-syntax/src/resolve.rs
ranjitjhala: Good point, have fixed -- btw, note that if we had a refinements, we should be able to prove this arm is unreachable...

nilehmann (approved):
I was worried about name clashes between flux aliases and rust types, but I think annotating rust aliases fixes this issue. I like it!

PR #92, opened 2022-05-16
Support for unit

nilehmann:
This is in preparation to use mir after analysis instead of the optimized version. Unit is represented as a nullary tuple as in rustc, but full support for tuples is not implemented yet.


ranjitjhala:
haha nice, this is surprisingly exciting :-)

PR #94, opened 2022-05-20
Support match statements

ranjitjhala:
Very basic support for `match` see `enum-match-00.rs`

* Changed the "control" from `Option<Expr>` to `Guard` -- proved unnecessary so far, but may come handy,
* Added a type for the result of `discriminator(...)`  rvalue, perhaps unnecessary
* Add a "memo" table to `pack_ty` and friends to avoid duplicate names.
* Remove unreachable names post-`join` (see `remove_dead_params`)


liquid-rust-typeck/src/type_env.rs
```diff
@@ -384,21 +390,26 @@ impl TypeEnvInfer {
...
             BaseTy::Int(_) | BaseTy::Uint(_) | BaseTy::Bool => bty.clone(),
         }
     }
 
-    pub fn join(&mut self, genv: &GlobalEnv, mut other: TypeEnv) -> bool {
+    /// join(self, genv, other) consumes the bindings in other, to "update"
+    /// 'self' in place, and returns 'true' if there was an actual change
+    /// or 'false' indicating no change (i.e. a fixpoint was reached).
+
+    pub fn join(&mut self, genv: &GlobalEnv, mut other: TypeEnv, target: BasicBlock) -> bool {
```
nilehmann, line 412:
```suggestion
    pub fn join(&mut self, genv: &GlobalEnv, mut other: TypeEnv) -> bool {
```

liquid-rust-typeck/src/checker.rs
```diff
@@ -84,6 +84,16 @@ pub struct Check<'a> {
     bb_envs: FxHashMap<BasicBlock, BasicBlockEnv>,
     kvars: &'a mut KVarStore,
 }
+/// A 'Guard' describes extra "control" information that holds at the start
+/// of the successor basic block
+enum Guard {
```
nilehmann, line 89: hmm I'm ambivalent about this, it is a good abstraction if we ever need to track variants, but it makes the code a bit complicated in some parts. I'd personally remove it, but no strong opinions.

liquid-rust-typeck/src/ty.rs
```diff
@@ -87,6 +87,8 @@ pub enum TyKind {
     Ref(RefKind, Ty),
     Param(ParamTy),
     Never,
+    /// Used internally to represent result of `discriminant` RVal
+    Discr,
```
nilehmann, line 91: I think in rustc a discriminant has an integer type and you can get the discriminant for a type with https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.discriminant_ty, but having a separate `Discr` type sounds like a better design to me, so let's leave it. worst case won't be too hard to remove this.

liquid-rust-typeck/src/ty.rs
```diff
@@ -395,6 +404,28 @@ impl TyS {
         matches!(self.kind(), TyKind::Uninit(..))
     }
 
+    pub fn vars_worker(&self, vars: &mut FxHashSet<Name>) {
```
nilehmann, line 407: Maybe `gather_vars`? Also, let's make this private, I think it's only being used in this file.
```suggestion
    fn gather_vars(&self, vars: &mut FxHashSet<Name>) {
```

liquid-rust-typeck/src/ty.rs
```diff
@@ -705,24 +745,25 @@ impl ExprS {
...
-                }
-                ExprKind::BinaryOp(_, e1, e2) => {
-                    go(e1, vars);
-                    go(e2, vars);
-                }
-                ExprKind::UnaryOp(_, e) => go(e, vars),
-                ExprKind::Proj(e, _) => go(e, vars),
-                ExprKind::Tuple(exprs) => exprs.iter().for_each(|e| go(e, vars)),
-                ExprKind::Var(Var::Bound(_)) | ExprKind::Constant(_) => {}
+    fn vars_worker(&self, vars: &mut FxHashSet<Name>) {
```
nilehmann, line 748: Same here
```suggestion
    fn gather_vars(&self, vars: &mut FxHashSet<Name>) {
```

liquid-rust-typeck/src/checker.rs
```diff
@@ -481,20 +499,24 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
                 TyKind::Indexed(bty @ (BaseTy::Int(_) | BaseTy::Uint(_)), exprs) => {
                     Expr::binary_op(BinOp::Eq, exprs[0].clone(), Expr::from_bits(bty, bits))
                 }
+                TyKind::Discr => Expr::bool(true),
```
nilehmann, line 502:
```suggestion
                TyKind::Discr => Expr::tt(),
```

liquid-rust-typeck/src/ty.rs
```diff
@@ -621,6 +657,10 @@ impl Expr {
         ExprKind::Constant(c).intern()
     }
 
+    pub fn bool(b: bool) -> Expr {
+        Expr::constant(Constant::Bool(b))
+    }
```
nilehmann, line 662: We dont' need this if we use `Expr::tt()` above

```suggestion
```

liquid-rust-typeck/src/type_env.rs
```diff
@@ -350,28 +354,30 @@ impl TypeEnvInfer {
                     .zip(sorts)
                     .map(|(e, sort)| {
                         let has_free_vars = !scope.contains_all(e.vars());
-                        if has_free_vars {
+                        if has_free_vars && !names.contains_key(e) {
```
nilehmann, line 357: Shouldn't we use the existing name if `has_free_vars && names.contains_key(e)`?

nilehmann (commented):
Mostly stylistics comments, except for one the one in `pack_ty`.

liquid-rust-typeck/src/type_env.rs
ranjitjhala: how odd -- my code has that explicit `!names.contains_key(e)` ... wonder why it didn't show up in the PR?

ranjitjhala: ah sorry, misunderstood. Yes let me think ...

ranjitjhala: ok fixed. take a look!

ranjitjhala:
Ok I think I fixed the `pack_ty` thing, and reverted `Guard` to `Option<Expr>` (left it as a type alias) with comment.

nilehmann (approved):
Lgtm

PR #102, opened 2022-06-03
Allow enum constructors

ranjitjhala:
See 

* [pos](liquid-rust/liquid-rust-tests/tests/pos/structs/enum-match-02.rs)
* [neg](liquid-rust/liquid-rust-tests/tests/neg/structs/enum-match-02.rs)

Main trick is to 

1. add the `Vec<ParamTy>` to hold the type-params of the enum
2. use those to compute the `variant_sig`.


liquid-rust-middle/src/core.rs
```diff
@@ -34,6 +35,7 @@ pub enum AdtDefKind {
 
 #[derive(Debug)]
 pub struct VariantDef {
+    pub def_id: DefId,
```
nilehmann, line 38: Let's remove this because we are not sure we are populating it with the right `def_id`

```suggestion
```

liquid-rust-desugar/src/desugar.rs
```diff
@@ -53,12 +54,12 @@ pub fn desugar_struct_def(
             .into_iter()
             .map(|ty| cx.desugar_ty(ty.unwrap()))
             .try_collect_exhaust()?;
-        let variants = Some(IndexVec::from_raw(vec![Some(VariantDef { fields })]));
+        let variants = Some(IndexVec::from_raw(vec![Some(VariantDef { def_id, fields })]));
```
nilehmann, line 57:
```suggestion
        let variants = Some(IndexVec::from_raw(vec![Some(VariantDef { fields })]));
```

liquid-rust-middle/src/global_env.rs
```diff
@@ -94,7 +94,18 @@ impl<'tcx> GlobalEnv<'tcx> {
             .iter()
             .map(|variant| self.default_variant_def(variant))
             .collect();
-        ty::AdtDef::transparent(adt_def.did, vec![], variants)
+        let generics = self.adt_def_generics(adt_def.did);
+        ty::AdtDef::transparent(adt_def.did, generics, vec![], variants)
+    }
+    pub fn adt_def_generics(&self, def_id: DefId) -> Vec<rustc_middle::ty::ParamTy> {
```
nilehmann, line 100: There should be a space between these functions. I wonder why rustfmt is not complaining 🤔 

```suggestion
    }
    
    pub fn adt_def_generics(&self, def_id: DefId) -> Vec<rustc_middle::ty::ParamTy> {
```

liquid-rust-middle/src/ty/mod.rs
```diff
@@ -276,10 +278,17 @@ impl AdtDef {
         &self.0.sorts
     }
 
+    pub fn generics(&self) -> &Vec<ParamTy> {
```
nilehmann, line 281: a `&Vec<T>` is rarely more useful than a `&[T]` so let's return the slice to hide implementation details.

```suggestion
    pub fn generics(&self) -> &[ParamTy] {
```

nilehmann (approved):
@ranjitjhala LGTM, a couple of small details. Also I think you didn't push the pos case.

PR #111, opened 2022-06-23
Post-condition error span

ranjitjhala:
Mainly: Don't report whole function, instead try to localize to last statement/expression where the post-condition was checked.

Two additional things

* adds qualifiers `v = x + y` and `v = x - y` to builtin set

*  allows parsing specs of form `x : & mut ...` (for symmetry with other kinds of binders)


flux-typeck/src/checker.rs
```diff
@@ -308,6 +319,31 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
         Ok(())
     }
 
+    fn is_stmt_nop(stmt: &Statement) -> bool {
```
nilehmann, line 322: I'd define this as a method of `Statement` and call it `is_nop`. Also, you can follow clippy suggestion.

```rust
fn is_nop(&self) -> bool {
    matches!(self.kind, StatementKind::Nop)
}
```

flux-typeck/src/checker.rs
```diff
@@ -308,6 +319,31 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+        self.body.basic_blocks[bb]
+            .statements
+            .iter()
+            .all(|stmt| Self::is_stmt_nop(stmt))
+    }
+
+    fn is_bb_ret(&self, bb: BasicBlock) -> bool {
+        match &self.body.basic_blocks[bb].terminator {
+            None => false,
+            Some(term) => matches!(term.kind, TerminatorKind::Return),
```
nilehmann, line 339: I'd define a method in `Terminator`

```rust
fn is_return(&self) -> bool {
    matches!(self.kind, TerminatorKind::Return)
}
```

flux-typeck/src/checker.rs
```diff
@@ -308,6 +319,31 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+        }
+    }
+
+    fn is_exit(&self, bb: BasicBlock) -> bool {
+        self.is_bb_nop(bb) && self.is_bb_ret(bb)
```
nilehmann, line 344: these tiny little functions don't feel idiomatic to me. I'd inline them

```rust
fn is_exit_block(&self, bb: BasicBlock) -> bool {
    let data = self.body.basic_bloks[bb];
    let is_no_op = data.statements.iter().all(Statement::is_nop);
    let is_ret = data.terminator.map(Terminator::is_return).unwrap_or(false);
    is_no_op && is_ret
}
```

flux-typeck/src/checker.rs
```diff
@@ -379,8 +422,9 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
         &mut self,
         rcx: &mut RefineCtxt,
         env: &mut TypeEnv,
+        tag: Tag,
```
nilehmann, line 425: I'd pass the `source_info` as an argument here and create the tag inside `check_ret`

nilehmann (approved):
@ranjitjhala looking good, just some stylistic comments.

flux-typeck/src/checker.rs
ranjitjhala: done.

PR #115, opened 2022-07-15
Refined Traits

ranjitjhala:
(Replaces https://github.com/liquid-rust/flux/pull/114)

### Refined Iterators

* [Rng](flux-tests/lib/range.rs) and 
* [RVec](flux-tests/lib/rvec.rs)

### Positive Tests

* [iter00](flux-tests/pos/surface/iter00.rs)
* [iter01](flux-tests/pos/surface/iter01.rs)

### Negative Tests

* [neg](flux-tests/neg/surface/iter00.rs)
* [neg](flux-tests/neg/surface/iter01.rs)

## Limitations

Currently, we only support trait-impls where

* we can resolve the exact impl at the call-site and 
* have a refined/signature without any `ProjectionTy`

## Notes

**Lowering**

- Define `struct Instance { def_id: DefId, substs: GenericArgs }`
- Change the lowered `Call` to have an field `instance: Option<Instance>`
- At lowering time, compute the `instance` field using the [Instance API](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/instance/struct.Instance.html)

**Checking**

- If a `Call` has an `instance` field, then use that instead of the original `def_id` and `substs`

**Extra Stuff**

This PR does some work that is somewhat unnecessary at the moment

- `ParamEnv` inside the lowered mir
- Original `SubstRef` inside each `Call` (in a `CallSubsts`)

but perhaps will be needed to enable trait projection normalization
for when we do not have an instance as described
[here](https://internals.rust-lang.org/t/accessing-the-defid-of-a-trait-implementation/17001)


flux-middle/src/rustc/mir.rs
```diff
@@ -40,21 +41,35 @@ pub struct LocalDecl {
...
     Return,
     Call {
         func: DefId,
-        substs: List<GenericArg>,
+        substs: CallSubsts<'tcx>,
         args: Vec<Operand>,
         destination: Place,
         target: Option<BasicBlock>,
         cleanup: Option<BasicBlock>,
+        instance: Option<Instance>,
```
nilehmann, line 72: I wonder if it is cleaner to just store the most specific "Instance" (DefId/Substs pair) we can find here instead of both, then the rest of the checking will proceed as normal. But at the same time, I guess having more information is better than having less information 🤔

flux-desugar/src/zip_resolver.rs
```diff
@@ -102,16 +102,23 @@ fn zip_path(path: Path, rust_ty: &rustc_ty::Ty) -> Path<Res> {
...
         .collect();
+
+    // TODO(RJ): LIFETIME-ARGS
+    // if path_args_len < rust_args.len() {
+    // for rust_arg in &rust_args[path_args_len..] {
+    // let ty = _default_generic_arg(rust_arg, path.span);
+    // args.push(ty)
+    // }
+    // }
+
```
nilehmann, line 121: remove
```suggestion
```

flux-desugar/src/zip_resolver.rs
```diff
@@ -120,3 +127,25 @@ fn zip_generic_arg(arg: Ty, rust_arg: &rustc_ty::GenericArg) -> Ty<Res> {
...
+// let args = substs
+// .iter()
+// .map(|arg| default_generic_arg(arg, span))
+// .collect();
+// let path = Path { ident, args, span };
+// Ty { kind: TyKind::Path(path), span }
+// }
+// _ => todo!(),
+// }
+// }
```
nilehmann, line 151: remove

```suggestion
```

flux-tests/tests/pos/surface/range.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub fn new(lo: i32, hi: i32) -> Self {
+        Self { _lo: lo, hi, cur: lo }
+    }
+}
+
+impl Iterator for Rng {
+    type Item = i32;
+
+    #[flux::sig(fn(self: &mut Rng[@lo, @hi]) -> Option<i32{v:lo <= v && v < hi}>)]
+    fn next(&mut self) -> Option<i32> {
```
nilehmann, line 25: I think we had this conversation, but what else is needed to give this a stronger type that says the return value is exactly `Some(cur)` as long as the range hasn't been exhausted?

Not for this PR.

nilehmann (approved):
@ranjitjhala lgtm, just a couple of comments and thoughts

flux-middle/src/rustc/mir.rs
ranjitjhala: Yes, unclear to me too, but didn't want to erase the original info...

flux-tests/tests/pos/surface/range.rs
ranjitjhala: Well, we could technically use the `cur`-indexed iterator and then just say that its `cur`? 

```
#[flux::sig(fn(self: &mut RngIter[@lo, @cur, @hi]) -> Option<i32[cur]>]
```

ranjitjhala:
ok have removed comments, will merge if this builds!

flux-tests/tests/pos/surface/range.rs
nilehmann: We would also need to make the reference strong and update `curr` at the output, but you should only update it when returning `Some`.

This also makes me think to what extent is sound to make it a strong reference here but a mutable reference in the trait definition.

PR #116, opened 2022-07-19
Add support for `flux::ignore` pragma to skip analyzing a module or whole crate

ranjitjhala:
See example tests

Pos

* [00](flux-tests/tests/neg/surface/ignore00.rs)

Neg

* [00](flux-tests/tests/neg/surface/ignore00.rs)
* [01](flux-tests/tests/neg/surface/ignore01.rs)

### DIFF
--- a/flux-driver/src/callbacks.rs
+++ b/flux-driver/src/callbacks.rs
@@ -80,9 +80,16 @@ struct CrateChecker<'a, 'genv, 'tcx> {
     assume: FxHashSet<LocalDefId>,
     ignores: Ignores,
 }
+
+/// `is_ignored` transitively follows the `def_id` 's parent-chain to check if
+/// any enclosing mod has been marked as `ignore`
 fn is_ignored(tcx: &TyCtxt, ignores: &Ignores, def_id: &LocalDefId) -> bool {
     let parent_def_id = tcx.parent_module_from_def_id(*def_id);
-    ignores.contains_key(&Some(parent_def_id))
+    if parent_def_id == *def_id {
+        false
+    } else {
+        ignores.contains_key(&Some(parent_def_id)) || is_ignored(tcx, ignores, &parent_def_id)
+    }
 }
 
 impl<'a, 'genv, 'tcx> CrateChecker<'a, 'genv, 'tcx> {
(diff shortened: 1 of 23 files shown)


flux-driver/src/collector.rs
```diff
@@ -27,12 +27,19 @@ pub(crate) struct SpecCollector<'tcx, 'a> {
     error_guaranteed: Option<ErrorGuaranteed>,
 }
 
+/// An IgnoreKey is either `Some(module_def_id)` or `None` indicating the entire crate
+pub type IgnoreKey = Option<LocalDefId>;
```
nilehmann, line 31: Maybe better to have an explicit enum

```suggestion
#[derive(PartialEq, Eq, Hash)]
pub enum IgnoreKey {
    Crate,
    Module(LocalDedIf)
}
```

flux-driver/src/collector.rs
```diff
@@ -27,12 +27,19 @@ pub(crate) struct SpecCollector<'tcx, 'a> {
...
+
+/// Set of module (`LocalDefId`) that should be ignored by flux
+pub type Ignores = FxHashMap<IgnoreKey, ()>;
```
nilehmann, line 34: Make this a set

```suggestion
pub type Ignores = FxHashSet<IgnoreKey>;
```

flux-driver/src/callbacks.rs
```diff
@@ -137,8 +162,13 @@ impl<'a, 'genv, 'tcx> CrateChecker<'a, 'genv, 'tcx> {
         let aliases = specs.aliases;
 
         // Function signatures
-        specs
+        let fn_specs: Vec<(LocalDefId, crate::collector::FnSpec)> = specs
             .fns
+            .into_iter()
+            .filter(|(def_id, _)| !is_ignored(&genv.tcx, &specs.ignores, def_id))
+            .collect();
```
nilehmann, line 169: I’d follow clippy suggestion here. You can either add this filter to the chain below or early return inside the `try_for_each_exhaust`

nilehmann (approved):
@ranjitjhala lgtm, just some stylistics comments

flux-driver/src/callbacks.rs
ranjitjhala: Where do you see the clippy suggestion? (I had to do this split as the borrow checker was yelling about `tcx` being consumed by the filter...?

ranjitjhala: nvm pushed into the `try_for_each_exhaust`

PR #117, opened 2022-07-20
Add support for `Constr` types 

ranjitjhala:
Add's support for specifications to use types of the form `{T : p}` which let us mix 
refinement constraints and parameter binders, e.g.

[**Function Signatures**](/flux-tests/tests/pos/surface/constr00.rs)

```rust
fn(&i32[@a], { &i32[@b] : b <= a } ) -> i32[a-b]
```

[**Struct Invariants**](/flux-tests/tests/pos/surface/constr01.rs)

```rust
#[flux::refined_by(a: int, b: int)]
struct Foo {
    #[flux::field({i32[@a]: 0 <= a})]
    x: i32,
    #[flux::field({i32[@b]: a < b})]
    y: i32,
}
```


flux-middle/src/ty/mod.rs
```diff
@@ -88,6 +88,7 @@ pub enum TyKind {
     Uninit,
     Ptr(Path),
     Ref(RefKind, Ty),
+    Constr(Pred, Ty),
```
nilehmann, line 91: Maybe let's make this an expression. In theory, we should allow kvars here, but fixpoint makes it really hard to have a kvar if it's not directly associated with a binder and we probably shouldn't fight against it right now. I guess there's a practical difference between a normal (existential) refinement type and the composition of an existential + a constr

```suggestion
Constr(Expr, Ty)
```

flux-typeck/src/constraint_gen.rs
```diff
@@ -220,6 +220,12 @@ fn subtyping(genv: &GlobalEnv, constr: &mut ConstrBuilder, ty1: &Ty, ty2: &Ty, t
                 subtyping(genv, constr, ty1, ty2, tag);
             }
         }
+
+        (_, TyKind::Constr(p2, ty2)) => {
```
nilehmann, line 224: I think we can also have `Constr` on the left when recursively checking subtyping on type arguments. The following should work for example:

```
Option<{i32[a] : a > 0}> <: Option<i32{v : v > 0 }>
```

 To implement this we need to expose `assume_pred` on `ConstrBuilder`

flux-typeck/src/wf.rs
```diff
@@ -141,6 +141,10 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
                     .try_for_each_exhaust(|ty| self.check_type(env, ty))
             }
             core::Ty::Never | core::Ty::Param(_) | core::Ty::Float(_) => Ok(()),
+            core::Ty::Constr(pred, ty) => {
```
nilehmann, line 144: nit: I'd move this up so the "catch all" arm is at the end.

flux-typeck/src/type_env/paths_tree.rs
```diff
@@ -152,6 +152,7 @@ impl PathsTree {
                     }
                     PlaceElem::Deref => {
                         let ty = node.expect_ty();
+                        let ty = ty.unconstr();
```
nilehmann, line 155: Was this failing in one of the examples you tried? I feel it should be covered by a call to `unpack` somewhere else. Everything going on in this file is very subtle and I'm not entirely sure if this line is sound so I'd prefer to remove it if it's not strictly necessary for the examples we have so far.

flux-middle/src/ty/fold.rs
```diff
@@ -253,6 +254,10 @@ impl TypeFoldable for Ty {
             | TyKind::Discr(..)
             | TyKind::Float(_)
             | TyKind::Uninit => {}
+            TyKind::Constr(pred, ty) => {
```
nilehmann, line 257: nit: I'd move this up so the "catch all" arm is at the end.

flux-middle/src/ty/fold.rs
```diff
@@ -232,6 +232,7 @@ impl TypeFoldable for Ty {
             | TyKind::Param(_)
             | TyKind::Never
             | TyKind::Discr(..) => self.clone(),
+            TyKind::Constr(pred, ty) => Ty::constr(pred.fold_with(folder), ty.fold_with(folder)),
```
nilehmann, line 235: nit: I'd move this up so the "catch all" arm is at the end.

nilehmann (changes requested):
@ranjitjhala nice! I really like this type.

I think there are a couple of issues we need to address before merging though

flux-middle/src/ty/mod.rs
ranjitjhala: Yes this is easy enough -- went back and forth on this a bit, let me make it `Expr`.

flux-typeck/src/constraint_gen.rs
ranjitjhala: Yes, this is also easy BUT I couldn't think of a way to actually trigger such a subtyping via a test so had left it for now

flux-typeck/src/type_env/paths_tree.rs
ranjitjhala: Aha, so yes this _was_ failing a test which necessitated this code, BUT it turns out that was _before_ I handled the `unpack` properly, and now in fact, it is not needed. good catch!

flux-typeck/src/constraint_gen.rs
ranjitjhala: ah figured it out. Anyways, you should **closely review** the code for `constr.push_guard(p)` in the updated PR!

PR #120, opened 2022-07-29
Impl Index and IndexMut for RVec

nilehmann:
* Implement `Index` and `IndexMut` for `RVec`
* Update tests using `RVec` to use square brackets syntax


ranjitjhala:
Awesome! I didn’t realize this was locked behind traits.

PR #121, opened 2022-08-01
Support for `Box`

ranjitjhala:
This adds support for the `Box` type by

* Adding trivially refined versions of _default_ parameters (e.g. `Box<i32>` is actually `Box<i32, Global>` )
* Allowing _deref_-ing `Box` values
* Allowing index-parameters to occur inside e.g. `Box<i32[@n]>`


flux-typeck/src/type_env/paths_tree.rs
```diff
@@ -160,7 +162,10 @@ impl PathsTree {
                             TyKind::Ref(mode, ty) => {
                                 return self.lookup_place_iter_ty(rcx, gen, *mode, ty, place_proj);
                             }
-                            _ => panic!("YIKES: {elem:?} {ty:?}"),
+                            TyKind::Indexed(bty, _) | TyKind::Exists(bty, _) if bty.is_box() => {
+                                return self.lookup_place_box_ty(bty);
+                            }
```
nilehmann, line 167: We could match on the `bty` to avoid the panic in `lookup_place_box_ty`, but I don't have strong feelings for it. I always feel this type of deep matching sounds good conceptually, but not it doesn't look so good once you write it...

```suggestion
                            TyKind::Indexed(BaseTy::Adt(def, substs), _) | TyKind::Exists(BaseTy::Adt(def, substs), _) if def.is_box() => {
                                return LookupResult::Box(substs[0].clone());
                            }
```

flux-typeck/src/type_env/paths_tree.rs
```diff
@@ -189,6 +204,7 @@ impl PathsTree {
                     return match self.lookup_place_iter(rcx, gen, ptr_path.clone(), proj) {
                         LookupResult::Ptr(_, ty2) => LookupResult::Ref(rk, ty2),
                         LookupResult::Ref(rk2, ty2) => LookupResult::Ref(rk.min(rk2), ty2),
+                        LookupResult::Box(ty2) => LookupResult::Box(ty2),
```
nilehmann, line 207: We should return `LookupResult::Ref(rk, ty2)` here. This code is exercised when you have `x: &Box<T>` and then do `**x`, the result is of type `T` but it's behind a shared reference so we keep track of it.

flux-desugar/src/zip_resolver.rs
```diff
@@ -90,26 +98,41 @@ fn zip_ty(ty: Ty, rust_ty: &rustc_ty::Ty) -> Ty<Res> {
...
+    // zip the supplied args
+    let mut args: Vec<Ty<Res>> = iter::zip(path.args, rust_args)
+        .map(|(arg, rust_arg)| zip_generic_arg(tcx, arg, rust_arg))
         .collect();
 
+    // copy over the 'default' args with trivial refinements
+    if 0 < default_args_len {
+        match res {
+            Res::Adt(def_id) => {
+                let generics = &tcx.generics_of(def_id).params[path_args_len..];
```
nilehmann, line 120: I feel it'd be better if we add the missing default parameters in the lowering (`core::Ty -> ty::Ty`) and here we just resolve the parameters we do have. The reason I think it'd be better is that in the lowering we can use https://github.com/liquid-rust/flux/blob/main/flux-middle/src/global_env.rs#L207 to get the trivially refined type. But I'm not sure if that will cause any other problems, so I'm ok with revisiting it later.

In general, I'm not happy with generating default types in `surface` or `core`.

flux-tests/tests/todo/list.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(Debug)]
+pub enum List<T> {
+    Nil,
+    Cons(T, Box<List<T>>),
+}
+
+pub fn len<T>(l: &List<T>) -> i32 {
+    match l {
+        List::Nil => 0,
+        List::Cons(_, tl) => 1 + len(&*tl),
```
nilehmann, line 15: I think the following will also work, but I always get deref coercion wrong. Also, the auto-deref version may generate different code.

```suggestion
        List::Cons(_, tl) => 1 + len(tl),
```

flux-tests/tests/todo/list.rs
```diff
@@ -0,0 +1,67 @@
...
+    match l {
+        List::Nil => never(0),
+        List::Cons(h, _) => h,
+    }
+}
+
+pub fn tail<T>(l: &List<T>) -> &List<T> {
+    match l {
+        List::Nil => never(0),
+        List::Cons(_, t) => &*t,
```
nilehmann, line 29: Same here about deref coercion.
```suggestion
        List::Cons(_, t) => t,
```

flux-desugar/src/desugar.rs
```diff
@@ -429,9 +429,14 @@ impl ParamsCtxt<'_> {
             }
             surface::TyKind::Constr(_, ty) => self.ty_gather_params(ty, adt_sorts),
 
-            surface::TyKind::Path(_) | surface::TyKind::Exists { .. } | surface::TyKind::Unit => {
+            surface::TyKind::Path(path) => {
+                for ty in &path.args {
+                    let _ = self.ty_gather_params(ty, adt_sorts);
```
nilehmann, line 434: We shouldn't ignore this error
```suggestion
                    self.ty_gather_params(ty, adt_sorts)?;
```

nilehmann, line 434: So what's your take about the following definition now?

```rust
fn foo(Vec<i32[@n]>) {}
```

should we fail at the call site (fail to infer parameter) or at the definition site (you can't have a binder there). If we agree on the latter, maybe we can leave a TODO for adding a check somewhere to catch this situation.

flux-typeck/src/type_env.rs
```diff
@@ -125,8 +125,8 @@ impl TypeEnv {
             LookupResult::Ref(RefKind::Mut, _) => {
                 panic!("cannot move out of `{place:?}`, which is behind a `&` reference")
             }
-            LookupResult::Ref(RefKind::Shr, _) => {
-                panic!("cannot move out of `{place:?}`, which is behind a `&mut` reference")
+            LookupResult::Ref(RefKind::Shr, _) | LookupResult::Box(_) => {
```
nilehmann, line 128: It's ok to move out of `Box`, e.g., 


```rust
fn foo(x: Box<Vec<i32>>) -> Vec<i32> {
    *x
}
```

nilehmann (changes requested):

flux-desugar/src/zip_resolver.rs
ranjitjhala: There are 3 options

1. Add the params here (i.e. `zip_resolver`)
2. Ignore the params and just deal with the mismatched args as the appear
3. Add the params later (e.g. in `lowering`)

(2) seems dicey and a bit ad-hoc (we might miss some cases and cause odd silent failures.)

(3) is challenging at that point (`lowering`) we don't have the original `rustc` signature so we 
don't know what extra params to add.

Hence (1) seemed the least worst for now...

flux-typeck/src/type_env/paths_tree.rs
ranjitjhala: ok!

flux-desugar/src/zip_resolver.rs
nilehmann: I think 3 can be done, but getting the default type is more circuitous. I [found this today](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/generics/struct.GenericParamDef.html#method.default_value) (I randomly read rustc source in my free time) which I think will serve the purpose. But I'm not terribly convinced this will end up being cleaner or that the function I found will get us what we need.

flux-desugar/src/desugar.rs
ranjitjhala: I added the code to check for the illegal binder at the definition site -- so now the above should be rejected (see test `param_infer00.rs`)

flux-typeck/src/type_env.rs
ranjitjhala: hmm -- ok so what should this code be? `move_place` ?

flux-desugar/src/zip_resolver.rs
ranjitjhala: OK let me add a link to this discussion in the code, and we can revisit later. [ IMO doing it in 3 is going to duplicate a bunch of the resolving work... ]

flux-typeck/src/type_env.rs
ranjitjhala: Changing the code to the following seems to work?

```rust
    pub fn move_place(&mut self, rcx: &mut RefineCtxt, gen: &mut ConstrGen, place: &Place) -> Ty {
        match self.bindings.lookup_place(rcx, gen, place) {
            LookupResult::Ptr(path, ty) => {
                self.bindings.update(&path, Ty::uninit());
                ty
            }
            LookupResult::Ref(RefKind::Mut, _) => {
                panic!("cannot move out of `{place:?}`, which is behind a `&mut` reference")
            }
            LookupResult::Ref(RefKind::Shr, _) => {
                panic!("cannot move out of `{place:?}`, which is behind a `&` reference")
            }
            LookupResult::Box(ty) => ty,
        }
    }
```

ranjitjhala: (also added a test for this in `box01.rs`)

nilehmann (approved):
@ranjitjhala let's leave a comment about moving out of `Box` as we discussed in slack and ship it as it is. Also, clippy is giving you a suggestion.

flux-typeck/src/type_env.rs
ranjitjhala: added comment!

PR #138, opened 2022-09-05
Add a poly-opt test for enums

ranjitjhala:


flux-tests/tests/pos/enums/opt01.rs
```diff
@@ -0,0 +1,39 @@
...
+#[flux::sig(fn(i32{v:false}) -> T)]
+pub fn never<T>(_x: i32) -> T {
+    loop {}
+}
+
+#[flux::sig(fn(Opt<T>[true]) -> T)]
+pub fn unwrap<T>(x: Opt<T>) -> T {
+    match x {
+        Opt::Some(v) => v,
+        Opt::None => unwrap(x),
```
nilehmann, line 37: Shouldn't we call never here?

ranjitjhala: oops. yes, fixed!

nilehmann (approved):

PR #139, opened 2022-09-11
Check Signatures

### LINKED ISSUE #137: Odd crash with polymorphic enum
The code below 

```rust
#![feature(register_tool)]
#![register_tool(flux)]

#[flux::refined_by(b:bool)]
pub enum Opt<T> {
    #[flux::variant(Opt<T>[false])]
    None,
    #[flux::variant(Opt<T>[true])]
    Some(T),
}

#[flux::sig(fn(Opt<i32>[true]) -> i32)]
pub fn unwrap_i32(x: Opt<i32>) -> i32 {
    match x {
        Opt::Some(v) => v,
        Opt::None => 0,
    }
}
```

panics with

```
thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', flux-typeck/src/type_env/paths_tree.rs:495:22
```

@nilehmann  The error is in a part of `paths_tree` I'm unfamiliar with so would be good if you can take a look!

ranjitjhala:
This PR adds a bunch of code to check that the `flux` and `rust` signatures align (the earlier combinations of `zip_resolver` or `table_resolver` either skipped checking altogether or erased the flux type.) Still not nice as we have to use the table-resolver for string resolution and `adt_def_sym` is a hack, but nevertheless an improvement, IMO.

See the new kinds of errors in

* flux-tests/tests/neg/error_messages/fun_sig_error.rs 
* flux-tests/tests/neg/error_messages/variant_sig_error.rs 

fixes #137


flux-syntax/src/surface_grammar.lalrpop
```diff
@@ -65,12 +65,12 @@ pub FnSig: surface::FnSig = {
 }
 
 pub Variant: surface::VariantDef = {
-     <tys:Fields?> <ret:Ty> => {
+    <lo:@L> <tys:Fields?> <ret:Ty> <hi:@L> => {
```
nilehmann, line 68:
```suggestion
    <lo:@L> <tys:Fields?> <ret:Ty> <hi:@R> => {
```

flux-driver/src/callbacks.rs
```diff
@@ -171,7 +171,7 @@ impl<'a, 'genv, 'tcx> CrateChecker<'a, 'genv, 'tcx> {
             .enums
             .into_iter()
             .try_for_each_exhaust(|(def_id, enum_def)| {
-                let enum_def = desugar::desugar_enum_def(genv, &adt_sorts, enum_def)?;
+                let enum_def = desugar::desugar_enum_def(genv, &adt_sorts, def_id, enum_def)?;
```
nilehmann, line 174: I think the EnumDef has the LocalDefId as a field so you don’t need to pass it as an extra argument here.

flux-desugar/src/lib.rs
```diff
@@ -14,35 +14,52 @@ mod table_resolver;
...
-    let mut resolver = table_resolver::Resolver::from_adt(genv, enum_def.def_id)?;
-    let enum_def = resolver.resolve_enum_def(enum_def)?;
+    let rust_adt_def = genv.tcx.adt_def(def_id.to_def_id());
+    let resolver = table_resolver::Resolver::from_adt(genv, def_id)?;
+    let rust_enum_def = rustc::lowering::lower_enum_def(genv.tcx, rust_adt_def)?;
+    let enum_def = zip_resolver::ZipResolver::new(genv.sess, &resolver)
+        .zip_enum_def(enum_def, &rust_enum_def)?;
     desugar::desugar_enum_def(genv.sess, &genv.consts, adt_sorts, enum_def)
+    // // let resolver = table_resolver::Resolver::from_adt(genv, enum_def.def_id)?;
+    // // let enum_def = resolver.resolve_enum_def(enum_def)?;
```
nilehmann, line 49:
```suggestion
```

nilehmann (approved):
I’m going to accept this because I think it’s an improvement to fail gracefully instead of crashing but I’m in general against this direction (see below). I left a couple of small comments and also there’s an empty file (`not_well_formed.rs`)


I don’t like mixing the resolution of strings into DefId with checking whether a flux signature is a refinement of the rust signature. Ideally, I’d like the resolution to be completely independent of the thing you are refining. Checking whether the flux signature is a refinement of the rust signature should happen later in an independent step against the resolved flux signature (maybe even against the desugared flux signature). The only advantage I see to mixing them up is to support `_` as in LH to avoid some keystrokes. We could probably deal with this as a special case anyways. Also I think `_`  is the wrong way to avoid keystrokes. The way to avoid keystrokes IMO is with a proc macro like so:

```rust
#[flux]
fn foo(x: i32{x > 0}) -> i32{v : v > 0} {
    x + 1
}
```

ranjitjhala:
Yes, its not pretty but definitely a local improvement; have merged in the spirit of "not letting the perfect be the enemy of the good" :-) 

Doing as a separate phase is tricky with the current (`zip_resolver`) setup because the zipping ignores the user's specification (e.g. `bool`) and just uses the plain rust type (e.g. `i32`) which
is undesirable and thwarts any downstream check. Of course, we could revert to the old `table_resolver` method to avoid this, but presumably there was some reason why we moved to the zip thing that I now forget?

The macro idea is definitely great but has the catch of breaking "backwards compatibility"? (i.e. code should compile with plain rust without any macro trickery)

Anyways, the stuff here should be easy enough to remove/revert once we have a better replacement.

PR #145, opened 2022-09-22
Basic support for arrays

nilehmann:
Add basic support for `TyKind::Array` and `AggregateValue::Array`. The typing rule for for `AggregateValue` is very simple, it creates a fresh kvariables for the type and checks that all the arguments are a subtype, this should allow us to prove that all elements in the array satisfy some property. The way the array len is implemented is a bit hacky, but this should allow us to make some progress. There is still work left to be done to support the examples in https://github.com/liquid-rust/flux/issues/142.


ohpauleez:
Thanks so much for this!

PR #148, opened 2022-09-24
Add support for dot-access in refinements

ranjitjhala:
See 

* flux-tests/tests/pos/structs/dot00.rs
* flux-tests/tests/neg/structs/dot00.rs

most action is in desugar but its getting a bit gross. 

anyways, may as well start reviewing -- i will add some more test cases for graceful error messages etc.


nilehmann (approved):

ranjitjhala:
Nice, thanks!!! In brief it looks like the key was to get rid of the impl and replace with plain function call so you can specify lifetimes?

nilehmann:
More concretely, the reason everything is sound is that `self.sorts` and `self.fresh_dot_params` access disjoint subsets of the fields in `self`. Normally, you would have to refactor both into free functions passing explicitly the subset of fields they access (or further decoupling and put the different subsets inside others structs). In this case, because we hold onto AdtSorts for a longer lifetime it's only necessary to make `sorts` into a free function. There are some ideas going around to add support for specifying which fields are accessed by a method https://smallcultfollowing.com/babysteps//blog/2021/11/05/view-types/

ranjitjhala:
Interesting! I definitely would not have been able to figure that out!

PR #149, opened 2022-09-24
Add error messages for incorrect indices

### LINKED ISSUE #143: Fail with error message when missing index
The code below 

```rust
#![feature(register_tool)]
#![register_tool(flux)]

#[flux::refined_by(x:int, y:int)]
pub struct Pair {
    #[flux::field(i32[@x])]
    pub x: i32,
    #[flux::field(i32[@y])]
    pub y: i32,
}

#[flux::sig(fn(Pair[@x]) -> i32)]
pub fn mytuple(p:Pair) -> i32 {
    p.x
}
```

just panics with

```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', /Users/rjhala/research/rust/flux/flux-desugar/src/desugar.rs:444:17
``` 

but instead we should pinpoint the offending spot.

ranjitjhala:
Fixes #143


flux-desugar/src/desugar.rs
```diff
@@ -480,7 +485,7 @@ impl<'a> ParamsCtxt<'a> {
         fields: &[Symbol],
         sorts: &[Sort],
     ) -> Vec<(Symbol, Param)> {
-        assert_eq!(sorts.len(), fields.len()); // TODO error message
+        assert_eq!(sorts.len(), fields.len());
```
nilehmann, line 488: Typically, we are using the `debug_assert*` family for this type of checks.

flux-tests/tests/todo/struct_types/I143.rs
```diff
@@ -9,8 +9,12 @@ pub struct Pair {
     pub y: i32,
 }
 
-#[flux::sig(fn(Pair[@x]) -> i32)]
-pub fn mytuple(p:Pair) -> i32 {
+#[flux::sig(fn(Pair[@p,@q,@r]) -> i32[p])]
```
nilehmann, line 12: This file has already been considered in other tests so we can remove it, no?

nilehmann (approved):
@ranjitjhala Couple of comments but looks good. Also take a look at the clippy suggestions

flux-tests/tests/todo/struct_types/I143.rs
ranjitjhala: Yes let me remove.

Issue #154, opened 2022-10-07
How to write a refinement on a single variable of a struct?


cole-k:
Ranjit asked me to post this question, although it seems rather simple (and perhaps ill-posed) to me.

I was playing around with flux and tried to extend the definition of `RangeIter` in `range.rs` to include the current index `cur`.

Original code:
https://github.com/liquid-rust/flux/blob/4e440050fe87c8edbd22271a761378eae3ca683e/flux-tests/tests/lib/range.rs#L7-L8

My change:
```rust
     #[flux::field({i32[@cur]:lo <= cur && cur <= hi})]
     cur: i32,
```

There are some other changes too, which I can post if desired.

What I ran into was an issue of not knowing how to express myself.
For the type signature of the function `next` I wanted to add an `ensures` clause that reflects the fact that in the updated `RangeIter`, the new `cur` is now either equal to `cur + 1` or `hi`.

Original code:
https://github.com/liquid-rust/flux/blob/4e440050fe87c8edbd22271a761378eae3ca683e/flux-tests/tests/lib/range.rs#L36

My attempted change:
```rust
     #[flux::sig(fn(self: &strg RngIter[@lo, @hi, @cur]) -> Option<i32{v:lo <= v && v < hi && v == cur}> ensures self: {RngIter[lo, hi, x]: x == (cur + 1) || x == hi})]
```

Flux says this is wrong because `x` is unbound, which makes sense to me. But I can't seem to figure out how to introduce an existential like it.

I can think of three explanations:
1. I don't know the right syntax.
2. The right syntax doesn't exist, but could.
3. The right syntax doesn't exist, and this shouldn't be possible.

ranjitjhala:
Consider the spec for `RangeIter`

```rust
#[flux::refined_by(lo: int, hi: int)]
pub struct RngIter {
    #[flux::field(i32[@lo])]
    _lo: i32,
    #[flux::field(i32[@hi])]
    hi: i32,
    #[flux::field(i32{v:lo <= v && v <= hi})]
    cur: i32,
}
```

What this says is that in "refinement-land" any `RangeIter` is associated with (i.e. abstracted by) _two_ indices -- respectively `lo` and `hi` which are really the _only_ things you can talk about.

(This is _very_ different from LH...)

So when you say `RangeIter[x, y]` in some spec, it means a `RangeIter` whose `lo` index is `x` and `hi` index is `y` -- something like three params doesn't even make sense as a `RangeIter` only has _two_ indices.

[...]

PR #155, opened 2022-10-07
Add more refinements to range.rs

cole-k:
# High-level details

Refines the types more in `range.rs` to serve as an example of how one might refine part of a struct.

See #154 for more information.

# Notes

I don't see this being run as a part of `cargo test -p flux-tests`. Not sure if it is deprecated or just not being tested. I could submit a PR to add it.

### DIFF
--- a/flux-tests/tests/lib/range.rs
+++ b/flux-tests/tests/lib/range.rs
@@ -1,15 +1,15 @@
-#[flux::refined_by(lo: int, hi: int)]
+#[flux::refined_by(lo: int, hi: int, cur: int)]
 pub struct RngIter {
     #[flux::field(i32[@lo])]
     _lo: i32,
     #[flux::field(i32[@hi])]
     hi: i32,
-    #[flux::field(i32{v:lo <= v && v <= hi})]
+    #[flux::field({i32[@cur]: lo <= cur && cur <= hi})]
     cur: i32,
 }
 
 impl RngIter {
-    #[flux::sig(fn(lo:i32, hi:i32{lo <= hi}) -> RngIter[lo, hi])]
+    #[flux::sig(fn(lo:i32, hi:i32{lo <= hi}) -> RngIter[lo, hi, lo])]
     pub fn new(lo: i32, hi: i32) -> RngIter {
         Self { _lo: lo, hi, cur: lo }
     }
@@ -33,7 +33,12 @@ impl Rng {
 impl Iterator for RngIter {
     type Item = i32;
 
-    #[flux::sig(fn(self: &mut RngIter[@lo, @hi]) -> Option<i32{v:lo <= v && v < hi}>)]
+    #[flux::sig(fn(self: &strg RngIter[@in]) 
+                -> Option<i32{nextV: in.lo <= nextV && nextV < in.hi 
+                                  && nextV == in.cur}> 
+                ensures self: RngIter{out: out.lo == in.lo && out.hi == in.hi 
+                                        && out.cur == (in.cur + 1) || out.cur == in.hi})
+     ]
     fn next(&mut self) -> Option<i32> {
         let cur = self.cur;
         let hi = self.hi;
@@ -50,7 +55,7 @@ impl IntoIterator for Rng {
     type Item = i32;
     type IntoIter = RngIter;
 
-    #[flux::sig(fn(Rng[@lo, @hi]) -> RngIter[lo, hi])]
+    #[flux::sig(fn(Rng[@lo, @hi]) -> RngIter[lo, hi, lo])]
     fn into_iter(self) -> RngIter {
         RngIter::new(self.lo, self.hi)
     }


flux-tests/tests/lib/range.rs
```diff
@@ -33,7 +33,12 @@ impl Rng {
 impl Iterator for RngIter {
     type Item = i32;
 
-    #[flux::sig(fn(self: &mut RngIter[@lo, @hi]) -> Option<i32{v:lo <= v && v < hi}>)]
+    #[flux::sig(fn(self: &strg RngIter[@in]) 
+                -> Option<i32{nextV: in.lo <= nextV && nextV < in.hi 
+                                  && nextV == in.cur}> 
+                ensures self: RngIter{out: out.lo == in.lo && out.hi == in.hi 
+                                        && out.cur == (in.cur + 1) || out.cur == in.hi})
+     ]
```
cole-k, line 41: I don't know what the standards are for formatting these signatures so I gave it my best go.

ranjitjhala: Sweet! Can you think of a new client that can be verified using this more precise spec? Then you can add a new test in `flux-tests/tests/pos/surface/iter01` that exercises the spec? Perhaps something like:

```
pub fn test_next() {
    let mut rng = RngIter::new(0, 10);
    let x0 = rng.next();
    let x1 = rng.next();
    let x2 = rng.next();
    assert ( x0 == 0 && x1 == 1 && x2 == 2);
}
```

nilehmann: also wondering if it's not enough to type the output like `Option<i32[in.curr]>`. If not I would like to understand why.

ranjitjhala:
Good point! I think Cole was just modifying the old spec but having exposed
cur we can do this.

flux-tests/tests/lib/range.rs
cole-k: `Option<i32[in.curr]>` works. That did not cross my mind.

cole-k:
I see now that this is a library being used as part of the tests – actually it is causing a test failure in https://github.com/liquid-rust/flux/blob/main/flux-tests/tests/pos/surface/iter00.rs so I can look into that later today.

```
error[FLUX]: precondition might not hold
  --> $DIR/iter00.rs:14:9
   |
14 |         assert(10 <= val && val < 15)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[FLUX]: precondition might not hold
  --> $DIR/iter00.rs:20:9
   |
20 |         assert(110 <= val && val < 115)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'll have to think about whether it's possible this precondition might be false.

nilehmann:
Tried it out a bit and the simplest spec I could come up with that also works with the tests we have is the following

```rust
#[flux::refined_by(lo: int, hi: int)]
pub struct RngIter {
    #[flux::field(i32[@lo])]
    lo: i32,
    #[flux::field({i32[@hi] : lo <= hi})]
    hi: i32,
}

impl Iterator for RngIter {
    type Item = i32;

    #[flux::sig(
      fn(self: &strg RngIter[@in]) -> Option<i32{v: in.lo < in.hi && v == in.lo}>
      ensures self: RngIter{v: v.hi == in.hi && (v.lo == in.lo + 1 || v.lo == in.hi)}
    )]
    fn next(&mut self) -> Option<i32> {
        let cur = self.lo;
        if cur < self.hi {
            self.lo += 1;
            Some(cur)
        } else {
            None
        }
    }
}
```

it needs extra qualifiers for flux-tests/tests/pos/surface/iter00.rs though

cole-k:
> Tried it out a bit and the simplest spec I could come up with that also works with the tests we have is the following
> 
> ```rust
> #[flux::refined_by(lo: int, hi: int)]
> pub struct RngIter {
>     #[flux::field(i32[@lo])]
>     lo: i32,
>     #[flux::field({i32[@hi] : lo <= hi})]
>     hi: i32,
> }
> 
> impl Iterator for RngIter {
>     type Item = i32;
> 
>     #[flux::sig(
>       fn(self: &strg RngIter[@in]) -> Option<i32{v: in.lo < in.hi && v == in.lo}>
>       ensures self: RngIter{v: v.hi == in.hi && (v.lo == in.lo + 1 || v.lo == in.hi)}
>     )]
>     fn next(&mut self) -> Option<i32> {
>         let cur = self.lo;
>         if cur < self.hi {
>             self.lo += 1;
>             Some(cur)
>         } else {
>             None
>         }
>     }
> }
> ```
> 
> it needs extra qualifiers for flux-tests/tests/pos/surface/iter00.rs though

Could you explain why it needs to do away with `cur`? Intuitively the code should be the same with that additional variable.

nilehmann:
There are a couple of things going on. The code as you send it with the extra `cur` and the signature 

```rust
    #[flux::sig(fn(self: &strg RngIter[@in]) 
                -> Option<i32{nextV: in.lo <= nextV && nextV < in.hi 
                                  && nextV == in.cur}> 
                ensures self: RngIter{out: out.lo == in.lo && out.hi == in.hi 
                                        && out.cur == (in.cur + 1) || out.cur == in.hi})
     ]
```
should pass [iter00.rs](https://github.com/liquid-rust/flux/blob/05abbccdd7bc71de507a83c3921906b8ab5bf3b6/flux-tests/tests/pos/surface/iter00.rs). The problem though is that by taking a strong reference the code generates more inference variables and it's missing qualifiers. If you add 

```
#![flux::qualifier(Ge10(a: int) : 10 <= a)]
#![flux::qualifier(Le15(a: int) : a <= 15)]
```
`test_next` should pass.

More concretely, before adding the strong reference, the type of `rng` was `RngIter[10, 15]` and didn't change inside the loop, so there was nothing to infer. After adding the strong reference the type of `rng` can change and flux has to infer a strong enough type at the head of the loop, i.e., a type `RngIter{v : $k(v)}` for some `$k`. For annoying reasons flux and fixpoint don't realize that they can use 10 and 15 to instantiate qualifiers and it fails to solve for `$k`.

ranjitjhala:
Nico since you have this paged in -- can you run with `LR_DUMP_TRACE=1` and
paste in the constraint file? either the "lrc" or "smt2".

nilehmann:
@ranjitjhala 
## New spec
```rust
fn(self: &strg RngIter[@in]) -> Option<i32{v: in.lo < in.hi && v == in.lo}>
ensures self: RngIter{v: v.hi == in.hi && (v.lo == in.lo + 1 || v.lo == in.hi)}
 ```
 ```
 (10 ≤ 15) ~ Call(12:19: 12:39)
$k1(10) ~ Goto(12:9: 12:16, bb2)
∀ a0: int, a1: int.
  $k1(a0) ⇒
    ∀ a2: int, a3: int.
      (a3 = a1 ∧ (a2 = a0 + 1 ∨ a2 = a1)) ⇒
        ∀ a4: int.
          (a4 < a1 ∧ a4 = a0) ⇒
            (a4 < a1 ∧ a4 = a0) ~ Goto(13:15: 13:24, bb5)
            ∀ a5: int.
              (a5 < a1 ∧ a5 = a0) ⇒
                ¬(10 ≤ a5) ⇒ $k0(false) ~ Goto(14:16: 14:37, bb9)
                (10 ≤ a5) ⇒ $k0(a5 < 15) ~ Goto(14:16: 14:37, bb9)
                ∀ a6: bool.
                  $k0(a6) ⇒
                    (a6 = true) ~ Call(14:9: 14:38)
                    (a5 < a1 ∧ a5 = a0) ~ Goto(13:5: 15:6, bb2)
                    $k1(a2) ~ Goto(13:5: 15:6, bb2)
 ```
 
## Old spec

```rust
fn(self: &mut RngIter[@lo, @hi]) -> Option<i32{v:lo <= v && v < hi}>
```
```
(10 ≤ 15) ~ Call(12:19: 12:39)
∀ a0: int.
  (10 ≤ a0 ∧ a0 < 15) ⇒
    (10 ≤ a0 ∧ a0 < 15) ~ Goto(13:15: 13:24, bb5)
    ∀ a1: int.
      (10 ≤ a1 ∧ a1 < 15) ⇒
        ¬(10 ≤ a1) ⇒ $k0(false) ~ Goto(14:16: 14:37, bb9)
        (10 ≤ a1) ⇒ $k0(a1 < 15) ~ Goto(14:16: 14:37, bb9)
        ∀ a2: bool.
          $k0(a2) ⇒
            (a2 = true) ~ Call(14:9: 14:38)
            (10 ≤ a1 ∧ a1 < 15) ~ Goto(13:5: 15:6, bb2)
```

nilehmann:
If we refine `Option` with a boolean we can give the following spec that captures the exact behavior. Sadly, since we can't refine external types and functions yet, we can't implement `Iterator`.

```rust
#![feature(register_tool, custom_inner_attributes)]
#![register_tool(flux)]
#![flux::qualifier(Le10(x: int) : x <= 10)]

#[flux::refined_by(lo: int, hi: int)]
pub struct RngIter {
    #[flux::field(i32[@lo])]
    lo: i32,
    #[flux::field({ i32[@hi] : lo <= hi })]
    hi: i32,
}

#[flux::refined_by(b: bool)]
pub enum Opt<T> {
    #[flux::variant(Opt<T>[false])]
    None,
    #[flux::variant({T} -> Opt<T>[true])]
    Some(T),
}

#[flux::sig(fn() -> T requires false)]
pub fn unreachable<T>() -> T {
    unreachable!()
}

#[flux::sig(fn(bool[true]) -> ())]
pub fn assert(_: bool) {}

impl<T> Opt<T> {
    #[flux::sig(fn(Opt<T>[true]) -> T)]
    pub fn unwrap(self) -> T {
        match self {
            Opt::Some(v) => v,
            Opt::None => unreachable(),
        }
    }
}

impl RngIter {
    #[flux::sig(
      fn(self: &strg RngIter[@lo, @hi]) -> Opt<i32[lo]>[lo < hi]
      ensures self: RngIter[if lo < hi { lo + 1 } else { hi }, hi]
    )]
    pub fn next(&mut self) -> Opt<i32> {
        let cur = self.lo;
        if cur < self.hi {
            self.lo += 1;
            Opt::Some(cur)
        } else {
            Opt::None
        }
    }
}

pub fn test() {
    let mut iter = RngIter { lo: 0, hi: 2 };

    let fst = iter.next().unwrap();
    let snd = iter.next().unwrap();

    assert(fst == 0);
    assert(snd == 1);
}

pub fn test_iter() {
    let mut iter = RngIter { lo: 0, hi: 10 };
    while let Opt::Some(v) = iter.next() {
        assert(0 <= v);
        assert(v < 10);
    }
}
```

(I implemented if-then-else in https://github.com/liquid-rust/flux/pull/161)

cole-k:
So I've worked my way through the refinement types tutorial and I'm a little confused by the VC for the old spec. I'm not really ready to read through Flux's full spec so I might just be missing a nuance there.

This part looks like it's coming from the `if` statement

```
    ∀ a1: int.
      (10 ≤ a1 ∧ a1 < 15) ⇒
        ¬(10 ≤ a1) ⇒ $k0(false, a0, a1) ~ Goto(14:16: 14:37, bb9)
        (10 ≤ a1) ⇒ $k0(a1 < 15, a0, a1) ~ Goto(14:16: 14:37, bb9)
```

but `10 ≤ a1` is not what's being checked in the `if` statement. It's `a1 < 15`.

Here's the function definition:

```rust
#[flux::sig(fn(self: &mut RngIter[@lo, @hi]) -> Option<i32{v:lo <= v && v < hi}>)]
fn next(&mut self) -> Option<i32> {
    let cur = self.cur;
    let hi = self.hi;
    if cur < hi {
        self.cur = cur + 1;
        Some(cur)
    } else {
        None
    }
}
```

If I were to guess what the VC would be for the function with an argument of type `RngIter[10,15]`, it would look something like

```
∀ self.lo: int.
  (self.lo = 10) ⇒
    ∀ self.hi: int.
      (self.hi = 15) ⇒
        ∀ self.cur: int.
          (self.lo ≤ self.cur ∧ self.cur < self.hi) ⇒
            ∀ cur: int.
              (cur = self.cur) ⇒
              ∀ hi: int.
                (hi = self.hi) ⇒
                ∀ hi: int.
                  ¬(cur < hi) ⇒
                   true (trivial)
                  (cur < hi) ⇒
                    ∀ x: int.
                      (x = cur + 1) ⇒ (self.lo ≤ x ∧ x < self.hi)
```

Obviously there's a lot of guesswork here, especially with mutable variables and structs, but I'm still a little confused how we end up with that assertion.

nilehmann:
@cole-k yeah I think there has been some confusion. the constraints I posted are for the verification of this function https://github.com/liquid-rust/flux/blob/main/flux-tests/tests/pos/surface/iter00.rs#L11 against the spec of `RngIter`. The constraint you get for the verification of `RngIter::next` (with the old spec) is the following

```
∀ a0: int, a1: int, a2: int.
  (a0 ≤ a2 ∧ a2 ≤ a1) ⇒
    ¬(a2 < a1) ⇒
      ∀ a3: int.
        $k0(a3, a0, a1, a2) ⇒ (a0 ≤ a3 ∧ a3 < a1) ~ RetAt(44:13: 44:17)
    (a2 < a1) ⇒
      (a0 ≤ a2 + 1 ∧ a2 + 1 ≤ a1) ~ Assign(41:13: 41:31)
      $k1(a2, a0, a1, a2) ~ Call(42:13: 42:22)
      ∀ a3: int.
        $k1(a3, a0, a1, a2) ⇒ (a0 ≤ a3 ∧ a3 < a1) ~ RetAt(42:13: 42:22)
```

which would look something like this using the original names

```
∀ lo: int, hi: int, cur: int.
  (lo ≤ cur ∧ cur ≤ hi) ⇒
    ¬(cur < hi) ⇒
      ∀ a3: int.
        $k0(a3, lo, hi, cur) ⇒ (lo ≤ a3 ∧ a3 < hi) ~ RetAt(44:13: 44:17)
    (cur < hi) ⇒
      (lo ≤ cur + 1 ∧ cur + 1 ≤ hi) ~ Assign(41:13: 41:31)
      $k1(cur, lo, hi, cur) ~ Call(42:13: 42:22)
      ∀ a3: int.
        $k1(a3, lo, hi, cur) ⇒ (lo ≤ a3 ∧ a3 < hi) ~ RetAt(42:13: 42:22)
```

nilehmann:
I guess this are the interesting bits from the constraint 

```
∀ a3: int.
    $k0(a3, lo, hi, cur) ⇒ (lo ≤ a3 ∧ a3 < hi) ~ RetAt(44:13: 44:17)
```
This comes from the else branch and the type of `None`. We assign `None` type `Option<i32{v : $k0(v, lo,hi,cur)}>` for some `$k0` that needs to be inferred (and can mention the variables in scope `lo`, `hi` and `cur`). Because we are returning this `None` the subtyping constraint `Option<i32{v : $k0(v,...)}> <: Option<i32{v : lo ≤ v ∧ v < hi} >` needs to hold which in turn generates the constraint above


```
(lo ≤ cur + 1 ∧ cur + 1 ≤ hi) ~ Assign(41:13: 41:31)
```
Because `curr` needs to be between `lo` and `hi` we have to check this when assigning `self.cur = cur + 1`

```
$k1(cur, lo, hi, cur) ~ Call(42:13: 42:22)
∀ a3: int.
    $k1(a3, lo, hi, cur) ⇒ (lo ≤ a3 ∧ a3 < hi) ~ RetAt(42:13: 42:22)
```
This comes from the type of `Some`. `Some` is a constructor which we give the type `i32{v : $k1(v, lo, hi, cur)} -> Option<i32{v : $k1(v, lo, hi, cur)}> `. When calling the constructor we need to check the type of `cur` is a subtype of the `i32{v : $k1(v, lo, hi, cur)}`, hence the first constraint. Then the type of `Some(cur)` is `Option<i32{v : $k1(v, lo, hi, cur)}> ` which needs to be a subtype of `Option<i32{v : lo ≤ v ∧ v < hi} >`, hence the second constraint.

nilehmann:
Closing this for inactivity

PR #166, opened 2022-10-12
Unchecked invariants

nilehmann:
This is part of https://github.com/liquid-rust/flux/issues/165. 

Adds a new flux annotation to specify invariants for enums. Invariants are assumed every time a value of that type is in scope. They are currently not verified and could introduce unsoundness.

### DIFF
--- a/flux-desugar/src/desugar.rs
+++ b/flux-desugar/src/desugar.rs
@@ -27,9 +27,8 @@ pub fn desugar_qualifier(
     sess: &FluxSession,
     consts: &[ConstInfo],
     qualifier: surface::Qualifier,
-    adt_sorts: &AdtSorts,
 ) -> Result<Qualifier, ErrorGuaranteed> {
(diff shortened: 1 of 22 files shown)


ranjitjhala:
Very nice! I think it should be possible to slap on a check at each constructor where:

1. You get to _assume_ the invariant holds for all the input values
2. You have to _guarantee_ the invariant holds for the output. 

For example, given

```rust
#[flux::refined_by(n:int)]
#[flux::invariant(n >= 0)]
pub enum List {
    #[flux::variant(List[0])]
    Nil,
    #[flux::variant({i32, Box<List[@n]>} -> List[n+1])]
    Cons(i32, Box<List>),
}
```

You get two constraints:

```
// Nil [subst output index into invariant formula]
0 >= 0 

// Cons
forall n. n >= 0 => n+1 >= 0
```

?

PR #171, opened 2022-10-14
Implement default return

nilehmann:
Closes https://github.com/liquid-rust/flux/issues/110


ranjitjhala:
sweet!

ranjitjhala:
I took the liberty of merging as I just hit something that would benefit from this! Thanks and apologies!

nilehmann:
but that's the best part. It's like unwrapping someone else's present

Issue #165, opened 2022-10-11
Verified implied constraints on ADT indices


nilehmann:
Indices on ADT sometimes have implied constraints. For example, we know by [its definition](https://github.com/liquid-rust/flux/blob/d119f9a17620205905bf05b2372ef920665d4fe5/flux-tests/tests/pos/enums/list00.rs#L9) that the length of a list is always non-negative. These constraints are sometimes necessary to prove the safety of a program, for example, in the following function we can only prove `n != 0`  in the second branch if we know the length of the tail is non-negative (`n >=0 => n + 1 != 0`):

```rust
#[flux::sig(fn(&List[@n]) -> bool[n == 0])]
pub fn empty(l: &List) -> bool {
    match l {
        List::Nil => true,
        List::Cons(_, _) => false,
    }
}
```
The current workaround is for a variant to [return a constrained type](https://github.com/liquid-rust/flux/blob/d119f9a17620205905bf05b2372ef920665d4fe5/flux-tests/tests/pos/enums/list00.rs#L13). This constraint is however not verified and could introduce unsoundness. We should verify these constraints.

PR #174, opened 2022-10-15
Verify invariants

nilehmann:
closes https://github.com/liquid-rust/flux/issues/165

Verify that `#[flux::invariant(...)]` annotations actually hold. Closes https://github.com/liquid-rust/flux/issues/165

To verify if an invariant holds for an ADT, we check for each variant/constructor that the invariant holds for the indices in the output, assuming all the preconditions that can be extracted from the arguments (this calls the unpack function in RefineCtxt). This includes invariants on the input types (which can be recursive). Each invariant is proven in isolation, but when extracting the preconditions other invariants on the input types can be assumed. I'm unsure if this is actually sound when there are complex interactions between invariants and/or mutually recursive types.


nilehmann:
@ranjitjhala can you comment on the strategy?

nilehmann:
I'm just going to merge this as I think it is already an improvement over completely unverified invariants. We can revisit later and now I'm mostly worried about getting in the necessary infrastructure to verify the invariants.

ranjitjhala:
👍🏽

ranjitjhala:
@nilehmann shouldn't we also check e.g. that the output type is _not_ constrained? AFAICT this is silently still allowed?


```rust
#![feature(register_tool)]
#![register_tool(flux)]

#[flux::refined_by(n:int)]
pub enum List<T> {
    #[flux::variant(List<T>[0])]
    Nil,
    #[flux::variant({T, Box<List<T>[@n]>} -> {List<T>[n+1]: 100 <= n})]
    Cons(T, Box<List<T>>),
}
```

Should we keep #165 open till this is plugged?

nilehmann:
@ranjitjhala yeah, even more restrictive the output can only be an indexed type. I think we can enforce that syntactically. I'll open a new issue.

nilehmann:
As a side note of we enforce it syntactically I think we could get rid of the curly braces and use normal parentheses.

PR #193, opened 2022-10-20
Rename `core` to `fhir`

nilehmann:
The name core was confusing as it made it sound like it was the lowest of the levels in the pipeline, when in fact we have even lower levels. I borrowed the name hir from rustc to refer to something a bit lower than the surface syntax.


ranjitjhala:
Haha, good! It was also puzzling because IIRC there were _multiple_ things
called `core` … even within `flux` ?

nilehmann:
I don't know within flux, but this was also confusing because there's also the `core` from the rust standard library.

PR #200, opened 2022-10-22
Inst bug

ranjitjhala:
Fix for odd param-inst bug


nilehmann (approved):
@ranjitjhala looks good

PR #207, opened 2022-10-22
Disable incremental compilation

nilehmann:
If incremental compilation is enabled then the rust borrow check won't run and we will fail to retrieve the mir causing a panic. This removes the `-C incremental=1` flag if found in the args.

This fixes the problem with running flux inside vscode with the `RUSTC_WRAPPER` hack.


ranjitjhala:
Awesome thanks!!!

Issue #184, opened 2022-10-18
Support slice and array indexing


nilehmann:
Implement support for [ProjectionElem::Index](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/syntax/enum.ProjectionElem.html#variant.Index). This requires a bit of refactoring to the code in https://github.com/liquid-rust/flux/blob/0520bc222f31fa9c3293f93eebe9088c5031b65f/flux-typeck/src/type_env/paths_tree.rs.

PR #223, opened 2022-11-10
Some progress on array support

ranjitjhala:
Issue #184 now we allow reading (but no bounds checks yet)

### DIFF
--- a/flux-middle/src/fhir.rs
+++ b/flux-middle/src/fhir.rs
@@ -134,7 +134,7 @@ pub enum Ty {
     Str,
     Char,
     Ptr(Ident),
-    Ref(RefKind, Box<Ty>),
+    Ref(WeakKind, Box<Ty>),
     Param(ParamTy),
     Tuple(Vec<Ty>),
     Array(Box<Ty>, ArrayLen),
@@ -145,9 +145,10 @@ pub enum Ty {
 pub struct ArrayLen;
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
-pub enum RefKind {
+pub enum WeakKind {
     Shr,
     Mut,
+    Arr,
 }
 
 pub struct Indices {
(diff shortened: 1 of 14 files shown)


flux-typeck/src/type_env/paths_tree.rs
```diff
@@ -262,6 +262,21 @@ impl PathsTree {
                             _ => panic!("Unsupported Deref: {elem:?} {ty:?}"),
                         }
                     }
+                    PlaceElem::Index(_) => {
+                        let ty = ptr.borrow().expect_owned();
+                        match ty.kind() {
+                            TyKind::Indexed(BaseTy::Array(arr_ty, _), _)
+                            | TyKind::Exists(BaseTy::Array(arr_ty, _), _) => {
```
nilehmann, line 269: `TyKind::Exists` shouldn't be possible here because the type should be already unpacked. If it is not, then there's a bug somewhere else.

flux-typeck/src/type_env/paths_tree.rs
```diff
@@ -262,6 +262,21 @@ impl PathsTree {
...
+                                let (rk, ty) =
+                                    Self::lookup_ty(genv, rcx, RefKind::Shr, arr_ty, place_proj)?;
```
nilehmann, line 271: Ok, this is the main issue. This is technically wrong. Right now `LookupKind` can be either `Ref` or `Node` reflecting the fact that the place can either point to a _path_ in the tree (something you can take a strong reference to) or nested inside a reference (mutable or shared). We need to add an extra case saying the place can also point inside an array/slice. I'd do the following.

First, modify `LookupKind`

```rust
enum LookupKind {
    Strg(Path, NodePtr),
    Weak(WeakKind, Ty),
}
    
enum WeakKind {
    Shr,
    Mut,
    Array,
}
```

Second, modify `lookup_ty` to take a `WeakKind` instead of a `RefKind` (`FoldResult` should be modified accordingly). Note that `WeakKind` doesn't affect how refinement checking works as we only care whether we know the exact location or not (so we know if we can strongly borrow it or not), but we keep track of it to avoid shooting ourselves in the foot.

I don't love the term `Weak`, but I don't have anything better (actually the main reason I haven't implemented this is that I don't have a better name XD)

We should also add tests for taking references into the array:

```rust
fn foo() {
    let mut arr = [1, 2, 3];
    let x = &arr[0];
    let y = &mut arr[0];
}    
```

nilehmann (commented):
@ranjitjhala some comments

flux-typeck/src/type_env/paths_tree.rs
nilehmann: re weak vs strg terminology:

I think using strg vs weak is misleading. I'm just realizing this after writing the paper, but the relevant distinction is whether we know the exact location or not (strong updates are a consequence of knowing the exact location). I which we had some terminology for that.

ranjitjhala:
@nilehmann I made the various edits you suggested. Take a look -- it does result in a bunch of `Ref(WeakKind::Arr,...)` which seems like they should be unreachable?

flux-middle/src/fhir.rs
```diff
@@ -145,9 +145,10 @@ pub enum Ty {
 pub struct ArrayLen;
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
-pub enum RefKind {
+pub enum WeakKind {
```
nilehmann, line 148: Oh, I think I explained myself wrong. We should not modify `RefKind`, but create a new type `WeakKind` which is only used in `PathsTree`

ranjitjhala:
@nilehmann am sending a new PR as I messed this up with my git incompetence...

PR #224, opened 2022-11-10
Tuple

### LINKED ISSUE #109: Surface support for tuples
We should have enough support for tuples in [flux-typeck](https://github.com/liquid-rust/flux/tree/main/flux-typeck), but they are not yet exposed in the surface syntax. Implementing this should require the following steps

1. Add a new `Tuple` variant to the surface syntax (https://github.com/liquid-rust/flux/blob/main/flux-syntax/src/surface.rs#L100).
1. Add code for desugaring the new variant (https://github.com/liquid-rust/flux/blob/main/flux-desugar/src/desugar.rs#L205).
1. Modify the parser to support tuples (https://github.com/liquid-rust/flux/blob/main/flux-syntax/src/surface_grammar.lalrpop). 
1. Remove the `Unit` variant from surface (https://github.com/liquid-rust/flux/blob/main/flux-syntax/src/surface.rs#L121).
1. Add some [pos](https://github.com/liquid-rust/flux/tree/main/flux-tests/tests/pos/structs) and [neg](https://github.com/liquid-rust/flux/tree/main/flux-tests/tests/neg/structs) tests, e.g.,
    ```rust
    #[flux::sig(fn((Nat, i32)) -> Nat)]
    fn test(pair: (i32, i32)) -> i32 {
        pair.0
    }
    ```
    The support in [flux-typeck](https://github.com/liquid-rust/flux/tree/main/flux-typeck) should be sufficient by now, but maybe we will find some surprices and will need to fix some bugs.

ranjitjhala:
Fixes #109


flux-desugar/src/annot_check.rs
```diff
@@ -186,9 +186,16 @@ impl<'genv, 'tcx> ZipChecker<'genv, 'tcx> {
...
             }
-            (TyKind::Unit, rustc_ty::TyKind::Tuple(tys)) if tys.is_empty() => Ok(()),
             (TyKind::Array(ty, _), rustc_ty::TyKind::Array(rust_ty, _)) => self.zip_ty(ty, rust_ty),
             (TyKind::Slice(ty), rustc_ty::TyKind::Slice(rust_ty)) => self.zip_ty(ty, rust_ty),
+            (TyKind::Tuple(tys), rustc_ty::TyKind::Tuple(rust_tys))
+                if tys.len() == rust_tys.len() =>
+            {
+                for (ty, rust_ty) in tys.iter().zip(rust_tys) {
+                    self.zip_ty(ty, rust_ty)?;
+                }
```
nilehmann, line 196: Let's do a `try_for_each_exhaust` here to report all the errors

flux-desugar/src/desugar.rs
```diff
@@ -346,6 +345,13 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
                 fhir::Ty::Array(Box::new(ty), fhir::ArrayLen)
             }
             surface::TyKind::Slice(ty) => fhir::Ty::Slice(Box::new(self.desugar_ty(None, *ty)?)),
+            surface::TyKind::Tuple(tys) => {
+                fhir::Ty::Tuple(
+                    tys.into_iter()
+                        .map(|ty| self.desugar_ty(None, ty))
+                        .collect::<Result<_, _>>()?,
```
nilehmann, line 352: you can use `try_collect` from `itertools` (or `try_collect_exhaust` if we want to report all the errors)

nilehmann (commented):
@ranjitjhala some small comments

flux-desugar/src/annot_check.rs
ranjitjhala: ok, see if I did it right?

flux-desugar/src/desugar.rs
ranjitjhala: done! (but see above!)

flux-desugar/src/annot_check.rs
nilehmann: looks good 👍🏽

PR #225, opened 2022-11-11
New184

ranjitjhala:
replaces https://github.com/liquid-rust/flux/pull/223

### DIFF
--- a/flux-middle/src/fhir.rs
+++ b/flux-middle/src/fhir.rs
@@ -157,18 +157,11 @@ pub enum WeakKind {
     Arr,
 }
 
-impl WeakKind {
-    pub fn from_ref_kind(rk: RefKind) -> WeakKind {
+impl From<RefKind> for WeakKind {
+    fn from(rk: RefKind) -> WeakKind {
         match rk {
             RefKind::Shr => WeakKind::Shr,
             RefKind::Mut => WeakKind::Mut,
-        }
-    }
-    pub fn to_ref_kind(rk: WeakKind) -> RefKind {
-        match rk {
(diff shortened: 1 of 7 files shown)


flux-middle/src/fhir.rs
```diff
@@ -150,6 +150,29 @@ pub enum RefKind {
...
+            RefKind::Mut => WeakKind::Mut,
+        }
+    }
+    pub fn to_ref_kind(rk: WeakKind) -> RefKind {
+        match rk {
+            WeakKind::Mut => RefKind::Mut,
+            WeakKind::Shr => RefKind::Shr,
+            WeakKind::Arr => RefKind::Shr,
+        }
+    }
```
nilehmann, line 173: This is not being used.

```suggestion
```

flux-middle/src/fhir.rs
```diff
@@ -150,6 +150,29 @@ pub enum RefKind {
...
+    Arr,
+}
+
+impl WeakKind {
+    pub fn from_ref_kind(rk: RefKind) -> WeakKind {
+        match rk {
+            RefKind::Shr => WeakKind::Shr,
+            RefKind::Mut => WeakKind::Mut,
+        }
+    }
```
nilehmann, line 166: The idiomatic way is to `impl From<RefKind> for WeaKind`

nilehmann (approved):
Looks good now.

flux-middle/src/fhir.rs
ranjitjhala: aha, thanks for the tip!

PR #238, opened 2022-12-03
Allow syntactic abstraction of refinements via "Interpreted" Functions `dfn`

ranjitjhala:
I've keenly felt the absence of this feature for awhile -- it lets you lift common patterns into flux-level functions which are really macros that get inlined into the relevant places in the specifications. See 

* `flux-tests/tests/pos/surface/ealias00.rs`
* `flux-tests/tests/pos/surface/ealias01.rs`
* `flux-tests/tests/pos/surface/ealias02.rs`

(and `neg` versions)

Mostly self contained in the `expand.rs` so we can easily revert later.

**Note:** The one thing I cannot figure out is the bizarre `DefinitionCycle` error - which is thrown if you have cyclic / mutually recursive `dfn` . Currently, I get some strange `cannot find fluent bundle` run-time error, so just using the `panic!` till this gets sorted out later.


flux-middle/src/expand.rs
```diff
@@ -0,0 +1,387 @@
+//! Expanding the expression alias `defn` in [`flux_middle::fhir`]
```
nilehmann, line 1: I think `fhir` is not the correct place to do the expansion. I'd rather do it during (right after) [`conv`](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/conv.rs) as a normalization step for types in [`rty`](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/mod.rs#L228). Some reasons why I think this would be better.

1. I imagine this sort of partial evaluation will only become more complicated over time. Complicated type manipulation should go in `rty` to avoid duplication.
2. We already have infrastructure to do [`"folding"`](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/fold.rs) over types in `rty` so we can reuse that. The type [`Binders`](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/mod.rs#L66) can be used to define abstractions, e.g., [here](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/subst.rs#L132) we apply an abstract refinement (of type `Binders<Pred>`).
3. Stuff in `fhir` is meant to capture what the user wrote but in a desugared form. In particular, it was designed to maintain spans. Doing the expansion here breaks this.
4. Doing the normalization eagerly is just one option and it's not clear to me it's the best option. I imagine we can do the normalization lazily in later steps. Having the logic implemented in `rty` keeps our options more open.

More concretely, I think we can `conv` definitions all the way down to `rty`. Then we can `normalize` definitions. We could implement a `normalize` method generically on `TypeFoldable` which I think is really neat.

side note: The function [`replace_bound_vars`](https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-middle/src/rty/mod.rs#L228) may be more complicated than what we need for this use case because it deals with abstract refinements which are substituted differently. Maybe we can have a dedicated function for when we are only substitution expressions

nilehmann (changes requested):
@ranjitjhala I have one major comment

flux-desugar/src/desugar.rs
```diff
@@ -496,22 +534,26 @@ impl<'a, 'tcx> ExprCtxt<'a, 'tcx> {
...
+                        .emit_err(errors::InvalidUnrefinedParam::new(def_ident, func)));
+                }
             }
-            (None, Some(uif)) => Ok(FuncRes::Uif(uif)),
-            (None, None) => Err(self.sess.emit_err(errors::UnresolvedVar::new(func))),
         }
+        if let Some(uif) = self.map.uif(func.name) {
+            return Ok(FuncRes::Uif(uif));
+        }
+        Err(self.sess.emit_err(errors::UnresolvedVar::new(func)))
```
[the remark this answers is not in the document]
ranjitjhala: Its somewhat related (iirc) as there was some function/logic that I wanted to reuse, and the previous code which did this double case split made it hard to reuse. So IIRC I factored it out and extended it.

flux-middle/src/expand.rs
ranjitjhala: Good point! To make sure I follow, we should 

1. Still do the "wf" checking and such at the `fhir` level (with the rest of the wf), but
2. Propagate the `Defn` down to the `rty` level and then 
3. Do the actual `exp_` at the `rty` level,

yes?

ranjitjhala:
btw, this PR was somewhat motivated by wanting to do this example

https://github.com/liquid-rust/flux/pull/238/commits/43bfa891dccf727df6358ef21d7141ffd69aa143#diff-b41377276614e16abfd86f1948720ceef91e72bae75241b28898b0b49cad0e9a

it would be nice I think at some future point to not have to "duplicate" the definitions of `is_leap_year` and such... maybe with some macro tricks?

flux-middle/src/expand.rs
nilehmann: Yes, that sounds good.

ranjitjhala:
@nilehmann - I shifted all the stuff to rty level - very civilized & pleasant to use the TypeFolder API!

ranjitjhala:
Still have to keep the Span to fail for cyclic defns - in theory could check that in FHir but that seems like gross duplication…

flux-middle/src/global_env.rs
```diff
@@ -30,21 +30,33 @@ pub struct GlobalEnv<'genv, 'tcx> {
...
         let check_asserts = CONFIG.check_asserts;
 
+        let mut defns: FxHashMap<Symbol, rty::Defn> = FxHashMap::default();
+        for defn in map.defns() {
+            let defn = rty::conv::conv_defn(defn);
+            defns.insert(defn.name, defn);
+        }
+        let defns = match Defns::new(defns) {
+            Ok(defns) => defns,
+            Err(_) => panic!("cyclic defns"),
```
ranjitjhala, line 47: Add TODO here to fix the ftl glitch

flux-middle/src/rty/mod.rs
```diff
@@ -93,6 +100,17 @@ pub struct Qualifier {
     pub expr: Expr,
 }
 
+pub struct Defn {
+    pub name: Symbol,
+    // pub args: Vec<(Name, Sort)>,
```
ranjitjhala, line 105: DELETE

flux-middle/src/fhir.rs
```diff
@@ -56,14 +56,15 @@ pub struct Qualifier {
...
-    assumes: FxHashSet<LocalDefId>,
+    pub(crate) uifs: FxHashMap<Symbol, UifDef>,
+    pub(crate) defns: FxHashMap<Symbol, Defn>,
+    pub(crate) consts: FxHashMap<Symbol, ConstInfo>,
+    pub(crate) qualifiers: Vec<Qualifier>,
+    pub(crate) adts: FxHashMap<LocalDefId, AdtDef>,
+    pub(crate) structs: FxHashMap<LocalDefId, StructDef>,
+    pub(crate) enums: FxHashMap<LocalDefId, EnumDef>,
+    pub(crate) fns: FxHashMap<LocalDefId, FnSig>,
+    pub(crate) assumes: FxHashSet<LocalDefId>,
```
nilehmann, line 67: I think we don't need to expose these anymore

```suggestion
    uifs: FxHashMap<Symbol, UifDef>,
    defns: FxHashMap<Symbol, Defn>,
    consts: FxHashMap<Symbol, ConstInfo>,
    qualifiers: Vec<Qualifier>,
    adts: FxHashMap<LocalDefId, AdtDef>,
    structs: FxHashMap<LocalDefId, StructDef>,
    enums: FxHashMap<LocalDefId, EnumDef>,
    fns: FxHashMap<LocalDefId, FnSig>,
    assumes: FxHashSet<LocalDefId>,
```

flux-middle/src/fhir.rs
```diff
@@ -285,7 +287,7 @@ pub struct AdtDef {
     pub refined_by: RefinedBy,
     pub invariants: Vec<Expr>,
     pub opaque: bool,
-    sorts: Vec<Sort>,
+    pub(crate) sorts: Vec<Sort>,
```
nilehmann, line 290: Do we still need this visibility? This was purposely hidden because it's kind of like an optimization since the sort information is already present in `refined_by`

```suggestion
    sorts: Vec<Sort>,
```

flux-middle/src/rty/mod.rs
```diff
@@ -373,6 +391,54 @@ impl AdtDef {
...
+impl TypeFoldable for Invariant {
+    fn super_fold_with<F: TypeFolder>(&self, folder: &mut F) -> Self {
+        let pred = self.pred.fold_with(folder);
+        Invariant { pred }
+    }
+
+    fn super_visit_with<V: TypeVisitor>(&self, visitor: &mut V) {
+        self.pred.visit_with(visitor);
+    }
+}
```
nilehmann, line 440: These implementations should go in the `fold` module for consistency

flux-tests/tests/neg/error_messages/dfn_err.rs
```diff
@@ -0,0 +1,20 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
+#![feature(custom_inner_attributes)]
+#![flux::dfn(nat(x: int) -> bool { 0 <= x })]
```
nilehmann, line 4: I prefer `def` over `dfn`. Mentally, I parse `dfn` as an acronym. But no strong feelings about this one.

note: For a while, I've been feeling that it wasn't such a good idea to have separate attributes for everything. Maybe `flux::sig`,  `flux::field`, `flux::variant` and `flux::refined_by` work, but for these top-level definitions we could have a single attribute and parse different items, e.g.,

```rust
#[flux {
    fn nat(x: int) -> bool {
        x >= 0
    }
    
    fn uninterpreted(x: int) -> bool;
    
    qualif MyQ1(x: int, a: int)  {
        x == a + FORTY_TOO
    }
}]
```

nilehmann (approved):
@ranjitjhala A couple of suggestions, but otherwise looks good. 

re keeping spans in rty: In theory, we still have the span in the `fhir::Map`, so we could look it up for error reporting. This is perhaps cleaner, but more cumbersome. We do something similar for reporting errors on invariants https://github.com/liquid-rust/flux/blob/e03dd11fe7ee909a9cdc7b0c92a06b435eb87026/flux-typeck/src/invariants.rs#L21 No strong feelings about this one.

flux-middle/src/rty/mod.rs
ranjitjhala: done!

flux-tests/tests/neg/error_messages/dfn_err.rs
ranjitjhala: interesting thought! It does seem a _lot_ cleaner. 

btw, if you saw the `date.rs` test there is a bunch of "duplication" so I wonder if instead we should use a macro like Prusti so:

```rust
flux::def! { fn nat(x: i32) -> bool { 0 <= x } } 
```

Generates both the flux-level `def` and the rust-level `fn` ?

Something to ponder...

ranjitjhala: Ah, yes I think I picked `dfn` vs `ufn` but I just realized your point is you can just use `fn` and disambiguate based on presence of definition! Definitely better.

ranjitjhala:
I missed the bit about `re keeping spans in rty` -- by "this is cleaner" do you mean (a) "keeping the Span in rty` or (b) using the `Map` -- I totally forgot about the `Map` thing, so yes we can definitely use that, but didn't follow which one you thought is cleaner -- (I assume you meant using the `Map` to avoid duplicating the `Span`?)

Btw, the real puzzler for me is the "missing fluent bundle" thing, I could never figure that out so have left the cycle error as a panic for now :-( ...

nilehmann:
I think (b) is cleaner, but if it makes the code much more complicated having the `Span` in `rty` is not that big of a deal I guess.

Were you emitting the error with `FluxSession::emit` or with `tcx.sess().emit()`? this is one reason I can think of you could be getting a `missing fluent bundle`. You have to use the the emit in `FluxSession`

PR #242, opened 2022-12-09
No span in rty

ranjitjhala:
Cleanup from #238 -- in particular remove the `Span` from `rty::Defn` and use `Map` instead.

Sadly, I still can't get the fluent-bundle thing to work :-( 

To reproduce:

1. Toggle the `1 + 1 < 0` in `sorted_defns` and
2. Run `flux --crate-type=rlib flux-tests/tests/todo/dfn_cycle.rs`


ranjitjhala:
@nilehmann  you said to use the `FluxSession` yes? (It looks like that's what I'm using?)

nilehmann:
@ranjitjhala oh I know what it is. The entry in the `ftl` file doesn't have a `.label`. You could either add a `.label` there or remove the `#[label]` annotation in the error definition.

ranjitjhala:
Aha, ok thank you 🙏 !!!

nilehmann (approved):

PR #243, opened 2022-12-09
RVec Macro

ranjitjhala:
A small macro to create `rvec` literals (without the tedious `push`-ing)"

```rust
pub fn vec_push() -> usize {
    let v = rvec![0, 1];
    let r = v[0];
    let r = r + v[1];
    let r = r + v[2]; //~ ERROR precondition might not hold
    r
}
```

### DIFF
--- a/flux-tests/tests/lib/rvec.rs
+++ b/flux-tests/tests/lib/rvec.rs
@@ -1,14 +1,13 @@
 #![allow(dead_code)]
-#![allow(unused_mut)]
 
 #[macro_export]
 macro_rules! rvec {
-    () => { RVec::new() };
-    ($($e:expr),+) => {
-       { let mut res = RVec::new();
-         $( res.push($e); )*
-         res }
-    }
+    ($($e:expr),*$(,)?) => {{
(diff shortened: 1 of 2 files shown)


nilehmann (approved):

ranjitjhala:
I suppose one can have separate patterns for the empty sequence?

nilehmann:
I was mostly worried about the optional trailing comma, but I think we should also accept empty vectors for consistency. We can ignore the warning like so:

```
#[macro_export]
macro_rules! rvec {
    ($($e:expr),*$(,)?) => {{ 
        #[allow(unused_mut)]
        let mut res = Vec::new();
        $( res.push($e); )*
        res 
    }}
}
```

nilehmann:
Maybe splitting the patterns is a better idea because we should accept this `rvec![,]` (`vec!` doesn't accept it). So something like:

```rust
#[macro_export]
macro_rules! rvec {
    () => { RVec::new() };
    ($($e:expr),+$(,)?) => {{ 
        let mut res = RVec::new();
        $( res.push($e); )+
        res 
    }}
}
```

PR #244, opened 2022-12-09
Various fixes

nilehmann:
* Add syntax for `!=` in refinements
* Fix single case for `Aggregate` in `resolve_func`
* Check the correct use of abstract refinements in predicate abstractions
* Correctly convert body of abstract predicates as a `Pred`


ranjitjhala:
sweet! I have missed "not".

PR #247, opened 2022-12-10
Add rslice example

nilehmann:
An attempt at creating an API for "refined slices" that can be borrowed mutably multiple times at non-overlapping subslices. The generated constraints are not very inference friendly so the api can only be used in straightforward cases. I think this could be improved if we implement variance for abstract refinements as that would eliminate cycles. This also makes me wonder if one could possibly opt out of "constrained-based" inference for abstract refinement and unify them syntactically as normal refinement parameters.

I believe a similar API could be used for a memory buffer with maybe uninitialized memory, i.e., `RawVec`.

### DIFF
--- a/flux-tests/tests/lib/rslice.rs
+++ b/flux-tests/tests/lib/rslice.rs
@@ -1,0 +1,30 @@
+use std::marker::PhantomData;
+
+use super::RVec;
+
+#[flux::opaque]
+#[flux::refined_by(len: int, valid: (int, int) -> bool)]
+pub struct RSlice<'a, T> {
+    data: *mut T,
+    len: usize,
+    _marker: PhantomData<&'a mut [T]>,
+}
+
+impl<'a, T> RSlice<'a, T> {
+    #[flux::assume]
+    #[flux::sig(fn(vec: &mut RVec<T>[@n]) -> RSlice<T>[n, |i,j| true])]
+    pub fn from_vec(vec: &'a mut RVec<T>) -> RSlice<T> {
+        let inner = &mut vec.inner;
+        RSlice { data: inner.as_mut_ptr(), len: inner.len(), _marker: PhantomData }
+    }
+
+    #[flux::assume]
+    #[flux::sig(
+        fn(self: &strg RSlice<T>[@len, @valid], left: usize, right: usize) -> RSlice<T>[right - left + 1, |i,j| true]
+        requires left <= right && right < len && valid(left, right)
+        ensures self: RSlice<T>[len, |i, j| valid(i, j) && (right < i || j < left)]
+    )]
+    pub fn subslice(&mut self, left: usize, right: usize) -> RSlice<'a, T> {
+        unsafe { RSlice { data: self.data.add(left), len: right - left + 1, _marker: PhantomData } }
+    }
+}
--- a/flux-tests/tests/lib/rvec.rs
+++ b/flux-tests/tests/lib/rvec.rs
@@ -1,5 +1,7 @@
 #![allow(dead_code)]
 
+pub mod rslice;
+
 #[macro_export]
 macro_rules! rvec {
     () => { RVec::new() };
--- a/flux-tests/tests/neg/surface/rslice00.rs
+++ b/flux-tests/tests/neg/surface/rslice00.rs
@@ -1,0 +1,13 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
+
+#[path = "../../lib/rvec.rs"]
+mod rvec;
+use rvec::{rslice::RSlice, RVec};
+
+#[flux::sig(fn(&mut RVec<T>[10]))]
+fn test00<T>(vec: &mut RVec<T>) {
+    let mut s = RSlice::from_vec(vec);
+    let s1 = s.subslice(0, 3);
+    let s2 = s.subslice(2, 5); //~ ERROR precondition
+}
--- a/flux-tests/tests/pos/surface/rslice00.rs
+++ b/flux-tests/tests/pos/surface/rslice00.rs
@@ -1,0 +1,13 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
+
+#[path = "../../lib/rvec.rs"]
+mod rvec;
+use rvec::{rslice::RSlice, RVec};
+
+#[flux::sig(fn(&mut RVec<T>[10]))]
+fn test00<T>(vec: &mut RVec<T>) {
+    let mut s = RSlice::from_vec(vec);
+    let s1 = s.subslice(0, 3);
+    let s2 = s.subslice(4, 5);
+}


ranjitjhala:
Nice! I was a bit confused by `valid` then I realized its more like `available`-to-slice.

Do you have an example of what you mean by variance for abs-ref that could eliminate cycles, or where the "syntactic" unification would work? (I wonder if the latter is an efficiency thing or an expressiveness?)

nilehmann:
@ranjitjhala If you just add one more subslicing, then it cannot prove the subslices are non-overlapping

```rust
fn test00<T>(vec: &mut RVec<T>) {
    let mut s = RSlice::from_vec(vec);
    let s1 = s.subslice(0, 3);
    let s2 = s.subslice(4, 5);  //~ ERROR precondition might not hold
    let s3 = s.subslice(7, 8); 
}
```

It works if I add the qualifier `Q1(a0: int, a1: int) { 3 < a1 || a0 < 0 }`, which took me a while to find.

The generated constraint is 

```
(0 ≤ 3 ∧ 3 < 10 ∧ $k0(0, 3)) ~ Call(18:14: 18:30)
∀ a0: int, a1: int.
  true ⇒ $k0(a0, a1) ~ Call(18:14: 18:30)
  $k0(a0, a1) ⇒ true ~ Call(18:14: 18:30)
(4 ≤ 5 ∧ 5 < 10 ∧ $k1(4, 5)) ~ Call(19:14: 19:30)
∀ a0: int, a1: int.
  ($k0(a0, a1) ∧ 3 < a0 ∨ a1 < 0) ⇒ $k1(a0, a1) ~ Call(19:14: 19:30)
  $k1(a0, a1) ⇒ ($k0(a0, a1) ∧ 3 < a0 ∨ a1 < 0) ~ Call(19:14: 19:30)
(7 ≤ 8 ∧ 8 < 10 ∧ $k2(7, 8)) ~ Call(20:14: 20:30)
∀ a0: int, a1: int.
  ($k1(a0, a1) ∧ 5 < a0 ∨ a1 < 4) ⇒ $k2(a0, a1) ~ Call(20:14: 20:30)
  $k2(a0, a1) ⇒ ($k1(a0, a1) ∧ 5 < a0 ∨ a1 < 4) ~ Call(20:14: 20:30)
```

This is just expressing equivalences so I believe syntactic unification would avoid the kvars altogether. Having variance would also eliminate the cycles. Something that may be affecting this as well is that the constraint we throw at fixpoint is more complicated because we generate two kvars for each kvar. For example, we define `$k0(a0, a1) := $k0_0(a0) ∧ $k0_1(a1, a0)`.

```
(constraint
  (and
    (forall ((a0 int) (a0 = 0))
      (forall ((a1 int) (a1 = 3))
        (and
          (tag (0 <= 3) "0")
          (tag (3 < 10) "0")
          (and
            (tag ($k0 a0) "0")
            (tag ($k1 a1 a0) "0")
          )
        )
      )
    )
    (forall ((a2 int) (true))
      (forall ((a3 int) (true))
        (and
          (forall ((_ Unit) (true))
            (and
              (tag ($k0 a2) "0")
              (tag ($k1 a3 a2) "0")
            )
          )
          (forall ((_ Unit) (and ($k0 a2) ($k1 a3 a2)))
            (tag (true) "0")
          )
        )
      )
    )
    (forall ((a4 int) (a4 = 4))
      (forall ((a5 int) (a5 = 5))
        (and
          (tag (4 <= 5) "1")
          (tag (5 < 10) "1")
          (and
            (tag ($k2 a4) "1")
            (tag ($k3 a5 a4) "1")
          )
        )
      )
    )
    (forall ((a6 int) (true))
      (forall ((a7 int) (true))
        (and
          (forall ((_ Unit) (and (and ($k0 a6) ($k1 a7 a6)) ((3 < a6) || (a7 < 0))))
            (and
              (tag ($k2 a6) "1")
              (tag ($k3 a7 a6) "1")
            )
          )
          (forall ((_ Unit) (and ($k2 a6) ($k3 a7 a6)))
            (and
              (and
                (tag ($k0 a6) "1")
                (tag ($k1 a7 a6) "1")
              )
              (tag ((3 < a6) || (a7 < 0)) "1")
            )
          )
        )
      )
    )
    (forall ((a8 int) (a8 = 7))
      (forall ((a9 int) (a9 = 8))
        (and
          (tag (7 <= 8) "2")
          (tag (8 < 10) "2")
          (and
            (tag ($k4 a8) "2")
            (tag ($k5 a9 a8) "2")
          )
        )
      )
    )
    (forall ((a10 int) (true))
      (forall ((a11 int) (true))
        (and
          (forall ((_ Unit) (and (and ($k2 a10) ($k3 a11 a10)) ((5 < a10) || (a11 < 4))))
            (and
              (tag ($k4 a10) "2")
              (tag ($k5 a11 a10) "2")
            )
          )
          (forall ((_ Unit) (and ($k4 a10) ($k5 a11 a10)))
            (and
              (and
                (tag ($k2 a10) "2")
                (tag ($k3 a11 a10) "2")
              )
              (tag ((5 < a10) || (a11 < 4)) "2")
            )
          )
        )
      )
    )
  )
)
```

ranjitjhala:
Ah ok, the hassle is with cycles and qualifiers - let’s stare at the
constraints for this, it’s strange that it’s NOT straight line (Ie I
haven’t grokked why exactly a cycle shows up here…)

ranjitjhala:
Just to make sure I'm following. Consider this stripped down API

```rust
s0 = new(20); // create "init" slice of 0...20
s1 = s0.sub(i0, j0);
s2 = s1.sub(i1, j1);
s3 = s2.sub(i2, j2);
```

Now I would expect the constraint to look like

```rust
$k0(0, 20) /\
(forall i, j. $k0(i, j) /\ ( outside i0, j0 ) => $k1(i, j) ) /\ 
(forall i, j. $k1(i, j) /\ ( outside i1, j1 ) => $k2(i, j) ) /\ 
(forall i, j. $k2(i, j) /\ ( outside i2, j2 ) => $k3(i, j) )
``` 

but the CYCLES are showing up because we ALSO have subtyping 
as the type of `s0` is a subtype of `s1` (as `s0` is passed in as an 
argument to get the `s1`) so contra-variance gives us the extra 
"backwards edges" constraints of the form

```rust
$k1 => $k0 
$k2 => $k1 
```

etc.?

ranjitjhala:
[If so, something seems *wrong* about the above ... ]

ranjitjhala:
What is also interesting here is that those "back edges" are really spurious. Meaning, if you have constraints like

```
  ($k0(a0, a1) ∧ 3 < a0 ∨ a1 < 0) ⇒ $k1(a0, a1) 
  $k1(a0, a1) ⇒ ($k0(a0, a1) ∧ STUFF)
```

are "logically equivalent to"

```
  ($k0(a0, a1) ∧ 3 < a0 ∨ a1 < 0) ⇒ $k1(a0, a1) 
  $k1(a0, a1) ⇒ STUFF
```

as long as this is the ONLY occurrence of `$k1` on the RHS. 

Though I wonder if there is some way to exploit this fact in the spec itself to avoid creating these spurious cycles in the constraints...

ranjitjhala:
I'm not following what #248 and #249 are doing but I'm guessing that 
the `@valid` inference (synthesis) is done purely via subtyping at least 
in _this_ PR (i.e. prior to 248 and 249?)  And further, by

> opt out of "constrained-based" inference for abstract refinement 
> and unify them syntactically as normal refinement parameters

you mean somehow carry the "lambdas" for the abs-ref around as
expressions in flux itself? so returning to the example above, `typeck`
would "solve"

* `valid0` (aka $k0) to `|i, j| true` 
* `valid1` (aka $k1) to `|i, j| true /\ ( outside i1, j1)`
* `valid2` (aka $k2) to `|i, j| true /\ ( outside i1, j1) /\ (outside i2, j2)` 

etc.? 

It could work I think as simple syntactic substitutions suffice to 
compute the new lambdas -- as happens here. I wonder if it might 
even work in general without "real" cycles... It would also possibly
allow for some/many of the use cases for "refinement bounds"
in a more "lightweight" fashion e.g. taking union/intersection 
of abstract refinements ... Probably worth a shot!

nilehmann:
Let me try to elaborate on how the constraint is generated after https://github.com/liquid-rust/flux/pull/249. 

Consider the following simplified API

```rust
fn new() -> Slice[|i,j| true];

fn subslice(self: Slice[@valid], a: usize, b: usize{ valid(a, b) }) 
    -> (Slice[|i,j| valid(i,j) && (a < ia || j < b)], Slice[|i,j| true])
```

I'm making `subslice` take ownership of the slice and then return it to avoid dealing with locations and strong references.

```
new : fn() -> Slice[|i,j| true]
subslice: forall a: int, b: int, valid: (int, int) -> bool. fn(Slice[valid], usize[a], {usize[b] : valid(a, b) }) 
              -> (Slice[|i, j| valid(i, j) && (a < i || j < b)], Slice[|i, j| true])
```

Now consider this code.

```rust
let s1 = new();
let (s1, s2) = subslice(s1, 0, 1)
```

At the beginning `s1` has type `Slice[|i,j|true]`. When calling `subslice(s1, 0, 1)` we first generate fresh evars for regular parameters and fresh kvars for abstract refinements. After this the signature for `slice` looks like:

```
fn(Slice[|i,j| $k0(i, j)], usize[?e0], {usize[?e1] : valid(?e0, ?e1)) 
    -> (Slice[|i, j| $k0(i, j) && (a < i || j < b)], Slice[|i, j| true])
```

Then we check subtyping between arguments. This gives us the following subtyping relations

```
Slice[|i,j|true] <: Slice[|i,j| $k0(i,j)]
usize[0] <: usize[?e0]
usize[1] <: {usize[?e1] : $k0(?e0, ?e1) }
```

Since we assume all refinement parameters are invariants the first relation generates

```
forall i, j. true => $k0(i, j) ∧ forall i, j. $k0(i, j) => true
```

The other relations generate the expected constraints plus the solution `?e0 -> 0, ?e1 -> 1`, which we substitute in the return type and in the constraint before continuing (as opposed to kvars which stay in the constraint and the return type). You can see how iterating this will generate cyclic constraints.

**Option A:** Making the subtyping of `Slice` covariant (or contravariant whichever is the right one) will eliminate the cyclic constraints. I don't think is too hard to associate variance with particular parameters. The part I'm unsure about is how we check the annotated variance is actually sound.

**Option B:** We could treat the abstract refinement like a "normal" parameter and generate an evar for it instead of a kvar. So the signature would look like this:

```
fn(Slice[|?e2|], usize[?e0], {usize[?e1] : valid(?e0, ?e1)) 
    -> (Slice[|i, j| ?e2(i, j) && (a < i || j < b)], Slice[|i, j| true])
```

That will generate the following subtyping relation

```
Slice[|i,j| true] <: Slice[?e2]
```

From which we can extract the solution `?e2 -> |i,j| true`. When replacing the solution in the return type we will then get

```
(Slice[|i,j| true && (a < i || j < b)], slice[|i,j| true])
```

ranjitjhala:
Yup - this all makes sense, did you already implement the ***@***.***`
approach? I guess my main concern is

1. It's a bit pesky to have two user-facing mechanisms
2. I worry a bit that the "refinement instantiation" will get complex.

Still, IMO "option B" is an interesting idea worth pursuing -- unless it
gets very hairy for some reason,
I'm very curious to see where it goes, because the "p1 and p2 can appear
freely" may give us a _lot_ of
the benefits of bounded refinements but without the complexity...

(e.g. I'm wondering if it would have sufficed for the STORM API...)

PR #249, opened 2022-12-11
Implement parameter inference using evars

nilehmann:


flux-tests/tests/lib/rvec.rs
```diff
@@ -116,13 +117,15 @@ impl<T> Iterator for RVecIter<T> {
 impl<T> std::ops::Index<usize> for RVec<T> {
     type Output = T;
 
+    #[flux::assume]
```
ranjitjhala, line 120: Why did these have to become `assume`?

nilehmann: There are many tests that import this file so this code is checked multiple times. This makes it very annoying if you ever break this code because many tests will spuriously fail.

PR #251, opened 2022-12-14
Evar inference for abstract refinements

nilehmann:
Implements the strategy discussed [here](https://github.com/liquid-rust/flux/pull/247#issuecomment-1345780147), where abstract refinements can be inferred in two different ways: via unification or via fixpoint inference. The implementation could be improved but this already shows the benefits of unification-based inference. For example, the constraint generated for [rslice00.rs](https://github.com/liquid-rust/flux/blob/5ccf67504f606c21ab77c5b91224bcd8ab0dbaf7/flux-tests/tests/pos/surface/rslice00.rs) looks like

```
(0 ≤ 3 ∧ 3 < 10) ~ Call(11:14: 11:30)
∀ a0: int, a1: int. true
(4 ≤ 5 ∧ 5 < 10 ∧ 3 < 4 ∨ 5 < 0) ~ Call(12:14: 12:30)
∀ a0: int, a1: int. true
```

It can't get any better than that. 

Having both inference modes also allows us to support simultaneously the example here https://github.com/liquid-rust/flux/blob/5ccf67504f606c21ab77c5b91224bcd8ab0dbaf7/flux-tests/tests/pos/abstract_refinements/test00.rs.

note: this PR does not remove the syntactic restrictions on abstract refinement marked for unification-based inference, which still must appear in horn positions. That would require further changes but it must be possible.


ranjitjhala:
🤌 wonderful! what is POC?

nilehmann:
proof of concept

PR #252, opened 2022-12-17
More support for Array and Slice

### LINKED ISSUE #184: Support slice and array indexing

ranjitjhala:
Fixes #184 maybe?

The main big change is factoring `Array` and `Slice` as `BaseTy` to allow refining with size.

### DIFF
--- a/flux-desugar/src/desugar.rs
+++ b/flux-desugar/src/desugar.rs
@@ -1079,7 +1079,7 @@ impl Binder {
         match bty {
             surface::BaseTy::Path(path) => Binder::from_res(name_gen, map, path.ident),
             surface::BaseTy::Array(_, _) | surface::BaseTy::Slice(_) => {
-                Binder::from_res(name_gen, map, Res::Uint(surface::UintTy::Usize))
+                Binder::Single(name_gen.fresh(), fhir::Sort::Int)
             }
         }
     }
--- a/flux-desugar/src/desugar.rs
+++ b/flux-desugar/src/desugar.rs
@@ -1085,7 +1085,9 @@ impl Binder {
     ) -> Binder {
         match base {
             surface::BaseTy::Path(path) => Binder::from_res(name_gen, map, path.ident),
-            surface::BaseTy::Array(_, _) | surface::BaseTy::Slice(_) => Binder::Unrefined,
+            surface::BaseTy::Array(_, _) | surface::BaseTy::Slice(_) => {
+                Binder::from_res(name_gen, map, Res::Uint(surface::UintTy::Usize))
+            }
         }
     }
 
--- a/flux-desugar/src/desugar.rs
+++ b/flux-desugar/src/desugar.rs
@@ -293,9 +293,9 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
         ty: surface::Ty<Res>,
     ) -> Result<fhir::Ty, ErrorGuaranteed> {
         match ty.kind {
-            surface::TyKind::Base(base) => self.desugar_base_bind(bind, base),
-            surface::TyKind::Indexed { base, indices } => {
-                match self.desugar_base(base)? {
+            surface::TyKind::Base(bty) => self.desugar_bty_bind(bind, bty),
+            surface::TyKind::Indexed { bty, indices } => {
+                match self.desugar_bty(bty)? {
                     BtyOrTy::Bty(bty) => Ok(fhir::Ty::Indexed(bty, self.desugar_indices(indices)?)),
                     BtyOrTy::Ty(_) => {
                         Err(self.sess.emit_err(errors::ParamCountMismatch::new(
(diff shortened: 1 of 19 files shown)


flux-desugar/src/desugar.rs
```diff
@@ -1058,6 +1078,19 @@ impl Binder {
...
 
+    fn from_base(
+        name_gen: &IndexGen<fhir::Name>,
+        map: &fhir::Map,
+        base: &surface::BaseTy<Res>,
+    ) -> Binder {
+        match base {
+            surface::BaseTy::Path(path) => Binder::from_res(name_gen, map, path.ident),
+            surface::BaseTy::Array(_, _) | surface::BaseTy::Slice(_) => {
+                Binder::from_res(name_gen, map, Res::Uint(surface::UintTy::Usize))
```
nilehmann, line 1089: I would inline this call as
```suggestion
                Binder::Single(name_gen.fresh(), fhir::Sort::Int)
```

flux-middle/src/rty/mod.rs
```diff
@@ -575,6 +567,16 @@ impl BaseTy {
         BaseTy::Adt(adt_def, substs.into())
     }
 
+    pub fn array(ty: Ty, c: Const) -> BaseTy {
+        BaseTy::Array(ty, c)
+        // Ty::indexed(BaseTy::Array(ty, c), RefineArgs::empty())
```
nilehmann, line 572:
```suggestion
```

flux-middle/src/rty/mod.rs
```diff
@@ -575,6 +567,16 @@ impl BaseTy {
...
+    }
+
+    pub fn slice(ty: Ty) -> BaseTy {
+        BaseTy::Slice(ty)
+        //Ty::indexed(BaseTy::Slice(ty), RefineArgs::empty())
```
nilehmann, line 577:
```suggestion
```

flux-desugar/src/desugar.rs
```diff
@@ -870,6 +878,13 @@ impl Binders {
         Ok(())
     }
 
+    // fn base_res(base: &surface::BaseTy<Res>) -> Res {
+    //     match base {
+    //         surface::BaseTy::Path(path) => path.ident,
+    //         _ => panic!("FIXME"),
+    //     }
+    // }
+
```
nilehmann, line 887:
```suggestion
```

flux-typeck/src/checker.rs
```diff
@@ -669,14 +682,36 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+        env: &mut TypeEnv,
+        src_info: SourceInfo,
+        place: &Place,
+    ) -> Result<Ty, CheckerError> {
+        let gen = &mut self.phase.constr_gen(self.genv, rcx, Tag::Other);
+        let ty = env
+            .lookup_place(rcx, gen, place)
+            .map_err(|err| CheckerError::from(err).with_src_info(src_info))?;
+        match ty.kind() {
+            TyKind::Indexed(BaseTy::Array(_, _), ixs) | TyKind::Indexed(BaseTy::Slice(_), ixs) => {
```
nilehmann, line 705: We typically use `idxs`

flux-typeck/src/checker.rs
```diff
@@ -669,14 +682,36 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+                Ok(Ty::indexed(BaseTy::Uint(UintTy::Usize), ixs.clone()))
+            }
+            TyKind::Exists(BaseTy::Array(_, _), ixs) | TyKind::Exists(BaseTy::Slice(_), ixs) => {
+                Ok(Ty::exists(BaseTy::Uint(UintTy::Usize), ixs.clone()))
+            }
```
nilehmann, line 710: This case is not necessary. Existentials should always be unpacked before going into the `TypeEnv`, if they are not, then there’s a bug somewhere else.

```suggestion
```

nilehmann (approved):
My initial reaction was that indexing arrays by the length would cause inconsistencies as you could write an array where the indexed length is different than the const length

```rust
#[flux::sig(fn([i32; _][2]))]
fn myarr(a: [i32; 1]) {
    a[1]
}
```
But if array literal are the only way to construct arrays then I think this should be ok and it just declaring an inhabited type. However, not connecting both lengths makes code cumbersome. For example, it makes it impossible to use the result of a function like https://doc.rust-lang.org/std/primitive.u32.html#method.to_be_bytes without explicitly checking the length, even though you already know the length is 4. Ideally, arrays shouldn’t be indexed but rather we the const length should be automatically _reflected_ into the logic when needed, i.e., the typing rule for `Len` should be something like

```
|- a: [T; n]         reflect(n) = a
———————————————————————————--———-——-
          |- Len(a): usize[a]
```
I’m ok with leaving the implementation like this for now as doing this reflection requires further design.

flux-desugar/src/desugar.rs
ranjitjhala: done!

flux-desugar/src/desugar.rs
ranjitjhala: done!

flux-typeck/src/checker.rs
ranjitjhala: Yes, that was my recollection too -- let me remove!

ranjitjhala:
Merging this in, let me make a separate issue for the `Array` thing.

PR #255, opened 2022-12-19
Array

### LINKED ISSUE #254: Add support for rust Array length `Const`
My initial reaction was that indexing arrays by the length would cause inconsistencies as you could write an array where the indexed length is different than the const length

```rust
#[flux::sig(fn([i32; _][2]))]
fn myarr(a: [i32; 1]) {
    a[1]
}
```
But if array literal are the only way to construct arrays then I think this should be ok and it just declaring an inhabited type. However, not connecting both lengths makes code cumbersome. For example, it makes it impossible to use the result of a function like https://doc.rust-lang.org/std/primitive.u32.html#method.to_be_bytes without explicitly checking the length, even though you already know the length is 4. Ideally, arrays shouldn’t be indexed but rather we the const length should be automatically _reflected_ into the logic when needed, i.e., the typing rule for `Len` should be something like

```
|- a: [T; n]         reflect(n) = a
———————————————————————————--———-——-
          |- Len(a): usize[a]
```
I’m ok with leaving the implementation like this for now as doing this reflection requires further design.

_Originally posted by @nilehmann in https://github.com/liquid-rust/flux/pull/252#pullrequestreview-1221901911_

ranjitjhala:
fixes #254 (maybe?)


flux-syntax/src/surface_grammar.lalrpop
```diff
@@ -147,6 +147,15 @@ ArgTyKind: surface::TyKind = {
...
 
     "{" <ty:Ty> ":" <pred:Level1> "}" => surface::TyKind::Constr(pred, Box::new(ty)),
+    "[" <ty:Ty> ";" <lo:@L> <lit:Lit> <hi:@R> "]" =>? {
+        let span = mk_span(lo, hi);
+        if let surface::LitKind::Integer = lit.kind {
+            if let Ok(val) = lit.symbol.as_str().parse::<usize>() {
+                return Ok(surface::TyKind::Array(Box::new(ty), surface::ArrayLen { val, span }))
+            }
+        }
+        return Err(ParseError::User { error: UserParseError::UnexpectedToken(lo, hi) })
```
nilehmann, line 157: Maybe we can use a let-else chain

```suggestion
        if let surface::LitKind::Integer = lit.kind && let Ok(val) = lit.symbol.as_str().parse::<usize>() {
            Ok(surface::TyKind::Array(Box::new(ty), surface::ArrayLen { val, span }))
        } else {
            Err(ParseError::User { error: UserParseError::UnexpectedToken(lo, hi) })
        }
```

flux-tests/tests/pos/surface/array01.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub fn read_u16(bytes: [u8; 2]) -> u16 {
+    u16::from_le_bytes(bytes)
+}
+
+// #[flux::assume]
+// fn to_bytes(n: u32) -> [u8; 4] {
+//     n.to_be_bytes()
+// }
+
```
nilehmann, line 12:
```suggestion
```

flux-middle/src/rustc/lowering.rs
```diff
@@ -622,7 +622,13 @@ pub fn lower_ty<'tcx>(tcx: TyCtxt<'tcx>, ty: rustc_ty::Ty<'tcx>) -> Result<Ty, U
             let tys = List::from_vec(tys.iter().map(|ty| lower_ty(tcx, ty)).try_collect()?);
             Ok(Ty::mk_tuple(tys))
         }
-        rustc_ty::Array(ty, _) => Ok(Ty::mk_array(lower_ty(tcx, *ty)?, Const)),
+        rustc_ty::Array(ty, len) => {
+            let param_env = ParamEnv::empty().with_reveal_all_normalized(tcx);
+            let val = len.try_eval_usize(tcx, param_env).unwrap_or_else(|| {
```
nilehmann, line 627: Ok, so this will fail if the length is a const generic

```rust
fn foo<const n: usize>(x: [i32; n]) { }
```
I think in general, instead of evaluating the length we should normalize the type with the appropriate `ParamEnv`. I'm not entirely sure, but I believe normalizing the type will evaluate the expression when possible. The correct `ParamEnv` will be either a generic one when checking the body of a function or an instantiated one when checking a function call. This is also related to supporting type projections. We can discuss this offline.

flux-typeck/src/checker.rs
```diff
@@ -701,12 +701,18 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
             .lookup_place(rcx, gen, place)
             .map_err(|err| CheckerError::from(err).with_src_info(src_info))?;
-        match ty.kind() {
-            TyKind::Indexed(BaseTy::Array(_, _), ixs) | TyKind::Indexed(BaseTy::Slice(_), ixs) => {
-                Ok(Ty::indexed(BaseTy::Uint(UintTy::Usize), ixs.clone()))
+
+        let idxs = match ty.kind() {
+            TyKind::Array(_, len) => {
+                let c =
+                    flux_fixpoint::Constant::Int(flux_fixpoint::Sign::Positive, len.val as u128);
```
nilehmann, line 708:
```suggestion
                    rty::Constant::from(len.val);
```

nilehmann (approved):
Ok, maybe I overreacted and this should work. The only thing I find a bit strange is that we only support the normalized version while the rust signature does not need to be normalized, i.e., you have to write

```rust 
use std::mem::size_of;
#[flux::sig(fn([i32; 4]))]
fn test01(x: [i32; 2 + 2]) { }

#[flux::sig(fn([i32; 4]))]
fn test02(x: [i32; size_of::<u32>()]) { }
```
instead of
```rust 
use std::mem::size_of;

#[flux::sig(fn([i32; 2 + 2]))]
fn test01(x: [i32; 2 + 2]) { }

#[flux::sig(fn([i32; size_of::<u32>()]))]
fn test02(x: [i32; size_of::<u32>()]) { }
```

I don't know if we ever want to allow the non-normalized version, but we can ponder that later. In any case, I would add pos and neg tests for the above.

flux-middle/src/rustc/lowering.rs
ranjitjhala: Yes this is what I was referring to as the `ParamEnv` hack ...

ranjitjhala: (So accept this for now and defer to later?) Let me add a TODO and link to this.

nilehmann: Yes, that sounds good, let's merge it as is for now.

PR #257, opened 2022-12-20
Macro for `from_elem_n`

ranjitjhala:
So we can write stuff like `rvec![ 100; 4 ]`


ranjitjhala:
Hopefully nothing untoward here!

PR #262, opened 2022-12-21
Naive join for tuples (?)

ranjitjhala:
I was getting an odd `panic!` in the `flux-wave` code that seems to come from here 

https://github.com/liquid-rust/examples/blob/main/wave/flux/src/iov.rs#L9-L11

I'm not 100% sure as the `SourcePos` is from inside the `unwrap_result` macro but changing it to use a custom `Pair` type made the error go away. 

So I implemented this "naive" fix which seems to pass the tests and make this example work, but I don't know if there are some more subtle invariants lurking that I'm unaware of?

LMK!

### DIFF
--- a/flux-typeck/src/type_env.rs
+++ b/flux-typeck/src/type_env.rs
@@ -609,9 +609,9 @@ impl TypeEnvInfer {
                 Ty::param(*param_ty1)
             }
             (TyKind::Tuple(tys1), TyKind::Tuple(tys2)) => {
-                let tys: Vec<Ty> = iter::zip(tys1, tys2)
+                let tys = iter::zip(tys1, tys2)
                     .map(|(ty1, ty2)| self.join_ty(ty1, ty2, src_info))
-                    .collect();
+                    .collect_vec();
                 Ty::tuple(tys)
             }
             _ => unreachable!("`{ty1:?}` -- `{ty2:?}`"),
--- a/flux-typeck/src/type_env.rs
+++ b/flux-typeck/src/type_env.rs
@@ -609,11 +609,10 @@ impl TypeEnvInfer {
                 Ty::param(*param_ty1)
             }
             (TyKind::Tuple(tys1), TyKind::Tuple(tys2)) => {
-                assert!(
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Oddly, attempts to reproduce as a minimal example like the code below seem to work just fine... 🤔 

```rust
#![feature(register_tool)]
#![register_tool(flux)]

fn baz() -> Result<(u32, u32), ()> {
    Err(())
}

pub fn baz_call() -> Result<u32, ()> {
    let p = baz();
    unwrap_result!(p);
    let (x, y) = p;
    Ok(x + y)
}

#[macro_export]
macro_rules! unwrap_result {
    ($p:ident) => {
        let $p = match $p {
            Ok(oc) => oc,
            Err(e) => {
                return Err(e);
            }
        };
    };
}
```

flux-typeck/src/type_env.rs
```diff
@@ -609,11 +609,10 @@ impl TypeEnvInfer {
...
             }
             (TyKind::Tuple(tys1), TyKind::Tuple(tys2)) => {
-                assert!(
-                    tys1.is_empty() && tys2.is_empty(),
-                    "join of non-empty tuples is not supported yet {src_info:?}"
-                );
-                Ty::tuple(vec![])
+                let tys: Vec<Ty> = iter::zip(tys1, tys2)
+                    .map(|(ty1, ty2)| self.join_ty(ty1, ty2, src_info))
+                    .collect();
```
nilehmann, line 614:
```suggestion
                let tys = iter::zip(tys1, tys2)
                    .map(|(ty1, ty2)| self.join_ty(ty1, ty2, src_info))
                    .collect_vec();
```

nilehmann (approved):
You need a join point to trigger that code. You could add the following as a test

```rust
#[flux::sig(fn(bool) -> i32{v: v >= 0})]
fn test00(b: bool) -> i32 {
    let p = if b { (0, 0) } else { (0, 1) };
    p.0 + p.1
}
```

PR #264, opened 2022-12-22
Support (some) `Projection` by normalizing them away

### LINKED ISSUE #260: Add support for `?` in calls
Currently the following code fails with

```rust
#![feature(register_tool)]
#![register_tool(flux)]

#[flux::sig(fn() -> Result<i32[0], ()>)]
fn baz() -> Result<i32, ()> {
    Ok(0)
}

#[flux::sig(fn() -> Result<i32[1], ()>)]
pub fn baz_call() -> Result<i32, ()> {
    let r = baz()?;
    Ok(r + 1)
}
```

which fails with

```
error[FLUX]: unsupported type in function call
    --> flux-tests/tests/pos/surface/result00.rs:11:13
     |
11   |     let r = baz()?;
     |             ^^^^^^
     |
note: function defined here
    --> /Users/rjhala/.rustup/toolchains/nightly-2022-11-07-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:2091:5
     |
2091 |     fn branch(self) -> ControlFlow<Self::Residual, Self::Output> {
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = note: unsupported type `<std::result::Result<T, E> as std::ops::Try>::Residual`

error: aborting due to previous error
```

ranjitjhala:
fixes #260 

The main trick is to build on the support for `instance` (c.f. #115) and when given a `def_id` with `Projection` in it, use the `rustc` API to `normalize` away the projections prior to `lower_fn_sig` etc.

### DIFF
--- a/flux-middle/src/rustc/lowering.rs
+++ b/flux-middle/src/rustc/lowering.rs
@@ -5,6 +5,7 @@ use itertools::Itertools;
 use rustc_const_eval::interpret::ConstValue;
 use rustc_errors::ErrorGuaranteed;
 use rustc_hir::def_id::DefId;
+use rustc_infer::infer::TyCtxtInferExt;
 use rustc_middle::{
     mir as rustc_mir,
     ty::{
@@ -13,6 +14,7 @@ use rustc_middle::{
         ParamEnv, TyCtxt, TypeVisitable,
     },
 };
+use rustc_trait_selection::traits::query::normalize::AtExt;
 
 use super::{
     mir::{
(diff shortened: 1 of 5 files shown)


flux-middle/src/rustc/lowering.rs
```diff
@@ -558,6 +558,20 @@ pub fn lower_variant_def(
...
+        match tcx.try_normalize_erasing_regions(param_env, fn_sig) {
+            Ok(fn_sig) => fn_sig,
+            Err(_) => {
+                return Err(errors::UnsupportedFnSig {
+                    span,
+                    reason: "Sorry, projections are not yet supported!".to_string(),
+                })
+            }
+        }
+    };
```
nilehmann, line 574: We are not doing anything interesting with regions yet, but we want to, so we shouldn't be erasing them. For example, the following fails right now

```rust
trait Trait {
    type A;
}

impl Trait for i32 {
    type A = i32;
}

fn foo(x: &i32) -> <i32 as Trait>::A {
    *x
}
```

I don't understand exactly what's going on with the obligations and InferCtxt but the following seems to work without erasing the regions

```rust
pub fn lower_fn_sig_of(tcx: TyCtxt, def_id: DefId) -> Result<PolyFnSig, errors::UnsupportedFnSig> {
    let fn_sig = tcx.fn_sig(def_id);
    let span = tcx.def_span(def_id);
    let param_env = tcx.param_env(def_id);
    match tcx
        .infer_ctxt()
        .build()
        .at(&ObligationCause::dummy(), param_env)
        .normalize(fn_sig)
    {
        Ok(fn_sig) => {
            lower_fn_sig(tcx, fn_sig.value)
                .map_err(|err| errors::UnsupportedFnSig { span, reason: err.reason })
        }
        Err(_) => {
            return Err(errors::UnsupportedFnSig {
                span,
                reason: "Sorry, projections are not yet supported!".to_string(),
            })
        }
    }
}
```

nilehmann (changes requested):

ranjitjhala:
Nice so the advantage here is the regions are NOT erased?

nilehmann:
Yeah, it seems that calling normalize on `At` doesnt erase regions, but you need and `InferCtxt` and an `ObligationCause` to get ab `At|`, which I don't quite understand.

ranjitjhala:
Interesting, these were alluded to in the replies here


https://internals.rust-lang.org/t/accessing-the-defid-of-a-trait-implementation/

Still lets just use `normalize` as it seems like a strict and local
improvement over the status quo!

ranjitjhala:
Hmm I can't get this to compile, something about adding 

   `rustc_infer::infer::TyCtxtInferExt`

to the dependencies?

ranjitjhala:
NVM figured it out, have applied your patch.

nilehmann (approved):

PR #265, opened 2022-12-23
Drop extra `fn` from `ufn` syntax

ranjitjhala:
... to make it more symmetric with the `dfn` syntax.

A more aggressive/ uniform thing would be to write

1. `flux::fn(name(x1:t1...) -> t)` for **uninterpreted** functions
2. `flux::fn(name(x1:t1...) -> t { e }` for **defined** (inlined) functions


nilehmann (changes requested):
I want to accept the changes, but you have mixed changes from the other PR

ranjitjhala:
Whoops -- looks like I started this off the wrong branch, my bad. Anyways, have nuked it and replaced with the unified `def` version.

PR #267, opened 2022-12-26
Check Target

ranjitjhala:
Adds an env-variable `LR_CHECK_TARGET=str` which focuses checking on `DefId` that contain `str` as a substring (so by default, `""` which checks everything.)

Rationale below (**11.56s vs 1.68s**)

```
rjhala@pozole ~/r/flux-wave (main) [101]> time RUSTFLAGS=-Awarnings cargo-flux check
   Compiling libc v0.2.138
   Compiling paste v1.0.11
    Checking flux-wave v0.1.0 (/Users/rjhala/research/flux-wave)
error[FLUX]: precondition might not hold
   --> src/os/mod.rs:550:13
    |
550 |     let r = os_futimens(os_fd, specs);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `flux-wave` due to previous error

________________________________________________________
Executed in   11.56 secs    fish           external
   usr time   13.76 secs    5.64 millis   13.76 secs
   sys time    0.90 secs    2.16 millis    0.90 secs
```

vs. with a specific target

```
$ time RUSTFLAGS=-Awarnings LR_CHECK_TARGET=futimens cargo-flux check
   Compiling libc v0.2.138
   Compiling paste v1.0.11
    Checking flux-wave v0.1.0 (/Users/rjhala/research/flux-wave)
error[FLUX]: precondition might not hold
   --> src/os/mod.rs:550:13
    |
550 |     let r = os_futimens(os_fd, specs);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `flux-wave` due to previous error

________________________________________________________
Executed in    1.68 secs    fish           external
   usr time    4.20 secs  195.00 micros    4.20 secs
   sys time    0.27 secs  806.00 micros    0.27 secs
```

### DIFF
--- a/flux-common/src/config.rs
+++ b/flux-common/src/config.rs
@@ -42,7 +42,7 @@ pub struct Config {
     pub check_asserts: AssertBehavior,
     pub dump_mir: bool,
     pub pointer_width: u64,
-    pub unsound_incremental: bool,
+    pub check_target: String,
 }
 
 pub static CONFIG: LazyLock<Config> = LazyLock::new(|| {
@@ -55,7 +55,7 @@ pub static CONFIG: LazyLock<Config> = LazyLock::new(|| {
             .set_default("check_asserts", "assume")?
             .set_default("check_asserts", "assume")?
             .set_default("pointer_width", 64)?
-            .set_default("unsound_incremental", false)?
+            .set_default("check_target", "")?
             .add_source(Environment::with_prefix("LR").ignore_empty(true))
             .build()?
             .try_deserialize()
--- a/flux-common/src/config.rs
+++ b/flux-common/src/config.rs
@@ -42,7 +42,7 @@ pub struct Config {
     pub check_asserts: AssertBehavior,
     pub dump_mir: bool,
     pub pointer_width: u64,
-    pub check_target: String,
+    pub check_def: String,
 }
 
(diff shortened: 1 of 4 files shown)


flux-common/src/config.rs
```diff
@@ -54,6 +55,7 @@ pub static CONFIG: LazyLock<Config> = LazyLock::new(|| {
             .set_default("check_asserts", "assume")?
             .set_default("check_asserts", "assume")?
             .set_default("pointer_width", 64)?
+            .set_default("check_target", "")?
```
nilehmann, line 58: I don't like the word target because it can be confused with the target architecture/os (that's what I thought when first reading the title). Maybe something with `filter` or `def`, `LR_CHECK_DEF`, `LR_DEF_FILTER`

ranjitjhala: Yes `LR_CHECK_DEF` is good!

ranjitjhala:
I also tweaked this to use the `DefPath` instead of the raw debug output.

https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.def_path_str

nilehmann (approved):

PR #270, opened 2022-12-26
More `Span` to decipher panics

ranjitjhala:


flux-typeck/src/constraint_gen.rs
```diff
@@ -29,6 +29,12 @@ pub struct ConstrGen<'a, 'tcx> {
     tag: Tag,
 }
 
+impl ConstrGen<'_, '_> {
+    pub fn span(&self) -> Option<Span> {
+        self.tag.span()
+    }
+}
+
```
nilehmann, line 37: I'd put this in the main `impl block`

nilehmann (approved):

flux-typeck/src/constraint_gen.rs
ranjitjhala: Ah good point -- somehow I missed it the first time round!

PR #278, opened 2022-12-30
Add `Sort::Adt` to `fhir` and defer index-flattening to `conv`

ranjitjhala:
This PR is a real mess. There's probably / hopefully some way to clean up the abstractions but right now its a bit gross to support the cases where:

* Adt with a single index used without fields (e.g. `RVec` refined by `len:int`)
* Adt with multiple-indices used as packed `Pair[@p]` or flattened `Pair[x, y]`

Anyways, lmk your thoughts @nilehmann!

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,4 +166,6 @@ to see what is happening.
 ## TODO: dot-in-def
 
 - [] test
+- `build_fhir_map` is where the action is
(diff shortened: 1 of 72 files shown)


nilehmann:
superseded by #296

PR #279, opened 2022-12-31
Refactor conv

nilehmann:
* Decouple the environment for bound vars (de Bruijn) and free vars (named), by defining an `ExprConvCtxt` trait
* Avoid threading the number of binders by explicitly tracking _layers_ in the bound variable environment


nilehmann:
I hope this isn't too annoying @ranjitjhala, but I think these changes simplify the code quite a bit and they make it easier to implement the changes in https://github.com/liquid-rust/flux/pull/278

ranjitjhala:
No worries I think I will probably circle back to take an entirely fresh
crack at #278

PR #288, opened 2023-01-03
Implement user-definable opaque sorts

nilehmann:
* Declare opaque sorts using `#![flux::sort(opaque SortName)]`
* Use opaque sort to implement a version of [`GhostCell`](https://docs.rs/ghost-cell/0.2.3/ghost_cell/) that doesn't require rank-2 lifetime polymorphism.


flux-fixpoint/src/lib.rs
```diff
@@ -109,6 +111,10 @@ impl<Tag: fmt::Display> fmt::Display for Task<Tag> {
         writeln!(f, "(data Pair 2 = [| Pair {{ fst: @(0), snd: @(1) }} ])")?;
         writeln!(f, "(data Unit 0 = [| Unit {{ }}])")?;
 
+        for sort in &self.sorts {
+            writeln!(f, "(data {sort} 0 = [\n| {sort}0 {{ }} \n| {sort}1 {{ }}])")?;
```
nilehmann, line 115: @ranjitjhala I would like this to be a `declare-sort` in z3, but I can't find a way to do it through `fixpoint`. More generally, I'm looking for a way to generate types with a unique id such that values of different types, but with the same id could be linked together. This is to encode the `GhostCell` API, but I also believe it will be useful for many other examples. Any thoughts?

PR #289, opened 2023-01-03
More Spans at Panic

ranjitjhala:


nilehmann (approved):
Maybe we could use a [thread_local](https://doc.rust-lang.org/std/thread/struct.LocalKey.html) to pass down the span without threading it.

ranjitjhala:
oh good idea, maybe another good "entry" issue?

PR #293, opened 2023-01-03
Allow UIF to appear in `def` 

ranjitjhala:


ranjitjhala:
why does it say 10 commits and not 2 or 3?

nilehmann:
maybe you didn't branch from main? In any case since we are squashing the commits it doesn't really matter

nilehmann (approved):

PR #300, opened 2023-01-06
Add simple profiling

cole-k:
# TL;DR

Set `LR_DUMP_TIMINGS=true` when you run `flux` to have flux write timing diagnostics to `./log/timings`.

Right now this is _extremely_ simple, it just provides some details for the spans under `flux_typeck` and `flux_driver`.

# Sample output

Below is a sample output for an invocation of `cargo-flux check` that took 19 seconds. The missing 2 seconds approximately accounts for the time it takes for `cargo check` to run.

Note that `check_crate` contains everything running under `check_top`, which is why the sum of the spans is greater than 19 seconds.

```
check_top
  Checker::infer
    num events:   205
    min non-zero: 0.52ms
    1st quartile: 0.52ms
    2nd quartile: 1.05ms
    3rd quartile: 2.62ms
    max:          24.12ms
    total time:   229.64ms
  Checker::check
    num events:   205
    min non-zero: 0.52ms
    1st quartile: 0.52ms
    2nd quartile: 1.05ms
    3rd quartile: 5.24ms
    max:          159.91ms
    total time:   2028.47ms
  FixpointCtx::check
    num events:   205
    min non-zero: 22.02ms
    1st quartile: 26.21ms
    2nd quartile: 28.31ms
    3rd quartile: 40.37ms
    max:          1867.51ms
    total time:   9106.36ms
total time: 11364.47ms

check_crate
  Callbacks::check_wf
    num events:   1
    min non-zero: 18.35ms
    1st quartile: 18.87ms
    2nd quartile: 18.87ms
    3rd quartile: 18.87ms
    max:          18.87ms
    total time:   18.87ms
  Callbacks::check_crate
    num events:   1
    min non-zero: 16986.93ms
    1st quartile: 16995.32ms
    2nd quartile: 16995.32ms
    3rd quartile: 16995.32ms
    max:          16995.32ms
    total time:   16995.32ms
total time: 17014.19ms
```

Edit: Updated the output, which previously showed 0ms for both `infer` and `check` due to too low a resolution on the timing histogram.

Edit 2: Updated the output again.

# Code changes

1. Added a variable (`dump_timings`) to the config.
2. Added a `check_top_span!` macro + some `info!` traces.
3. Updated `logger.rs` to conditionally add layers corresponding to `LR_DUMP_CHECKER_TRACE` and `LR_DUMP_TIMINGS`. It now returns a closure that generates the timing output (see comments for why I chose to do this).

### DIFF
--- a/flux-driver/src/callbacks.rs
+++ b/flux-driver/src/callbacks.rs
@@ -81,6 +81,8 @@ fn check_crate(tcx: TyCtxt, sess: &FluxSession) -> Result<(), ErrorGuaranteed> {
         let map = build_fhir_map(tcx, sess, &mut specs)?;
         check_wf(tcx, sess, &map)?;
 
+        tracing::info!("Callbacks::check_wf");
+
         let mut genv = GlobalEnv::new(tcx, sess, map)?;
         // Assert behavior from Crate config
         // TODO(atgeller) rest of settings from crate config
--- a/flux-driver/src/callbacks.rs
+++ b/flux-driver/src/callbacks.rs
@@ -69,40 +69,45 @@ impl Callbacks for FluxCallbacks {
 }
 
 fn check_crate(tcx: TyCtxt, sess: &FluxSession) -> Result<(), ErrorGuaranteed> {
-    let mut specs = SpecCollector::collect(tcx, sess)?;
+    tracing::info_span!("check_crate").in_scope(|| {
+        tracing::info!(event = "check_crate start");
+        let mut specs = SpecCollector::collect(tcx, sess)?;
 
-    // Ignore everything and go home
-    if specs.ignores.contains(&IgnoreKey::Crate) {
-        return Ok(());
-    }
+        // Ignore everything and go home
+        if specs.ignores.contains(&IgnoreKey::Crate) {
+            return Ok(());
+        }
 
-    // Do defn-expansion _after_ the WF check, so errors are given at user-specification level
-    let map = build_fhir_map(tcx, sess, &mut specs)?;
-    check_wf(tcx, sess, &map)?;
+        // Do defn-expansion _after_ the WF check, so errors are given at user-specification level
+        let map = build_fhir_map(tcx, sess, &mut specs)?;
+        check_wf(tcx, sess, &map)?;
 
-    let mut genv = GlobalEnv::new(tcx, sess, map)?;
-    // Assert behavior from Crate config
-    // TODO(atgeller) rest of settings from crate config
-    if let Some(crate_config) = specs.crate_config {
-        let assert_behavior = crate_config.check_asserts;
-        genv.register_assert_behavior(assert_behavior);
-    }
+        let mut genv = GlobalEnv::new(tcx, sess, map)?;
+        // Assert behavior from Crate config
+        // TODO(atgeller) rest of settings from crate config
+        if let Some(crate_config) = specs.crate_config {
+            let assert_behavior = crate_config.check_asserts;
+            genv.register_assert_behavior(assert_behavior);
+        }
 
-    let ck = CrateChecker::new(&mut genv, specs.ignores);
+        let ck = CrateChecker::new(&mut genv, specs.ignores);
 
-    if ck.ignores.contains(&IgnoreKey::Crate) {
-        return Ok(());
-    }
+        if ck.ignores.contains(&IgnoreKey::Crate) {
+            return Ok(());
+        }
 
-    let crate_items = tcx.hir_crate_items(());
-    let items = crate_items.items().map(|item| item.owner_id.def_id);
-    let impl_items = crate_items
-        .impl_items()
-        .map(|impl_item| impl_item.owner_id.def_id);
+        let crate_items = tcx.hir_crate_items(());
+        let items = crate_items.items().map(|item| item.owner_id.def_id);
+        let impl_items = crate_items
+            .impl_items()
+            .map(|impl_item| impl_item.owner_id.def_id);
 
-    items
-        .chain(impl_items)
-        .try_for_each_exhaust(|def_id| ck.check_def(def_id))
+        tracing::info!(event = "check_crate end");
+
+        items
+            .chain(impl_items)
+            .try_for_each_exhaust(|def_id| ck.check_def(def_id))
+    })
 }
 
 struct CrateChecker<'genv, 'tcx> {
--- a/flux-driver/src/callbacks.rs
+++ b/flux-driver/src/callbacks.rs
@@ -70,7 +70,6 @@ impl Callbacks for FluxCallbacks {
 
 fn check_crate(tcx: TyCtxt, sess: &FluxSession) -> Result<(), ErrorGuaranteed> {
     tracing::info_span!("check_crate").in_scope(|| {
-        tracing::info!(event = "check_crate start");
         let mut specs = SpecCollector::collect(tcx, sess)?;
 
         // Ignore everything and go home
@@ -102,11 +101,13 @@ fn check_crate(tcx: TyCtxt, sess: &FluxSession) -> Result<(), ErrorGuaranteed> {
             .impl_items()
             .map(|impl_item| impl_item.owner_id.def_id);
 
-        tracing::info!(event = "check_crate end");
-
-        items
+        let result = items
             .chain(impl_items)
-            .try_for_each_exhaust(|def_id| ck.check_def(def_id))
+            .try_for_each_exhaust(|def_id| ck.check_def(def_id));
+
+        tracing::info!("Callbacks::check_crate");
+
+        result
     })
 }
 
--- a/flux-typeck/src/dbg.rs
+++ b/flux-typeck/src/dbg.rs
@@ -29,16 +29,6 @@ macro_rules! _check_top_span {
 pub use crate::_check_top_span as check_top_span;
 
 #[macro_export]
-macro_rules! _fixpoint_span {
-    ($tcx:expr, $def_id:expr) => {{
-        let path = $tcx.def_path($def_id);
-        let def_id = path.data.iter().join("::");
-        tracing::info_span!("fixpoint", def_id = def_id.as_str())
-    }};
-}
-pub use crate::_fixpoint_span as fixpoint_span;
-
-#[macro_export]
 macro_rules! _basic_block_start {
     ($bb:expr, $rcx:expr, $env:expr) => {{
         tracing::debug!(event = "basic_block_start", bb = ?$bb, rcx = ?$rcx, env = ?$env)
--- a/flux-typeck/src/dbg.rs
+++ b/flux-typeck/src/dbg.rs
@@ -19,6 +19,16 @@ macro_rules! _check_span {
 pub use crate::_check_span as check_span;
 
 #[macro_export]
+macro_rules! _fixpoint_span {
+    ($tcx:expr, $def_id:expr) => {{
+        let path = $tcx.def_path($def_id);
+        let def_id = path.data.iter().join("::");
+        tracing::info_span!("fixpoint", def_id = def_id.as_str())
+    }};
+}
+pub use crate::_fixpoint_span as fixpoint_span;
+
+#[macro_export]
 macro_rules! _basic_block_start {
     ($bb:expr, $rcx:expr, $env:expr) => {{
         tracing::debug!(event = "basic_block_start", bb = ?$bb, rcx = ?$rcx, env = ?$env)
--- a/flux-typeck/src/dbg.rs
+++ b/flux-typeck/src/dbg.rs
@@ -19,6 +19,16 @@ macro_rules! _check_span {
 pub use crate::_check_span as check_span;
 
 #[macro_export]
+macro_rules! _check_top_span {
+    ($tcx:expr, $def_id:expr) => {{
+        let path = $tcx.def_path($def_id);
+        let def_id = path.data.iter().join("::");
+        tracing::info_span!("check_top", def_id = def_id.as_str())
+    }};
+}
+pub use crate::_check_top_span as check_top_span;
+
+#[macro_export]
 macro_rules! _fixpoint_span {
     ($tcx:expr, $def_id:expr) => {{
         let path = $tcx.def_path($def_id);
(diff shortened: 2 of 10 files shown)


flux-driver/src/callbacks.rs
```diff
@@ -69,40 +69,45 @@ impl Callbacks for FluxCallbacks {
 }
 
 fn check_crate(tcx: TyCtxt, sess: &FluxSession) -> Result<(), ErrorGuaranteed> {
-    let mut specs = SpecCollector::collect(tcx, sess)?;
+    tracing::info_span!("check_crate").in_scope(|| {
```
cole-k, line 72: Not sure if I set things up wrong, but I don't see this in the output when I run it on a sample repo.

Also, I wasn't sure whether it would make more sense to use the `#[instrument]` macro here.

flux-typeck/src/fixpoint.rs
```diff
@@ -122,56 +124,63 @@ where
         did: DefId,
         constraint: fixpoint::Constraint<TagIdx>,
     ) -> Result<(), Vec<Tag>> {
-        let kvars = self
-            .fixpoint_kvars
-            .into_iter_enumerated()
-            .map(|(kvid, sorts)| fixpoint::KVar(kvid, sorts))
-            .collect_vec();
+        fixpoint_span!(self.genv.tcx, did).in_scope(|| {
+            tracing::debug!(event = "fixpoint start");
```
cole-k, line 128: It seems like `tracing_timing` requires events to be emitted since those are what it collects in its histograms.

Putting "start" and "end" events seems a little inelegant though, so I wonder if I'm doing something wrong. Perhaps we only need the "end" event?

flux/src/logger.rs
```diff
@@ -1,23 +1,85 @@
...
         let writer = BoxMakeWriter::new(Arc::new(file));
-        let fmt_layer = tracing_subscriber::fmt::layer()
-            .with_writer(writer)
-            .json()
-            .with_filter(Targets::new().with_target("flux_typeck::checker", Level::DEBUG));
+        Some(
+            tracing_subscriber::fmt::layer()
+                .with_writer(writer)
+                .json()
+                .with_filter(Targets::new().with_target("flux_typeck::checker", Level::DEBUG)),
```
cole-k, line 33: `Option<Layer>` behaves as a toggle for the layer, see [this example](https://docs.rs/tracing-subscriber/0.3.16/tracing_subscriber/layer/index.html#runtime-configuration-with-layers).

The types don't work otherwise.

flux/src/main.rs
```diff
@@ -4,7 +4,7 @@ use std::{env, io, process::exit};
 const CMD_RUSTC: &str = "rustc";
 
 fn main() -> io::Result<()> {
-    logger::install()?;
+    let resolve_logs = logger::install()?;
```
cole-k, line 7: I feel like the proper thing to do here is to extract the `Dispatch` used and then pass it to a function which resolves the logs. However, the full type of the `Dispatch` was like two lines long and I didn't feel particularly enthused about putting that into the return type of `logger::install()`. So instead I returned a closure that would do the cleanup - this closure is to be called after the main work is done.

ranjitjhala:
🤩 awesome - what test did you run that takes 13 (or 130?) seconds?

nilehmann (changes requested):
If I follow [this](https://docs.rs/tracing-timing/0.6.0/tracing_timing/#interpreting-the-output) correctly, then you are getting zero time in `infer` and `check` because you are not emitting any events inside those scopes. I think it would make more sense to add a scope for `flux_typeck::check` and emit the events there. Something like:

```rust
tracing::info_span!("something", || {
    let bb_envs = Checker::infer(genv, ...
    trace!("Checker::infer")
    ...
        Checker::check(genv, ...
    trace!("Checker::check")
    refine_tree.simplify()
    ...
    fcx.check(def_id, constraint)
    trace!("FixpointCtxt::check")
    ...
})
```

flux/src/main.rs
nilehmann: I think that's fine.

flux/src/logger.rs
```diff
@@ -1,23 +1,85 @@
...
-use tracing::Level;
+use tracing::{Dispatch, Level};
 use tracing_subscriber::{filter::Targets, fmt::writer::BoxMakeWriter, prelude::*, Registry};
+use tracing_timing::TimingLayer;
 
-pub fn install() -> io::Result<()> {
-    if CONFIG.dump_checker_trace {
-        let log_dir = &CONFIG.log_dir;
+const ONE_HUNDRED_MILLIS: u64 = 100_000_000;
+const ONE_MINUTE: u64 = 60_000_000_000;
```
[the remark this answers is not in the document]
cole-k: [`with_histogram`](https://docs.rs/tracing-timing/0.6.0/tracing_timing/struct.TimingSubscriber.html#method.with_histograms) returns a `u64` and [`Duration::as_nanos`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_nanos) returns a `u128` so I was avoiding doing the casting (even though I know it's safe).

Your thoughts?

cole-k:
> If I follow [this](https://docs.rs/tracing-timing/0.6.0/tracing_timing/#interpreting-the-output) correctly, then you are getting zero time in `infer` and `check` because you are not emitting any events inside those scopes. I think it would make more sense to add a scope for `flux_typeck::check` and emit the events there. Something like:
> 
> ```rust
> tracing::info_span!("something", || {
>     let bb_envs = Checker::infer(genv, ...
>     trace!("Checker::infer")
>     ...
>         Checker::check(genv, ...
>     trace!("Checker::check")
>     ...
>     fcx.check(def_id, constraint)
>     trace!("FixpointCtxt::check")
>     ...
> })
> ```

Actually, we do emit events for those, see e.g. [`basic_block_start!`](https://github.com/cole-k/flux-develop/blob/81683f044f5443cb176beefd8f392b3e8dede8de/flux-typeck/src/dbg.rs#L34).

The issue was that the resolution of the histogram was too low (I was requiring 100ms minimum for durations), which I tinkered with since I was getting some timeout errors which I think were actually caused by not waiting properly for the synchronization. I've updated the PR with the new runtimes.

While we're talking about the events emitted, I was wondering whether there was a way to name them. This would allow for finer grained profiling since right now all of the events have `""` as their name.

Referring back to [`basic_block_start!`](https://github.com/cole-k/flux-develop/blob/81683f044f5443cb176beefd8f392b3e8dede8de/flux-typeck/src/dbg.rs#L34), I tried to remove the `event = `, but the compiler yelled about not understanding what `?` meant in `bb = ?$bb` when I did that. From what I read in [the docs](https://docs.rs/tracing/0.1.37/tracing/index.html#events-1), you should be able to name events - and that seems to be the name being used by `tracing-timing` to identify them.

flux/src/logger.rs
```diff
@@ -1,23 +1,85 @@
...
+                })
+                .with_filter(
+                    Targets::new()
+                        .with_target("flux_typeck::checker", Level::DEBUG)
+                        .with_target("flux_typeck::fixpoint", Level::DEBUG),
+                ),
+        )
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
cole-k: It seems I need to do

`let mut timing_layer = None;`

for this to work.

I'll do it though, seems cleaner.

flux/src/logger.rs
```diff
@@ -1,23 +1,85 @@
...
+                for (span_name, events_map) in spans_map {
+                    let mut total_time = 0;
+                    for (_event_name, event_histogram) in events_map {
+                        event_histogram.refresh();
+                        for iteration_value in event_histogram.iter_recorded() {
+                            total_time += iteration_value.value_iterated_to();
+                        }
+                    }
+                    let message =
+                        format!("{}: total time {}ms\n", span_name, total_time / 1_000_000);
```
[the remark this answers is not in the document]
cole-k: In order to implement this I had to change the outer loop in order to collect the errors. Wondering if I did the right thing or there is something more idiomatic.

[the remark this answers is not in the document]
nilehmann: You can use `?`

```rust
timing_layer.with_histograms(|spans_map| {
    for (span_name, events_map) in spans_map {
        ⋮
        write!(&mut file, "{}: total time {}ms\n", span_name, total_time / 1_000_000);
    } 
    Ok(())
})?;
```

You'll probably need to annotate the return type of the closure though.

cole-k:
https://github.com/liquid-rust/flux/pull/300/commits/90da41dbb519fff39f7727b6c3b83d05589a2373 has some pretty big changes. I didn't update the PR description to reflect them.

They are briefly summarized in the new output, rendered below. Total time taken is 18s.

```
check_top
  Checker::infer
    num events:   205
    min non-zero: 0.52ms
    1st quartile: 0.52ms
    2nd quartile: 1.05ms
    3rd quartile: 2.62ms
    max:          24.12ms
    total time:   227.02ms
  Checker::check
    num events:   205
    min non-zero: 0.52ms
    1st quartile: 0.52ms
    2nd quartile: 1.05ms
    3rd quartile: 6.29ms
    max:          159.38ms
    total time:   1890.58ms
  FixpointCtx::check
    num events:   205
    min non-zero: 25.17ms
    1st quartile: 26.21ms
    2nd quartile: 28.31ms
    3rd quartile: 40.37ms
    max:          1893.73ms
    total time:   9479.13ms
total time: 11596.73ms

check_crate
  Callbacks::check_crate
    num events:   1
    min non-zero: 17297.31ms
    1st quartile: 17314.09ms
    2nd quartile: 17314.09ms
    3rd quartile: 17314.09ms
    max:          17314.09ms
    total time:   17314.09ms
total time: 17314.09ms
```

flux-driver/src/callbacks.rs
```diff
@@ -69,40 +69,46 @@ impl Callbacks for FluxCallbacks {
...
-    let items = crate_items.items().map(|item| item.owner_id.def_id);
-    let impl_items = crate_items
-        .impl_items()
-        .map(|impl_item| impl_item.owner_id.def_id);
+        tracing::info!("Callbacks::check_crate");
 
-    items
-        .chain(impl_items)
-        .try_for_each_exhaust(|def_id| ck.check_def(def_id))
+        result
```
nilehmann, line 110: The result is not relevant here, so maybe better like so:

```rust
        items
            .chain(impl_items)
            .try_for_each_exhaust(|def_id| ck.check_def(def_id))?;

        tracing::info!("Callbacks::check_crate");

        Ok(())
```

flux-typeck/src/lib.rs
```diff
@@ -48,24 +49,35 @@ pub fn check<'tcx>(
...
+        let mut kvars = fixpoint::KVarStore::new();
+        let mut refine_tree =
+            Checker::check(genv, body, def_id, &mut kvars, bb_envs).emit(genv.sess)?;
+        refine_tree.simplify();
 
-    match fcx.check(def_id, constraint) {
-        Ok(_) => Ok(()),
-        Err(tags) => report_errors(genv, body.span(), tags),
-    }
+        tracing::info!("Checker::check");
```
nilehmann, line 62: Not very important but I would swap this and the previous line

flux-typeck/src/dbg.rs
```diff
@@ -18,6 +18,26 @@ macro_rules! _check_span {
...
+
+#[macro_export]
+macro_rules! _fixpoint_span {
+    ($tcx:expr, $def_id:expr) => {{
+        let path = $tcx.def_path($def_id);
+        let def_id = path.data.iter().join("::");
+        tracing::info_span!("fixpoint", def_id = def_id.as_str())
+    }};
+}
+pub use crate::_fixpoint_span as fixpoint_span;
```
nilehmann, line 39: I think this is not used anymore

flux-typeck/src/dbg.rs
```diff
@@ -18,6 +18,26 @@ macro_rules! _check_span {
 }
 pub use crate::_check_span as check_span;
 
+#[macro_export]
+macro_rules! _check_top_span {
```
nilehmann, line 22: We really need a better name for this function, check is too generic, but can't think of anything else.

nilehmann (approved):
@cole-k looking good!  just a final round of comments, but feel free to merge after you address them.

flux-driver/src/callbacks.rs
cole-k: I chose not to do this because we won't log duration in the case that it fails (i.e. we'll only trace if we get an overall `Ok`).

flux-typeck/src/dbg.rs
cole-k: Neither could I. As long as nothing downstream depends on the terminology we can change it later.

cole-k:
@nilehmann I don't have write access, so merge at your leisure.

nilehmann:
Ah thought you had. Just gave you write access. @cole-k

ranjitjhala:
am merging! thanks @cole-k and @nilehmann !

ranjitjhala:
btw `flux` is getting large enough now that its probably useful to have a searchable `mkDocs` type of document -- (if not just dumping some of this info into the README). So for example, whats the quick summary of how to get these nice profiles? [if you point me to it i can draft what i mean by the doc - by scraping the relevant text from this conversation] thanks!

nilehmann:
@ranjitjhala the rust community typically uses https://rust-lang.github.io/mdBook/ for this sort of stuff

ranjitjhala:
Sweet let me look into that!

cole-k:
@ranjitjhala RE:

> ... So for example, whats the quick summary of how to get these nice profiles? [if you point me to it i can draft what i mean by the doc - by scraping the relevant text from this conversation] thanks!

The quick summary is in the body of this PR. I just updated it to reflect the final changes before merging.

> # TL;DR
> Set LR_DUMP_TIMINGS=true when you run flux to have flux write timing diagnostics to ./log/timings.
> Right now this is extremely simple, it just provides some details for the spans under flux_typeck and flux_driver.

If you want a little more details, each event under the spans that are timed timed includes the following summary:

```
num events:   number
min non-zero: time in ms
1st quartile: time in ms
2nd quartile: time in ms
3rd quartile: time in ms
max:          time in ms
total time:   time in ms
```

ranjitjhala:
Awesome thanks @cole-k !!!

PR #302, opened 2023-01-07
Add support for localizing qualifier usage to a function

ranjitjhala:
When defining a qualifier, specify it is NOT global as

```rust
#![flux::qualifier(local, MyQ1(x: int, y: int, z: int) { x + y <= z + 10 })]
```

To then use it at a particular function write

```rust
#[flux::qualifiers(MyQ1)]
#[flux::sig(fn (a: i32) -> RVec<Pair{v : v.x + v.y <= a + 10 }>)]
pub fn mk_pairs_with_bound(a: i32) -> RVec<Pair> { ... }
```

see `tests/{pos,neg}/surface/local_qual00.rs`


flux-middle/src/global_env.rs
```diff
@@ -27,6 +27,7 @@ pub struct GlobalEnv<'genv, 'tcx> {
     qualifiers: Vec<rty::Qualifier>,
     uifs: FxHashMap<Symbol, rty::UifDef>,
     fn_sigs: RefCell<FxHashMap<DefId, rty::PolySig>>,
+    fn_quals: FxHashMap<DefId, Vec<SurfaceIdent>>,
```
nilehmann, line 30: Maybe 
```suggestion
    fn_quals: FxHashMap<DefId, FxHashSet<String>>,
```
Also, perhaps a  doc comment saying this is the set of qualifiers for a function.

nilehmann (approved):

PR #304, opened 2023-01-08
Skip trivial constraints and do constant folding

ranjitjhala:
* Omit calling `fixpoint` when the constraint is trivial
* Do constant folding in `flux`

(shaves another ~4 / 18 seconds off `flux-wave`)


flux-fixpoint/src/constraint.rs
```diff
@@ -517,6 +543,95 @@ impl fmt::Display for Constant {
...
+    fn to_bool(&self) -> Option<bool> {
+        match self {
+            Constant::Bool(b) => Some(*b),
+            _ => None,
+        }
+    }
+
+    fn to_int(&self) -> Option<i128> {
+        match self {
+            Constant::Int(Sign::Positive, n) => Some(*n as i128),
```
nilehmann, line 556: This is wrong because `n as i128` can overflow. The reason ints are represented with a sign and a `u128` and not just an `i128`, is that we need to be able to represent both `u128::MAX` and `i128::MIN` as they are both valid rust literals. 

Initially, I thought about representing numbers with [BigInt](https://docs.rs/big_num/latest/big_num/struct.BigInt.html) since at this level we are actually dealing with mathematical integers, but I ended up going for the simpler representation because we only needed to represent valid rust literals. If we are going to do arithmetic maybe we need to reconsider (or be very careful with overflows).

flux-middle/src/rty/expr.rs
```diff
@@ -267,6 +267,26 @@ impl Expr {
...
+            BinOp::Div => c1.div(c2),
+            BinOp::Mod => c1.mod_(c2),
+            BinOp::Eq => Some(c1.eq(c2)),
+            BinOp::Ne => Some(c1.ne(c2)),
+        }
+    }
+
     /// Simplify expression applying some simple rules like removing double negation. This is
     /// only used for pretty printing.
     pub fn simplify(&self) -> Expr {
```
nilehmann, line 292: Note that this function is only used for pretty printing and it won't affect the constraint we throw at fixpoint.

nilehmann (changes requested):
@ranjitjhala I have some concerns with the constant folding part. Maybe we can drop that part from this PR and focus on the `is_concrete` part.

ranjitjhala:
Incidentally, I saw some smt2 files with constraints like 10000 != 0 (in the head) but no such clause in the lrc file - any idea why?

nilehmann:
> Incidentally, I saw some smt2 files with constraints like 10000 != 0 (in the head) but no such clause in the lrc file - any idea why?

Yeah, that's because when dumping the `lrc` file we do call simplify

flux-middle/src/rty/expr.rs
nilehmann: You probably read that comment in `Expr::is_trivially_true`.

Yeah, comparisons should be easier than general constant folding. We could call `Expr::simplify` [here](https://github.com/liquid-rust/flux/blob/9bf0d9cc21fb2998c7abaa24b09d8b3316b2da24/flux-refineck/src/refine_tree.rs#L382). But note that we still need to implement comparisons in the signs + usize representation, but I think if we define it as a struct with the fields in the right order then we could `#[derive(PartialOrd, Ord, PartialEq, Eq)]`. The alternative is to extend `Expr::is_trivially_true` to look for trivial comparisons. The case `1=1` is already handled there for example.

ranjitjhala:
@nilehmann  I've removed all the constant folding stuff from this PR to simplify matters. let me experiment with that separately to see if its worth it...

nilehmann (approved):
Ok, this looks good to me. 🚢 it!

PR #305, opened 2023-01-08
Add cargo-flux binary

cole-k:
# Edit: Usage and installation details are now in `flux-docs`

# Testing locally

Change https://github.com/liquid-rust/flux/pull/305/files#diff-e3670cdab561aaacb93c3521ffc0db64aa666949dcfc0e5e7f0148fd595ef5afR18 to your flux binary path.

```
cargo install --path flux/bin
```

Note that this will clobber a pre-existing binary, so if you put your `cargo-flux` shell script there, you'll need to copy it elsewhere if you don't want it to disappear.


ranjitjhala:
Nice! Can you update the relevant bit in the docs?

cole-k:
Haven't updated the docs yet but you should now be able to just do

```
cargo install --lib flux
cargo install --lib flux-bin
```

I'd love to just be able to do `cargo install`, but seems [that isn't possible](https://github.com/rust-lang/cargo/issues/4101) with the way things are set up.

@nilehmann any suggestions on how we might be able to reconfigure? Should I add a `[package]` and containing `[[bin]]`s to the main `Cargo.toml`?

cole-k:
We could also add a binary that generates the vscode plugin json.

```json
{
  "rust-analyzer.checkOnSave.extraEnv": {
    "RUSTC_WRAPPER": "/path/to/flux/target/release/flux",
    "RUSTUP_TOOLCHAIN": "nightly-2022-10-11",
    "LD_LIBRARY_PATH": "/path/to/.rustup/toolchains/nightly-2022-10-11-x86_64-apple-darwin/lib"
  }
}
```

Although personally I feel like it would make more sense to have this be generated when you run `cargo build`. Not sure how to do that or where to put it, though.

nilehmann:
I think this is good enough. You can do `cargo build --release` that will put all the binaries in `/path/to/flux/target/release` and then you can set up your vscode as

```
{
  "rust-analyzer.checkOnSave.overrideCommand": [
    "/path/to/flux/target/release/cargo-flux",
    "check",
    "--workspace",
    "--message-format=json"
  ]
}
```

flux-bin/src/utils.rs
```diff
@@ -0,0 +1,108 @@
...
+pub const LIB_PATH: &str = "PATH";
+#[cfg(target_os = "linux")]
+pub const LIB_PATH: &str = "LD_LIBRARY_PATH";
+#[cfg(target_os = "macos")]
+pub const LIB_PATH: &str = "DYLD_FALLBACK_LIBRARY_PATH";
+
+pub const EXIT_ERR: i32 = -1;
+
+// It's better than using .expect(), right?
+pub fn report_err(message: &str, err: impl Error) -> i32 {
```
nilehmann, line 13: Maybe better to just return `Result<T, Box<dyn std::error::Error>>` everywhere and then print at the top level. The `description` will probably be good enough (and maybe better than our custom messages)  so no point in categorizing all the errors.

cole-k: Yeah I think that makes sense. I'm rendering the `impl Error` in addition to outputting a message, but looking through the file, it seems like there isn't much helpful.

cole-k:
> I think this is good enough. You can do `cargo build --release` that will put all the binaries in `/path/to/flux/target/release` and then you can set up your vscode as
> 
> ```json
> {
>   "rust-analyzer.checkOnSave.overrideCommand": [
>     "/path/to/flux/target/release/cargo-flux",
>     "check",
>     "--workspace",
>     "--message-format=json"
>   ]
> }
> ```

If you ran `cargo install --lib flux/bin` would you be able to even write it as below?

```json
{
  "rust-analyzer.checkOnSave.overrideCommand": [
    "cargo-flux",
    "check",
    "--workspace",
    "--message-format=json"
  ]
}
```

flux-bin/src/utils.rs
```diff
@@ -0,0 +1,108 @@
...
+    toolchain_file
+        .toolchain()
+        .spec()
+        .ok_or_else(|| report_message("No spec in rust-toolchain file"))?
+        .channel()
+        .ok_or_else(|| report_message("No channel in rust-toolchain file"))
+        .map(|channel| channel.name().to_string())
+}
+
+pub fn get_dyld_fallback_library_path(rust_toolchain: &str) -> Result<PathBuf, i32> {
```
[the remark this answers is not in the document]
cole-k: I picked the linux name (`ld_library_path`) since it seems the most descriptive, but I'm open to suggestions.

nilehmann:
If  `cargo-flux` is in your path and `flux` is in the same directory then yes you would be able to write it like that. 

But what is this `cargo install --lib`? It doesn't work for me. I can do `cargo install --path flux-bin` and `cargo install --path flux`.

cole-k:
> If `cargo-flux` is in your path and `flux` is in the same directory then yes you would be able to write it like that.
> 
> But what is this `cargo install --lib`? It doesn't work for me. I can do `cargo install --path flux-bin` and `cargo install --path flux`.

Yeah sorry, I meant `cargo install --path`. Fixed.

flux-bin/src/utils.rs
cole-k: We now use `anyhow` to collect the errors.

flux-bin/src/utils.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .map_err(anyhow::Error::from);
+            }
+        }
+    }
+    Err(anyhow!("Could not read Rustup toolchains folder"))
+}
+
+pub fn get_rustup_home() -> Result<PathBuf> {
+    env::var("RUSTUP_HOME")
+        .map(|home_dir| PathBuf::from(&home_dir))
```
nilehmann, line 71:
```suggestion
        .map(PathBuf::from)
```

flux-bin/src/utils.rs
```diff
@@ -0,0 +1,88 @@
...
+pub const LIB_PATH: &str = "PATH";
+#[cfg(target_os = "linux")]
+pub const LIB_PATH: &str = "LD_LIBRARY_PATH";
+#[cfg(target_os = "macos")]
+pub const LIB_PATH: &str = "DYLD_FALLBACK_LIBRARY_PATH";
+
+pub const EXIT_ERR: i32 = -1;
+
+pub fn get_flux_path() -> Result<PathBuf> {
+    let mut flux_path = env::current_exe().map(|path| path.with_file_name("flux"))?;
```
nilehmann:
```suggestion
    let mut flux_path = env::current_exe()?.with_file_name("flux"));
```

flux-bin/src/utils.rs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                _ => Err(anyhow::Error::from(e)),
+            }
+        })
+}
+
+/// Prepends the path so that it's the first checked.
+pub fn extend_env_var_with_path(var_name: &str, new_path: PathBuf) -> Result<OsString> {
+    let mut paths = env::var(var_name)
+        .map(|paths_str| env::split_paths(&paths_str).collect::<Vec<_>>())
```
nilehmann, line 87: I prefer `collect_vec`, but you would need to add `itertools` as a dependency.

flux/Cargo.toml
```diff
@@ -2,6 +2,13 @@
 edition = "2021"
 name = "flux"
 version = "0.1.0"
+default-run = "flux"
```
nilehmann, line 5: 👍🏽

cole-k:
I've updated the readme with the new instructions. I tried to follow them myself but it would help if one of you could confirm that it works.

I also added a `FLUX_PATH` env variable so you can run a different version of flux if desired. I feel like there isn't really a difference between running

```bash
cargo install --lib flux
```

and


```bash
cargo build
```

but maybe if you have the muscle memory for the latter, you will find that env variable helpful.

nilehmann (commented):
@cole-k the code looks good. Just left you some small stylistics suggestions. 

@ranjitjhala made this page https://liquid-rust.github.io/flux/. We you should add the changes to the docs there. The page can be edited here https://github.com/liquid-rust/flux/tree/main/flux-docs/src. We should also probably remove most of the text from the README and just put links to the doc

flux-bin/src/utils.rs
cole-k: The annotation isn't even necessary, I think my intellisense inserted it before it knew the whole type.

cole-k:
Moved the readme's contents to `flux-docs`.

flux-docs/src/use/install.md
```diff
@@ -23,8 +23,17 @@ To build the source you need a nightly version of rustc.
...
+```
+
+Note that you should not call `flux` directly.
+
+Finally, install `rustc-flux` and `cargo-flux` using
+
+```bash
+cargo install --path flux-bin
+```
+```
```
ranjitjhala, line 39: extra ``` ?

cole-k: Thanks for that catch

nilehmann:
@cole-k I think this is ready to go as soon as you resolve the conflict

nilehmann (approved):

PR #316, opened 2023-01-11
Cache fixpoint queries

ranjitjhala:
This PR caches the _safe_ fixpoint queries for each `DefId`  in `log/cache.json` and then on subsequent runs, skips queries that are already in the "safe" cache, makes it possible to work with `flux-wave` rather interactively.

### DIFF
--- a/flux-fixpoint/src/constraint.rs
+++ b/flux-fixpoint/src/constraint.rs
@@ -151,7 +151,7 @@ impl<Tag> Constraint<Tag> {
         }
     }
 
-    pub fn myhash(&self) -> u64 {
+    pub fn hash_with_default(&self) -> u64 {
         let mut hasher = DefaultHasher::new();
         self.hash(&mut hasher);
         hasher.finish()
--- a/flux-fixpoint/src/constraint.rs
+++ b/flux-fixpoint/src/constraint.rs
@@ -155,9 +155,11 @@ impl<Tag> Constraint<Tag> {
 
 impl<Tag> Hash for Constraint<Tag> {
     fn hash<H: Hasher>(&self, state: &mut H) {
+        let tag = std::mem::discriminant(self);
+        tag.hash(state);
         match self {
             Constraint::Pred(p, _) => p.hash(state),
-            Constraint::Conj(cs) => cs.iter().for_each(|c| c.hash(state)),
+            Constraint::Conj(cs) => cs.hash(state),
             Constraint::Guard(p, c) => {
                 p.hash(state);
                 c.hash(state)
--- a/flux-fixpoint/src/constraint.rs
+++ b/flux-fixpoint/src/constraint.rs
@@ -1,5 +1,7 @@
 use std::{
+    collections::hash_map::DefaultHasher,
     fmt::{self, Write},
+    hash::{Hash, Hasher},
     sync::LazyLock,
 };
 
@@ -14,7 +16,7 @@ pub enum Constraint<Tag> {
     ForAll(Name, Sort, Pred, Box<Self>),
 }
 
-#[derive(Clone)]
+#[derive(Clone, Hash)]
 pub enum Sort {
     Int,
     Bool,
@@ -24,17 +26,19 @@ pub enum Sort {
     Func(FuncSort),
 }
 
-#[derive(Clone)]
+#[derive(Clone, Hash)]
 pub struct FuncSort {
     pub inputs_and_output: Vec<Sort>,
 }
 
+#[derive(Hash)]
 pub enum Pred {
     And(Vec<Self>),
     KVar(KVid, Vec<Name>),
     Expr(Expr),
 }
 
+#[derive(Hash)]
 pub enum Expr {
     Var(Name),
     Constant(Constant),
@@ -47,12 +51,13 @@ pub enum Expr {
     Unit,
 }
 
+#[derive(Hash)]
 pub enum Func {
     Var(Name),
     Uif(String),
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Hash)]
 pub enum Proj {
     Fst,
     Snd,
@@ -143,6 +148,31 @@ impl<Tag> Constraint<Tag> {
             Constraint::Conj(cs) => cs.iter().any(|c| c.is_concrete()),
             Constraint::Guard(_, c) | Constraint::ForAll(_, _, _, c) => c.is_concrete(),
             Constraint::Pred(p, _) => p.is_concrete() && !p.is_trivially_true(),
+        }
+    }
+
+    pub fn myhash(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        self.hash(&mut hasher);
+        hasher.finish()
+    }
+}
+
+impl<Tag> Hash for Constraint<Tag> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Constraint::Pred(p, _) => p.hash(state),
+            Constraint::Conj(cs) => cs.iter().for_each(|c| c.hash(state)),
+            Constraint::Guard(p, c) => {
+                p.hash(state);
+                c.hash(state)
+            }
+            Constraint::ForAll(x, t, p, c) => {
+                x.hash(state);
+                t.hash(state);
+                p.hash(state);
+                c.hash(state)
+            }
         }
     }
 }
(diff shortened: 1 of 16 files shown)


ranjitjhala:
One question is whether we want flags for this or caching to be the default?

At any rate, one must be careful to do `rm flux-tests/**/cache.json` before running `cargo test`!

ranjitjhala:
yikes occurs to me that we should compute the hash AFTER "closing" with constants...

flux-fixpoint/src/constraint.rs
```diff
@@ -145,6 +150,31 @@ impl<Tag> Constraint<Tag> {
...
+        self.hash(&mut hasher);
+        hasher.finish()
+    }
+}
+
+impl<Tag> Hash for Constraint<Tag> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Constraint::Pred(p, _) => p.hash(state),
+            Constraint::Conj(cs) => cs.iter().for_each(|c| c.hash(state)),
```
nilehmann, line 165: I don't fully understand the problem but maybe we should beware of prefix collision here (https://doc.rust-lang.org/std/hash/trait.Hash.html#prefix-collisions)

flux-middle/src/cache.rs
```diff
@@ -0,0 +1,55 @@
...
+        QueryCache { entries: FxHashMap::default() }
+    }
+
+    pub fn insert<Tag>(&mut self, did: DefId, constr_hash: u64) {
+        let str = crate::pretty::def_id_to_string(did);
+        self.entries.insert(str, constr_hash);
+    }
+
+    pub fn is_safe<Tag>(&self, did: DefId, constr_hash: u64) -> bool {
+        let str = crate::pretty::def_id_to_string(did);
```
nilehmann, line 28: This string representation for `DefId` is not guaranteed to be unique, so I wouldn't use it to hash stuff. Why not only store the constraint hash?

flux-middle/src/cache.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub fn new() -> Self {
+        QueryCache { entries: FxHashMap::default() }
+    }
+
+    pub fn insert<Tag>(&mut self, did: DefId, constr_hash: u64) {
+        let str = crate::pretty::def_id_to_string(did);
+        self.entries.insert(str, constr_hash);
+    }
+
+    pub fn is_safe<Tag>(&self, did: DefId, constr_hash: u64) -> bool {
```
nilehmann, line 27: Is there a reason to parameterize this by the tag?

nilehmann (changes requested):

flux-fixpoint/src/constraint.rs
ranjitjhala: hmm I don't follow that comment either -- but perhaps doesn't matter here because of the `and` ?

nilehmann:
> One question is whether we want flags for this or caching to be the default?
> 
> At any rate, one must be careful to do `rm flux-tests/**/cache.json` before running `cargo test`!

We should probably add a flag to turn caching on an off and disabled it for testing. What prusti does is define one flag to set the path for the cache directory and if it is empty then  it disables the cache.

flux-fixpoint/src/constraint.rs
nilehmann: I think the problems applies because if we have the following constraints

```
c1 = Constraint::Conj("a > 0", "b > 0");
c2 = Constraint::Conj("c > 0");
c3 = Constraint::Conj("a > 0");
c2 = Constraint::Conj("b > 0", "c > 0");
```

Then the the hash for the tuples `(c1, c2)` and `(c3, c2)` will be the same no?

nilehmann: Being this strict will maybe miss equivalent queries that could be normalized to the same query, but better be safe than sorry.

flux-middle/src/cache.rs
ranjitjhala: I was using the `fn` name as a form of "garbage" collection -- namely just store (the most recent) hash for each function, as opposed to every query since the last creation of the file. 

I changed it to use [`def_path_str`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.def_path_str)

which seems stable and unique and "fully qualified" e.g.

* `runtime::<impl types::VmCtx>::translate_path` 
* `wrappers::wasi_path_readlink`

flux-middle/src/cache.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn path() -> Result<PathBuf, std::io::Error> {
+        let dir = &CONFIG.log_dir;
+        std::fs::create_dir_all(dir)?;
+        Ok(dir.join("cache.json"))
+    }
+
+    pub fn save(&self) -> Result<(), std::io::Error> {
+        let path = Self::path()?;
+        let mut file = File::create(path).unwrap();
+        serde_json::to_writer(&mut file, &self.entries).unwrap();
```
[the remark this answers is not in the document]
ranjitjhala: I like the fact that we can read the json for debugging but we can switch to this if its faster?

flux-middle/src/cache.rs
ranjitjhala: oops, vestige.

flux-middle/src/cache.rs
nilehmann: ok, that's fair we don't want the cache growing indefinitely. I think we can at some point implement a lru cache or some other form of eviction strategy.

ranjitjhala:
Ok I like that I'll add a `LR_CACHE=file` flag and only do the caching if that is set. (Maybe that's related to the failed test? do you know if those are persisted across CI?)

nilehmann:
> 
Oh, I know what it is! We are not adding the qualifiers to the cache, so the cached query for the `pos` case which contains the local qualifier wins over the `neg` one.

nilehmann:
Maybe we should hash the `flux-fixpoint::Task`

flux-middle/src/cache.rs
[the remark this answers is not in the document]
nilehmann: It should be faster and it should also generate smaller files. We could write a small tool if we ever want to pick at the content of the file.

ranjitjhala:
@nilehmann I made all the edits except the `bincode` (which threw some scary SIGABORT run-time error!)

* `LR_CACHE=file` switches it on

cole-k:
Might I suggest that instead we have two env vars

* `LR_CACHE=1` enables caching
* `LR_CACHE_FILE=/some/file` is the absolute path to a file to cache to (defaults to `~/.flux/cache`)

This suggestion doesn't need to block this PR though.

ranjitjhala:
Hmm I like this - @nico any thoughts?

ranjitjhala:
I implemented @cole-k 's suggestion...

nilehmann:
I would either have the two env variables as @cole-k suggests or make `LR_CACHE` mean the path to the file and not just the name (with empty meaning disabled). Both options sound good. But the default CANNOT be `~/.flux/cache.json` as the cache should be per _project_.

Also @ranjitjhala you haven't considered these changes https://github.com/liquid-rust/flux/pull/316/files#r1067353829.

cole-k:
> I would either have the two env variables as @cole-k suggests or make `LR_CACHE` mean the path to the file and not just the name (with empty meaning disabled). Both options sound good. But the default CANNOT be `~/.flux/cache.json` as the cache should be per _project_.

Whoops, yeah I meant `./.flux/cache.json`!

ranjitjhala:
Odd I totally didn’t even see those changes - some UI glitch . Let me take
a look…

flux-fixpoint/src/constraint.rs
ranjitjhala: @nilehmann I just updated this can you take a look and see if this is what you meant?

ranjitjhala:
Just updated the `hash` for constraints @nilehmann

nilehmann (approved):
Ok, so maybe let's merge this, and then @cole-k can put a PR with his suggestions.

ranjitjhala:
@nilehmann any idea why the `check-fmt` and `check-tests` seem blocked?

ranjitjhala:
looks like some odd gh glitch (the actions seem to be  "green" ...)

PR #319, opened 2023-01-11
#205 Add more binops

### LINKED ISSUE #205: Implement initial support for `BinOp::BitOr`, `BinOp::Shl` and `BinOp::Shr`
We could easily implement some basic support for these operations. For `BitOr` we should do the same as with `BitAnd` and have a refined signature for the boolean case, and an unrefined signature for the integral case. `BinOp::Shl` and `BinOp::Shr` should just have an unrefined signature for now.

cole-k:
Adds support for `BitOr`, `Shl`, and `Shr`.

Resolves #205.

### DIFF
--- a/flux-refineck/src/checker.rs
+++ b/flux-refineck/src/checker.rs
@@ -933,18 +933,13 @@ fn type_of_shift<'a>(
     op: mir::BinOp,
 ) -> (BaseTy, &'a rty::RefineArg, &'a rty::RefineArg, sigs::Sig<2>) {
     match (num_ty.kind(), shift_by_ty.kind()) {
-        (Int!(num_base_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
-            (
-                BaseTy::Int(*num_base_ty),
-                num_idxs.nth(0),
-                shift_by_idxs.nth(0),
-                sigs::signed_bin_ops(op),
-            )
+        (Int!(int_ty, int_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
+            (BaseTy::Int(*int_ty), int_idxs.nth(0), shift_by_idxs.nth(0), sigs::signed_bin_ops(op))
         }
-        (Uint!(num_base_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
+        (Uint!(uint_ty, uint_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
             (
-                BaseTy::Uint(*num_base_ty),
-                num_idxs.nth(0),
+                BaseTy::Uint(*uint_ty),
+                uint_idxs.nth(0),
                 shift_by_idxs.nth(0),
                 sigs::unsigned_bin_ops(op),
             )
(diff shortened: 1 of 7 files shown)


flux-tests/tests/neg/surface/binop.rs
```diff
@@ -0,0 +1,43 @@
...
+// Logical BinOps
+
+// Should be {v: v == a || v == b}
+#[flux::sig(fn(a: bool, b: bool) -> bool{v: v == a})]
+pub fn logical_or(a: bool, b: bool) -> bool {
+    a | b //~ ERROR postcondition
+}
+
+// Should be v == true
+#[flux::sig(fn(bool{v: v == false}, bool{v: v == true}) -> bool{v: v == false})]
```
nilehmann, line 34: For any `T` you can write `T{v: v == e}`  as `T[e]`.

```suggestion
#[flux::sig(fn(bool[false], bool[true]) -> bool[false])]
```

flux-refineck/src/checker.rs
```diff
@@ -923,6 +927,37 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
     }
 }
 
+fn type_of_shift<'a>(
```
nilehmann, line 930: yikes, I didn't know shift operators could be used with mixed types. We need a more general infrastructure to declare types for operators, but this will do for now.

flux-refineck/src/checker.rs
```diff
@@ -923,6 +927,37 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+    num_ty: &'a Ty,
+    shift_by_ty: &'a Ty,
+    op: mir::BinOp,
+) -> (BaseTy, &'a rty::RefineArg, &'a rty::RefineArg, sigs::Sig<2>) {
+    match (num_ty.kind(), shift_by_ty.kind()) {
+        (Int!(num_base_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
```
nilehmann, line 936: We call this an `int_ty`.

```suggestion
        (Int!(num_int_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
```

flux-refineck/src/checker.rs
```diff
@@ -923,6 +927,37 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
+            (
+                BaseTy::Int(*num_base_ty),
+                num_idxs.nth(0),
+                shift_by_idxs.nth(0),
+                sigs::signed_bin_ops(op),
+            )
+        }
+        (Uint!(num_base_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
```
nilehmann, line 944: Same here we call this a `uint_ty`

```suggestion
        (Uint!(num_uint_ty, num_idxs), Int!(_, shift_by_idxs) | Uint!(_, shift_by_idxs)) => {
```

nilehmann (approved):
Looking good! just some small comments.

flux-refineck/src/checker.rs
cole-k: Yeah, it makes sense to me to allow `i32 << u32` but i don't really understand why you would allow a signed shift operand.

PR #321, opened 2023-01-12
Rename ENV variables to use `FLUX_` prefix and enable config file.

cole-k:
See the new `flags.md` for more details.

### DIFF
--- a/flux-common/src/config.rs
+++ b/flux-common/src/config.rs
@@ -1,8 +1,11 @@
 use std::{io::Read, path::PathBuf, sync::LazyLock};
 
-use config::Environment;
+use config::{Environment, File};
 use serde::Deserialize;
 pub use toml::Value;
+
+const FLUX_ENV_VAR_PREFIX: &str = "FLUX";
+const FLUX_CONFIG_ENV_VAR: &str = "FLUX_CONFIG";
 
 #[derive(Debug, Deserialize, Copy, Clone)]
 #[serde(rename_all = "lowercase")]
@@ -35,6 +38,7 @@ pub struct CrateConfig {
 
 #[derive(Deserialize)]
 pub struct Config {
+    pub path: Option<PathBuf>,
     pub log_dir: PathBuf,
     pub dump_constraint: bool,
     pub dump_checker_trace: bool,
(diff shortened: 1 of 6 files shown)


flux-common/src/config.rs
```diff
@@ -35,6 +38,7 @@ pub struct CrateConfig {
 
 #[derive(Deserialize)]
 pub struct Config {
+    pub path: Option<PathBuf>,
```
nilehmann, line 41: I'd add this to the `set_default` to make it more explicit that it defaults to `None`.

nilehmann (approved):
Nice!

flux-common/src/config.rs
cole-k: So I did that but it wants me to annotate the `Option`, and then complains when I try to specify it as `PathBuf`

```
error[E0277]: the trait bound `ValueKind: From<PathBuf>` is not satisfied
   --> flux-common/src/config.rs:57:34
    |
57  |             .set_default("path", None::<PathBuf>)?
    |              -----------         ^^^^^^^^^^^^^^^ the trait `From<PathBuf>` is not implemented for `ValueKind`
    |              |
    |              required by a bound introduced by this call
    |
```

I put `None::<String>` instead and it seems to work. It's a little weird that Rust's type system requires you to annotate `None::<PathBuf>`, is there any case where the annotation actually makes a difference when running code?

nilehmann: Did you test setting a custom `Path`, if it doesn't let you `set_default` with `Node::<PathBuf>` then I'm worried that it won't allow you to set the path because it doesn't know how to parse a string into a `PathBuf`. In that case, maybe we should give `path` type `Option<String>` and convert to `PathBuf` latter.

nilehmann: (`set_default` takes any `T: ValueKind` so there's no way to know the type of `None` alone by just using it as an argument to `set_default`. And yes the value `None` can have a different layout depending on the type of `T` in `Option<T>`. If `T` has a niche, for example, `None` will have the same size as `T`. The optimization is called [num discriminant elision](https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#discriminant-elision-on-option-like-enums))

cole-k: It deserializes fine. I think `serde` would complain at compile time if it didn't know how to deserialize a `PathBuf`.

PR #328, opened 2023-01-14
Implement macros for reporting bugs

nilehmann:
* Implement macros to report bugs in a structured way
* Track span in thread local to avoid threading it down


ranjitjhala:
This is terrific @nilehmann! it might be good to extract a small example of
how to use this new infrastructure and pop into the docs as reference?

PR #329, opened 2023-01-14
Change syntax to support declaring multiple items in a single block

nilehmann:


ranjitjhala:
Awesome! As added bonus now we get fn keyword for syntax highlighting.

PR #330, opened 2023-01-15
Existentials in function output position

nilehmann:
Implement existentials in function output position using the `#n` syntax as discussed in https://github.com/liquid-rust/flux/issues/239

Closes https://github.com/liquid-rust/flux/issues/239


ranjitjhala:
Whoa!!! Awesome!!!

Issue #332, opened 2023-01-17
Unexpected panic


nilehmann:
The following code

```rust
#[flux::refined_by(n: int)]
#[flux::invariant(n >= 0)]
pub enum List<T> {
    #[flux::variant(List<T>[0])]
    Nil,
    #[flux::variant((T, Box<List<T>[@n]>) -> List<T>[n+1])]
    Cons(T, Box<List<T>>),
}

impl<T> List<T> {
    #[flux::sig(
        fn(self: &strg List<T>[@n1], List<T>[@n2])
        ensures self: List<T>[n1 + n2]
    )]
    pub fn append(&mut self, other: List<T>) {
        if let List::Cons(_, tl) = self {
            tl.append(other)
        }
    }
}
```
is panicking with
```
thread 'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: UnsolvedEvar { evar: ?e0#1 }', flux-refineck/src/type_env/paths_tree.rs:676:26
```

The implementation of append is wrong, but the code shouldn't panic.

ranjitjhala:
Yikes -- I just slammed into this myself! Is there some new protocol for threading `Span` around to show at the panic site? (I thought there was something in a PR)

PR #333, opened 2023-01-17
Span for unwrap error

ranjitjhala:
For #332


ranjitjhala:
merging as super local!

PR #336, opened 2023-01-18
More spans and better error message 

ranjitjhala:
... when type invariant fails to hold when a place is folded.

### DIFF
--- a/flux-errors/locales/en-US/refineck.ftl
+++ b/flux-errors/locales/en-US/refineck.ftl
@@ -23,13 +23,10 @@ refineck_param_inference_error =
     parameter inference error at function call
 
 refineck_fold_error =
-    fold error
+    type invariant may not hold (when place is folded)
 
 refineck_unknown_error =
-    unknown error
-
-refineck_invariant_error =
-    type invariant may not hold
+    cannot prove this code safe
 
 refineck_overflow_error =
     arithmetic operation may overflow
--- a/flux-errors/locales/en-US/refineck.ftl
+++ b/flux-errors/locales/en-US/refineck.ftl
@@ -28,6 +28,9 @@ refineck_fold_error =
 refineck_unknown_error =
     unknown error
 
+refineck_invariant_error =
+    type invariant may not hold
+
 refineck_overflow_error =
     arithmetic operation may overflow
 
--- a/flux-refineck/src/checker.rs
+++ b/flux-refineck/src/checker.rs
@@ -74,7 +74,7 @@ pub trait Phase: Sized {
         ck: &mut Checker<Self>,
         rcx: RefineCtxt,
         env: TypeEnv,
-        src_info: Option<SourceInfo>,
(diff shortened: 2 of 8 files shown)


flux-refineck/src/checker.rs
```diff
@@ -583,18 +583,15 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
...
-            let tag = match src_info {
-                Some(info) => Tag::RetAt(info.span),
-                None => Tag::Ret,
-            };
+            let tag = Tag::RetAt(src_info.span);
             self.phase
                 .constr_gen(self.genv, &rcx, tag)
                 .check_ret(&mut rcx, &mut env, &self.output)
-                .map_err(|err| err.with_src_info_opt(src_info))
+                .map_err(|err| err.with_src_info_opt(Some(src_info)))
```
nilehmann, line 594: I think we have a `with_src_info`
```suggestion
                .map_err(|err| err.with_src_info(src_info))
```

flux-refineck/src/constraint_gen.rs
```diff
@@ -54,7 +54,8 @@ pub enum Tag {
     Rem(Span),
     Goto(Option<Span>, BasicBlock),
     Overflow(Span),
-    Other,
+    Other(Span),
+    Invariant(Span),
```
nilehmann, line 58: I realized we already have `Tag::Fold` so we don't this extra case. I'd make all the cases you marked with `Tag::Invariant` a `Tag::Fold`.  If you want you can change the message for `Tag::Fold` to say something about an invariant instead of fold. I see how saying fold in the error can be confusing for the user, but I would still like to refer to it as a folding error internally.

flux-errors/locales/en-US/refineck.ftl
```diff
@@ -28,6 +28,9 @@ refineck_fold_error =
 refineck_unknown_error =
     unknown error
```
nilehmann, line 29: while we are at it, maybe we should change this to something less intimidating. Something like "the safety of this statement cannot be proven.

nilehmann (approved):
Some suggestions, but otherwise it looks good. I like that we have less `Option<Span>` now. 

I think I'll  revamp the whole `Tag` infrastructure. The Checker doesn't always have the full context to provide a reason, only the span associated with it, so the tag should be generated further down at the moment you actually generate a constraint.

flux-errors/locales/en-US/refineck.ftl
nilehmann: Btw, I like `Tag::Other` for prototyping, but eventually, everything should be moved out of it. But no reason to completely throw off users while we transition.

flux-refineck/src/constraint_gen.rs
ranjitjhala: ah yes, ok am reverting to `Fold` but changing the message for `Fold`.

PR #340, opened 2023-01-18
Const Eval

### LINKED ISSUE #338: Const bug
There is something odd going on with the translation of `const` into the fixpoint query.

The following code:

```rust

#![feature(register_tool)]
#![register_tool(flux)]

#[flux::constant]
const INITIAL_CAPACITY: usize = 7; // 2^3 - 1

#[flux::constant]
const MAXIMUM_ZST_CAPACITY: usize = 1 << (usize::BITS - 1); // Largest possible power of two

#[flux::sig(fn (n: usize{n < MAXIMUM_ZST_CAPACITY}))]
fn foo(n: usize) {}

pub fn bar() {
    foo(INITIAL_CAPACITY)
}
```

generates the smt2 file

```smt2
(constraint
  (forall ((_ Unit) (a1 = -9223372036854775808))
    (forall ((_ Unit) (a0 = 7))
      (tag (7 < a1) "0")
    )
  )
)
```

(which is rejected of course...)

ranjitjhala:
Fixes #338

### DIFF
--- a/flux-middle/src/const_eval.rs
+++ b/flux-middle/src/const_eval.rs
@@ -1,4 +1,3 @@
-use flux_fixpoint::Sign::{Negative, Positive};
 use rustc_middle::ty::{self as rustc_ty, ParamEnv, TyCtxt, TyKind};
 
 use crate::rustc::mir::Constant;
@@ -10,7 +9,7 @@ pub fn scalar_int_to_rty_constant<'tcx>(
     match ty.kind() {
         TyKind::Int(_) => {
             let i = scalar_to_int(tcx, scalar, ty)?;
-            crate::rty::Constant::from(i)
+            Some(crate::rty::Constant::from(i))
         }
         TyKind::Uint(_) => {
             let u = scalar_to_uint(tcx, scalar, ty)?;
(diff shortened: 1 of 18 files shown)


flux-middle/src/const_eval.rs
```diff
@@ -0,0 +1,87 @@
...
+    ty: rustc_middle::ty::Ty<'tcx>,
+) -> Option<crate::rty::Constant> {
+    match ty.kind() {
+        TyKind::Int(_) => {
+            let i = scalar_to_int(tcx, scalar, ty)?;
+            if 0 <= i {
+                Some(crate::rty::Constant::Int(Positive, i as u128))
+            } else {
+                Some(crate::rty::Constant::Int(Negative, (0 - i) as u128))
+            }
```
nilehmann, line 17: Use `Constant::from(i)`
```suggestion
            Constant::from(i)
```

flux-middle/src/const_eval.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+        TyKind::Uint(_) => {
+            let u = scalar_to_uint(tcx, scalar, ty)?;
+            Some(crate::rty::Constant::Int(Positive, u))
```
nilehmann, line 21:
```suggestion
            Some(crate::rty::Constant::from(u))
```

nilehmann (approved):
There's one small bug, but otherwise looks good.

flux-middle/src/const_eval.rs
nilehmann: ups, this was supposed to be part of the review.

PR #342, opened 2023-01-19
Add _very_ basic support for `Ptr`

ranjitjhala:
Essentially treat `Ptr` as opaque base type with no refinements.

### DIFF
--- a/flux-middle/src/rty/mod.rs
+++ b/flux-middle/src/rty/mod.rs
@@ -20,7 +20,7 @@ use itertools::Itertools;
 use rustc_hash::FxHashMap;
 use rustc_hir::def_id::DefId;
 use rustc_index::{bit_set::BitSet, newtype_index};
-use rustc_middle::mir::Field;
+use rustc_middle::mir::{Field, Mutability};
 pub use rustc_middle::ty::{AdtFlags, FloatTy, IntTy, ParamTy, ScalarInt, UintTy};
 use rustc_span::Symbol;
 pub use rustc_target::abi::VariantIdx;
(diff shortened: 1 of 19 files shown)


flux-middle/src/rty/mod.rs
```diff
@@ -194,6 +194,7 @@ pub enum BaseTy {
     Slice(Ty),
     Adt(AdtDef, Substs),
     Float(FloatTy),
+    Ptr(Ty, Mutability),
```
nilehmann, line 197: We should call this `RawPtr` to differentiate from `rty::TyKind::Ptr`

```suggestion
    RawPtr(Ty, Mutability),
```

flux-middle/src/rustc/ty.rs
```diff
@@ -95,6 +95,7 @@ pub enum TyKind {
     Tuple(List<Ty>),
     Uint(UintTy),
     Slice(Ty),
+    Ptr(Ty, Mutability),
```
nilehmann, line 98: Same here, if it's called a `RawPtr` in rustc we should also call it a `RawPtr` here
```suggestion
    RawPtr(Ty, Mutability),
```

flux-middle/src/rustc/ty.rs
```diff
@@ -171,6 +172,10 @@ impl Ty {
         TyKind::Slice(ty).intern()
     }
 
+    pub fn mk_ptr(ty: Ty, mutbl: Mutability) -> Ty {
```
nilehmann, line 175:
```suggestion
    pub fn mk_raw_ptr(ty: Ty, mutbl: Mutability) -> Ty {
```

nilehmann (approved):
I think this is the right direction. Just some notes about names.

 I like that `RawPtr` is a `BaseTy` such that eventually, we make it refined by an opaque sort you can attach uninterpreted functions to. That should allow us to encode the stuff for ffi.

PR #343, opened 2023-01-21
More on (Raw) Pointers

ranjitjhala:
casts and borrows -- please take a very close look at this @nilehmann ...


flux-tests/tests/pos/surface/ptr.rs
```diff
@@ -3,15 +3,23 @@
...
+pub fn get_mut(vec: &mut Vec<i32>, off: usize) -> Option<&mut i32> {
+    unsafe { Some(&mut *vec.as_mut_ptr().add(off)) }
+}
+
+pub unsafe fn copy_nonoverlapping(vec: &mut Vec<i32>, dst: usize, src: usize, len: usize) {
+    unsafe {
+        ptr::copy_nonoverlapping(vec.as_mut_ptr().add(src), vec.as_mut_ptr().add(dst), len);
+    }
+}
+
```
nilehmann, line 25: Let's create a file with simple tests something like:

```rust
// MutToConstPointer
unsafe fn test00(p: *mut i32) {
    let x = p as *const i32
}

// Read from a raw pointer
unsafe fn test01(p: *[mut|const] i32) -> i32 {
     *p
}

// Pointer subtyping
unsafe fn test01(p: *[mut|const] i32) -> *[mut|const] i32 {
     p
}

// Borrow from a pointer
unsafe fn test01(p: *[mut|const] i32) -> &[mut] i32 {
     &mut *p
}
```

nilehmann (changes requested):

flux-refineck/src/type_env.rs
```diff
@@ -119,6 +119,7 @@ impl TypeEnv {
                 debug_assert!(WeakKind::from(rk) <= result_rk);
                 Ty::mk_ref(rk, ty)
             }
+            FoldResult::Raw(ty) => Ty::mk_ref(rk, ty), // TODO(RJ): is this legit?
```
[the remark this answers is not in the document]
ranjitjhala: ok, so I'm replacing `self.refine_ty(ty, mk_pred)` with `self.refine_ty_true(ty)` , yes?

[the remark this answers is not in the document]
nilehmann: yes, exactly.

flux-tests/tests/pos/surface/ptr.rs
ranjitjhala: Ok! I added a new test file with the various legal (?) combinations

flux-refineck/src/checker.rs
```diff
@@ -794,7 +794,7 @@ impl<'a, 'tcx, P: Phase> Checker<'a, 'tcx, P> {
                     }
                 }
             }
-            CastKind::FloatToInt | CastKind::IntToFloat => {
+            CastKind::FloatToInt | CastKind::IntToFloat | CastKind::Pointer(_) => {
```
nilehmann, line 797: Better to be explicit here to avoid problems in the future

```suggestion
            CastKind::FloatToInt | CastKind::IntToFloat | CastKind::Pointer(PointerCast::MutToConstPointer) => {
```

nilehmann (approved):
Looks good just a final comment.

PR #345, opened 2023-01-23
Support loading flux annotations from external crates

nilehmann:
This PR adds support for loading refinement types for dependencies that were annotated with flux. Note that this doesn't allow us to annotate external dependencies for which we don't have control (e.g., `std`), you can only annotate crates you created and then use them as dependencies for other crates.

## Strategy

The strategy is very simple. There are two sides to it

### Generate metadata
If we detect `flux` is called with the flag `--emit=metadata`, which is the standard flag `cargo` uses to instruct `rustc` to generate a metadata file, then we will honor it and dump our own file with the necessary metadata for cross-crate refinement checking. We put the file in the same path rustc would put its metadata but using the extension `fluxmeta` instead of `rmeta`.

### Read metadata for dependencies
When `flux` is called with the flag `--extern crate_name=/path/to/lib.rmeta`, which is the standard flag `cargo` uses to instruct `rustc` to read the metadata for a dependency, then we will honor it and read the flux metadata from the same path but replacing the extension `rmeta` with `fluxmeta`.

## Limitations

Opaque sort declarations, flux defns (uninterpreted and interpreted), constants, and qualifiers cannot be used cross-crate. Those are currently identified with a string so further work needs to be done to disambiguate them between crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -563,6 +563,7 @@ version = "0.1.0"
 dependencies = [
  "flux-common",
  "flux-errors",
+ "flux-macros",
  "flux-middle",
  "itertools",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -564,6 +564,7 @@ dependencies = [
  "flux-common",
  "flux-errors",
  "flux-middle",
+ "itertools",
 ]
 
 [[package]]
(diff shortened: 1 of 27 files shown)


ranjitjhala:
May be good to add the text for the PR (ie "Strategy" and "Limitations") somewhere in the docs?

PR #348, opened 2023-01-25
Support for rust aliases

nilehmann:
tldr, we now accept this

```rust
type MyResult<T> = Result<T, i32>;

#[flux::sig(fn() -> MyResult<i32{v: v > 0}>)]
fn test01() -> MyResult<i32> {
    Ok(1)
}
```


ranjitjhala:
Hallelujah 🎊

Issue #323, opened 2023-01-12
Separate namespaces for constants and uifs


ranjitjhala:
**Name clash in qualifiers**: we could make constants leave in a different _namespace_, i.e., define a new `newtype_index!` for them. And while we are at it maybe we should unify constants and uif. Just realized that we could also have name clashes with uifs too because they can have an arbitrary name so if someone names on `a0`, `a1`, ... we will have a clash.

_Originally posted by @nilehmann in https://github.com/liquid-rust/flux/pull/322#pullrequestreview-1246038740_

cole-k:
Resolved by https://github.com/liquid-rust/flux/pull/380

PR #380, opened 2023-02-21
Separate namespaces for constants and uifs

cole-k:
# Before

* Uif: `c{n}`
* Constant: `a{n}`
* Variable: `a{n}`

# After

* Uifs: `c{n}`
* Constant: `c{n}`
* Variable `a{n}`

### DIFF
--- a/flux-fixpoint/src/lib.rs
+++ b/flux-fixpoint/src/lib.rs
@@ -16,7 +16,7 @@ use std::{
 };
 
 pub use constraint::{
-    BinOp, Const, Constant, Constraint, Expr, Func, FuncSort, KVid, Name, ConstName, Pred, Proj,
+    BinOp, Const, ConstName, Constant, Constraint, Expr, Func, FuncSort, KVid, Name, Pred, Proj,
     Qualifier, Sign, Sort, UifDef, UnOp,
 };
 use flux_common::{cache::QueryCache, format::PadAdapter};
--- a/flux-fixpoint/src/lib.rs
+++ b/flux-fixpoint/src/lib.rs
@@ -167,7 +167,7 @@ impl<Tag: fmt::Display> fmt::Display for Task<Tag> {
         writeln!(f, "(data Unit 0 = [| Unit {{ }}])")?;
 
         for (name, sort) in &self.constants {
-            write!(f, "(constant {name:?} {sort:?})")?;
+            write!(f, "(constant $c{name:?} {sort:?})")?;
         }
         for uif_def in &self.uifs {
             writeln!(f, "{uif_def}")?;
--- a/flux-fixpoint/src/lib.rs
+++ b/flux-fixpoint/src/lib.rs
@@ -172,7 +172,7 @@ impl<Tag: fmt::Display> fmt::Display for Task<Tag> {
         writeln!(f, "(data Unit 0 = [| Unit {{ }}])")?;
 
         for cinfo in &self.constants {
-            write!(f, "{cinfo}")?;
+            writeln!(f, "{cinfo}")?;
         }
 
         for kvar in &self.kvars {
--- a/flux-fixpoint/src/lib.rs
+++ b/flux-fixpoint/src/lib.rs
@@ -16,8 +16,8 @@ use std::{
 };
 
 pub use constraint::{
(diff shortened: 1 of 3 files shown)


nilehmann (changes requested):
Looks good, but the idea was to unify `uif`s and constants. If you see, both [uifs](https://github.com/liquid-rust/flux/blob/014c5c511fa69497449d2ecace40205ec376fc2e/flux-fixpoint/src/lib.rs#L203) and [constants](https://github.com/liquid-rust/flux/blob/main/flux-fixpoint/src/lib.rs#L170)  are encoded the same way in fixpoint, so it makes sense to keep them together. I think we can remove [`UifDef`](https://github.com/liquid-rust/flux/blob/main/flux-fixpoint/src/constraint.rs#L75) and pass everything in the same vector to `Task`.

cole-k:
> Looks good, but the idea was to unify `uif`s and constants. If you see, both [uifs](https://github.com/liquid-rust/flux/blob/014c5c511fa69497449d2ecace40205ec376fc2e/flux-fixpoint/src/lib.rs#L203) and [constants](https://github.com/liquid-rust/flux/blob/main/flux-fixpoint/src/lib.rs#L170) are encoded the same way in fixpoint, so it makes sense to keep them together. I think we can remove [`UifDef`](https://github.com/liquid-rust/flux/blob/main/flux-fixpoint/src/constraint.rs#L75) and pass everything in the same vector to `Task`.

Done.

flux-refineck/src/fixpoint.rs
```diff
@@ -388,17 +402,30 @@ impl<'a> KVarGen for Box<dyn KVarGen + 'a> {
...
+            let name = const_name_gen.fresh();
+            let cinfo = ConstInfo {
+                name,
+                sym: const_info.sym,
+                sort: fixpoint::Sort::Int,
+                val: Some(const_info.val),
+            };
+            (Key::Const(const_info.def_id), cinfo)
+        });
+    let uifs = genv.uifs().map(|uif_def| {
```
nilehmann, line 421: We should sort `uifs` here the same way consts are being sorted. We use the resulting list of fixpoint consts as part of the hash for a `Task` which we use for caching, so we want to have a consistent ordering here.

flux-refineck/src/fixpoint.rs
```diff
@@ -174,11 +185,15 @@ where
...
             .const_map
             .values()
-            .map(|const_info| (const_info.name, fixpoint::Sort::Int))
+            .map(|const_info| {
+                fixpoint::ConstInfo {
+                    name: const_info.name,
+                    orig: const_info.sym,
+                    sort: const_info.sort.clone(),
+                }
+            })
```
nilehmann, line 194: Use `into_values` instead of `values` to avoid cloning.

```suggestion
            .into_values()
            .map(|const_info| {
                fixpoint::ConstInfo {
                    name: const_info.name,
                    orig: const_info.sym,
                    sort: const_info.sort,
                }
            })
```

flux-fixpoint/src/lib.rs
```diff
@@ -198,9 +200,9 @@ impl fmt::Display for KVar {
     }
 }
 
-impl fmt::Display for UifDef {
+impl fmt::Display for ConstInfo {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "(constant {} {})", self.name, self.sort)
+        write!(f, "(constant {0:?} {1:?})", self.name, self.sort)
```
nilehmann, line 205:
```suggestion
        write!(f, "(constant {:?} {:?})", self.name, self.sort)
```

nilehmann: Actually, I just thought we can add a comment here to know what's the origin the same way we do for kvars.

```suggestion
        write!(f, "(constant {:?} {:?}) // orig: {}", self.name, self.sort, self.orig)
```

nilehmann (approved):
@cole-k looks good. Mostly minor suggestions. The only important one is that we should sort uifs the same way we sort `ConstDefId`s.

Feel free to merge after addressing my comments.

cole-k:
Resolved #323

PR #381, opened 2023-02-24
Refined type variables

nilehmann:
This PR implements initial support for *refined type variables*. The design is a combination of the proposals outlined in [System F(lux) v2](https://hackmd.io/sSer5Q-uTeSd-K3JElWtlQ?here) and [System F(flux) v3](https://hackmd.io/O7Ywiy2ORq6OfrpvMSjTOQ?view). Like v2, type variables can be of kind `base`  or `type`. Following v3, when instantiating a type variable of kind `base`, we can pass a *type constructor*.

## Instantiation of type variables of kind `base`

Consider the following definition
```rust
fn id<T: base>[x: sortof(T)](T[x]) -> ∃ν:sortof(T). T[ν] { x }
```
That is, `id` is polymorphic on a type variable `T` of kind `base` and a refinement variable `x` of sort `sortof(T)`, where `sortof(T)` is just notation for the sort of a type variable of kind `base`.

Suppose we call `id` with an argument of type `i32`. Since indices for `i32` must be of sort `int` we have to instantiate the type variable with a type constructor taking values of sort `int`, i.e., a lambda of kind `int -> type`. We do it as follows by generating a fresh kvar `κ`[^1]
```rust
id<λν:int. {i32[ν] ∧ κ(ν)}>
```
This gives us the following definition after substituting 
```rust
fn id[x: int]({ i32[x] | κ(x) }) -> ∃ν:int. { i32[ν] | κ(ν) }
```
Note, that `sortof(T)` got replaced by `int` and `T` by the result of applying the lambda to `x`.

## Picking the right kind

The kind of a type variable is automatically inferred from its usage and there's no surface syntax to specify it explicitly. The strategy to pick the kind of a type variable is as follows

* A type variable used in a struct/enum definition is always of kind `type`
* A type variable used in a function is always of kind `base`

For example, in the following definition the kind of `T` is `type`

```rust
struct S1<T> {
    f: T,
}
```

Conversely, in the following definition the kind of `T` is `base`

```rust
fn id<T>(x: T) -> T { x }
```

Thus, they desugar to something like

```rust
struct S1<T: type> {
    f: T,
}

fn id<T: base>[x: sortof(T)](T[x]) -> ∃ν:sortof(T). T[ν] { x }
```

The rationale behind the strategy is that users never provide the instantiation for a function type variable explicitly, but they do it for structs/enums, i.e., a user can write `S1<i32{v: v > 0}>`. This means that for functions we can hide the detail that the instantiation must be a type constructor.

In practice, this means that users may get an error when defining a struct/enum if they try to refine a type variable [^2]

```rust
struct S2<T> {
    f: T{v: v > 0} //~ ERROR type variable cannot be refined
}
```

When writing a function you never get an error because a type variable of kind `base` can always be "lifted" to a variable of kind `type` by wrapping it in an existential. For example, the following definition
```rust
fn test<T>(t: T, s: S1<T>) {}
```
desugars to
```rust
fn test<T: base>[t: sortof(T)](T[t], S1<∃ν. T[ν]>) { }
```
Thus, we can transparently use `T` of kind `base` as an argument to `S1` by lifting `T` to the type `∃ν:sortof(T). T[ν]`.

## Examples

There's not a whole lot you can do with a refined type variable because a value of sort `sortof(T)` cannot be meaningfully used in refinements. You can basically just "move it around". The most important use case is that we can specify `swap` as follows (this was the main reason I wanted to implement refined type variables)
```rust
fn swap<T>(x: &strg T[@n], y: &strg T[@m]) 
ensures x: T[m], y: T[n]
```
See [this test](https://github.com/liquid-rust/flux/blob/8e4481785c46528bb4fe9d41e9a099f91bbc75e4/flux-tests/tests/pos/surface/refined_type_var00.rs) for the full example.

## Alternative

I explored the idea of giving kind `base` to all type variables and then automatically converting surface types into type constructors when used as arguments for structs/enums, e.g., we can convert `S1<i32{v: v > 0}>` into `S1<λν:int. {i32[ν] ∧ ν > 0}>` and `S1<i32[0]>` into `S1<λν:int. {i32[ν] ∧ ν == 0}>`.  This turned out to be problematic because we need to be able to compute the subtyping and the "join" between pairs of types of the form `S1<λv:σ₁. T1>` `S1<λv:σ₂. T2>`.  We can only do this if `T1` and `T2` are syntactically restricted and these restrictions are still unclear to me. The problem is exacerbated because when a user writes `S1<T>`, the `T` can contain type aliases and type aliases may expand to exotic types like `∃a,b.(i32[a], {i32[b] | b > a})`

## Future extensions

To make refined type variables more useful we can come up with a way to specify "sort bounds". I'm thinking something like

```rust
fn choose<T>(x: T, y: T) -> T{v: v == x && v == y}
requires HasEq(sortof(T))

fn max<T>(x: T, y: T) -> T[if x > y  { x } else { y } ]
requires HasCmp(sortof(T))
```

Note that this is a restriction on the sort associated to `T` and not `T` itself. This is why I'm choosing the syntax `HasEq` instead of something rustier like `sortof(T): Eq` because this is not directly related to the `Eq` trait. However, we could maybe close the loop by requiring the `Eq` trait to be implemented on types `T` for which `HasEq(sortof(T))` holds. Then we should be able to infer `HasEq(sortof(T))` from `T: Eq`.

[^1]: Remember that `{t ∧ p}` is syntax for a constrained type, i.e., values of type `t` and also `p` holds.
[^2]: This is equivalent to the restriction in sprite that a type variable of kind `*` can only be used as `T{v: true}`


ranjitjhala:
@nilehmann I think for PRs for significant features like this, instead of (only) leaving this documentation in the PR text, we should maybe also dump it somewhere inside `flux-docs` (some `design.md` or similar to be easily searchable with the rustbook stuff?)

PR #383, opened 2023-03-01
Add a flag for overflow check

cole-k:
### `FLUX_CHECK_OVERFLOW=0`

Behavior as previously (only check underflow on subtraction of unsigned integers).

### `FLUX_CHECK_OVERFLOW=1`

Checks overflow and underflow on arithmetic (plus/minus/times) operations for signed and unsigned integers.

### Tests

I added tests but left them commented because we can't configure the flag on a per-test level (see https://github.com/cole-k/flux-develop/blob/b8339c091f7aadf85306c843f95322daa5e9f3b0/flux-driver/src/collector.rs#L657-L663)

### DIFF
--- a/flux-config/src/lib.rs
+++ b/flux-config/src/lib.rs
@@ -129,7 +129,7 @@ static CONFIG: LazyLock<Config> = LazyLock::new(|| {
             .set_default("check_def", "")?
             .set_default("cache", false)?
             .set_default("cache_file", "cache.json")?
-            .set_default("check_overflow", true)?;
+            .set_default("check_overflow", false)?;
         // Config comes first, enviroment settings override it.
         if let Some(config_path) = CONFIG_PATH.as_ref() {
             config_builder = config_builder.add_source(File::from(config_path.to_path_buf()));
--- a/flux-config/src/lib.rs
+++ b/flux-config/src/lib.rs
@@ -57,6 +57,10 @@ pub fn driver_path() -> Option<&'static PathBuf> {
     CONFIG.driver_path.as_ref()
 }
 
+pub fn check_overflow() -> bool {
+    CONFIG.check_overflow
+}
+
 #[derive(Debug)]
 pub struct CrateConfig {
     pub log_dir: PathBuf,
@@ -78,6 +82,7 @@ struct Config {
     check_def: String,
     cache: bool,
     cache_file: String,
+    check_overflow: bool,
 }
 
 #[derive(Copy, Clone, Deserialize)]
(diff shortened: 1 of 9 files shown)


flux-config/src/lib.rs
```diff
@@ -78,6 +82,7 @@ struct Config {
     check_def: String,
     cache: bool,
     cache_file: String,
+    check_overflow: bool,
```
nilehmann, line 85: The option in `rustc` to enable disable overflow checks is `-Coverflow-checks`. Not sure if we should be consistent with that and call the flag `FLUX_OVERFLOW_CHECKS`. I think I generally prefer `check overflows` as if you are telling the tool `do (not) check overflows`, but maybe being consistent is better. We can always change it later though, so no biggie.

flux-refineck/src/sigs.rs
```diff
@@ -159,6 +218,59 @@ fn mk_signed_bin_ops() -> impl Iterator<Item = (mir::BinOp, Sig<2>)> {
...
+                let Int = BaseTy::Int(int_ty);
+            }
+            let bit_width: u128 = int_ty.bit_width().unwrap_or(flux_config::pointer_width().bits()).into();
+            [
+                // ARITH
+                (Add, s!(fn(a: Int, b: Int) -> Int[a + b]
+                            requires E::and(
+                                [E::le(&a + &b, int_max(bit_width)),
+                                E::ge(a + b, int_min(bit_width))]
+                            ) => ConstrReason::Overflow)
```
nilehmann, line 238: Nit: I would format it like so

```suggestion
                (Add, s!(fn(a: Int, b: Int) -> Int[a + b]
                            requires E::and([
                                         E::le(&a + &b, int_max(bit_width)),
                                         E::ge(a + b, int_min(bit_width)),
                                     ]) => ConstrReason::Overflow)
```

nilehmann (approved):
Looks good, just some stylistic suggestions.

We still need to add the corresponding invariants [here](https://github.com/flux-rs/flux/blob/00d2aa92c8c6a77bc57ac5211537bcf651940ffd/flux-middle/src/rty/mod.rs#L795). That can come in a different PR though

flux-config/src/lib.rs
cole-k: We're already diverging with Rust with the way we handle overflow checking, right? I agree that our name is more consistent (both with intuition and also the naming convention of our flags).

flux-fixpoint/src/constraint.rs
```diff
@@ -632,6 +632,18 @@ impl Constant {
         let n2 = other.to_int()?;
         Some(Constant::Bool(n1 >= n2))
     }
+
+    pub fn int_min(bit_width: u128) -> Constant {
```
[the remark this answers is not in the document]
cole-k: Switched to `u32` due to requirements from exponentiation (also it turns out `^` is XOR so I'm glad you suggested this...). We downcast the `u64` unsafely but I think I'll trust that `bit_width` won't return anything actually bigger than `u32`.

PR #394, opened 2023-03-13
Support extern specs for functions

cole-k:
## Current work

- [x] Macro to generate dummy functions from stubs (like Prusti's)
- [ ] WONTDO (Nico and I talked about this) Hack to make macro available in `(rustc|cargo)-flux`
- [x] Replace the `DefId` of a function marked `extern` with that of a dummy function with an assumed spec
- [x] Address review comments (most notably move the macro to a new crate)

## Overview

Here's my understanding of how extern specs are implemented in Prusti, which I am using as a guide for our own implementation.

### Invocation stage

[Prusti adds the compilation flag `--extern prusti_contracts=path/to/libprusti_contracts.rlib`](https://github.com/viperproject/prusti-dev/blob/d1d4dccb179e8be4769045cd3956e1d98be69c52/prusti-launch/src/bin/prusti-rustc.rs#L85-L105) so that macros are available whenever you check with it.

### Macro stage

The user writes code like

```rust
use prusti_contracts::*;

#[extern_spec(std::mem)]
#[ensures(*a == old(*b) && *b == old(*a))]
pub fn swap(a: &mut i32, b: &mut i32);
```

The `extern_spec` macro expands it to something like

```rust
#[prusti::extern_spec]
#[ensures(*a == old(*b) && *b == old(*a))]
pub fn prusti_extern_spec_swap(a: &mut i32, b: &mut i32) {
    std::mem::swap(a, b)
}
```

### Type checking stage

When Prusti sees the `prusti::extern_spec` annotation, it traverses the body of its associated function until it finds a function call expression. It then extracts the `DefId` of the function (in this case `std::mem::swap`). It then registers this `DefId` as having the same spec as `prusti_extern_spec_swap`.

### Why does Prusti do this?

This is very janky, but after discussing with Nico and thinking about it myself it unfortunately seems like the easiest solution.

There are several explanations for the jank:

* Passing an `--extern` flag to `cargo`/`rustc` makes it so that you can check things without introducing an additional dep, most notably standalone files.
* Going through the whole rigamarole of having a macro expand the spec is the best way to get typing information about the external function. There doesn't seem any particularly good way to take a string like `std::mem::swap` and resolve it to the right `DefId` without running the whole compiler.*
* However, there doesn't seem to be a straightforward way of writing out a path and telling Rust "hey, get the `DefId` of this but otherwise ignore it." So that's why they make a whole dummy function and traverse it to extract the path.
* Finally, there are some advantages to having a dummy function, most notably that they can just pretend as if the external spec was a trusted local definition - this prevents some duplication of logic.

\* I'd be interested in seeing if it's possible to generate a new top-level AST node to extend the `TyCtx` with and run the typechecking/namespace resolution algorithm on it. This essentially would do the macro expansion as part of the flux driver callback. However I don't have high hopes for this method since I'm guessing that flux is operating after these resolutions have finished.

## What I've added

See [Current work](#current-work) for the checklist.

[See the tests for examples of the syntax](https://github.com/flux-rs/flux/pull/394/files#diff-cc7bc02cac0f0f5f5805f12a4778af6acabdbc02c5526e5f148cb459075e7b25).

### Invocation stage

`cargo-flux` passes the feature `flux-attrs-proc-macros/enabled` which pulls in the deps of `flux-attrs-proc-macros` (without this feature, the `extern_spec` proc macro is just a pass-through function and therefore does not require any dependency on `flux-attrs`).

We do _not_ pass the `--extern` flag. Nico and I talked about this and it's too much of a hack and only is beneficial for invocations of `rustc-flux`.

### Macro

Here's how it works. We start with an empty function definition.

```rust
use flux_macros::extern_spec;

#[extern_spec(std::mem)]
#[flux::sig(fn(&mut i32[@a], &mut i32{v : a < v }) -> ())]
fn swap(a: &mut i32, b: &mut i32);
```

The macro parses the `std::mem` into a module path and then transforms the function into

```rust
#[flux::extern_spec]
#[flux::sig(fn(&mut i32[@a], &mut i32{v : a < v }) -> ())]
#[allow(unused, dead_code)]
fn swap(a: &mut i32, b: &mut i32) {
    std::mem::swap(a, b)
}
```

Note that `extern_spec` and `flux::extern_spec` are not the same - the latter is an "internal" attribute that expects the function to be defined in a particular way.

### Type checking

I've added an attribute `#[flux::extern_spec]`. This attribute requires an associated `#[flux::sig]` attribute and that the dummy function it's an attribute of is in the exact form of the function Prusti's `extern_spec` macro generates. The attribute makes the associated dummy function trusted.

I added an `extern_fns` attribute to the various contexts. It maps the `DefId` of an external function to the `DefId` of its local dummy function. Now when we resolve a function `DefId` during type checking, we first check if it's an external spec, returning the associated dummy function's signature if it is.

The nice thing about having the dummy function is that well-formedness checks and identifier resolution are handled just like they would be for a normal function, so we don't need to do any extra work.

## Limitations

When we add support for `extern_spec` macros, users will have to import the macro from `flux_macros` unqualified. Perhaps there is a way to change the import so that they can call it `flux::extern_spec` - I'm not too sure about how name resolution works.

Extern specs are only implemented for functions right now.

## Acknowledgements

[Prusti](https://github.com/viperproject/prusti-dev/)'s handling of extern specs was instrumental in helping me figure out what code I needed to write. For every line of code I wrote, I probably read 10 lines of their code.

### DIFF
--- a/flux-attrs/src/lib.rs
+++ b/flux-attrs/src/lib.rs
@@ -1,0 +1,137 @@
+#![feature(proc_macro_diagnostic)]
+
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote_spanned};
+use syn::{
+    parse_quote_spanned, punctuated::Punctuated, spanned::Spanned, Expr, FnArg, GenericArgument,
+    GenericParam, Token,
+};
+
+pub fn extern_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
+    transform_extern_spec(attr, tokens).unwrap_or_else(|err| err.to_compile_error())
+}
+
+fn transform_extern_spec(attr: TokenStream, tokens: TokenStream) -> syn::Result<TokenStream> {
+    let mod_path: Option<syn::Path> =
+        if !attr.is_empty() { Some(syn::parse2(attr)?) } else { None };
+    let (sig, attrs) = extract_sig_from_stub(tokens)?;
+    create_dummy_fn(mod_path, sig, attrs)
+}
+
+/// Takes a function stub, i.e. something that looks like
+/// ```ignore
+///     pub fn swap(a: &mut i32, b: &mut i32) -> ();
+/// ```
+/// and extracts its signature and any attributes.
+///
+/// This function actually tries to parse the stub as a `syn::TraitItemMethod`
+/// and requires that it not have a default implementation.
+fn extract_sig_from_stub(
+    tokens: TokenStream,
+) -> syn::Result<(syn::Signature, Vec<syn::Attribute>)> {
+    let span = tokens.span();
+    match syn::parse2::<syn::TraitItemMethod>(tokens) {
+        // Parse must succeed and it must not have a default implementation (i.e. it must end with a ';')
+        Err(_) | Ok(syn::TraitItemMethod { default: Some(_), .. }) => Err(syn::Error::new(
+            span,
+            "Invalid function stub: extern specs expect a function stub like: fn swap(a: &mut i32, b: &mut i32);"
+        )),
+        Ok(trait_item_method) => {
+            Ok((trait_item_method.sig, trait_item_method.attrs))
+        }
+    }
+}
+
+/// Creates a dummy function whose inner body is just a pass-through call to the
+/// extern function. Takes the signature and attributes
+/// ```ignore
+/// swap(a: &mut i32, b: &mut i32) -> ()
+/// ```
+/// and produces a function that looks like.
+/// ```ignore
+///     pub fn flux_extern_spec_swap(a: &mut i32, b: &mut i32) -> () {
+///         swap(a,b)
+///     }
+/// ```
+///
+/// The big idea:
+/// 1. Mangle the name of the function into flux_extern_spec_{fn_ident} so that it doesn't
+///    conflict with an existing function. TODO: add some random number to ensure it.
+/// 2. Create a new function with the mangled name and same signature.
+/// 3. Replace its body with a single expression that is a function call.
+/// 4. The function call is of the form mod_path::fn_ident<generic_args>(params)
+///    where the generic args and params have been modified so that they function as
+///    arguments to the call instead of being a part of the function signature.
+/// 5. Finally, we suppress dead_code/unused warnings and add the flux::extern_spec
+///    so that flux knows this is the generated dummy function for the external function
+///    that is being called.
+fn create_dummy_fn(
+    mod_path: Option<syn::Path>,
+    sig: syn::Signature,
+    attrs: Vec<syn::Attribute>,
+) -> syn::Result<TokenStream> {
+    let span = sig.span();
+    let mut mangled_sig = sig.clone();
+    let ident = sig.ident;
+    mangled_sig.ident = format_ident!("flux_extern_spec_{}", ident);
+    let generic_call_args = transform_generic_params_to_call_args(sig.generics.params);
+    let call_args = transform_params_to_call_args(sig.inputs);
+    let fn_path: syn::Path = if let Some(mod_path) = mod_path {
+        parse_quote_spanned!( ident.span() => #mod_path :: #ident)
+    } else {
+        parse_quote_spanned!( ident.span() => #ident )
+    };
+    Ok(quote_spanned! { span =>
+                        #[flux::extern_spec]
+                        #(#attrs)*
+                        #[allow(unused, dead_code)]
+                        #mangled_sig {
+                            #fn_path :: < #generic_call_args > ( #call_args )
+                        }
+    })
+}
+// Cribbed from Prusti's extern_spec_rewriter
+fn transform_generic_params_to_call_args(
+    generic_params: Punctuated<GenericParam, Token!(,)>,
+) -> Punctuated<GenericArgument, Token!(,)> {
+    Punctuated::from_iter(
+        generic_params
+            .iter()
+            .flat_map(|param| -> Option<GenericArgument> {
+                let span = param.span();
+                match param {
+                    GenericParam::Type(syn::TypeParam { ident, .. }) => {
+                        Some(parse_quote_spanned! { span => #ident })
+                    }
+                    GenericParam::Lifetime(_) => None,
+                    GenericParam::Const(syn::ConstParam { ident, .. }) => {
+                        Some(parse_quote_spanned! {span => #ident })
+                    }
+                }
+            }),
+    )
+}
+
+// Cribbed from Prusti's extern_spec_rewriter
+fn transform_params_to_call_args(
+    params: Punctuated<FnArg, Token!(,)>,
+) -> Punctuated<Expr, Token!(,)> {
+    Punctuated::from_iter(params.iter().map(|param| -> Expr {
+        let span = param.span();
+        match param {
+            FnArg::Typed(pat_type) => {
+                match pat_type.pat.as_ref() {
+                    syn::Pat::Ident(ident) => {
+                        parse_quote_spanned! {span => #ident }
+                    }
+                    _ => {
+                        unimplemented!(
+                            "extern specs don't support patterns other than simple identifiers"
+                        )
+                    }
+                }
+            }
+            FnArg::Receiver(_) => parse_quote_spanned! {span => self},
+        }
+    }))
+}
--- a/flux-attrs/src/lib.rs
+++ b/flux-attrs/src/lib.rs
@@ -90,6 +90,7 @@ fn create_dummy_fn(
                         }
     })
 }
+
 // Cribbed from Prusti's extern_spec_rewriter
 fn transform_generic_params_to_call_args(
     generic_params: Punctuated<GenericParam, Token!(,)>,
--- a/flux-driver/src/callbacks.rs
+++ b/flux-driver/src/callbacks.rs
@@ -378,6 +378,13 @@ fn build_fhir_map(early_cx: &mut EarlyCtxt, specs: &mut Specs) -> Result<(), Err
         .err()
         .or(err);
 
+    // Extern Fns
+    std::mem::take(&mut specs.extern_fns)
+        .into_iter()
+        .for_each(|(extern_def_id, local_def_id)| {
+            early_cx.map.insert_extern_fn(extern_def_id, local_def_id);
+        });
+
     if let Some(err) = err {
         Err(err)
     } else {
--- a/flux-driver/src/collector.rs
+++ b/flux-driver/src/collector.rs
@@ -14,7 +14,10 @@ use rustc_ast::{
 };
 use rustc_errors::{ErrorGuaranteed, IntoDiagnostic};
 use rustc_hash::{FxHashMap, FxHashSet};
-use rustc_hir::{def_id::LocalDefId, EnumDef, ImplItemKind, Item, ItemKind, VariantData};
+use rustc_hir::{
+    def_id::{DefId, LocalDefId},
+    EnumDef, ImplItemKind, Item, ItemKind, VariantData,
+};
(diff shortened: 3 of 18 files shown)


flux-driver/src/collector.rs
```diff
@@ -358,11 +373,42 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
...
+                                    {
+                                        return Ok(def_id);
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
nilehmann, line 408: I'd use [if-let-chains](https://rust-lang.github.io/rfcs/2497-if-let-chains.html)

flux-driver/src/collector.rs
```diff
@@ -358,11 +373,42 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
...
+        def_id: LocalDefId,
+    ) -> Result<DefId, ErrorGuaranteed> {
+        use rustc_hir::{def, ExprKind, Node};
+        if let Node::Item(i) = self.tcx.hir().find_by_def_id(def_id).unwrap() {
+            if let ItemKind::Fn(_rust_fn_sig, _generics, body_id) = &i.kind {
+                if let Node::Expr(e) = self.tcx.hir().find(body_id.hir_id).unwrap() {
+                    if let ExprKind::Block(b, _) = e.kind {
+                        if let Some(e) = b.expr {
+                            if let ExprKind::Call(callee, _args) = e.kind {
+                                if let ExprKind::Path(ref qself) = callee.kind {
```
nilehmann, line 395: Ref patterns went out of fashion after [match-ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html)

```suggestion
                                if let ExprKind::Path(qself) = &callee.kind {
```

flux-driver/src/collector.rs
```diff
@@ -358,11 +373,42 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
...
 
+    // In Prusti they suggested looking into doing this instead of using a Visitor...
+    // it seems more brittle but I guess conversely their version is a little permissive.
+    fn extract_extern_def_id_from_extern_spec_fn(
+        &mut self,
+        def_id: LocalDefId,
+    ) -> Result<DefId, ErrorGuaranteed> {
+        use rustc_hir::{def, ExprKind, Node};
+        if let Node::Item(i) = self.tcx.hir().find_by_def_id(def_id).unwrap() {
+            if let ItemKind::Fn(_rust_fn_sig, _generics, body_id) = &i.kind {
```
nilehmann, line 390: I typically avoid defining names if we are not going to use them to avoid the extra noise.

```suggestion
            if let ItemKind::Fn(_, _, body_id) = &i.kind {
```

flux-middle/src/global_env.rs
```diff
@@ -45,6 +45,7 @@ pub struct GlobalEnv<'sess, 'tcx> {
     adt_variants: RefCell<VariantMap>,
     defns: Defns,
     queries: Queries,
+    extern_fns: RefCell<FxHashMap<DefId, DefId>>,
```
nilehmann, line 48: I don't think this is ever mutated, so we can drop the `RefCell`

```suggestion
    extern_fns: FxHashMap<DefId, DefId>,
```

flux-middle/src/global_env.rs
```diff
@@ -123,14 +131,21 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
...
 
     pub fn lookup_fn_sig(&self, def_id: DefId) -> Result<rty::PolySig, UnsupportedFnSig> {
-        match self.fn_sigs.borrow_mut().entry(def_id) {
+        // If it's an extern_fn, resolve it to its local fn_sig's def_id
+        let did =
+            if let hash_map::Entry::Occupied(entry) = self.extern_fns.borrow_mut().entry(def_id) {
+                *entry.get()
+            } else {
+                def_id
+            };
```
nilehmann, line 140: It's an overkill to use the entry API here a simple `get` would do

```suggestion
        let did = *self.extern_fns.borrow_mut().get(&def_id).unwrap_or(&def_id);
```

flux-middle/src/global_env.rs
```diff
@@ -123,14 +131,21 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
     }
 
     pub fn lookup_fn_sig(&self, def_id: DefId) -> Result<rty::PolySig, UnsupportedFnSig> {
-        match self.fn_sigs.borrow_mut().entry(def_id) {
+        // If it's an extern_fn, resolve it to its local fn_sig's def_id
```
nilehmann, line 134: This is great! At some point I thought we would need to change everything in the `fhir::Map` to use `DefId` instead of `LocalDefId` but keeping track of just this mapping is brilliant

flux-driver/src/callbacks.rs
```diff
@@ -411,6 +411,15 @@ fn build_fhir_map(early_cx: &mut EarlyCtxt, specs: &mut Specs) -> Result<(), Err
         .err()
         .or(err);
 
+    err = std::mem::take(&mut specs.extern_fns)
+        .into_iter()
+        .try_for_each_exhaust(|(extern_def_id, local_def_id)| {
+            early_cx.map.insert_extern_fn(extern_def_id, local_def_id);
+            Ok(())
+        })
```
nilehmann, line 419: This doesn't fail so you can use a normal `for_each`

flux-macros/src/lib.rs
```diff
@@ -46,6 +48,28 @@ decl_derive!(
...
+//     match item {
+//         syn::Item::Verbatim(stub_tokens) => {
+//
+//         }
+//         _ => {
+//             Err(syn::Error::new(Span::call_site(), "Extern specs only may be used on function stubs"))
+//         }
+//     }
+//
+// }
```
nilehmann, line 75: We need to create a new macro crate. This one is for internal development and it shouldn't be exposed to users.

nilehmann: Also, this new macro crate should eventually be published in crates.io to make it easy to use with `cargo-flux`

flux-middle/src/global_env.rs
cole-k: The `get` requires a mutable borrow unless I'm mistaken. I did have it like that at first.

nilehmann: You are mistaken, the entry api does but you shouldn't be using the entry api.

cole-k:
Added the `extern_spec` as a macro in `flux-attrs-proc-macros` and tested it locally - testing in our tests would require an overhaul of how we do tests and would be generally annoying so I have not done that yet.

flux-middle/src/queries.rs
```diff
@@ -200,15 +200,18 @@ impl<'tcx> Queries<'tcx> {
 
     pub(crate) fn fn_sig(&self, genv: &GlobalEnv, def_id: DefId) -> QueryResult<rty::PolySig> {
         run_with_cache(&self.fn_sig, def_id, || {
-            if let Some(local_id) = def_id.as_local() {
+            // If it's an extern_fn, resolve it to its local fn_sig's def_id
+            let did =
+                if let Some(entry) = genv.extern_fns().get(&def_id) { *entry } else { def_id };
```
nilehmann, line 205: maybe instead of exposing the entire `extern_fns` map we could only expose a function that tells you if the `DefId` is an extern spec

```rust
impl GlobalEnv {
   ...
   pub fn is_extern_spec(&self, def_id: DefId) -> Option<DefId> { ... }
 }
 ```
 Then  we could use it here like so
 
 ```rust
 let did = genv.is_extern_spec(def_id).unwrap_or_else(def_id);
```

flux-tests/tests/pos/surface/extern_spec_macro.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::mem::swap;
+
+#[extern_spec]
+#[flux::sig(fn(&mut i32[@a], &mut i32{v : a < v }) -> ())]
+fn swap(a: &mut i32, b: &mut i32);
+
+pub fn test() {
+  let mut x = 1;
+  let mut y = 2;
+  swap(&mut x, &mut y);
```
nilehmann, line 15: You should probably enable format on save. I've seen you pushing commits just to fix formatting issues

```suggestion
    let mut x = 1;
    let mut y = 2;
    swap(&mut x, &mut y);
```

flux-driver/src/collector.rs
```diff
@@ -358,11 +373,35 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
...
+    ) -> Result<DefId, ErrorGuaranteed> {
+        use rustc_hir::{def, ExprKind, Node};
+        if let Node::Item(i) = self.tcx.hir().find_by_def_id(def_id).unwrap()
+            && let ItemKind::Fn(_, _, body_id) = &i.kind
+            && let Node::Expr(e) = self.tcx.hir().find(body_id.hir_id).unwrap()
+            && let ExprKind::Block(b, _) = e.kind
+            && let Some(e) = b.expr
+            && let ExprKind::Call(callee, _) = e.kind
+            && let ExprKind::Path(ref qself) = callee.kind {
+                let typeck_result = self.tcx.typeck(def_id);
```
nilehmann, line 396: There's still no consensus on how rustfmt will format this but I've been formatting it like so to be consistent with how rustfmt formats multiline if conditions

```suggestion
            && let ExprKind::Path(ref qself) = callee.kind 
        {
                let typeck_result = self.tcx.typeck(def_id);
```

flux-attrs/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+                        #[flux::extern_spec]
+                        #(#attrs)*
+                        #[allow(unused, dead_code)]
+                        #mangled_sig {
+                            #fn_path :: < #generic_call_args > ( #call_args )
+                        }
+    })
+}
+// Cribbed from Prusti's extern_spec_rewriter
+fn transform_generic_params_to_call_args(
```
nilehmann, line 94:
```suggestion

// Cribbed from Prusti's extern_spec_rewriter
fn transform_generic_params_to_call_args(
```

flux-attrs/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    Punctuated::from_iter(
+        generic_params
+            .iter()
+            .flat_map(|param| -> Option<GenericArgument> {
+                let span = param.span();
+                match param {
+                    GenericParam::Type(syn::TypeParam { ident, .. }) => {
+                        Some(parse_quote_spanned! { span => #ident })
+                    }
+                    GenericParam::Lifetime(_) => None,
```
nilehmann, line 106: Why are we filtering out lifetimes here?

nilehmann (approved):
Awesome job @cole-k! Looks really good. Just minor comments.

flux-attrs/src/lib.rs
cole-k: We're transforming the function signature's arguments into arguments to pass to the external function. Do we need to explicitly indicate the lifetimes when they're call arguments?

nilehmann: ah makes sense 👍🏽

flux-tests/tests/pos/surface/extern_spec_macro.rs
cole-k: Not sure it would fix this one since I wrote this test on the command line 😛

flux-attrs/src/lib.rs
cole-k: Surprised neither formatter adds a newline

PR #399, opened 2023-03-16
Early Binder

nilehmann:
* Introduce `EarlyBinder` as in `rustc` to mark explicitly when generics need to be replaced.
* Introduce the notion of early bound refinement variables and implement parenthetical parameters for types aliases as early bound parameters.


flux-desugar/src/desugar.rs
nilehmann, line 1: Testing comment on file

PR #411, opened 2023-03-31
Enable proc macros in tests

nilehmann:
* Add dummy `flux-attrs-proc-macros-build` crate to pre-build the `flux_attrs-proc-macros` crate with the flux feature enabled
* Add a `--extern` flag in `compilete.rs` to make proc macros available in tests
* Fix extern specs tests to be valid refinements of functions
* Enable extern specs tests that use the `extern_spec` proc macro
* Fix the `extren_spec` proc macro to return an empty stream of tokens if the feature is disabled.

I need someone to fill this [`todo!`](https://github.com/flux-rs/flux/blob/e726bc3e657191124c0f0d0085fb260531f885f3/flux-tests/tests/compiletest.rs#L33) in mac. I believe the name of the shared library will be different.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,6 +508,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "flux-attrs-proc-macros-build"
+version = "0.1.0"
+dependencies = [
+ "flux-attrs-proc-macros",
+]
+
+[[package]]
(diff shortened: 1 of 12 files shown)


cole-k (approved):
👍 LGTM

PR #414, opened 2023-04-03
Syntax: use `|` instead of `:` for constraint types

nilehmann:
This is to make the distinction between an existential type (`i32{v: v > 0}`) and a constraint type (`{i32[n] | n > 0}`) a bit clearer.


ranjitjhala:
Looking at the diffs the `|` does look cleaner, but I can't tell if that's because I'm used to the `|` from LH...

PR #428, opened 2023-04-11
Enable overflow tests

nilehmann:
* Add config field to `Checker` to control overflow checking dynamically
* Add crate config option to enable overflow checking on a per-crate basis.
* Enable ignored overflow checks tests


ranjitjhala:
@nilehmann is this something that is user-configurable? If so would be great to update the `docs` on how to configure appropriately...

nilehmann:
good point, will do.

PR #435, opened 2023-04-25
Automatically scrape qualifiers from constraint bodies

ranjitjhala:
Depends on https://github.com/ucsd-progsys/liquid-fixpoint/pull/651

### DIFF
--- a/flux-tests/tests/neg/surface/scrape00.rs
+++ b/flux-tests/tests/neg/surface/scrape00.rs
@@ -1,5 +1,7 @@
 #![feature(register_tool)]
 #![register_tool(flux)]
+#![feature(custom_inner_attributes)]
+#![flux::cfg(scrape_quals = false)]
 
 // test that the (fixpoint) `--scrape` mechanism suffices to get
 // the qualifier needed for the loop invariant below.


ranjitjhala:
occurs to me that the CI will probably fail until the fixpoint PR is merged into develop?

flux-tests/tests/neg/surface/scrape00.rs
```diff
@@ -0,0 +1,29 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
```
nilehmann, line 2: I'd add an explicit `#![flux::cfg(scrape_quals = false)]` here

flux-tests/tests/pos/surface/scrape00.rs
```diff
@@ -0,0 +1,31 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
+#![feature(custom_inner_attributes)]
+#![flux::cfg(scrape_quals = "true")]
```
nilehmann, line 4: I think we can write it without the quotes

```suggestion
#![flux::cfg(scrape_quals = true)]
```

flux-fixpoint/src/lib.rs
```diff
@@ -156,6 +158,9 @@ impl KVar {
 
 impl<Tag: fmt::Display> fmt::Display for Task<Tag> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.scrape_quals {
+            writeln!(f, "(fixpoint \"--scrape=both\")")?;
```
nilehmann, line 162: Is `both` necessary to verify the tests we have?

nilehmann (approved):
LGTM

flux-fixpoint/src/lib.rs
ranjitjhala: Good point -- not really. `--both` verifies these examples if you remove ALL the default qualifiers, but `--head` suffices with the default qualifers (as we need something like `x <= y`)

ranjitjhala: So for now, I'd like to leave `both` as default. Another option is to make `scrape_quals` not a `bool` but something like `No|Head|Both` ?

PR #438, opened 2023-05-03
Implement type holes

nilehmann:
Support for writing `_` in a signature signifying a hole. The hole is filled during `annot_check` when zipping against the default lifted `fhir`.


ranjitjhala:
Nice! Maybe good to have some docs / examples illustrating _what_ you can put in a hole e.g.

* only the `rust` types e.g. you can write `RVec<_>` for `RVec<i32>` ?
* also the refinements e.g. you can write `RVec<_>` for `RVec<i32{v: 0 <= v}>`  or `i32[10]` ?

The second example makes me think its _also_ refinements ... ?

PR #449, opened 2023-05-18
Add `Set<int>` and related operations as index sort

ranjitjhala:
See [list01.rs](flux-tests/tests/pos/enums/list01.rs)

PR #450, opened 2023-05-18
Cleanup ctor_arity

ranjitjhala:
Left this out of the `sets` PR #449


ranjitjhala:
(Went ahead and merged as it was a leftover suggestion from the previous PR)

PR #475, opened 2023-07-19
Update Cargo package versions

jepotter1:
This fixes build failure since Flux depended on a different version of `hashbrown` than one of the other dependencies, creating a dependency conflict. Also, it's good to have updated packages.

I have verified that the test suite passes after implementing this change.

@nilehmann

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,6 +20,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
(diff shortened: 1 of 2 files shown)


nilehmann:
Thanks, @jepotter1. I still don't understand what the issue was as the project compiled for us both locally and in the CI. So I assume there must be something different in your local configuration. Are you in windows?

Still, I don't see any problems with merging this.

PR #478, opened 2023-07-25
Very preliminary code to check trait-bound obligations

ranjitjhala:
For example, the following now works:

**NOTE** there are all sorts of short-cuts with binders, next: extend this to allow the trait-predicates to refer to refinement parameters.

```rust
#[flux::sig(fn(bool[true]))]
pub fn assert(_b: bool) {}

#[flux::sig(fn(it: I) where I: Iterator<Item = i32{v:10<=v}>)]
pub fn foo<I>(it: I)
where
    I: Iterator<Item = i32>,
{
    for x in it {
        assert(0 <= x)
    }
}

#[flux::sig(fn(v: Vec<i32{v:70<=v}>))]
pub fn test_ok(v: Vec<i32>) {
    let zig = v.into_iter();
    foo(zig);
}

#[flux::sig(fn(v: Vec<i32{v:7<=v}>))]
pub fn test_err(v: Vec<i32>) {
    let zig = v.into_iter();
    foo(zig); //~ ERROR: refinement type
}
```


flux-middle/src/rty/projections.rs
```diff
@@ -67,3 +83,220 @@ impl TypeFolder for WithPredicates {
...
+        Ok(e) => bug!("invalid selection for {oblig:?} = {e:?}"),
+        Err(e) => bug!("error selecting {oblig:?}: {e:?}"),
+    };
+    impl_source
+}
+
+/// Given an an `impl_rty` e.g. `std::vec::IntoIter<Nat>` and an `elem` e.g. `std::iter::Iterator::Item`,
+/// returns the component of the `impl_rty` that corresponds to the `elem`, e.g. `Nat`.
+
+pub fn resolve_impl_projection(
```
nilehmann, line 262: Instead of having a separate `resolve_impl_projection` this should be integrated into `normalize_projections`. I think like rustc we need to have an equivalent to `normalize_projection_type` (https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/project/fn.normalize_projection_type.html) that takes an `AliasTy` (+ all the necessary context) and returns the normalized type. That function first looks for a candidate in `ParamEnv` and if it can't find one then it looks into impl sources (see this function to see how rustc looks for candidates in different places https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_trait_selection/traits/project.rs.html#1463)

nilehmann (commented):
This is going in the right direction, but I have one big comment about integrating normalization form param env and from impl sources into a single function.

nilehmann:
re skipping existentials to infer the mapping for type variables:

Just thinking out loud. We probably shouldn't worry about this right now.

The generic existential syntax allows us to an *inner* type to depend on the index of an *outer* type. In terms of LH, in flux we can have a type `{v: T | p}` where `v` is in scope inside `T`. In particular `v` is in scope for type arguments used to construct `T`. I know this has caused problems in LH, but it hasn't caused problems for us so far and it allows to express cool types. For example, you can express a vector that contains vectors whose elements are valid indices into  itself

```
RVec<{n. RVec<i32{v: v < n}>[n]}>
```

Then the following works

```rust
#[flux::sig(fn() -> RVec<{n. RVec<usize{v: v < n}>[n]}>)]
fn foo() -> RVec<RVec<usize>> {
    let mut vec = RVec::new();
    vec.push(range(100));
    vec.push(range(10));
    vec.push(range(42));
    vec
}

#[flux::sig(fn(n: usize) -> RVec<usize{v: v < n}>[n])]
fn range(n: usize) -> RVec<usize> {
    let mut i = 0;
    let mut vec = RVec::new();
    while i < n {
        vec.push(i);
        i += 1;
    }
    vec
}
```

Now, in terms of traits we can get into a situation where for `SomeTrait` we have an implementation:

```rust
impl<T> SomeTrait for RVec<RVec<T>> {
    type Assoc = T;
}
```

And then we want to normalize `<RVec<{n. RVec<{v. i32[v] | v < n}>[n]}> as SomeTrait>::Assoc`. Now the question is whether this is something we can make sense of, or whether we will always have to fail.

nilehmann:
Maybe one way to think about it is that `RVec<{n. RVec<{v. i32[v] | v < n}>[n]}> as SomeTrait` will never be true because `impl<T> SomeTrait for RVec<RVec<T>>` is implicitly saying that `T` needs to be well-formed in the empty environment. 

I guess one conclusion from all this is that we can safely match type variables inside existential as long as the matching type doesn't have escaping variables (https://github.com/flux-rs/flux/blob/8262936f8dedbf4790dd6e41e6dc4b0324b128da/flux-middle/src/rty/fold.rs#L142)

flux-middle/src/rty/projections.rs
```diff
@@ -67,3 +83,220 @@ impl TypeFolder for WithPredicates {
...
+    }
+
+    fn insert(&mut self, pty: &ParamTy, ty: &Ty) {
+        match self.map.insert(*pty, ty.clone()) {
+            None => (),
+            Some(_) => bug!("duplicate insert"),
+        }
+    }
+
+    fn base_ty(ty: &Ty) -> BaseTy {
```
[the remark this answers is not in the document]
nilehmann: Alternatively, following my ramble below, we can also skip existentials and fail if we ever match a type with escaping vars.

flux-middle/src/rty/projections.rs
```diff
@@ -67,3 +83,220 @@ impl TypeFolder for WithPredicates {
...
+    impl_rty: &Ty,
+    callsite_def_id: DefId,
+) -> ImplSourceUserDefinedData<'tcx, Obligation<'tcx, rustc_middle::ty::Predicate<'tcx>>> {
+    // 1a. build up the `Obligation` query
+    let trait_def_id = genv.tcx.parent(elem);
+    let predicate = TraitPredicate {
+        trait_ref: TraitRef::new(genv.tcx, trait_def_id, vec![into_rustc_ty(&genv.tcx, impl_rty)]),
+        constness: rustc_middle::ty::BoundConstness::NotConst,
+        polarity: rustc_middle::ty::ImplPolarity::Positive,
+    };
```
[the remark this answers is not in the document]
ranjitjhala: wow nice, good catch! thanks!

flux-middle/src/rty/projections.rs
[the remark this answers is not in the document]
ranjitjhala: done!

ranjitjhala:
@nilehmann - ok I think I've addressed the issues, in particular look at the new 

* `projections::normalize`
* `projections::normalize_alias_ty` 

which internally use the `ProjectionTable::normalize` method, that sequences the `normalize_with_preds` and `normalizes_with_impl`

flux-middle/src/rty/mod.rs
```diff
@@ -92,10 +92,24 @@ pub enum ClauseKind {
...
+            ty.clone()
+        } else {
+            bug!("expected `GenericArg::Ty`")
+        }
+    }
+
+    pub fn elem(&self) -> DefId {
+        self.alias_ty.def_id
+    }
+}
```
nilehmann, line 111: If I'm reading correctly, I think we are not using these methods.

```suggestion
```

flux-middle/src/rty/projections.rs
```diff
@@ -1,36 +1,75 @@
...
+    def_id: DefId,
+    preds: FxHashMap<AliasTy, Ty>,
+}
 
-impl ProjectionTable {
-    pub fn new(predicates: GenericPredicates) -> Self {
-        let mut res = FxHashMap::default();
+impl<'sess, 'tcx> ProjectionTable<'sess, 'tcx> {
+    fn new(genv: &'sess GlobalEnv<'sess, 'tcx>, def_id: DefId) -> Result<Self, QueryErr> {
+        let predicates = callsite_predicates(genv, def_id)?;
```
nilehmann, line 41: Always use `skip_binder` (or `instantiate_identity`)

```suggestion
        let predicates = genv.predicates_of(def_id)?.skip_binder();
```

flux-middle/src/rty/projections.rs
```diff
@@ -48,22 +87,254 @@ fn without_constrs<T: TypeFoldable>(t: &T) -> T {
...
+    fn infer_from_arg(&mut self, src: &rustc_middle::ty::GenericArg, dst: &GenericArg) {
+        if let GenericArg::Ty(dst) = dst {
+            self.infer_from_ty(&src.as_type().unwrap(), dst)
+        }
+    }
+
+    fn infer_from_ty(&mut self, src: &rustc_middle::ty::Ty, dst: &Ty) {
+        use rustc_middle::ty;
+        match src.kind().clone() {
+            ty::TyKind::Param(pty) => self.insert(&pty, dst),
```
nilehmann, line 180: I think we need to check for escaping vars here when we are actually matching it with a type variable.

flux-refineck/src/constraint_gen.rs
```diff
@@ -236,13 +237,22 @@ impl<'a, 'tcx> ConstrGen<'a, 'tcx> {
             }
         }
 
+        // check (non-closure) obligations -- the closure ones are handled in `checker` since
+        // as we have to recursively walk over their def_id bodies.
+        for pred in &obligs {
+            if let rty::ClauseKind::Projection(projection_pred) = pred.kind().skip_binder() {
+                let impl_elem =
+                    normalize_alias_ty(infcx.genv, callsite_def_id, &projection_pred.alias_ty)?;
```
nilehmann, line 245: Not sure about this, but instead of opening the predicate and then normalizing the `AliasTy` we could normalize the predicate (which is `TypeFoldable`) and when opened everything will be formalized.

flux-refineck/src/constraint_gen.rs
```diff
@@ -236,13 +237,22 @@ impl<'a, 'tcx> ConstrGen<'a, 'tcx> {
...
+                infcx.subtyping(rcx, &impl_elem, &projection_pred.term);
```
nilehmann, line 246: I believe this is unsound in general. We should do subtyping both ways

```suggestion
                infcx.subtyping(rcx, &impl_elem, &projection_pred.term);
                infcx.subtyping(rcx, &projection_pred.term, &impl_elem);
```

flux-tests/tests/pos/surface/trait01a.rs
```diff
@@ -1,9 +1,21 @@
...
 
+// TODO: see todo/trait02.rs
+// #[flux::sig(fn(it: IntoIter<i32{v:100<=v}>))]
+// pub fn goo<I>(it: IntoIter<i32>) {
+//     // fn next<J><J:Iterator) -> Option<J::Item>
+//     //  query: what is J::Item when J is IntoIter<Nat>
+//     for x in it {
+//         assert(10 <= x)
+//     }
+// }
```
nilehmann, line 17: Clean this up

flux-tests/tests/neg/surface/trait02.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[flux::sig(fn(it: I) where I: Iterator<Item = i32{v:10<=v}>)]
+pub fn foo<I>(it: I)
+where
+    I: Iterator<Item = i32>,
+{
+    // let x = it.next()
+    // fn next<J><J:Iterator) -> Option<J::Item>
+    //  query: what is J::Item when J is I
```
nilehmann, line 23: Clean this up

flux-refineck/src/checker.rs
```diff
@@ -1199,9 +1193,9 @@ pub(crate) mod errors {
             Self { kind: CheckerErrKind::OpaqueStruct(def_id), span }
         }
 
-        pub fn query(query_error: QueryErr, span: Span) -> Self {
-            Self { kind: CheckerErrKind::Query(query_error), span }
-        }
+        // pub fn query(query_error: QueryErr, span: Span) -> Self {
+        //     Self { kind: CheckerErrKind::Query(query_error), span }
+        // }
```
nilehmann, line 1198: I think we can get rid of this

```suggestion
        // pub fn query(query_error: QueryErr, span: Span) -> Self {
        //     Self { kind: CheckerErrKind::Query(query_error), span }
        // }
```

flux-refineck/src/checker.rs
```diff
@@ -495,46 +488,48 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
...
+                fn_trait_pred.to_closure_sig(*def_id, tys.clone()),
+                self.config,
+            )?;
         }
-        fn_trait_preds
+        Ok(())
     }
 
-    fn check_obligs(
+    /// This checks obligations related to closures; the remainder are checked in `check_non_closure_obligs`
```
nilehmann, line 521: Update comment. `check_non_closure_obligs` doesn't seem to be the name.

flux-refineck/src/checker.rs
```diff
@@ -476,11 +468,12 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
         args: &[Operand],
     ) -> Result<Ty, CheckerError> {
         let actuals = self.check_operands(rcx, env, terminator_span, args)?;
-        let predicates = self.callsite_predicates(terminator_span)?;
+        // let predicates = self.callsite_predicates(terminator_span)?;
         //  println!("TRACE: check_call 1: {did:?} {fn_sig:?}");
```
nilehmann, line 472:
```suggestion
```

flux-middle/src/rustc/lowering.rs
```diff
@@ -758,7 +754,7 @@ pub(crate) fn lower_generic_predicates<'tcx>(
     let mut fn_output_proj = FxHashMap::default();
     let mut predicates = vec![];
 
-    // println!("TRACE: lower_generic_predicates 1: {def_id:?} {generics:?}");
+    // println!("TRACE: lower_generic_predicates 1: {generics:?}");
```
nilehmann, line 757:
```suggestion
```

flux-middle/src/rty/projections.rs
```diff
@@ -48,22 +87,254 @@ fn without_constrs<T: TypeFoldable>(t: &T) -> T {
...
+        Err(e) => bug!("error selecting {oblig:?}: {e:?}"),
+    };
+    impl_source
 }
 
-impl TypeFolder for WithPredicates {
+/// QUERY: normalize <std::vec::IntoIter<Nat> as std::iter::Iterator::Item
+/// Given an an `impl_rty` e.g. `std::vec::IntoIter<Nat>` and an `elem` e.g. `std::iter::Iterator::Item`,
+/// returns the component of the `impl_rty` that corresponds to the `elem`, e.g. `Nat`.
+
```
nilehmann, line 257:
```suggestion
```

flux-middle/src/rty/projections.rs
```diff
@@ -48,22 +87,254 @@ fn without_constrs<T: TypeFoldable>(t: &T) -> T {
...
+        GenericArg::Lifetime(_) => todo!(),
+    }
+}
+fn into_rustc_bty<'tcx>(tcx: &TyCtxt<'tcx>, bty: &BaseTy) -> rustc_middle::ty::Ty<'tcx> {
+    match bty {
+        BaseTy::Int(i) => rustc_middle::ty::Ty::new_int(*tcx, *i), // rustc_middle::ty::Ty::mk_int(*tcx, int_ty),
+        BaseTy::Uint(i) => rustc_middle::ty::Ty::new_uint(*tcx, *i),
+        BaseTy::Param(pty) => pty.to_ty(*tcx),
+        BaseTy::Slice(ty) => rustc_middle::ty::Ty::new_slice(*tcx, into_rustc_ty(tcx, ty)),
+        BaseTy::Bool => todo!(),
```
nilehmann, line 112: I'm fine with filling these in as we go, but we can at least fill the easy ones.

flux-middle/src/rty/projections.rs
```diff
@@ -1,36 +1,75 @@
...
 };
 
-#[derive(Debug)]
-struct ProjectionTable(FxHashMap<AliasTy, Ty>);
+fn callsite_predicates(genv: &GlobalEnv, def_id: DefId) -> Result<GenericPredicates, QueryErr> {
+    match genv.predicates_of(def_id) {
+        Ok(eb) => Ok(eb.0),
+        Err(e) => Err(e),
+    }
+}
```
nilehmann, line 31:
```suggestion
```

nilehmann (approved):
Ok, looks fine. Just one important comment regarding subtyping both ways when checking the projection predicate.

ranjitjhala:
Hmm -- I'm not sure about the sub-typing -- what you propose seems too conservative. 

For example, `test_ok` would be rejected, and I don't think it should be?

```rust
#[flux::sig(fn(it: I) where I: Iterator<Item = i32{v:10<=v}>)]
pub fn foo<I>(it: I)
where
    I: Iterator<Item = i32>,
{
    for x in it {
        assert(0 <= x)
    }
}

#[flux::sig(fn(v: Vec<i32{v:70<=v}>))]
pub fn test_ok(v: Vec<i32>) {
    let zig = v.into_iter();
    foo(zig);
}
```

ranjitjhala:
Instead, I think we want semantics like this (which currently verifies)

```rust
#[flux::sig(fn (x:Vec<i32{v:10 <= v}>))]
pub fn lib(_x:Vec<i32>) {}

#[flux::sig(fn (x:Vec<i32{v:70 <= v}>))]
pub fn cli(y:Vec<i32>) { lib(y) }
```

flux-refineck/src/constraint_gen.rs
ranjitjhala: ah yes, I was trying to avoid exposing a second function, this is perfect, thanks!

nilehmann:
I don't think the example will get rejected. We have by subtyping when calling `into_iter`

`Vec<i32{v: v > 0}> <: Vec<i32{v: k0}>`  (which requires `10 <= v => k0`)

Then we have that type type of `zig` is `IntoIter<i32{v: k0}>`. After instantiating type parameters, the type of `foo` is `fn(IntoIter<i32{v: k1}>)` (importantly for a fresh `k1`), by argument subtyping we get

`IntoIter<i32{v: k0}> <: IntoIter<i32{v: k1}>` (which requires `k0 => k1`)

Then, after normalizing `<IntoIter<i32{v: k1}> as Iterator>::Item == i32{v: 0 <= v}` we get `i32{v: $k1} == i32{v: 0 <= v}`. Even if we interpret this as subtyping both ways the program type checks because we have

```
10 <= v => k0
k0 => k1
k1 => 0 <= v
0 <= v => k1
```

flux-refineck/src/constraint_gen.rs
ranjitjhala: Hmm no, that doesn't quite work because we want to turn the `AliasTy` into a `Ty` but the "folder" for `AliasTy` returns an `AliasTy` ...

ranjitjhala: I suppose we could turn the `alias_ty` into a proper `ty` and then just fold over that...

nilehmann: oh right, we would need a more general predicate for equality between arbitrary types. ok then, so let's leave it like this.

ranjitjhala:
You are right -- I made the check `invariant` to be conservative, but I think the cycle is ... perhaps not needed. unclear, but have added the other direction for now.

nilehmann:
yes, the cycle is not needed in some cases, but I believe we need it in general. My hunch is that we can interpret the projection constraint as subtyping when there are no mutable references involved, but I'm not certain. 

In any case, something I've been thinking is whether we can make fixpoint smarter to solve `k1` "syntactically" in constraints like this that have "easy" cycles

```
10 <= v => k0
k0 => k1
k1 => 0 <= v
0 <= v => k1
```

ranjitjhala:
wait my bad ... actually there is no cycle here. The graph is just k0 -> k1

flux-middle/src/rty/projections.rs
```diff
@@ -48,22 +80,247 @@ fn without_constrs<T: TypeFoldable>(t: &T) -> T {
...
 }
+fn into_rustc_bty<'tcx>(tcx: &TyCtxt<'tcx>, bty: &BaseTy) -> rustc_middle::ty::Ty<'tcx> {
+    match bty {
+        BaseTy::Int(i) => rustc_middle::ty::Ty::new_int(*tcx, *i), // rustc_middle::ty::Ty::mk_int(*tcx, int_ty),
+        BaseTy::Uint(i) => rustc_middle::ty::Ty::new_uint(*tcx, *i),
+        BaseTy::Param(pty) => pty.to_ty(*tcx),
+        BaseTy::Slice(ty) => rustc_middle::ty::Ty::new_slice(*tcx, into_rustc_ty(tcx, ty)),
+        BaseTy::Bool => rustc_middle::ty::Ty::new(*tcx, rustc_middle::ty::TyKind::Bool),
+        BaseTy::Char => rustc_middle::ty::Ty::new(*tcx, rustc_middle::ty::TyKind::Char),
+        BaseTy::Str => rustc_middle::ty::Ty::new_static_str(*tcx),
```
nilehmann, line 107: This is wrong, we need [`tcx.types.str_`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.CommonTypes.html#structfield.str_)

ranjitjhala: whoops. fixed here https://github.com/flux-rs/flux/pull/479

PR #480, opened 2023-08-01
Preliminary support for `async` functions

ranjitjhala:
The following now works, with *rather* bare-bones support for opaque-types, and `async` (which ends up connecting with the stuff we did for closures too!)

```rust
#[flux::sig(async fn (y:i32{v:0<=v}) -> i32{v:v < 1000 })]
pub async fn foo(y: i32) -> i32 {
    let z = if y > 10 { 1 } else { 0 };
    assert(z >= 0);
    assert(y >= 0);
    z + 999 //~ ERROR refinement type
}
```

Next:

- [] `await`
- [] dealing with dependencies and `Binder<...>`


flux-syntax/src/surface.rs
```diff
@@ -192,6 +194,8 @@ pub enum TyKind<R = ()> {
     Constr(Expr, Box<Ty<R>>),
     Tuple(Vec<Ty<R>>),
     Array(Box<Ty<R>>, ArrayLen),
+    /// The first `R` parameter is for the `DefId` corresponding to the hir OpaqueTy
+    Opaque(R, Option<Vec<GenericArg<R>>>, Bounds<R>),
```
nilehmann, line 198: Let's call this `ImplTrait` to be consistent with [`rustc_ast::TyKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.TyKind.html#variant.ImplTrait)

nilehmann, line 198: The `Option<Vec<GenericArg<R>>` looks sketchy and I think is wrong. If I understand correctly, the args in [`rustc_hir::TyKind::OpaqueDef`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.TyKind.html#variant.OpaqueDef) correspond to the lifetimes captured by the opaque type so we should be generating a list of lifetime holes instead of type holes.

In any case, I wouldn't deal with this during resolving but during desugaring. So, I would eliminate the `Option<Vec<GenericArg<R>>>` field here, and during resolving we just get the `DefId`. Then during desugaring, we *query* the `Generics` of the opaque type and generate corresponding holes there.

This is also related to opaque types depending on refinement parameters.

flux-syntax/src/surface.rs
```diff
@@ -368,3 +372,13 @@ impl fmt::Debug for BinOp {
         }
     }
 }
+
+pub fn async_return_ty(ty: Ty) -> Ty {
```
nilehmann, line 376: I would carry the asyncness and deal with this later (maybe during `conv`) but I guess this is fine for now.

flux-middle/locales/en-US.ftl
```diff
@@ -22,5 +22,5 @@ middle_unsupported_generic_bound =
 # Query Errors
 
 middle_query_unsupported_type =
-    unsupported type
+    unsupported type FOOASD
```
nilehmann, line 25:
```suggestion
    unsupported type
```

flux-desugar/src/table_resolver.rs
```diff
@@ -552,6 +599,10 @@ impl ResKey {
         let s = path.segments.iter().map(|segment| segment.ident).join("::");
         Ok(ResKey { s })
     }
+
+    fn from_lang_item(item: hir::LangItem) -> Self {
+        ResKey { s: format!("{:?}", item) }
```
nilehmann, line 604: Let's not rely on the debug format
```suggestion
        ResKey { s: item.name().to_string() }
```

flux-desugar/src/table_resolver.rs
```diff
@@ -277,6 +278,7 @@ impl<'sess> Resolver<'sess> {
 
     fn resolve_path(&self, path: Path) -> Result<Path<Res>, ErrorGuaranteed> {
         let Some(res) = self.table.get(&ResKey::from_path(&path)) else {
+            // panic!("Unresolved path {:?} {:?}", path.segments.iter().join("::"), path.generics);
```
nilehmann, line 281:
```suggestion
```

flux-desugar/src/table_resolver.rs
```diff
@@ -462,33 +500,42 @@ impl<'sess> NameResTable<'sess> {
                 };
                 self.collect_from_path(path)
             }
+            hir::TyKind::OpaqueDef(item_id, args, _) => {
+                self.opaque = Some((item_id.owner_id.def_id, args.len()));
```
nilehmann, line 504: We should report an error or at least panic here if `self.opaque` is already `Some`.

flux-middle/src/fhir/lift.rs
```diff
@@ -176,14 +176,16 @@ pub fn lift_fn_sig(
...
         params: vec![],
         requires: vec![],
         args,
         output,
         lifted: true,
         span: fn_sig.span,
-    })
+    };
+    Ok(fn_sig)
+    // Ok(fhir::FnInfo { fn_sig, fn_preds, fn_impls })
```
nilehmann, line 188:
```suggestion
```

flux-middle/src/fhir/lift.rs
```diff
@@ -297,9 +303,14 @@ impl<'a, 'tcx> LiftCtxt<'a, 'tcx> {
...
         };
+        let generics = match path.segments.last().unwrap().args {
+            Some(args) => self.lift_generic_args(args.args)?,
+            None => vec![],
+        };
+
         let path = fhir::Path {
             res,
-            generics: self.lift_generic_args(path.segments.last().unwrap().args)?,
+            generics, // : self.lift_generic_args(path.segments.last().unwrap().args)?,
```
nilehmann, line 313:
```suggestion
            generics,
```

flux-middle/src/fhir/lift.rs
```diff
@@ -323,29 +334,30 @@ impl<'a, 'tcx> LiftCtxt<'a, 'tcx> {
 
     fn lift_generic_args(
         &self,
-        args: Option<&hir::GenericArgs>,
+        // args: Option<&hir::GenericArgs>,
```
nilehmann, line 337:
```suggestion
```

flux-middle/src/fhir/lift.rs
```diff
@@ -323,29 +334,30 @@ impl<'a, 'tcx> LiftCtxt<'a, 'tcx> {
...
-                        let ty = self.lift_ty(ty)?;
-                        lifted.push(fhir::GenericArg::Type(ty));
-                    }
-                    hir::GenericArg::Const(_) => {
-                        return self.emit_unsupported("const generics are not supported")
-                    }
-                    hir::GenericArg::Infer(_) => {
-                        bug!("unexpected inference generic argument");
-                    }
+        // if let Some(args) = args {
```
nilehmann, line 341:
```suggestion
```

flux-middle/src/fhir/lift.rs
```diff
@@ -323,29 +334,30 @@ impl<'a, 'tcx> LiftCtxt<'a, 'tcx> {
...
+                }
+                hir::GenericArg::Const(_) => {
+                    return self.emit_unsupported("const generics are not supported")
+                }
+                hir::GenericArg::Infer(_) => {
+                    bug!("unexpected inference generic argument");
                 }
             }
         }
+        //}
```
nilehmann, line 360:
```suggestion
```

flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -306,11 +320,47 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
                 self.check_type(infcx, ty)?;
                 self.check_pred(infcx, pred)
             }
+            fhir::TyKind::OpaqueDef(_item_id, args, _) => {
+                // TODO: self.check_item_id(infcx, *item_id)?;
```
nilehmann, line 324:
```suggestion
```

nilehmann (changes requested):
Some comments

flux-desugar/src/desugar.rs
```diff
@@ -455,17 +473,22 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
...
-        Ok(fhir::ClauseKind::Projection(proj))
-    } else {
-        panic!("produce error for desugar_bound")
-    }
+        let trait_def_id = match path.res {
+            Res::Trait(trait_def_id) => trait_def_id,
+            Res::LangTrait(LangItem::Future) => {
+                self.early_cx.tcx.lang_items().future_trait().unwrap()
+            }
+            _ => panic!("unexpected trait {:?}", path.res),
```
nilehmann, line 481: Never use `panic!`, use either `bug!` or `span_bug!` if there's a relevant span in scope.
```suggestion
            _ => span_bug!(path.span, "unexpected trait {:?}", path.res),
```

flux-desugar/src/desugar.rs
```diff
@@ -455,17 +473,22 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
...
+        };
+        if let [surface::GenericArg::Constraint(ident, ty)] = path.generics.as_slice() {
+            let item_id = self.lookup_item_id(trait_def_id, *ident)?;
+            let projection_ty = fhir::AliasTy { args: substs.clone(), def_id: item_id };
+            let term = self.desugar_ty(None, ty, binders)?;
+            let proj = fhir::ProjectionPredicate { projection_ty, term };
+            Ok(fhir::ClauseKind::Projection(proj))
+        } else {
+            panic!("unexpected path in desugar_bound: {:?}", path.generics)
```
nilehmann, line 490:
```suggestion
            bug!("unexpected path in desugar_bound: {:?}", path.generics)
```

flux-middle/src/fhir.rs
```diff
@@ -120,7 +122,8 @@ pub struct AliasTy {
 #[derive(Default, Debug)]
 pub struct Map {
     generics: FxHashMap<LocalDefId, Generics>,
-    predicates: FxHashMap<LocalDefId, GenericPredicates>,
+    predicates: ItemPredicates,
+    item_bounds: ItemPredicates,
```
nilehmann, line 126: Do we need them both? Maybe we can store them both in the same map. Opaque types are also items that happen to have predicates and generics.

flux-refineck/src/fold_unfold.rs
```diff
@@ -315,6 +316,9 @@ impl<'a, 'tcx, M: Mode> FoldUnfoldAnalysis<'a, 'tcx, M> {
             TerminatorKind::Goto { target } => {
                 self.goto(bb, *target, env)?;
             }
+            TerminatorKind::Yield { resume, .. } => {
+                self.goto(bb, *resume, env)?;
```
nilehmann, line 320: You should do a `M::projection` of the corresponding place here

flux-refineck/src/type_env.rs
```diff
@@ -385,8 +395,8 @@ impl BasicBlockEnvShape {
             BaseTy::Slice(ty) => BaseTy::Slice(Self::pack_ty(scope, ty)),
             BaseTy::Ref(r, ty, mutbl) => BaseTy::Ref(*r, Self::pack_ty(scope, ty), *mutbl),
             BaseTy::Array(ty, c) => BaseTy::Array(Self::pack_ty(scope, ty), c.clone()),
-            BaseTy::Alias(..) => tracked_span_bug!("packing of `AliasTy` is not implemented"),
-            BaseTy::Int(_)
+            BaseTy::Alias(..) // => tracked_span_bug!("packing of `AliasTy` is not implemented"),
```
nilehmann, line 398: I'm not entirely sure what's the right thing to do here, but we should at least add an `assert` for `BaseTy::Alias` ensuring it doesn't have free variables, i.e., `assert!(scope.has_free_vars(bty))`.

nilehmann (commented):
More comments

Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [workspace]
 resolver = "2"
 
+edition = "2021"
```
nilehmann, line 4: This has no effect when specified in a workspace toml

```suggestion
```

flux-refineck/src/fold_unfold.rs
```diff
@@ -739,6 +758,16 @@ impl PlaceNode {
     }
 }
 
+fn generic_args_fields(substs: &Substs) -> Vec<PlaceNode> {
+    let fields = substs
+        .as_closure()
```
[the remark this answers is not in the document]
ranjitjhala: ok, have removed this function, and used `as_closure()` and `as_generator()` at the appropriate callsites.

flux-desugar/src/table_resolver.rs
ranjitjhala: Sadly had to use the `format` version here because the `to_string()` method returns 

```
TRACE: from_lang_item(Future) -> "future_trait"
```

which then fools the subsequent resolving (as we just write `Future<Item=...>`)

flux-refineck/src/constraint_gen.rs
```diff
@@ -539,8 +570,70 @@ impl<'a, 'tcx> InferCtxt<'a, 'tcx> {
...
+        let args = vec![GenericArg::Ty(refine_default(bty.clone()).skip_binder())];
+        let alias_ty = rty::AliasTy::new(def_id, args);
+        let proj_ty = refine_default(BaseTy::projection(alias_ty));
+        rty::projections::normalize(self.genv, self.def_id, &proj_ty, self.span())
+            .unwrap()
+            .skip_binder()
+    }
+
+    fn opaque_subtyping(&mut self, rcx: &mut RefineCtxt, bty: &BaseTy, opaque_def_id: DefId) {
+        // TODO(RJ): odd hack because a malformed predicate makes the `item_bounds` query return `Err` which cannot be `unwrapped` in `self.opaque_predicates`
```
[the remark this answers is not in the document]
ranjitjhala: Ok, so I 

1. made all the `subtyping` methods now have type `Result<(), CheckerError>`
2. made `check_crate_wf` _also_ check the `DefKind::OpaqueTy` so it ends up checking the bounds for the `impl trait` in a "strict" fashion, instead of lazily ...

flux-desugar/src/table_resolver.rs
ranjitjhala: @nilehmann check / advice ? ^^^

flux-refineck/src/fold_unfold.rs
ranjitjhala: @nilehmann I added this, but please check if its the right thing, as I don't know what `M::projection` is supposed to do!

ranjitjhala:
@nilehmann -- I think I addressed the issues above but you should check

* the `format!` thing 
* the `M::projection` thing

flux-refineck/src/fold_unfold.rs
nilehmann: Yes, this looks good

flux-desugar/src/table_resolver.rs
nilehmann: @ranjitjhala 

Oh, I see. This is then related to https://github.com/flux-rs/flux/pull/480#discussion_r1297462221. I think the problem comes from supporting

```rust
#[flux::sig(fn() -> impl Future<Output=i32>)]
async fn foo() - i32 {
    0
}
```

which makes it weird because we have to resolve `Future` even though the user didn't write it. If we carry the asyncness down and generate the `impl Future<..>` during desugaring then we wouldn't have to resolve `Future`. During desugaring, we can create the`impl Future` by generating a `fhir::Res::Struct` with the appropriate `DefId`  corresponding to the `future_trait` lang item. If we do this then we won't need `fhir::Res::LangTrait`.

flux-desugar/src/desugar.rs
```diff
@@ -569,8 +598,35 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
...
+            surface::TyKind::ImplTrait(res, bounds) => {
+                if let Res::OpaqueTy(def_id) = res &&
+                   let Some(local_def_id) = def_id.as_local()
+                {
+                    let owner_id = OwnerId { def_id: local_def_id };
+                    let item_id = hir::ItemId { owner_id };
+
+                    let arity = self.early_cx.tcx.generics_of(local_def_id).params.len();
+                    let args = Self::mk_generic_args_with_hole(arity, span);
+                    let args = self.desugar_generic_args(*res, &args, binders)?;
```
nilehmann, line 610: This is still wrong, we have to create `GenericArg`s matching the kind of the corresponding `GenericParam`s.  However, I think these are always guaranteed to be lifetime arguments so we can't just call `desugar_generic_args` with an empty slice and that will insert the appropriate lifetime holes.

```suggestion
                    let args = self.desugar_generic_args(*res, &[], binders)?;
```

flux-desugar/src/desugar.rs
```diff
@@ -569,8 +598,35 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
...
+
+                    let kind = fhir::TyKind::OpaqueDef(item_id, args, false);
+                    let self_ty =fhir::Ty { kind: kind.clone(), fhir_id: ty_fhir_id,  span };
```
nilehmann, line 613:
```suggestion
                    let self_ty = fhir::Ty { kind: kind.clone(), fhir_id: ty_fhir_id,  span };
```

flux-desugar/src/desugar.rs
```diff
@@ -569,8 +598,35 @@ impl<'a, 'tcx> DesugarCtxt<'a, 'tcx> {
...
+            }
         };
-        Ok(fhir::Ty { kind, fhir_id: self.next_fhir_id(), span })
+        Ok(fhir::Ty { kind, fhir_id: ty_fhir_id, span })
+    }
+
+    fn mk_generic_args_with_hole(arity: usize, span: Span) -> Vec<surface::GenericArg<Res>> {
+        (0..arity)
+            .map(|_| surface::GenericArg::Type(surface::Ty { kind: surface::TyKind::Hole, span }))
+            .collect()
```
nilehmann, line 629: Remove this

nilehmann (commented):
There are a couple of comments you missed, probably because they are hidden in the UI.

flux-desugar/src/table_resolver.rs
ranjitjhala: quite so, let me make this change!

flux-desugar/src/desugar.rs
ranjitjhala: By **can't just call** you mean **can just call** I presume?

nilehmann: ups, yeah, I mean we *can*

flux-desugar/src/table_resolver.rs
ranjitjhala: Ok, take a look -- I didn't use the `fhir::Res::Struct` because post-desugaring, the `async` output is directly represented as an

1. `Opaque` (with the appropriate impl's `DefId`) 
2. the associated predicates `item_bounds` for (1)

So I modified the `resolve` to track (1) and generate (2) during `desugar_ty` (see `desugar_async_out_ty`)

ranjitjhala:
Done, I think but see https://github.com/flux-rs/flux/pull/480#discussion_r1299288706

flux-desugar/src/table_resolver.rs
nilehmann: oh right, we have to resolve the `Opaque` type. This sounds good

nilehmann (approved):
LGTM

PR #481, opened 2023-08-17
Add support for _lowering_ `const-generics`

ranjitjhala:
To support signatures like [this](https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Vec.html#method.from_array)

TODO (perhaps in a later PR)

- [] add corresponding support in `fhir` so we can write such signatures in checked code
- [] add some way to refer to the `N` in refinements? (seems v. fancy?)


flux-middle/src/rustc/ty.rs
```diff
@@ -22,6 +22,7 @@ use rustc_span::{symbol::kw, Symbol};
 
 use self::subst::Subst;
 use crate::{
+    fhir::ArrayLen,
```
nilehmann, line 25: This module shouldn't depend on stuff from `fhir`

flux-middle/src/rustc/ty.rs
```diff
@@ -157,14 +159,39 @@ pub enum AliasKind {
...
+pub enum Const {
+    Param(ParamConst),
+    Value(ValueConst),
+}
+
+impl From<&ArrayLen> for Const {
+    fn from(len: &ArrayLen) -> Self {
+        Const::Value(ValueConst { val: len.val })
+    }
+}
```
nilehmann, line 182: Let's remover this to avoid the dependency on `fhir`.

```suggestion
```

flux-middle/src/rustc/ty/subst.rs
```diff
@@ -55,6 +55,15 @@ impl Subst for GenericArg {
         match self {
             GenericArg::Ty(ty) => GenericArg::Ty(ty.subst(substs)),
             GenericArg::Lifetime(re) => GenericArg::Lifetime(re.subst(substs)),
+            GenericArg::Const(c) => GenericArg::Const(c.subst(substs)),
+        }
+    }
+}
+impl Subst for Const {
```
nilehmann, line 62:
```suggestion

impl Subst for Const {
```

flux-tests/tests/neg/surface/const03.rs
```diff
@@ -1,18 +0,0 @@
-// Test definition and checking of const in struct
```
nilehmann, line 1: What happened with this test? It is no longer working?

flux-refineck/src/checker.rs
```diff
@@ -766,7 +765,11 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
 
         let idx = match ty.kind() {
             TyKind::Indexed(BaseTy::Array(_, len), _) => {
-                Index::from(Expr::constant(rty::Constant::from(len.val)))
+                if let Const::Value(value) = len {
+                    Index::from(Expr::constant(rty::Constant::from(value.val)))
+                } else {
+                    tracked_span_bug!("unexpected array length")
```
nilehmann, line 771: So this is the hard part. I've been thinking for a while about generic consts appearing in the refinements. I think supporting that would be very cool.

This is related to global consts (e.g., `const N: i32 = 3`) which we do support inside refinements with `ExprKind:: ConstDefId`. I was thinking that we can maybe represent generic consts with the same variant. Then, we only need to be careful when encoding into fixpoint to differentiate between global consts, for which we assume a value, and generic consts which we leave opaque (we could also have a different variant in `ExprKind`). This change should be relatively simple and it would allow us to remove the `tracked_span_bug` here.

The next step would be to support const generics in our surface syntax. For that, the tricky part is name resolution. It shouldn't be that complicated though, we just need to track those names when desugaring and decide what to do with shadowing.

All this can go into a different PR.

nilehmann (approved):
LGTM

flux-middle/src/rustc/ty.rs
ranjitjhala: done.

flux-refineck/src/checker.rs
ranjitjhala: Agreed! though the other thing would be that _presumably_ `rustc` has some mechanisms to do arithmetic or similar _on_ the const-generics? i.e. do you have some way to specify that `I < N` where both are const-generics?

nilehmann: No, you can't specify *bounds* on const generics (i.e., `I<N`). If you think about it, you end up needing arbitrary SMT in order to do that, that's what I think is really cool to be able to mention them in refinements, because you get it "for free"

flux-tests/tests/neg/surface/const03.rs
ranjitjhala: whoops this was just some glitch on my part, let me restore!

flux-refineck/src/checker.rs
ranjitjhala: Interesting -- so then what's the point of quantifying over the `N`? 

Strictly speaking you can do some lets say, _ad-hoc_ / "lightweight" arithmetic reasoning, e.g. the sort done in GHC's `Nat` solver, or stuff like this: 

https://discourse.llvm.org/t/rfc-enforcing-bounds-safety-in-c-fbounds-safety/70854

Maybe that you can say things like you get passed in two arrays of the same length `N`?

Issue #285, opened 2023-01-02
Unsupported cast when passing in a borrowed slice?


ranjitjhala:
I added a method to `rvec` 

```rust
    #[flux::trusted]
    #[flux::sig(fn(self: &strg RVec<T>[@n], other: &[T][@m]) -> () ensures self: RVec<T>[n + m])]
    pub fn extend_from_slice(&mut self, other: &[T])
    where
        T: Clone,
    {
        self.inner.extend_from_slice(other)
    }
```

but then the following code fails with `Unsupported Statement` 

```rust
#![feature(register_tool)]
#![register_tool(flux)]

#[path = "../lib/rvec.rs"]
mod rvec;
use rvec::RVec;

pub fn test() {
    let mut src: RVec<u8> = rvec![1, 2, 3];
    let bytes: [u8; 2] = [4, 5];
    src.extend_from_slice(&bytes)
}
```

It looks like the following cast needs to be supported?

```rust
 _11 = move _12 as &[u8] (Pointer(Unsize))
``` 

as we get the error

```
 
error[FLUX]: unsupported statement
  --> flux-tests/tests/todo/ext00.rs:13:27
   |
13 |     src.extend_from_slice(&bytes)
   |                           ^^^^^^
   |
   = note: unsupported cast
```

nilehmann:
I think it should relatively easy to support casting `&[T; n]` to `&[T][n]`.

PR #490, opened 2023-09-02
Misc additions

ranjitjhala:
to support various `const`ants and `cast`s, to support stuff like:

- https://github.com/stellar/soroban-examples/blob/main/token/src/balance.rs#L28
- https://github.com/stellar/soroban-examples/blob/main/token/src/event.rs#L9

### DIFF
--- a/flux-middle/src/rty/mod.rs
+++ b/flux-middle/src/rty/mod.rs
@@ -858,7 +858,7 @@ impl Index {
 
 impl From<ValueConst> for Index {
     fn from(value: ValueConst) -> Self {
-        let c = Constant::Int(flux_fixpoint::Sign::Positive, value.val.try_into().unwrap());
+        let c = Constant::from(value.val);
         let expr = Expr::constant(c);
         Index { expr, is_binder: TupleTree::Leaf(false) }
     }
--- a/flux-middle/src/rty/mod.rs
+++ b/flux-middle/src/rty/mod.rs
@@ -259,7 +259,7 @@ pub struct TyS {
     kind: TyKind,
 }
 
-#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
+#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable, Debug)]
 pub enum TyKind {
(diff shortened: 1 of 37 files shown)


flux-refineck/src/checker.rs
```diff
@@ -972,7 +972,8 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
             }
             CastKind::FloatToInt
             | CastKind::IntToFloat
-            | CastKind::Pointer(mir::PointerCast::MutToConstPointer) => {
+            | CastKind::Pointer(mir::PointerCast::MutToConstPointer)
+            | CastKind::Pointer(mir::PointerCast::Unsize) => {
```
nilehmann, line 976: This is unsound for `mir::PointerCast::Unsize`. As implemented, you can use it to coerce a `&mut [i32{v: v > 0}; 10]` to a `&mut [i32]`. I tried to implement this once but got caught up in the many details because there are other forms of [unsized coercions ](https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions). 

We should make sure this is only being applied for unsizing of arrays into slices (I think this is enough for panic) and that the semantics is

```
&mut [T; n] -> &mut [T][n]
&[T;n] -> &[T][n]
```

Related to https://github.com/flux-rs/flux/issues/285

nilehmann (changes requested):

flux-refineck/src/checker.rs
ranjitjhala: ok I made some changes -- see if its what you had in mind?

flux-middle/src/rty/mod.rs
```diff
@@ -856,6 +856,14 @@ impl Index {
     }
 }
 
+impl From<ValueConst> for Index {
+    fn from(value: ValueConst) -> Self {
+        let c = Constant::Int(flux_fixpoint::Sign::Positive, value.val.try_into().unwrap());
```
nilehmann, line 861:
```suggestion
        let c = Constant::from(value.val);
```

nilehmann (approved):
Yeah this looks good now, just one comment

PR #503, opened 2023-09-07
Move `fn` refinement params into EarlyBinders / `refparams_of`

ranjitjhala:
This is the first step in allowing stuff like

```
fn (x: i32, it: I) where I: Iterator<Item = i32{v: x < v}>)
```

The main changes:

0. Use `EarlyBound` names (instead of LateBound/ layer) in function signatures;
1. Stash the refinement-params `RefineParams` in the `genv` via query `refparams_of`;
2. Add the `instantiate_refparams` to ONLY instantiate the `EarlyBound` refparams.

@nilehmann you probably want to pay particularly close attention to the implementation of 
`instantate_refparams` which is done by tweaking the `GenericsSubstFolder` to 
take an `Option<[GenericArg]>` and suitably skipping the relevant cases when `None`.


flux-refineck/src/checker.rs
```diff
@@ -190,16 +190,30 @@ impl<'a, 'tcx, M: Mode> Checker<'a, 'tcx, M> {
...
+        let span = genv.tcx.def_span(def_id);
+
+        let body = genv.mir(def_id.expect_local()).with_span(span)?;
 
         let mut rcx = refine_tree.refine_ctxt_at_root();
 
         let rvid_gen = IndexGen::new();
+        // TODO-EARLY: this is where the "free" names are getting generated, want to also
+        // use those names in the predicates.
+        let params = genv.refparams_of(def_id.expect_local()).with_span(span)?;
```
nilehmann, line 205:
```suggestion
        let params = genv.refparams_of(def_id).with_span(span)?;
```

flux-fhir-analysis/src/lib.rs
```diff
@@ -142,6 +143,21 @@ fn item_bounds(
     Ok(rty::EarlyBinder(conv::conv_opaque_ty(genv, local_id.to_def_id(), opaque_ty, &wfckresults)?))
 }
 
+fn refparams_of(genv: &GlobalEnv, local_id: LocalDefId) -> QueryResult<rty::RefParams> {
```
nilehmann, line 146: I still think we don't need this and we can put it as an extra field in `rty::Generics`. We would avoid the extra query and type.

To do that we just need to pass a `&[fhir::RefineParam]` (the result of `genv.map().get_refine_params()`) to `conv::conv_generics`. Everywhere else we create a `rty::Generics` we just fill the extra field with an empty list.

For easy access we could still have a method in `GlobalEnv`

```rust
fn refine_params_of(self: &GlobalEnv, def_id: DefId) -> QueryResult<List<RefineParam>> {
    self.generics_of(def_id).refine_params.clone()
}
```

nilehmann (changes requested):
This is looking good! 

I have several comments, maybe the most controversial is that I think we should get rid of `RefParams` both in `fhir` and `rty`. 

Don't forget to uncollapse hidden comments.

flux-middle/src/rty/projections.rs
```diff
@@ -42,13 +42,15 @@ impl<'sess, 'tcx> ProjectionTable<'sess, 'tcx> {
...
+        param_env: &ParamEnv,
         t: &T,
     ) -> Result<Self, QueryErr> {
         let mut preds = FxHashMap::default();
 
         let mut vec = vec![];
         // 1. Insert generic predicates of the callsite `def_id`
-        vec.push(genv.predicates_of(def_id)?.skip_binder().predicates);
+        // TODO-EARLY vec.push(genv.predicates_of(def_id)?.skip_binder().predicates);
+        vec.push(param_env.predicates.clone());
```
nilehmann, line 53: See See https://github.com/flux-rs/flux/pull/503#discussion_r1319146083

I believe that instead of having this extra `vec` we could have the opaque type bounds already in the param env.

This can go into the next PR that deals with dependencies in opaque types. We are skipping the binder in the bounds a few lines down, so I wonder if that early binder should be eliminated together with the binder on function predicates.

flux-desugar/src/lib.rs
```diff
@@ -153,8 +154,12 @@ pub fn desugar_fn_sig(
         // See crate level comment
         genv.map().insert_generics(def_id, generics);
 
+        // DESUGAR FN_SIG (needs to happen AFTER desugaring generics)
         let (generic_preds, fn_sig) = cx.desugar_fn_sig(&fn_sig, &mut Binders::new())?;
 
+        let ref_params = fhir::RefParams { params: fn_sig.params.clone() };
+        genv.map().insert_refparams(def_id, ref_params);
```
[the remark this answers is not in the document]
ranjitjhala: Yes, this makes sense -- I wonder what your strategy is for the `rty` level though, because there are no explicit `.params` there, is your point that we should just add that as a field to the `rty::FnSig`?

flux-middle/src/rty/mod.rs
```diff
@@ -85,6 +103,9 @@ pub struct GenericPredicates {
     pub predicates: List<Clause>,
 }
 
+/// An alias for the "EarlyBinder-instantiated" GenericPredicates of a DefId
+pub type ParamEnv = GenericPredicates;
```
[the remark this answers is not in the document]
ranjitjhala: Actually, I *removed* the `ParamEnv` entirely, instead storing the `refine: &[Expr]` list and passing that into `normalize` in order to `instantiate_refine` the `EarlyBind<Predicates>`..

flux-desugar/src/lib.rs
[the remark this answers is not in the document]
nilehmann: For `rty` I would add a field to `rty::Generics` and put it there as described here https://github.com/flux-rs/flux/pull/503#discussion_r1319162997

flux-middle/src/rty/subst.rs
```diff
@@ -332,7 +333,14 @@ impl TypeFolder for GenericsSubstFolder<'_> {
     fn fold_ty(&mut self, ty: &Ty) -> Ty {
         match ty.kind() {
             TyKind::Param(param_ty) => self.ty_for_param(*param_ty),
-            TyKind::Indexed(BaseTy::Param(param_ty), idx) => self.bty_for_param(*param_ty, idx),
+            TyKind::Indexed(BaseTy::Param(param_ty), idx) => {
+                let idx = idx.try_fold_with(self).into_ok();
+                if self.generics.is_some() {
```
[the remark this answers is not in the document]
ranjitjhala: Yes, this was a bit delicate. 

I had a failing test `refined_type_var00` because of the following:

```rust
#[flux::sig(fn<T as base>(b: bool, x: T[@n], y: T[@m]) -> T[if b { n } else { m }])]
pub fn choose<T>(b: bool, x: T, y: T) -> T {
    if b {
        x
    } else {
        y
    }
}

pub fn test01() {
    assert(choose(true, 0, 1) == 0);
    assert(choose(false, 0, 1) == 1);
}
```

Now, at the callsite `choose(true, 0, 1)` there are *two* substitutions going on in the signature for `choose` i.e. for the `T[@n]` and `T[@m]`

1. `T` -> `i32{v: ...}`
2. earlybound `n, m` -> fresh-evars.

The trouble is that if you solely rely on the `bty_for_param` code below, then the `n, m` substitutions "get lost" and so those variables are not "solved for" and you get that pesky instantiation error. So instead, we _first_ substitute for the indices, and then let `bty_for_param` do its business.

Maybe I should add this as a comment to the code?

LMK if that makes sense.

flux-fhir-analysis/src/lib.rs
ranjitjhala: makes sense, will do!

flux-middle/src/rty/subst.rs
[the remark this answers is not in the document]
nilehmann: ah yeah, this is delicate. although I still don't follow. the problem you describe sounds related to not calling `fold_with` on the `idx` (which was a bug before). My question is why check for `is_some` here and then again in `bty_for_param`, I feel we can write it like

```rust
let idx = idx.fold_with(folder);
self.bty_for_param(*param_ty, &idx)
```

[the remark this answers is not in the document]
ranjitjhala: duh! of course, fixed.

flux-middle/src/rty/projections.rs
ranjitjhala: yes, leaving this for the `impl` PR... as will be subsumed by it.

ranjitjhala:
Ok, I think I addressed all the issues above *except* for one thing that I'll leave for you: 

```rustc
// TODO(RJ): using `Vec` instead of `List` due to some inscrutable rustc error
fn refine_params_of(genv: &GlobalEnv, local_id: LocalDefId) -> QueryResult<Vec<rty::RefineParam>> {
    Ok(generics_of(genv, local_id)?.refine_params)
}
```

I can't figure out how to use `List` instead of `Vec` here because I get some scary rustc error 😨

flux-middle/src/rty/subst.rs
```diff
@@ -329,10 +330,44 @@ impl TypeFolder for GenericsSubstFolder<'_> {
...
+    do its business.
+
+    */
     fn fold_ty(&mut self, ty: &Ty) -> Ty {
         match ty.kind() {
             TyKind::Param(param_ty) => self.ty_for_param(*param_ty),
-            TyKind::Indexed(BaseTy::Param(param_ty), idx) => self.bty_for_param(*param_ty, idx),
+            TyKind::Indexed(BaseTy::Param(param_ty), idx) => {
+                // See [NOTE:index-subst]
+                let idx = idx.try_fold_with(self).into_ok();
```
nilehmann, line 368:
```suggestion
                let idx = idx.fold_with(self);
```

flux-middle/src/rty/subst.rs
```diff
@@ -329,10 +330,44 @@ impl TypeFolder for GenericsSubstFolder<'_> {
...
+    2. earlybound `n, m` -> fresh-evars.
+
+    The trouble is that if you solely rely on the `bty_for_param` code below,
+    then the `n, m` substitutions "get lost" and so those variables are not
+    "solved for" and you get that pesky instantiation error.
+
+    Instead, we _first_ substitute for the indices, and then let `bty_for_param`
+    do its business.
+
+    */
```
nilehmann, line 362: Prefer line comments over block comments per the style guidelines https://rust-coding-guidelines.github.io/rust-coding-guidelines/safe-guides/code_style/comments/P.CMT.03.html

flux-fhir-analysis/src/conv.rs
```diff
@@ -117,7 +120,11 @@ pub(crate) fn conv_opaque_ty(
     wfckresults: &fhir::WfckResults,
 ) -> QueryResult<List<rty::Clause>> {
     let cx = ConvCtxt::new(genv, wfckresults);
-    let env = &mut Env::new(&[]);
+    let parent = genv.tcx.parent(def_id).as_local().unwrap();
```
nilehmann, line 123:
```suggestion
    let parent = genv.tcx.parent(def_id).expect_local();
```

flux-fhir-analysis/src/lib.rs
```diff
@@ -142,6 +143,11 @@ fn item_bounds(
     Ok(rty::EarlyBinder(conv::conv_opaque_ty(genv, local_id.to_def_id(), opaque_ty, &wfckresults)?))
 }
 
+// TODO(RJ): using `Vec` instead of `List` due to some inscrutable rustc error
+fn refine_params_of(genv: &GlobalEnv, local_id: LocalDefId) -> QueryResult<Vec<rty::RefineParam>> {
+    Ok(generics_of(genv, local_id)?.refine_params)
+}
```
nilehmann, line 149: We don't need this as a query anymore. I was thinking adding the method directly to `GlobalEnv`

flux-middle/src/rty/mod.rs
```diff
@@ -59,10 +59,17 @@ pub use crate::{
 #[derive(Debug, Clone)]
 pub struct Generics {
     pub params: List<GenericParamDef>,
+    pub refine_params: Vec<RefineParam>,
```
nilehmann, line 62: We `Generics` to be cheaply clonable so we need to make this a `List` 

```suggestion
    pub refine_params: List<RefineParam>,
```

flux-refineck/src/checker.rs
```diff
@@ -58,6 +58,8 @@ pub(crate) struct Checker<'ck, 'tcx, M> {
     def_id: DefId,
     /// [`Generics`] of the function being checked.
     generics: Generics,
+    /// [`Expr`]s used to instantiate EarlyBinders for signature of function being checked
+    refparams: Vec<Expr>,
```
nilehmann, line 62: This should become a list

```suggestion
    refparams: List<Expr>,
```

flux-middle/src/global_env.rs
```diff
@@ -107,6 +107,10 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
         self.queries.check_wf(self, flux_id.into())
     }
 
+    pub fn refparams_of(&self, def_id: impl Into<DefId>) -> QueryResult<Vec<rty::RefineParam>> {
+        self.queries.refine_params_of(self, def_id.into())
+    }
```
nilehmann, line 112: My point is that we can get rid of the query and make this

```suggestion
    pub fn refparams_of(&self, def_id: impl Into<DefId>) -> QueryResult<List<rty::RefineParam>> {
        Ok(self.generics_of(def_id)?.refine_params.clone())
    }
```

flux-middle/src/queries.rs
```diff
@@ -190,6 +193,21 @@ impl<'tcx> Queries<'tcx> {
...
+    ) -> QueryResult<Vec<rty::RefineParam>> {
+        run_with_cache(&self.refparams_of, def_id, || {
+            let def_id = genv.lookup_extern(def_id).unwrap_or(def_id);
+            if let Some(local_id) = def_id.as_local() {
+                (self.providers.refine_params_of)(genv, local_id)
+            } else {
+                Ok(vec![])
+            }
+        })
+    }
```
nilehmann, line 209:
```suggestion
```

flux-middle/src/queries.rs
```diff
@@ -94,6 +96,7 @@ pub struct Queries<'tcx> {
     adt_def: Cache<DefId, QueryResult<rty::AdtDef>>,
     generics_of: Cache<DefId, QueryResult<rty::Generics>>,
     predicates_of: Cache<DefId, QueryResult<rty::EarlyBinder<rty::GenericPredicates>>>,
+    refparams_of: Cache<DefId, QueryResult<Vec<rty::RefineParam>>>,
```
nilehmann, line 99:
```suggestion
```

flux-middle/src/queries.rs
```diff
@@ -75,6 +76,7 @@ impl Default for Providers {
             variants_of: |_, _| empty_query!(),
             fn_sig: |_, _| empty_query!(),
             generics_of: |_, _| empty_query!(),
+            refine_params_of: |_, _| empty_query!(),
```
nilehmann, line 79:
```suggestion
```

flux-middle/src/queries.rs
```diff
@@ -52,6 +52,7 @@ pub struct Providers {
     ) -> QueryResult<rty::Opaqueness<rty::EarlyBinder<rty::PolyVariants>>>,
     pub fn_sig: fn(&GlobalEnv, LocalDefId) -> QueryResult<rty::EarlyBinder<rty::PolyFnSig>>,
     pub generics_of: fn(&GlobalEnv, LocalDefId) -> QueryResult<rty::Generics>,
+    pub refine_params_of: fn(&GlobalEnv, LocalDefId) -> QueryResult<Vec<rty::RefineParam>>,
```
nilehmann, line 55:
```suggestion
```

nilehmann (approved):
Looks good. The last couple of comments. We should make the field `refine_params` in `Generics` a list because the struct needs to be cheaply clonable. Not sure what rust error you were getting into but if we get rid of the query I think it should work.

PR #510, opened 2023-09-12
Add `RefineArgs` to the Opaque/AliasTy

ranjitjhala:
To make this work

```rust
#[flux::sig(fn (x:i32) -> impl Iterator<Item = i32{v:x<=v}>)]
pub fn lib(x: i32) -> impl Iterator<Item = i32> {
    Some(x).into_iter()
}

#[flux::sig(fn (k:i32) -> Option<i32{v:k<=v}>)]
pub fn test_client(k: i32) -> Option<i32> {
    let mut it = lib(k);
    it.next()
}
```


ranjitjhala:
The main things to look at are:

* Change to `pub struct AliasTy` (addition of a `refine_args` field

* `NOTE:Opaque-Refine-Args]`

ranjitjhala:
@nilehmann -- I 

* removed the `OpaqueRefineArgs` hack, 
* added the `EVar` method instead,
* added `RefineArgs` to the `fhir::OpaqueDef`

so I think this is ready to review.

flux-middle/src/global_env.rs
```diff
@@ -111,6 +111,14 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
         Ok(self.generics_of(def_id)?.refine_params.clone())
     }
 
+    pub fn refparams_of_parent(
```
nilehmann, line 114: This will work for now, but technically refinement parameters should follow the parent relationship in `rty::Generics`. So what we want in all the places using this function is to iterate over all the refinement parameters refined in the item and the ones in all the transitive parents (in the reverse order, i.e., going from the root and then down). I'd do one of the following:

1. Add to `rty::Generics` a method 
    ```rust
    fn all_refine_params(&self, genv: &GlobalEnv) -> Vec<rty::RefineParam>
    ```
    Then to get all the refinement parameters of an item we do 
    ```rust
    genv.generics_of(def_id).all_refine_params(genv)
    ```

2. Have in `rty::Generics` a method 
    ```rust
    fn refine_param_at(&self, genv: &GlobalEnv, idx: usize) -> rty::RefineParam
    ``` 
    and a method 
    ```rust
    fn refine_param_count(&self) -> usize
    ```
    Then to iterate over all refinement parameters we do
    ```rust
    let generics = genv.generics_of(def_id);
    let refine_count = generics.refine_param_count(genv);
    (0..refine_count).map(|idx| generics.refine_param_at(genv, idx))
    ```
    This is very general and it's what rustc does which we replicate it for type parameters. See 
    https://github.com/flux-rs/flux/blob/594c59d67d27258e4dfc4ce71d4a08d68b9fb1fb/flux-middle/src/rty/mod.rs#L551
    To replicate the same api we will need to store the count of refine parameters in the parent like we do for type params
    https://github.com/flux-rs/flux/blob/main/flux-middle/src/rty/mod.rs#L64

nilehmann (approved):
this looks good. Various comments mostly stylistic

flux-middle/src/global_env.rs
ranjitjhala: I made a separate issue to clean this up!

PR #512, opened 2023-09-14
Add support for earlybinders/refparams for `async` functions

ranjitjhala:
Follow on PR for #510 [branched off that PR so should first deal with that...]


ranjitjhala:
@nilehmann  is there some way to "rebase" this on top of `main` -- this PR should be _very_ small but there's all the junk from `opaque-args` in it?

nilehmann:
@ranjitjhala maybe you can merge main into this branch. But I don't know exactly because I don't understand what's going on.

ranjitjhala:
Ah yes, ok merging in `main` did the trick -- now its a tiny PR (main change is the tests...) -- so you can take a look. [ Main thing is to pass in the `refparams` into the closures, rather nice that it all works out after that!

nilehmann (changes requested):
I don't like this. I feel the better way to solve this is to implement #513.

Closures and generators inherit the refinement parameters of the function they are defined in, so #513 should cleanly solve the issue here (I think).

nilehmann:
hmmm on second thought, checking the closure/generator with different free variables could mess up the generated constraint.

I'll sleep on it.

ranjitjhala:
The other #513 is orthogonal to this -- it is _essential_ that the closure/generator be checked with the same free variables, because those are also what is tucked inside the tupled-args for the closure, and in fact, for generators, there is no other explicit "call" so really, that's the only way you can "connect" the actual definitions with what is in the closure.

nilehmann (approved):
Yeah sorry, I got confused.

Still, it is somewhat related to #513, rustc has this notion of items inheriting type parameters from their parents which plays very nicely with the way they implicitly quantify over parameters when calling `instantiate_identity`.

PR #521, opened 2023-09-18
Allow single condition in `ensures` clause

ranjitjhala:
e.g. to write 

```rust
#[flux::sig(fn(b:bool) ensures b)]
pub fn assume(b: bool) {
    if !b {
        panic!("assume fails")
    }
}
```

### DIFF
--- a/crates/flux-syntax/src/surface.rs
+++ b/crates/flux-syntax/src/surface.rs
@@ -169,14 +169,6 @@ pub enum Constraint<R = ()> {
     Pred(Expr),
 }
 
-// CUT #[derive(Debug, Default)]
-// CUT pub enum Ensures<R = ()> {
-// CUT     Binds(Vec<(Ident, Ty<R>)>),
-// CUT     Cond(Expr),
-// CUT     #[default]
-// CUT     None,
-// CUT }
-
 #[derive(Debug)]
 pub enum FnRetTy<R = ()> {
     Default(Span),
--- a/crates/flux-syntax/src/surface.rs
+++ b/crates/flux-syntax/src/surface.rs
@@ -155,17 +155,19 @@ pub struct FnSig<R = ()> {
     pub returns: FnRetTy<R>,
     /// example: `*x: i32{v. v = n+1}`
     // pub ensures: Vec<(Ident, Ty<R>)>,
-    pub ensures: Option<Ensures<R>>,
+    pub ensures: Ensures<R>,
     /// example: `where I: Iterator<Item = i32{v:0<=v}>`
     pub predicates: Vec<WhereBoundPredicate<R>>,
     /// source span
     pub span: Span,
 }
 
-#[derive(Debug)]
-pub enum Ensures<R=()> {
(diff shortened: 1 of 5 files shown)


crates/flux-syntax/src/surface.rs
```diff
@@ -153,14 +153,22 @@ pub struct FnSig<R = ()> {
     pub args: Vec<Arg<R>>,
     /// example `i32{v:v >= 0}`
     pub returns: FnRetTy<R>,
-    /// example: `*x: i32{v. v = n+1}`
-    pub ensures: Vec<(Ident, Ty<R>)>,
+    /// example: `*x: i32{v. v = n+1}` or just `x > 10`
+    pub ensures: Ensures<R>,
```
nilehmann, line 157: Any reason to not make this a list of constraints (where a constraint is either `x: T` or `pred`) like in the lower levels?

crates/flux-tests/tests/neg/surface/assume00.rs
```diff
@@ -0,0 +1,20 @@
+#![feature(register_tool)]
+#![register_tool(flux)]
+
+#[flux::sig(fn(b:bool[true]))]
+pub fn assert(_b: bool) {}
```
nilehmann, line 5:
```suggestion
#[flux::sig(fn(bool[true]))]
pub fn assert(_: bool) {}
```

crates/flux-tests/tests/neg/surface/assume00.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[flux::sig(fn(b:bool) ensures b)]
+pub fn assume(b: bool) {
+    if !b {
+        panic!("assume fails")
+    }
+}
```
nilehmann, line 12: We should add a test where the `ensures` fails

```rust
#[flux::sig(fn(x: i32) ensures x > 0)]
pub fn fail_ensures(x: i32) { }
```

or maybe

```rust
#[flux::sig(fn(b: bool) ensures b)]
pub fn bad_assume(b: bool) { }
```

nilehmann:
This also makes me think about `panic` in general. Normally you would like to prove the absence of panics, but the soroban example may be a compelling use case where you want the opposite. In general, you probably want both

```rust
fn assume(b: bool) ensures b {
    if !b { panic!("failed assume") }
}

fn unreachable<T>() -> T requires false {   // or `fn unreachable() -> ! requires false` in nightly
    panic!("entered unreachable code")
}
```

crates/flux-syntax/src/surface.rs
ranjitjhala: duh. let me do that!

crates/flux-tests/tests/neg/surface/assume00.rs
ranjitjhala: good call!

ranjitjhala:
ok I think I've fixed the above.

nilehmann (approved):
:shipit:

PR #522, opened 2023-09-20
Convert `TraitRef` to `PolyTraitRef`

ranjitjhala:
To support using (lifting) signatures like

```rust
    pub fn fold<B, F>(&self, init: B, f: F) -> B
    where
        F: FnMut(B, &T) -> B,
    {
        self.inner.iter().fold(init, f)
    }
```

### DIFF
--- a/crates/flux-fhir-analysis/src/conv.rs
+++ b/crates/flux-fhir-analysis/src/conv.rs
@@ -327,7 +327,7 @@ impl<'a, 'tcx> ConvCtxt<'a, 'tcx> {
                 if let Some(closure_kind) = self.genv.tcx.fn_trait_kind_from_def_id(trait_id) {
                     self.conv_fn_bound(env, bounded_ty, trait_ref, closure_kind, clauses)
                 } else {
-                    let path = &trait_ref.path;
+                    let path = &trait_ref.trait_ref;
                     self.conv_trait_bound(env, bounded_ty, trait_id, &path.args, clauses)?;
                     self.conv_type_bindings(env, bounded_ty, trait_id, &path.bindings, clauses)
                 }
@@ -364,11 +364,11 @@ impl<'a, 'tcx> ConvCtxt<'a, 'tcx> {
         &self,
         env: &mut Env,
         self_ty: &rty::Ty,
-        trait_ref: &fhir::TraitRef,
+        trait_ref: &fhir::PolyTraitRef,
         kind: rty::ClosureKind,
         clauses: &mut Vec<rty::Clause>,
     ) -> QueryResult<()> {
-        let path = &trait_ref.path;
+        let path = &trait_ref.trait_ref;
         let pred = rty::FnTraitPredicate {
             self_ty: self_ty.clone(),
             tupled_args: self.conv_ty(env, path.args[0].expect_type())?,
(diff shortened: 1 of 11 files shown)


crates/flux-fhir-analysis/src/conv.rs
```diff
@@ -349,33 +350,66 @@ impl<'a, 'tcx> ConvCtxt<'a, 'tcx> {
...
+        env.push_layer(layer);
+
         let pred = rty::FnTraitPredicate {
             self_ty: self_ty.clone(),
             tupled_args: self.conv_ty(env, path.args[0].expect_type())?,
             output: self.conv_ty(env, &path.bindings[0].term)?,
             kind,
         };
-        clauses.push(rty::Clause::new(rty::ClauseKind::FnTrait(pred)));
+        let vars = trait_ref
```
nilehmann, line 407: Can you please leave a comment here
```suggestion
        // FIXME(nlehmann) We should use `tcx.late_bound_vars` here instead of trusting our lowering
        let vars = trait_ref
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -426,12 +426,12 @@ impl GenericArg {
...
 impl Clause {
-    pub fn new(kind: ClauseKind) -> Self {
-        Clause { kind }
+    pub fn new(vars: impl Into<List<BoundVariableKind>>, kind: ClauseKind) -> Self {
+        Clause { kind: Binder::new(kind, vars.into()) }
     }
 
     pub fn kind(&self) -> ClauseKind {
-        self.kind.clone()
+        self.kind.clone().skip_binder()
```
nilehmann, line 434:
```suggestion
        // FIXME(nilehmann) we should deal with the binder in all the places this is used instead of blindly skipping it
        self.kind.clone().skip_binder()
```

crates/flux-middle/src/fhir/lift.rs
```diff
@@ -221,7 +221,7 @@ impl<'a, 'tcx> LiftCtxt<'a, 'tcx> {
     ) -> Result<fhir::WhereBoundPredicate, ErrorGuaranteed> {
         if let hir::WherePredicate::BoundPredicate(bound) = pred {
             if !bound.bound_generic_params.is_empty() {
-                return self.emit_unsupported("higher-rank trait bounds are not supported 3");
+                return self.emit_unsupported("higher-rank trait bounds are not supported");
```
nilehmann, line 224: Btw, when you use `cargo xtask run` you'll see that we report the location an error was emitted, e.g.,

```
error[FLUX]: refinement type error
 --> attic/playground.rs:4:5
  |
4 |     0
  |     ^ a postcondition cannot be proved
-Ztrack-diagnostics: created at crates/flux-refineck/src/lib.rs:114:15   <------- this
```
You can also pass `-Ztrack-diagnostics=y` to enable it if you are not using `cargo xtask run`

nilehmann (approved):
LGTM, but please add the fixmes

crates/flux-middle/src/fhir/lift.rs
ranjitjhala: aha good to know -- does it apply to `panic` etc too? should add to the docs

nilehmann: Both `-Ztrack-diagnostic` and `panic` behavior depends on whether functions are annotated with [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute). I typically add it to functions when I encounter the errors and I think it will make the error more useful. For `panic` though you can always get the full story enabling the backtrace.

PR #531, opened 2023-10-02
Support polymorphic theory functions (set_cup, map_store) etc.

ranjitjhala:
just sets up the infrastructure internally, does not expose polymorphic function sorts to user code yet.

Also doesn't work with implicit coercions e.g. treating a `List` index as `Set<int>` by using the `.elems` selector, 
mostly because (a) I haven't figured out that code yet, and (b) not clear if we should add much/any complexity to 
support that use case...

@nilehmann - should we use some `Binder` like method instead of the `params: usize` business for `FuncSort`?

### DIFF
--- a/crates/flux-middle/src/rty/mod.rs
+++ b/crates/flux-middle/src/rty/mod.rs
@@ -157,10 +157,12 @@ pub enum Sort {
     Tuple(List<Sort>),
     Func(FuncSort),
     App(SortCtor, List<Sort>),
+    Var(usize),
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
 pub struct FuncSort {
+    params: usize,
     inputs_and_output: List<Sort>,
 }
 
@@ -667,7 +669,7 @@ impl Sort {
 impl FuncSort {
     pub fn new(mut inputs: Vec<Sort>, output: Sort) -> Self {
         inputs.push(output);
-        FuncSort { inputs_and_output: List::from_vec(inputs) }
+        FuncSort { params: 0, inputs_and_output: List::from_vec(inputs) }
     }
 
     pub fn inputs(&self) -> &[Sort] {
@@ -1713,6 +1715,7 @@ mod pretty {
                 Sort::Real => w!("real"),
                 Sort::BitVec(w) => w!("bitvec({})", ^w),
                 Sort::Loc => w!("loc"),
+                Sort::Var(n) => w!("@{}", ^n),
                 Sort::Func(sort) => w!("{:?}", sort),
                 Sort::Tuple(sorts) => {
                     if let [sort] = &sorts[..] {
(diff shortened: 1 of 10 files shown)


crates/flux-middle/src/rty/mod.rs
```diff
@@ -157,10 +157,12 @@ pub enum Sort {
     Tuple(List<Sort>),
     Func(FuncSort),
     App(SortCtor, List<Sort>),
+    Var(usize),
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
 pub struct FuncSort {
+    params: usize,
```
nilehmann, line 165: I think it's a good idea to add something like `Binder`, but it doesn't have to be as general, we can have it just for function sorts. Maybe

```rust
pub struct PolyFuncSort {
    params: usize,
    fsort: FuncSort
}
```

and then 

```rust
pub enum Sort {
    ...
    Func(PolyFuncSort),
    ...
}
```

I'd also make `Sort::subst` private and only expose it through `PolyFuncSort`, maybe:

```rust
impl PolyFuncSort {
    pub fn instantiate(&self,  args: &[Sort]) -> FuncSort {
        let inputs_and_output = func_sort
            .inputs_and_output
            .iter()
            .map(|sort| sort.subst(&args))
            .collect();
        fhir::FuncSort { inputs_and_output }
    }
}
```

We will probably also need a `skip_binder`.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -157,10 +157,12 @@ pub enum Sort {
     Tuple(List<Sort>),
     Func(FuncSort),
     App(SortCtor, List<Sort>),
+    Var(usize),
```
nilehmann, line 160: Maybe we can make a wrapper over this `usize` (akin to [`ParamTy`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.ParamTy.html)) and write some documentation around binders

```rust
enum SortVar {
    pub index: usize
}
```

or
 
```rust
newtype_index! {
    pub struct SortVar { }
}
```

crates/flux-tests/tests/lib/rset.rs
```diff
@@ -0,0 +1,41 @@
+#![allow(dead_code)]
+#![flux::defs {
+    fn set_add<T>(x: int, s: Set<T>) -> Set<T> { set_union(set_singleton(x), s) }
```
nilehmann, line 3: I was very confused about how this could possibly be working, but I think it is not being executed. Maybe move it to todo.

nilehmann (changes requested):
I think it's ok to not support coercions here, at least for now, but at some point we need to clean up the story around them. The code for inserting coercions is very messy, I think we need to refactor it to support it in more places. My intuition is that we need to solve the set of constraints as a whole instead of trying to do unification eagerly (in `try_equate`).

We can also decide to drop coercions entirely. We have them mostly for historical reasons so all the examples with `RVec` work. But you relying on them may suggest it's something it's expected to work?

ranjitjhala:
Ok I think I've made those edits.

PR #532, opened 2023-10-03
Eq sort

ranjitjhala:
Restrict sort-instantiation to equality-supporting sorts (i.e. not `FuncSort`) to avoid unpleasant downstream crashes in Fixpoint


nilehmann (approved):
That was easy.

PR #542, opened 2023-10-17
Towards polymorphic sorts

ranjitjhala:
Allow polymorphic sorts, see `rset.rs`, `rmap.rs` and `rmapk.rs` and clients.


ranjitjhala:
The `rset00.rs` tests work as expected. I'll probably add some more tests, and perhaps some WF/error message tests too, but I think you can take a look at this PR.

ranjitjhala:
Currently, we have to explicitly write the generic `<Tiger> { elems :  Set<Tiger>}` in the `refined_by` clause.

Should we require this? (Without it, you get a weird crash later.) But OTOH we could plausibly automatically just "scrape" them from the actual sorts appearing in the `refined_by` ? e.g. `Set<Tiger>` I suppose the trouble is how do you know which one is supposed to be a generic and which one is just some other rust type?


```rust
#[flux::refined_by(<Tiger> { elems: Set<Tiger> })]
// #[flux::refined_by( elems: Set<Tiger> )] //~ ERROR unbound generic `Tiger`
pub struct RSet<Tiger> {
    pub inner: std::collections::HashSet<Tiger>,
}
```

ranjitjhala:
Maybe for now, better to be explicit and throw a WF error if there is an unbound generic.

ranjitjhala:
I decided to go the other way -- I removed the explicit params and instead just `gather` them from the sorts themselves. 
The principle here -- perhaps inspired by the excessive bureaucracy at banks in India -- is we shouldn't be asking users to write down stuff that we can extract from other stuff they've _already_ written down!

ranjitjhala:
@nilehmann - I think I'm done with this, ready for review.

nilehmann:
I'm going to take a look at this now

nilehmann (changes requested):
First batch of comments. 

I've only looked until desugaring. Most comments are related to how we use the `SortResolver`. I think we could reuse it instead of creating a new one every time.

Also as a general comment about `SortResolver`. I think we never have a case where both `generics` and `sort_vars` are `Some` so we could expose that as two construction methods (not sure about the names)


```rust
impl<'a> SortResolver<'a> {
    fn with_sort_vars(sess: &'a FluxSession, sort_decls: &'a SortDecls, sort_vars: &[surface::Ident]) -> Self {
        ... 
    }
    
    // we grab `sort_decls` from `genv` and fetch generics of `owner_id` inside 
    fn with_generics(genv: &'a GlobalEnv<'a, '_>, owner_id: DefId) -> Self {
        ...
    }
}
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -1397,26 +1472,36 @@ impl Binders {
...
         params: impl IntoIterator<Item = &'a surface::RefineParam>,
     ) -> Result<Self> {
         let mut binders = Self::new();
         binders.push_layer();
-        binders.insert_params(genv, params)?;
+        binders.insert_params(genv, None, sort_vars, params)?;
         Ok(binders)
     }
 
     fn insert_params<'a>(
```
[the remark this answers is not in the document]
ranjitjhala: have made it a field as suggested!

crates/flux-middle/src/rty/mod.rs
```diff
@@ -771,6 +779,10 @@ impl<T> Binder<T> {
         self.value
     }
 
+    pub fn skip_binder_as_ref(&self) -> &T {
```
nilehmann, line 782: You can do the same with `.as_ref().skip_binder()`. That's just a couple of extra characters longer, I'm not sure we should ahve the extra method, but also not strongly against it

crates/flux-desugar/src/desugar.rs
```diff
@@ -1581,21 +1633,65 @@ fn index_sort(
...
+    match res {
+        fhir::Res::PrimTy(PrimTy::Int(_) | PrimTy::Uint(_)) => Some(fhir::Sort::Int),
+        fhir::Res::PrimTy(PrimTy::Bool) => Some(fhir::Sort::Bool),
+        fhir::Res::PrimTy(PrimTy::Float(..) | PrimTy::Str | PrimTy::Char) => Some(fhir::Sort::Unit),
+        fhir::Res::Def(DefKind::TyAlias { .. } | DefKind::Enum | DefKind::Struct, def_id) => {
+            // TODO: duplication with sort_of_path
+            let mut sort_args = vec![];
+            if let Some(generics) = genv.map().get_generics(def_id) {
+                for (param, arg) in generics.params.iter().zip(&path.generics) {
+                    if let GenericParamKind::SplTy = param.kind {
```
nilehmann, line 1660: When mixing types from different IRs in the same file prefer qualifying them.

```suggestion
                    if let fhir::GenericParamKind::SplTy = param.kind {
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -1581,21 +1633,65 @@ fn index_sort(
...
+                        sort_args.push(sort);
+                    }
+                }
+            };
+            Some(fhir::Sort::Record(def_id, List::from_vec(sort_args)))
+        }
+        fhir::Res::Def(DefKind::TyParam, def_id) => {
+            let param = genv.get_generic_param(def_id.expect_local());
+            match &param.kind {
+                fhir::GenericParamKind::BaseTy | fhir::GenericParamKind::SplTy => {
```
nilehmann, line 1675: `fhir::Sort::Param` represents the sort of a parameter of kind `BaseTy`. `SplTy` does not have a sort in this sense. I tested this and all test pass.

It should be possible to have a concept of a sort associated to a `SplTy` but that should be a different variant. Note that returning `None` for `SplTy` is conservative as we would be rejecting things that could potentially be accepted.


```suggestion
                fhir::GenericParamKind::BaseTy => {
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -1581,21 +1633,65 @@ fn index_sort(
...
+    let res = resolver_output.path_res_map[&path.node_id];
+
+    match res {
+        fhir::Res::PrimTy(PrimTy::Int(_) | PrimTy::Uint(_)) => Some(fhir::Sort::Int),
+        fhir::Res::PrimTy(PrimTy::Bool) => Some(fhir::Sort::Bool),
+        fhir::Res::PrimTy(PrimTy::Float(..) | PrimTy::Str | PrimTy::Char) => Some(fhir::Sort::Unit),
+        fhir::Res::Def(DefKind::TyAlias { .. } | DefKind::Enum | DefKind::Struct, def_id) => {
+            // TODO: duplication with sort_of_path
+            let mut sort_args = vec![];
+            if let Some(generics) = genv.map().get_generics(def_id) {
```
nilehmann, line 1658: This code is assuming that external types don't have polymorphic sorts. That's wrong and we should use `rty::Generics` instead. However, I'm ok with leaving it like this for now. My intention is to get rid of this function eventually as the logic is too subtle to have it duplicated.

crates/flux-refineck/src/constraint_gen.rs
```diff
@@ -146,6 +147,26 @@ impl<'a, 'tcx> ConstrGen<'a, 'tcx> {
...
+    fn check_generic_args(
+        &self,
+        did: DefId,
+        generic_args: &[GenericArg],
+    ) -> Result<(), CheckerErrKind> {
+        let generics = self.genv.generics_of(did)?;
+        for (idx, arg) in generic_args.iter().enumerate() {
+            let param = generics.param_at(idx, self.genv)?;
+            match param.kind {
+                GenericParamDefKind::SplTy | GenericParamDefKind::BaseTy => {
```
nilehmann, line 159: This feels weird. If the param is `GenericParamDefKind::BaseTy` we should only be checking that the argument is `GenericArg::BaseTy`. We only need to go into the type for parameters of kind `SplTy`

crates/flux-middle/src/global_env.rs
```diff
@@ -180,63 +183,114 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
...
-                self.sort_of_self_ty(self_ty)
-            }
-            fhir::Res::Def(DefKind::TyParam, def_id) => {
-                let param = self.get_generic_param(def_id.expect_local());
-                match &param.kind {
-                    fhir::GenericParamKind::BaseTy => Some(fhir::Sort::Param(def_id)),
-                    fhir::GenericParamKind::Type { .. } | fhir::GenericParamKind::Lifetime => None,
-                }
+                let mut sort_args = vec![];
+                if let Some(generics) = self.map().get_generics(def_id) {
```
nilehmann, line 217: Same comment here than for `sort_of_surface_path`. We can revisit later.

crates/flux-middle/src/global_env.rs
```diff
@@ -180,63 +183,114 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
...
+
+    pub fn sort_of_self_ty_alias(&self, alias_to: DefId) -> Option<fhir::Sort> {
+        let self_ty = self.tcx.type_of(alias_to).instantiate_identity();
+        self.sort_of_self_ty(alias_to, self_ty)
+    }
+
+    fn sort_of_generic_param(&self, def_id: DefId) -> Option<fhir::Sort> {
+        let param = self.get_generic_param(def_id.expect_local());
+        match &param.kind {
+            fhir::GenericParamKind::BaseTy | fhir::GenericParamKind::SplTy => {
```
nilehmann, line 244: Same comment here about `fhir::Sort::Param` not quite being the way we should represent the sort of a `SplTy`

nilehmann (changes requested):
@ranjitjhala  Another batch of comments. Looking better!

I still haven't gone through the changes in `flux-fhir-analysis` and wf.

crates/flux-desugar/src/desugar.rs
nilehmann: Another though. There's an implicit _synchronization_ here that maps the generic arguments of the path to the generic sorts of the index. Maybe we should make that mapping explicit. For example, instead of iterating over the generic args of the path, we could iterate over `RefinedBy::sort_params`. If we do that, only `RefinedBy` needs to be generic of the order/mapping (this makes https://github.com/flux-rs/flux/pull/542#discussion_r1376787104 obsolete)

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -382,11 +383,53 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
...
+    fn check_generic_args_kinds(
+        &self,
+        adt_id: Option<DefId>,
+        args: &[fhir::GenericArg],
+    ) -> Result<(), ErrorGuaranteed> {
+        if let Some(def_id) = adt_id &&
+           let Ok(generics) = self.genv.generics_of(def_id)
+        {
+           for (arg, param) in args.iter().zip(generics.params.iter()) {
+             if param.kind == GenericParamDefKind::SplTy {
```
nilehmann, line 416:
```suggestion
           for (arg, param) in iter::zip(args, &generics.params) {
```

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -382,11 +383,53 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
...
+             }
+           }
+        }
+        Ok(())
+    }
+
     fn check_generic_args(
         &mut self,
         infcx: &mut InferCtxt,
+        adt_id: Option<DefId>,
```
nilehmann, line 429: This function should take `DefId` instead of `Option<DefId>`. For anything with generic arguments, we should be able to get its generic parameters and check the kinds. I'll comment at the respective call site how this function should be called.

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -445,15 +488,20 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
...
             | fhir::Res::Def(..)
             | fhir::Res::PrimTy(..) => {}
         }
         let snapshot = self.xi.snapshot();
-        let args = self.check_generic_args(infcx, &path.args);
+        let adt_id = match &path.res {
+            // TODO:Enums
+            fhir::Res::Def(DefKind::Struct, did) => Some(*did),
+            _ => None,
+        };
```
nilehmann, line 503: `fhir::Res::PrimTy`, `fhir::Res::SelfTyAlias` and `SelfTyParam` should never have generic arguments, it's an instant error to have them, so we can skip calling `self.check_generic_args` in those cases. This assumption is verified in `annot_check` when zipping it against the rust signature.

Most `DefKind`s have generics (or cannot appear in a path resolution) so I would filter `DefKind::TyParam` and call `self.check_generic_args` on the rest of `fhir::Res::Def`

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -234,7 +235,7 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
         match bound {
             fhir::GenericBound::Trait(trait_ref, _) => self.check_path(infcx, &trait_ref.trait_ref),
             fhir::GenericBound::LangItemTrait(_, args, bindings) => {
-                self.check_generic_args(infcx, args)?;
+                self.check_generic_args(infcx, None, args)?;
```
nilehmann, line 238: We can get the `DefId` for the lang item like so: `let def_id = self.genv.tcx.require_lang_item(*lang_item, None);`

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -362,7 +363,7 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
             }
             fhir::TyKind::OpaqueDef(_, args, _refine_args, _) => {
                 // TODO sanity check the _refine_args (though they should never fail!) but we'd need their expected sorts
-                self.check_generic_args(infcx, args)
+                self.check_generic_args(infcx, None, args)
```
nilehmann, line 366: Call `check_generics_args` with the `item_id` of the `OpaqueDef` (`item_id.onwer_id.to_def_id()`)

nilehmann (commented):
Comments about wf

crates/flux-fhir-analysis/src/conv.rs
```diff
@@ -193,12 +194,24 @@ pub(crate) fn conv_generics(
     })
 }
 
+fn sort_args_for_adt(genv: &GlobalEnv, def_id: impl Into<DefId>) -> List<fhir::Sort> {
+    let mut sort_args = vec![];
+    for param in &genv.tcx.generics_of(def_id.into()).params {
+        if let rustc_middle::ty::GenericParamDefKind::Type { .. } = param.kind {
+            sort_args.push(fhir::Sort::Param(param.def_id));
```
nilehmann, line 201: This made me realize we are using `fhir::Sort::Param` for two unrelated reasons

1) The sort of a parameter of kind base
2) To bind the sort in a `refined_by` to a corresponding generic

I believe we should use `fhir::Sort::Var` for the second one. Let me elaborate. The annotation `refined_by(elems: Set<T>)` implicitly creates a sort `record RSet<S> { elems: Set<S> }`  this is analogous `fn foo<s>(....)` so it would be consistent to use `fhir::Sort::Var`. We will need a mapping between the `DefId` of the type parameter and the index of the sort it corresponds to so we can implement `sort_of_path` (related to https://github.com/flux-rs/flux/pull/542#discussion_r1376838230). As a nice side effect, I think we could get rid of all the `param_sort` methods in `fhir`.

crates/flux-middle/src/fhir.rs
```diff
@@ -497,7 +498,8 @@ pub enum Sort {
     Var(usize),
     /// A record sort corresponds to the sort associated with a type alias or an adt (struct/enum).
     /// Values of a record sort can be projected using dot notation to extract their fields.
-    Record(DefId),
+    /// the List<Sort> is for the type parameters of (generic) record sorts
+    Record(DefId, List<Sort>),
```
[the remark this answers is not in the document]
ranjitjhala: see #548

crates/flux-middle/src/fhir.rs
```diff
@@ -631,9 +644,8 @@ impl SortCtor {
 impl Ty {
     pub fn as_path(&self) -> Option<&Path> {
         match &self.kind {
-            TyKind::BaseTy(BaseTy {
-                kind: BaseTyKind::Path(QPath::Resolved(None, path)), ..
-            }) => Some(path),
+            TyKind::BaseTy(bty) | TyKind::Indexed(bty, _) => bty.as_path(),
+            TyKind::Constr(_, ty) | TyKind::Exists(_, ty) => ty.as_path(),
```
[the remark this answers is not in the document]
ranjitjhala: reverted!

crates/flux-middle/src/rty/mod.rs
```diff
@@ -205,7 +206,7 @@ pub struct AdtDef(Interned<AdtDefData>);
 #[derive(Debug, Eq, PartialEq, Hash, TyEncodable, TyDecodable)]
 pub struct AdtDefData {
     invariants: Vec<Invariant>,
-    sort: Sort,
+    sort: Sort, // TODO: Binder<Sort>
```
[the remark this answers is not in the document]
ranjitjhala: I thought we'd have to worry about `Var` inside those `Sort` eventually.

crates/flux-middle/src/fhir.rs
```diff
@@ -781,26 +816,83 @@ impl RefinedBy {
...
+
+        let mut params = vec![];
+        for param in &generics.params {
+            let def_id = param.def_id;
+            if let rustc_middle::ty::GenericParamDefKind::Type { .. } = param.kind &&
+               sort_params.contains(&def_id)
+            {
+                params.push(def_id);
+            }
+        }
```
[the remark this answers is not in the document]
ranjitjhala: may be redundant with using `Var`.

crates/flux-middle/src/rty/mod.rs
ranjitjhala: removed.

crates/flux-desugar/src/desugar.rs
ranjitjhala: done!

crates/flux-desugar/src/desugar.rs
ranjitjhala: done!

crates/flux-refineck/src/constraint_gen.rs
ranjitjhala: restricted the check to only `BaseTy`

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -382,11 +383,53 @@ impl<'a, 'tcx> Wf<'a, 'tcx> {
...
+           let Ok(generics) = self.genv.generics_of(def_id)
+        {
+           for (arg, param) in args.iter().zip(generics.params.iter()) {
+             if param.kind == GenericParamDefKind::SplTy {
+                if let fhir::GenericArg::Type(ty) = arg {
+                   self.check_ty_is_base(ty)?;
+                }
+             }
+           }
+        }
```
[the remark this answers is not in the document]
ranjitjhala: done!

crates/flux-fhir-analysis/src/wf/mod.rs
ranjitjhala: done.

crates/flux-fhir-analysis/src/wf/mod.rs
ranjitjhala: done

crates/flux-fhir-analysis/src/conv.rs
ranjitjhala: done. (I assume you meant use `subst` instead of `param_subst`? (and yes we can I removed all the `param_subst` stuff)

ranjitjhala:
Ok I think I addressed the various comments; including representing the `T` inside `refined_by` with `Var`.
One unpleasant thing is that to do so, I'm now requiring an *explicit* list of `sort_params` i.e.

```rust
#[flux::refined_by(<K,V>{keys: Set<K>, vals: Map<K, V>})]
```

instead of automatically finding the `K,V` from the generics (its a bit tedious as you have to first "gather"
the relevant `Vec<Ident>` using the generics -- still, not impossible by any means...)

Not clear if we want to write this explicitly -- more the user has to write, the more opportunity for error!

Any thoughts?

crates/flux-refineck/src/constraint_gen.rs
```diff
@@ -146,6 +147,26 @@ impl<'a, 'tcx> ConstrGen<'a, 'tcx> {
...
+        for (idx, arg) in generic_args.iter().enumerate() {
+            let param = generics.param_at(idx, self.genv)?;
+            match param.kind {
+                GenericParamDefKind::BaseTy => {
+                    if !arg.is_valid_base_arg() {
+                        return Err(CheckerErrKind::InvalidGenericArg);
+                    }
+                }
+                _ => continue,
+            }
```
nilehmann, line 165: This is still odd, but now I understand the problem. Turns out that implementing this is a bit more involved. 

In this test

```rust
fn make_eq_hash() -> impl Eq + Hash
    0
}

fn id<T as base>(x: T) -> T[x] {
    x
}


fn test00() {
    let z = mk_eq_hash();
    id(z);
}
```

At the point we call `id` we have `λ. impl Eq + Hash`  as the generic argument, i.e., a lambda with no parameters. That's already wrong at many levels. Arguments of kind base should be a lambda with exactly one parameter. So there's a problem when we generate the instantiation here https://github.com/flux-rs/flux/blob/main/crates/flux-middle/src/global_env.rs#L341. I think we need to do this check before we instantiate the arguments or make the instantiation itself check it.

We can ponder this later though. The change is a bit orthogonal to the rest of the changes in the PR. I would revert the changes and remove the associated tests, so we can tackle it later.

crates/flux-middle/src/fhir.rs
```diff
@@ -537,6 +538,17 @@ impl PolyFuncSort {
...
+    // fn mk_param_subst(&self, subst: &SortParamSubst) -> PolyFuncSort {
+    //     let params = self.params;
+    //     let args = self
+    //         .fsort
+    //         .inputs_and_output
+    //         .iter()
+    //         .map(|arg| arg.param_subst(subst))
+    //         .collect();
+    //     PolyFuncSort { params, fsort: FuncSort { inputs_and_output: args } }
+    // }
```
nilehmann, line 551:
```suggestion
```

crates/flux-middle/src/fhir.rs
```diff
@@ -725,10 +742,17 @@ impl Ident {
 /// in a definition.
 ///
 /// [early bound]: https://rustc-dev-guide.rust-lang.org/early-late-bound.html
+///
+/// Sort parameters e.g. #[flux::refined_by( elems: Set<T> )] tracks the mapping from
+/// bound Var -> Generic id. e.g. if we have RMap<K, V> refined_by(keys: Set<K>)
+/// then RMapIdx = forall #0. { keys: Set<#0> }
+/// and sort_params = vec![0]  i.e. maps Var(0) to Generic(0)
+
```
nilehmann, line 750:
```suggestion
```

crates/flux-middle/src/fhir.rs
```diff
@@ -725,10 +742,17 @@ impl Ident {
...
 #[derive(Clone, Debug, TyEncodable, TyDecodable)]
 pub struct RefinedBy {
     pub def_id: DefId,
     pub span: Span,
+    sort_params: Vec<usize>,
```
nilehmann, line 755: hmm, I don't like this representation too much. I thought this was going to be a `Vec<DefId>`. Related to https://github.com/flux-rs/flux/pull/542#issuecomment-1791049978 perhaps.

In any case, I'm not too worried about this. The code is probably going to have a major refactor anyways.

nilehmann (approved):
Ok, this looks good. There are still many details I'm not happy with, but I'm ok with merging so we can move forward.

About the syntax, as always, I'm not too worried about it, we can always change it later. I'm more concerned about having enough pieces of the puzzle in place. 

That being said, I do think is a bit redundant to annotate the parameters. It also makes it that you need to check more things. For example, [this line](https://github.com/flux-rs/flux/pull/542/files#diff-c78159aae1ce76438bf8e9bf9a7a6a2876ce7bc72420f8745500bf54aba699bcR131) will crash if you don't use the same names.

crates/flux-refineck/src/constraint_gen.rs
ranjitjhala: made a fresh issue

crates/flux-middle/src/fhir.rs
ranjitjhala: fixed.

crates/flux-refineck/src/constraint_gen.rs
ranjitjhala: made separate issue

PR #563, opened 2023-11-27
fix links in the book

asmarcz:


nilehmann:
Thanks for the PR @asmarcz!

nilehmann (approved):

PR #574, opened 2023-12-07
add support for lnot on ints and uints

### LINKED ISSUE #571: bitwise not crashes verification
```Rust
fn test(x: i32) -> i32 {
    !x
}
```
fails with the error 
```
thread 'rustc' panicked at crates/flux-refineck/src/sigs/mod.rs:125:18:
no entry found for key
```

enjhnsn2:
fixes #571

### DIFF
--- a/crates/flux-refineck/src/sigs/default.rs
+++ b/crates/flux-refineck/src/sigs/default.rs
@@ -171,12 +171,12 @@ pub(crate) fn mk_not() -> (mir::UnOp, Sig<1>) {
 pub(crate) fn mk_lnot() -> impl IntoIterator<Item = (mir::UnOp, Sig<1>)> {
     let int_lnots = INT_TYS.map(|int_ty| {
         define_btys! { let Int = BaseTy::Int(int_ty); };
-        (mir::UnOp::Not, s!(fn(a: Int) -> Int[E::tt()]))
+        (mir::UnOp::Not, s!(fn(a: Int) -> Int{v: E::tt()}))
     });
 
     let uint_lnots = UINT_TYS.map(|uint_ty| {
         define_btys! { let Uint = BaseTy::Uint(uint_ty); };
-        (mir::UnOp::Not, s!(fn(a: Uint) -> Uint[E::tt()]))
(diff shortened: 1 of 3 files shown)


crates/flux-refineck/src/sigs/default.rs
```diff
@@ -166,3 +167,17 @@ pub(crate) fn mk_not() -> (mir::UnOp, Sig<1>) {
     define_btys! { let bool = BaseTy::Bool; }
     (mir::UnOp::Not, s!(fn(a: bool) -> bool[a.not()]))
 }
+
+pub(crate) fn mk_lnot() -> impl IntoIterator<Item = (mir::UnOp, Sig<1>)> {
+    let int_lnots = INT_TYS.map(|int_ty| {
+        define_btys! { let Int = BaseTy::Int(int_ty); };
+        (mir::UnOp::Not, s!(fn(a: Int) -> Int[E::tt()]))
```
nilehmann, line 174: This needs to be

```suggestion
        (mir::UnOp::Not, s!(fn(a: Int) -> Int{ v: E::tt() }))
```


`Int[E::tt()]` would mean the integer is indexed by a `bool` which is not well-formed.

nilehmann (changes requested):
@enjhnsn2 looks good, but some errors need to be fixed.

We also need to add the signatures [here](https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/sigs/overflow.rs/#L30). We have a different set of signatures depending on whether overflow checking is enabled. Overflow is not relevant for `!` but we still need to duplicate the signature.

Also, can you add a test [here](https://github.com/flux-rs/flux/blob/main/crates/flux-tests/tests/pos/surface/binop.rs#L81)?

enjhnsn2:
it was my impression (from context) that `E::tt()` told Flux to treat the operorator as an uninterpreted function, as in the other bitwise operations:
```Rust
// BIT
(BitAnd, s!(fn(a: Int, b: Int) -> Int{v: E::tt()})),
(BitOr,  s!(fn(a: Int, b: Int) -> Int{v: E::tt()})),
```

nilehmann (approved):
Looks good to me. I'm going to mer it.

PR #575, opened 2023-12-13
Improvements to `extern_spec`

ranjitjhala:
That allow, refining (`impl` of) types with 

1. Default type parameters,
2. Bounds in type parameters,
3. Multiple `impl` blocks

See 

* (pos)[crates/flux-tests/tests/pos/surface/rset03.rs]
* (pos)[crates/flux-tests/tests/pos/surface/vec00.rs]
* (neg)[crates/flux-tests/tests/neg/surface/rset03.rs]

*Lots* of broken things remain (see TODO)


book/src/dev/develop.md
```diff
@@ -151,3 +151,21 @@ check_crate
...
+#[extern_spec]
+#[flux::refined_by(elems: Set<T>)]
+struct HashSet<T, S = RandomState>;
+```
+
+and you want to see what the `extern_spec` macro expands it out to, then run
+
+```
+cargo x run -- -Zunpretty=expanded path/to/file.rs 
+```
```
nilehmann, line 170: There's an xtask command for this `cargo x expand path/to/file.rs`

crates/flux-middle/src/queries.rs
```diff
@@ -194,10 +194,10 @@ impl<'tcx> Queries<'tcx> {
     pub(crate) fn generics_of(
         &self,
         genv: &GlobalEnv,
-        def_id: DefId,
+        def_id0: DefId,
```
nilehmann, line 197:
```suggestion
        def_id: DefId,
```

nilehmann (approved):
LGTM, just some small comments.

ranjitjhala:
@nilehmann I added some minor tweaks to support a bit of `vec` in case you want to check.

crates/flux-refineck/src/type_env.rs
```diff
@@ -508,8 +508,8 @@ impl BasicBlockEnvShape {
             (GenericArg::BaseTy(_), GenericArg::BaseTy(_)) => {
                 tracked_span_bug!("generic argument join for base types is not implemented")
             }
-            (GenericArg::Lifetime(re1), GenericArg::Lifetime(re2)) => {
-                debug_assert_eq!(re1, re2);
+            (GenericArg::Lifetime(re1), GenericArg::Lifetime(_re2)) => {
+                // CHECK: debug_assert_eq!(re1, re2);
```
nilehmann, line 512: Just realize this is inside join and not subtyping. This feels bad. Can you minimize an example that needs this change?

ranjitjhala: Yes, its rather tiny:

```rust
pub fn baz(v: &Vec<i32>) {
    for x in v {}
}
```

ranjitjhala:
The error also happens with iterating over RSet .. wonder why not with
RVec???

- Ranjit.

crates/flux-refineck/src/type_env.rs
nilehmann: Ok did some digging, this is indeed a bug somewhere else. 

When calling `Iter::next` we "downgrade" a pointer to a reference ([this line)](https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/constraint_gen.rs#L265). 

So, we start with 

```
r: ptr(x), x: Iter<'?11, i32>
``` 
Then, just before the call, we have 

```
r: &mut Iter<?20, i32>, x: †Iter<'?20, i32>
```

i.e., the pointer `r` is now a mutable reference, and the type of `x` has been strongly updated (and blocked). The issue is that lifetimes should never be strongly updated. The call to `block_with` should update everything but lifetimes. That's what we do in assignments https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/type_env.rs#L172.

The fix is a bit tricky because we are updating through path so we cannot do the same we do in `assign`.

nilehmann: > The error also happens with iterating over RSet .. wonder why not with RVec??? - Ranjit.
> […](#)

The problem happens when we iterate `Vec` by reference. We don't have a by-reference iterator on `RVec`.

ranjitjhala:
ok let me revert that `assert` and the associated tests; its orthogonal to this PR anyways, so I've made a separate issue.

PR #583, opened 2023-12-19
Add support for FreeBSD

jepotter1:
With this change, Flux builds and all tests pass on FreeBSD 13.2-RELEASE.


nilehmann (approved):
Thanks @jepotter1!

PR #586, opened 2023-12-27
Add support for `extern_spec` for `enum`

ranjitjhala:
The following now works

```rust
#[extern_spec]
#[flux::refined_by(b:bool)]
enum Option<T> {
    #[flux::variant(Option<T>[false])]
    None,
    #[flux::variant({T} -> Option<T>[true])]
    Some(T),
}

#[flux::sig(fn(x:Option<T>[true]) -> T)]
pub fn my_unwrap<T>(x: Option<T>) -> T {
    match x {
        Option::Some(v) => v,
        Option::None => never(0),
    }
}

pub fn test3() {
    let x = Option::Some(42);
    let y = my_unwrap(x);
    assert(y == 42);
}
```

### DIFF
--- a/crates/flux-fhir-analysis/src/annot_check.rs
+++ b/crates/flux-fhir-analysis/src/annot_check.rs
@@ -13,17 +13,16 @@ use flux_middle::{
     fhir::{
         self,
         lift::{self, LiftCtxt},
-        WfckResults,
+        Res, WfckResults,
     },
     global_env::GlobalEnv,
 };
 use rustc_data_structures::unord::UnordMap;
 use rustc_errors::IntoDiagnostic;
 use rustc_hir::OwnerId;
-use rustc_middle::ty::TyCtxt;
 
 pub fn check_fn_sig(
-    tcx: TyCtxt,
+    genv: &GlobalEnv,
     sess: &FluxSession,
     wfckresults: &mut WfckResults,
     owner_id: OwnerId,
@@ -32,13 +31,14 @@ pub fn check_fn_sig(
     if fn_sig.lifted {
         return Ok(());
     }
-    let self_ty = lift::lift_self_ty(tcx, sess, owner_id)?;
-    let expected_fn_sig = &lift::lift_fn(tcx, sess, owner_id)?.1.fn_sig;
-    Zipper::new(sess, wfckresults, self_ty.as_ref()).zip_fn_sig(fn_sig, expected_fn_sig)
+    let self_ty = lift::lift_self_ty(genv.tcx, sess, owner_id)?;
+    let expected_fn_sig = &lift::lift_fn(genv.tcx, sess, owner_id)?.1.fn_sig;
+    Zipper::new(sess, wfckresults, self_ty.as_ref()).zip_fn_sig(genv, fn_sig, expected_fn_sig)
 }
 
 pub fn check_alias(
-    tcx: TyCtxt,
+    genv: &GlobalEnv,
+    // tcx: TyCtxt,
     sess: &FluxSession,
     wfckresults: &mut WfckResults,
     ty_alias: &fhir::TyAlias,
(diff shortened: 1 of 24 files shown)


crates/flux-refineck/src/checker.rs
```diff
@@ -122,6 +122,19 @@ enum Guard {
...
+fn adt_def_num_variants(adt_def: &AdtDef) -> usize {
+    let def_id = adt_def.did();
+    let is_extern_enum = format!("{def_id:?}").contains("__FluxExternEnum");
+    let num_variants = adt_def.variants().len();
+    if is_extern_enum {
+        num_variants - 1
+    } else {
+        num_variants
+    }
+}
```
nilehmann, line 136: This should be fixed somewhere else. I believe the underlying problem is that we are creating the `AdtDef` in [`conv::adt_def_for_enum`](https://github.com/flux-rs/flux/blob/867519a6bc3fc7285ce36f82e3e19dee2b27bd2d/crates/flux-fhir-analysis/src/conv.rs#L220) with the fake enum rustc `AdtDef` instead of the real one. 

`adt_def_for_enum` should be aware of extern specs and adjust accordingly. Note that `adt_def_for_struct` has the same issue but doesn't cause problems because extern structs are always opaque.

Ideally, we could encode with an enum whether the spec is local or external. Then we could make the definition of `fhir::EnumDef` something like

```rust
enum LocalOrExtern {
    Local(OwnerId),
    Extern(OwnerId, DefId)
}

enum EnumDef {
    owner_id: LocalOrExtern,
    ...
 }
```

crates/flux-fhir-analysis/src/annot_check.rs
```diff
@@ -29,60 +31,68 @@ pub fn check_fn_sig(
...
             fields.iter().try_for_each_exhaust(|field| {
                 if field.lifted {
                     return Ok(());
                 }
-                let self_ty = lift::lift_self_ty(tcx, sess, struct_def.owner_id)?;
-                Zipper::new(sess, wfckresults, self_ty.as_ref())
-                    .zip_ty(&field.ty, &liftcx.lift_field_def_id(field.def_id)?.ty)
+                let self_ty = lift::lift_self_ty(genv.tcx, sess, struct_def.owner_id)?;
+                Zipper::new(sess, wfckresults, self_ty.as_ref()).zip_ty(
+                    genv,
```
nilehmann, line 68: Instead of threading down `genv` we could add it to a field in the zipper (instead of `FluxSession`)

crates/flux-fhir-analysis/src/annot_check.rs
```diff
@@ -29,60 +31,68 @@ pub fn check_fn_sig(
...
-    let (.., expected_ty_alias) = lift::lift_type_alias(tcx, sess, ty_alias.owner_id)?;
-    Zipper::new(sess, wfckresults, None).zip_ty(&ty_alias.ty, &expected_ty_alias.ty)
+    let (.., expected_ty_alias) = lift::lift_type_alias(genv.tcx, sess, ty_alias.owner_id)?;
+    Zipper::new(sess, wfckresults, None).zip_ty(genv, &ty_alias.ty, &expected_ty_alias.ty)
 }
 
 pub fn check_struct_def(
-    tcx: TyCtxt,
+    genv: &GlobalEnv,
     sess: &FluxSession,
```
nilehmann, line 55: We can remove `sess` if we pass `genv`.

crates/flux-fhir-analysis/src/annot_check.rs
```diff
@@ -283,61 +304,79 @@ impl<'zip> Zipper<'zip> {
...
-                self.zip_ty(self_ty, expected_self_ty)?;
-                self.zip_path(path, expected_path)
+                self.zip_ty(genv, self_ty, expected_self_ty)?;
+                self.zip_path(genv, path, expected_path)
             }
             _ => Err(self.emit_err(errors::InvalidRefinement::from_qpaths(qpath, expected_qpath))),
         }
     }
 
+    fn is_same_res(genv: &GlobalEnv, res: Res, expected: Res) -> bool {
```
nilehmann, line 343: Ok, this is a bit dumb. The only reason we need to check this is that we allow you to write the "return type" for a variant. If we instead had:

```rust
#[flux::refined_by(b: bool)]
enum Option<T> {
    #[flux::variant([false])]
    None,
    #[flux::variant((T) -> [true])]
    Some(T)
}
```
then, this wouldn't cause any problems. Not in this PR, but we should consider changing the syntax.

nilehmann (changes requested):
[One major comment](https://github.com/flux-rs/flux/pull/586#discussion_r1437103826), but overall it looks good. 

I don't mind the hack (I considered myself). I was settling into something more sophisticated and expanding the macro to:

```rust
#[flux::extern(enum)]
mod _extern_option {
    type Option<T> = core::option::Option<T>;
    
    enum FakeOption<T> {
        #[flux::variant(Option<T>[false])]
        None,
        #[flux::variant((T) -> Option<T>[true])
        Some(T)
    }
}
```
but that requires a bit more thought and in the end it's equally hacky.

crates/flux-refineck/src/checker.rs
ranjitjhala: Instead of the above, which messes up the `owner_id`, I just (equivalently) added an extra `extern_id: Option<DefId>` field to the struct which tracks if the spec is external.

```rust
#[derive(Debug)]
pub struct EnumDef {
    pub owner_id: OwnerId,
    pub params: Vec<RefineParam>,
    pub variants: Vec<VariantDef>,
    pub invariants: Vec<Expr>,
    /// Whether this is an extern_spec for some other enum
    pub extern_id: Option<DefId>,
}
```

crates/flux-fhir-analysis/src/conv.rs
```diff
@@ -225,7 +225,8 @@ pub(crate) fn adt_def_for_enum(
     let def_id = enum_def.owner_id;
     let sort_args = sort_args_for_adt(genv, def_id);
     let sort = rty::Sort::tuple(conv_sorts(genv, &genv.index_sorts_of(def_id, &sort_args)));
-    let adt_def = lowering::lower_adt_def(&genv.tcx.adt_def(enum_def.owner_id));
+    let is_extern = enum_def.extern_id.is_some();
+    let adt_def = lowering::lower_adt_def(&genv.tcx.adt_def(enum_def.owner_id), is_extern);
```
nilehmann, line 229: I was thinking something more like:

```suggestion
    let adt_def = if let Some(extern_id) = enum_def.extern_id {
        lowering::lower_adt_def(&genv.tcx.adt_def(extern_id))
    } else {
        lowering::lower_adt_def(&genv.tcx.adt_def(enum_def.owner_id))
    };
```

We should create the `AdtDef` with the (real) extern id if there's one.

ranjitjhala: oh, that's much better, thanks!

crates/flux-fhir-analysis/src/annot_check.rs
```diff
@@ -317,19 +317,33 @@ impl<'zip> Zipper<'zip> {
...
```
nilehmann, line 320: Now that the `GlobalEnv` is part of the `Zipper` you could make this a method.

nilehmann (approved):
LGTM

Issue #588, opened 2023-12-29
Unsound to use fresh KVar for generics with bound


ranjitjhala:
The code below currently verifies, but shouldn't (maybe we're generating fresh kvars for the `Self` at the call to `x.foo` ?)

```rust
pub trait MyTrait {
    fn foo(&self) -> Self;
}

// This "verifies" unsoundly due to the parametricity of `T`.
#[flux::sig(fn<T as base, refine q: T -> bool> (&T{v:q(v)}) -> T{v: q(v)})]
pub fn bar<T: MyTrait>(x: &T) -> T {
    x.foo() //~ ERROR refinement type
}
```

nilehmann:
Another example. The following code verifies because collect has type `forall B: FromIterator. fn(Self) -> B` and we are generating an (unconstrained) kvar for `B`

```rust
#[flux::sig(fn(Vec<i32>) -> Vec<i32[0]>)]
pub fn test(x: Vec<i32>) -> Vec<i32> {
    x.into_iter().collect()
}
```

nilehmann:
This was fixed in https://github.com/flux-rs/flux/pull/623 and a test was added in https://github.com/flux-rs/flux/commit/3869e7f3263d922a46a2755305fd8a9885c80345

PR #589, opened 2023-12-29
Allow refining `Self`

ranjitjhala:
Updates the `Generics` type to include `Self as base` for `trait`, so the following now works:

```rust
#[flux::generics(Self as base)]
pub trait MyTrait {
    #[flux::sig(fn<refine p: Self -> bool>(&Self{v: p(v)}) -> Self{v: p(v)})]
    fn foo(&self) -> Self;
}

#[flux::sig(fn<T as base, refine q: T -> bool> (&T{v:q(v)}) -> T{v: q(v)})]
pub fn bar<T: MyTrait>(x: &T) -> T {
    x.foo()
}
```

### DIFF
--- a/crates/flux-desugar/src/desugar.rs
+++ b/crates/flux-desugar/src/desugar.rs
@@ -200,20 +200,21 @@ enum QPathRes<'a> {
 
 fn self_res(genv: &GlobalEnv, owner: OwnerId) -> SelfRes {
     let def_id = owner.def_id;
-    let kind = genv.tcx.def_kind(def_id);
-    match kind {
-        DefKind::Trait => SelfRes::Param(def_id.to_def_id()),
-        DefKind::Impl { .. } => {
-            if let Some(alias_to) = genv.tcx.opt_parent(def_id.to_def_id())
-                && let Some(sort) = genv.sort_of_self_ty_alias(alias_to)
-            {
-                SelfRes::Alias(sort)
-            } else {
-                SelfRes::None
+    let parent_id = genv.tcx.opt_parent(def_id.to_def_id());
+    if let Some(alias_to) = parent_id {
+        match genv.tcx.def_kind(alias_to) {
+            DefKind::Trait => return SelfRes::Param(alias_to),
+            DefKind::Impl { .. } => {
+                if let Some(sort) = genv.sort_of_self_ty_alias(alias_to) {
+                    return SelfRes::Alias(sort);
+                } else {
+                    return SelfRes::None;
+                }
             }
+            _ => return SelfRes::None,
         }
-        _ => SelfRes::None,
     }
+    return SelfRes::None;
 }
 
 impl<'a, 'tcx> RustItemCtxt<'a, 'tcx> {
--- a/crates/flux-desugar/src/desugar.rs
+++ b/crates/flux-desugar/src/desugar.rs
@@ -420,6 +420,7 @@ impl<'a, 'tcx> RustItemCtxt<'a, 'tcx> {
             params: env.into_root().into_params(self),
             variants,
             invariants,
+            extern_id: enum_def.extern_id,
         })
     }
 
--- a/crates/flux-desugar/src/desugar.rs
+++ b/crates/flux-desugar/src/desugar.rs
@@ -227,14 +227,14 @@ enum QPathRes<'a> {
     NumConst(i128),
 }
 
-fn super_hack_is_impl(def_id: DefId) -> bool {
-    format!("{def_id:?}").contains("::{impl#")
+fn is_impl_item(genv: &GlobalEnv, owner_id: OwnerId) -> bool {
+    let hir_id = genv.tcx.hir().local_def_id_to_hir_id(owner_id.def_id);
+    let node = genv.tcx.hir().get(hir_id);
+    matches!(node, hir::Node::ImplItem { .. })
 }
 
-fn self_sort(genv: &GlobalEnv, parent_id: Option<DefId>) -> Option<fhir::Sort> {
-    if let Some(alias_to) = parent_id
-        && super_hack_is_impl(alias_to)
-    {
+fn self_sort(genv: &GlobalEnv, parent_id: Option<DefId>, is_impl: bool) -> Option<fhir::Sort> {
+    if is_impl && let Some(alias_to) = parent_id {
         genv.sort_of_self_ty_alias(alias_to)
     } else {
         None
@@ -249,10 +249,9 @@ impl<'a, 'tcx> RustItemCtxt<'a, 'tcx> {
         opaque_tys: Option<&'a mut UnordMap<LocalDefId, fhir::OpaqueTy>>,
     ) -> RustItemCtxt<'a, 'tcx> {
         let generics = genv.tcx.generics_of(owner);
-
-        let owner_id = owner.def_id.to_def_id();
-        let parent_id = genv.tcx.opt_parent(owner_id);
-        let self_sort = self_sort(genv, parent_id);
+        let is_impl = is_impl_item(genv, owner);
+        let parent_id = genv.tcx.opt_parent(owner.def_id.to_def_id());
+        let self_sort = self_sort(genv, parent_id, is_impl);
         let sort_resolver = SortResolver::with_generics(
             genv.sess,
             genv.map().sort_decls(),
@@ -284,14 +283,17 @@ impl<'a, 'tcx> RustItemCtxt<'a, 'tcx> {
         lifted_generics: fhir::Generics,
         generics: &surface::Generics,
     ) -> Result<fhir::Generics> {
+        // Step 1: desugar the surface generics by themselves
         let generics = self.desugar_surface_generics(generics)?;
 
+        // Step 2: map each (surface) generic to its specified kind
         let generic_kinds: FxHashMap<_, _> = generics
             .params
             .into_iter()
             .map(|param| (param.def_id, param.kind))
             .collect();
 
+        // Step 3: traverse lifted_generics, using the surface-kind, if specified, and lifted kind otherwise
         let params = lifted_generics
             .params
             .iter()
--- a/crates/flux-desugar/src/desugar.rs
+++ b/crates/flux-desugar/src/desugar.rs
@@ -17,7 +17,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use rustc_hir as hir;
 use rustc_hir::OwnerId;
 use rustc_span::{
-    def_id::{DefId, LocalDefId},
+    def_id::LocalDefId,
     sym::{self},
     symbol::kw,
     Span, Symbol,
@@ -29,7 +29,7 @@ use self::env::{Scope, ScopeId};
 use crate::{
     errors,
     resolver::ResolverOutput,
-    sort_resolver::{SortResolver, SORTS},
+    sort_resolver::{SelfRes, SortResolver, SORTS},
 };
 
 pub fn desugar_qualifier(
@@ -198,17 +198,21 @@ enum QPathRes<'a> {
     NumConst(i128),
 }
 
-fn is_impl_item(genv: &GlobalEnv, owner_id: OwnerId) -> bool {
-    let hir_id = genv.tcx.hir().local_def_id_to_hir_id(owner_id.def_id);
-    let node = genv.tcx.hir().get(hir_id);
-    matches!(node, hir::Node::ImplItem { .. })
-}
-
-fn self_sort(genv: &GlobalEnv, parent_id: Option<DefId>, is_impl: bool) -> Option<fhir::Sort> {
-    if is_impl && let Some(alias_to) = parent_id {
-        genv.sort_of_self_ty_alias(alias_to)
-    } else {
-        None
+fn self_res(genv: &GlobalEnv, owner: OwnerId) -> SelfRes {
+    let def_id = owner.def_id;
+    let kind = genv.tcx.def_kind(def_id);
+    match kind {
+        DefKind::Trait => SelfRes::Param(def_id.to_def_id()),
+        DefKind::Impl { .. } => {
+            if let Some(alias_to) = genv.tcx.opt_parent(def_id.to_def_id())
+                && let Some(sort) = genv.sort_of_self_ty_alias(alias_to)
+            {
+                SelfRes::Alias(sort)
+            } else {
+                SelfRes::None
+            }
+        }
+        _ => SelfRes::None,
     }
 }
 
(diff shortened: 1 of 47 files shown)


crates/flux-middle/src/fhir.rs
```diff
@@ -836,8 +839,8 @@ pub struct Defn {
 }
 
 impl Generics {
-    pub(crate) fn get_param(&self, def_id: LocalDefId) -> &GenericParam {
-        self.params.iter().find(|p| p.def_id == def_id).unwrap()
+    pub(crate) fn get_param(&self, def_id: LocalDefId) -> Option<&GenericParam> {
+        self.params.iter().find(|p| p.def_id == def_id)
```
nilehmann, line 843: You are hiding a problem somewhere else. I think the problem is this test you added

https://github.com/flux-rs/flux/pull/589/files#diff-951d1aeaa0d86a85e279fc5169ff954af98a94aa53e0a6ca824b3c4d3c44ff7cR11-R18

If you specify an abstract refinement you need to specify all the generics of the function. The following works

```rust
#[flux::sig(fn<T, refine q: T -> bool>(&T{v:q(v)}) -> T{v: q(v)})]
```

The issue is that we don't check the refined signature has enough generics. Alternatively, we could lift the rust generics if the flux signature only has refinement parameters.

nilehmann (changes requested):

crates/flux-middle/src/fhir.rs
ranjitjhala: I'm not following this comment. 

1. I put the `Option` in over there because otherwise we just get a strange crash,
2. (In `main`) if you just write the signature you wrote, you get

```
error[FLUX]: type cannot be refined
 --> crates/flux-tests/tests/todo/refined_fn_in_trait_01.rs:7:42
  |
7 | #[flux::sig(fn<T, refine q: T -> bool> (&T{v:q(v)}) -> T{v: q(v)})]
  |                                          ^
-Ztrack-diagnostics: created at crates/flux-desugar/src/desugar.rs:678:37
```

3. (In `main`) if you do write `T as base` then it works... ?

crates/flux-desugar/src/sort_resolver.rs
```diff
@@ -115,6 +123,10 @@ impl<'a> SortResolver<'a> {
             Ok(fhir::Sort::Bool)
         } else if ident.name == SORTS.real {
             Ok(fhir::Sort::Real)
+        } else if let Some(def_id) = self.parent_id
```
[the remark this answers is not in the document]
ranjitjhala: Hmm, do you mean I should add a case `fhir::Sort::SelfParam` for the case when we are using `Self` as a TyParam?

crates/flux-middle/src/fhir.rs
nilehmann: The crash happens because the test is "wrong". The refined signature is

```rust
#[flux::sig(fn<refine q: T -> bool>(&T{v:q(v)}) -> T{v: q(v)})]
```
 but it should be 
 ```rust
#[flux::sig(fn<T, refine q: T -> bool>(&T{v:q(v)}) -> T{v: q(v)})]
```
is missing the `T`.

If you put the angle brackets in the refined signature you have to specify all the generics. We only auto-generate the generics if there are no angle brackets.

The crash should be fixed somewhere else.

File: crates/flux-desugar/src/sort_resolver.rs (hunk 2 above)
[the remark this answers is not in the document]
Comment (nilehmann) [reply crates/flux-desugar/src/sort_resolver.rs:122]: > Hmm, do you mean I should add a case `fhir::Sort::SelfParam` for the case when we are using `Self` as a TyParam?

More specifically, we should also be able to implement the trait. The following

```rust
impl MyTrait for i32 {
    #[flux::sig(fn<refine p: Self -> bool>(&Self{v: p(v)}) -> Self{v: p(v)})]
    fn foo1(&self) -> Self {
        *self
    }
}
```
is currently failing because `Self` is resolved to `fhir::Sort::SelfParam` with the def id of the impl block which doesn't make any sense.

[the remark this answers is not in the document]
nilehmann: we could also not support `Self` in the impl. You have to write the sort explicitly 

```rust
impl MyTrait for i32 {
    #[flux::sig(fn<refine p: int -> bool>(&Self{v: p(v)}) -> Self{v: p(v)})]
    fn foo1(&self) -> Self {
        *self
    }
}
```

but then at some point we will need to be able to prove that `fn<refine p: int -> bool>(&Self{v: p(v)}) -> Self{v: p(v)})` is a valid refinement/specialization of `fn<refine p: Self -> bool>(&Self{v: p(v)}) -> Self{v: p(v)})`

crates/flux-middle/src/rty/mod.rs
```diff
@@ -274,6 +274,7 @@ pub struct FnSig {
     requires: List<Constraint>,
     args: List<Ty>,
     output: Binder<FnOutput>,
+    lifted: bool,
```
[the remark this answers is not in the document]
ranjitjhala: Yes it is -- i'd like to make it part of this PR so maybe we can sync about it tomorrow or soon?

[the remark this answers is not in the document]
nilehmann: > Yes it is -- i'd like to make it part of this PR so maybe we can sync about it tomorrow or soon?

let's sync tomorrow

crates/flux-middle/src/fhir.rs
ranjitjhala: Hmm - still not following. 

Even if you write the `fn<T, refine...>` as you suggest, you *still* get the `T cannot be refined error`. 

So why not just cut to the chase and make it return that error in this case, right away? 

I mean sure you could have some other logic to check that all `T` are specified in the angle brackets but 

(a) That seems overkill  _why_ should I have to write _everything_ including unnecessary stuff (e.g. `T`) just because I wanted to write _one_ (e.g. `refine q: T -> bool`) thing, and 

(b) Anyways, the _real_ problem here is you need `T` to be refinable?

ranjitjhala: I guess your point is 

1. we should just auto-generate the generics (to avoid the crash),
2. then the rest of the code would automatically point out the 'cannot refine' error?

ranjitjhala: I think adding the "check-signature-has-enough-generics" is a bit pointless, for the reasons above. So really, we should probably do (1) which may be easy or hard (IDK) but at least doing the `Option` seems a strict improvement over the current crash?

crates/flux-desugar/src/sort_resolver.rs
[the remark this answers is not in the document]
ranjitjhala: Do we currently support `Self` in the `impl`? (I can't recall...) 

The `refined_fn_in_trait.rs` supports it in the `trait` -- but not in the `impl` afaict?

I think ideally it would be nice to support it in `impl` so let me do this asap!

crates/flux-middle/src/fhir.rs
nilehmann: My point is that making `get_param` return `Option` is not fixing the underlying problem. If you revert that change and write `fn<T, refine...>` you still get the `T cannot be refined error`. 

The underlying problem is that we assume the refined signature has the same number of generics as the rust signature.

We can't avoid checking the list of generics in the refined signature matches the generics in the rust signature, e.g., the following should give an error (I think right now silently adds an extra generic)

```rust
#[flux::sig(fn<A, B>(x: A) )]
fn foo<A>(x: A) {}
```

The reason that generics can be specified in the refined signature is in case you want to specify they are of kind `base`

crates/flux-desugar/src/sort_resolver.rs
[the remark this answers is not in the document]
nilehmann: yes, we do support `Self` in an `impl` when used as a type.

crates/flux-middle/src/fhir.rs
ranjitjhala: Yes, it makes sense to write `<T as base,...>` my point is only that it doesn't make sense to ever (?) require the user to write `<T,...>` ...

crates/flux-desugar/src/sort_resolver.rs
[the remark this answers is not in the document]
ranjitjhala: Ok I have changed this to use the sort of `Self` appropriately for `impl` -- perhaps the code can be simplified more..

crates/flux-middle/src/fhir.rs
ranjitjhala: I fixed this in `desugar_generics` 
1. lifting the rust generics (to make sure we got all the type variables) and then
2. layering on the surface spec (computed by `desugar_surface_generics`) to mark variables as `base` if so specified.

crates/flux-middle/src/rty/mod.rs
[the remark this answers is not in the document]
ranjitjhala: Oh wait, this has nothing to do with #588 -- instead this is about using the trait-spec to verify the impl when the latter has no spec. e.g.

```rust
pub trait Silly {
    #[flux::sig(fn(&Self) -> i32{v:100 < v})]
    fn bloop(&self) -> i32;
}

impl Silly for i32 {
    fn bloop(&self) -> i32 {
        0
    }
}
```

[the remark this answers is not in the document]
ranjitjhala: but maybe its too orthogonal to this PR so let me make a separate issue.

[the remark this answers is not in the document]
ranjitjhala: have removed the `lifted` stuff from `FnSig`

ranjitjhala:
@nilehmann I think I've addressed the issues would be good to take a look

crates/flux-desugar/src/sort_resolver.rs
```diff
@@ -20,8 +22,11 @@ type Result<T = ()> = std::result::Result<T, ErrorGuaranteed>;
 pub(crate) struct SortResolver<'a> {
     sess: &'a FluxSession,
     sort_decls: &'a fhir::SortDecls,
-    generic_params: FxHashMap<Symbol, rustc_span::def_id::DefId>,
+    generic_params: FxHashMap<Symbol, DefId>,
     sort_params: FxHashMap<Symbol, usize>,
+    parent_id: Option<DefId>,
+    /// the sort of `Self` in an `impl` block, None otherwise
+    self_sort: Option<fhir::Sort>,
```
nilehmann, line 29: Instead of multiple fields I'd have a single field capturing how `Self` should be resolved. We could have an enum like so

```rust
enum SelfRes {
    /// A `Self` parameter in a trait definition.
    Param(DefId),
    /// An alias to another sort, e.g., when used inside an impl block
    Alias(fhir::Sort),
    /// It is not valid to use `Self`, e.g., when used in a free function
    None
}
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -227,6 +227,20 @@ enum QPathRes<'a> {
     NumConst(i128),
 }
 
+fn super_hack_is_impl(def_id: DefId) -> bool {
+    format!("{def_id:?}").contains("::{impl#")
+}
```
nilehmann, line 232: Use `tcx.def_kind` and test against `DefKind::Impl`

nilehmann (changes requested):
A couple of comments

crates/flux-desugar/src/sort_resolver.rs
ranjitjhala: yes, this is excellent. Done!

crates/flux-desugar/src/sort_resolver.rs
```diff
@@ -9,19 +9,35 @@ use rustc_errors::ErrorGuaranteed;
...
 pub(crate) struct SortResolver<'a> {
     sess: &'a FluxSession,
     sort_decls: &'a fhir::SortDecls,
-    generic_params: FxHashMap<Symbol, rustc_span::def_id::DefId>,
+    generic_params: FxHashMap<Symbol, DefId>,
     sort_params: FxHashMap<Symbol, usize>,
+    self_res: SelfRes,
+    // parent_id: Option<DefId>,
+    // /// the sort of `Self` in an `impl` block, None otherwise
+    // self_sort: Option<fhir::Sort>,
```
nilehmann, line 40:
```suggestion
```

crates/flux-middle/src/global_env.rs
```diff
@@ -168,6 +168,12 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
         self.queries.lower_late_bound_vars(self, def_id)
     }
 
+    // pub fn get_generic_param(&self, def_id: LocalDefId) -> Option<&fhir::GenericParam> {
+    //     let owner = self.hir().ty_param_owner(def_id);
+    //     let generics = self.map().get_generics(owner)?;
+    //     generics.get_param(def_id)
+    // }
```
nilehmann, line 175:
```suggestion
```

crates/flux-middle/src/global_env.rs
```diff
@@ -256,6 +262,17 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
         }
     }
 
+    pub fn sort_of_self_param(&self, owner: DefId) -> Option<fhir::Sort> {
```
nilehmann, line 265:
```suggestion
    fn sort_of_self_param(&self, owner: DefId) -> Option<fhir::Sort> {
```

crates/flux-refineck/src/checker.rs
```diff
@@ -150,6 +150,40 @@ impl<'a, 'tcx> Checker<'a, 'tcx, ShapeMode> {
...
+//         && is_lifted
+//     {
+//         let method_sig = genv
+//             .fn_sig(trait_method_id)
+//             .with_span(genv.tcx.def_span(trait_method_id))?;
+//         println!("TRACE: checker_sig {def_id:?} ==> {fn_sig:?} ==> {method_sig:?}");
+//         // next_how_to_instantiate_method_sig()
+//     }
+//     Ok(fn_sig)
+// }
```
nilehmann, line 185:
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -158,6 +192,7 @@ impl<'a, 'tcx> Checker<'a, 'tcx, RefineMode> {
         bb_env_shapes: ShapeResult,
         config: CheckerConfig,
     ) -> Result<(RefineTree, KVarStore), CheckerError> {
+        // TODO(check-trait-impl) let fn_sig = checker_sig(genv, def_id)?;
```
nilehmann, line 195:
```suggestion
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -198,6 +198,25 @@ enum QPathRes<'a> {
...
+                if let Some(sort) = genv.sort_of_self_ty_alias(alias_to) {
+                    return SelfRes::Alias(sort);
+                } else {
+                    return SelfRes::None;
+                }
+            }
+            _ => return SelfRes::None,
+        }
+    }
+    return SelfRes::None;
```
nilehmann, line 217:
```suggestion
            DefKind::Trait => SelfRes::Param(alias_to),
            DefKind::Impl { .. } => {
                if let Some(sort) = genv.sort_of_self_ty_alias(alias_to) {
                    SelfRes::Alias(sort)
                } else {
                    SelfRes::None
                }
            }
            _ => SelfRes::None,
        }
    }
    SelfRes::None
```

nilehmann (approved):
LGTM

nilehmann:
I'm going to merge this myself because I have some big changes that are going to conflict.

ranjitjhala:
thanks!

Issue #598, opened 2024-01-12
Check Associate-Pred signatures


ranjitjhala:
We should try making the following definition work without trusting it:

```rust
#[flux::sig(fn<T as base>(&{T[@x] | <T as MyTrait>::f(x)}))]
pub fn bob<T: MyTrait>(x: &T) {
    x.method();
}
```

We just need to encode the associated predicate as an uninterpreted function. We could generalize `refine_tree::NodeKind::Guard` to contain a `Pred` and then do something similar to `FixpointCtxt::collect_sorts` to collect the abstract refinements needed to encode the constraint.

_Originally posted by @nilehmann in https://github.com/flux-rs/flux/pull/597#pullrequestreview-1816072205_

ranjitjhala:
Hmm, I think this is a bit trickier, as ultimately, to do the **sort checking** (and determine the sort of the uninterpreted function in the fixpoint encoding) you want to "normalize" to determine that

1. In the _specification_ of `MyTrait`, we know `f : Self -> bool` and 
2. In this _instance_  `<T as MyTrait>` tells you that the `Self` is `T` and so
3. The `<T as MyTrait>::f`  is of type `T -> bool`.

A more lassez-faire approach would be to recover the type from the arguments, from `x` but that can get messy e.g. if we use the `<T as MyTrait>::f` at two different sorted index/arguments in the same signature. 

Am I missing something?

ranjitjhala:
Actually, now that I've written it all out, it doesn't seem so bad so I'll just do it! :-)

nilehmann:
I'm talking specifically about the cases where the associated predicate cannot be normalized away. In those cases both the predicate and the sort remain abstract.

nilehmann:
More concretely, given something like

```rust
#[flux::generics(Self as base)]
#[flux::predicate(f: (Self) -> bool)]
trait MyTrait {
    #[flux::sig(fn(self: &Self{v: <Self as MyTrait>::f(v)}) -> Self{v: <Self as MyTrait>::f(v)}]
    fn method(&self) -> Self;
}

#[flux::sig(fn<T as base>(&T{v: <T as MyTrait>::f(x)}) -> T{v: <T as MyTrait>::f(x)})]
pub fn bob<T: MyTrait>(x: &T) ->  T {
    x.method()
}
```
we should generate a constraint that looks like (in pseudo fixpoint)

```smt2
(declare-sort MyTrait)
(declare-fun f (MyTrait) Bool)

[...]

PR #599, opened 2024-01-14
WF Checks on Associated Predicates

ranjitjhala:
PR adds a bunch of WF/desugaring checks on Associated/Alias Predicates, which is a stepping stone to #598

The main technical bit is to 
1. Use the `Spec` of an `AssocPredicate` (i.e. the `Sort` given in the `trait`), to
2. Determine the `Impl`'s sort (by substituting the `GenericArgs` sorts).


ranjitjhala:
@nilehmann -- I've now added code to also so this PR (almost) fixes #598 -- the one little thing I couldn't figure out is how to get this

```rust
#[flux::generics(Self as base)]
#[flux::predicate(f: (Self) -> bool)]
pub trait MyTrait {
    #[flux::sig(fn(self: &Self{v: <Self as MyTrait>::f(v)}) -> Self{v: <Self as MyTrait>::f(v)})]
    fn method(&self) -> Self;
}
``` 

To work because it wasn't clear how to "collect" the name `MyTrait` -- the desugar/resolve thing currently barfs on the `MyTrait` ... I'm guessing this is quite easy?

ranjitjhala:
Also, keep an eye out for the `const_name_gen` shenanigans :-)

crates/flux-middle/src/rty/subst.rs
```diff
@@ -396,3 +396,23 @@ impl TypeFolder for SortSubst<'_> {
...
+
+impl TypeFolder for GenericSortSubst<'_> {
+    fn fold_sort(&mut self, sort: &Sort) -> Sort {
+        if let Sort::Param(var) = sort {
+            self.args[var.index as usize].clone()
+        } else {
+            sort.super_fold_with(self)
+        }
+    }
+}
```
nilehmann, line 418: This should go in `GenericsSubstFolder` so we only have one substitution for generics. We'll need to modify the logic in `GenericsSubstFolder::sort_for_param` to call `peel_out_sort`. Note that it's ok to panic if `peel_out_sort` returns `None` because we assume we only substitute on well-formed terms and that substitution preserves well-formedness. If substitution fails then the problem is somewhere else.

As a side note, the way I understand `Sort::Param` is as follows: if we have `|x: sortof(Self)| ...` and we substitute `Self` with `i32` we get `|x: sortof(i32)| ...` which then normalizes to `|x: int| ...`. Instead of having `x: sortof(i32)` as an intermediate step, we normalize it immediately as we apply the substitution, this way we don't need to represent `sortof` for arbitrary types, only for type parameters.

crates/flux-refineck/src/fixpoint_encoding.rs
```diff
@@ -439,11 +443,20 @@ where
...
+            rty::Pred::Expr(expr) => {
+                let mut bindings = vec![];
+                let mut preds = vec![];
+                self.pred_to_fixpoint_internal(expr, &mut bindings, &mut preds);
+                (bindings, preds)
+            }
+            rty::Pred::Alias(alias_pred, args) => {
+                let func = self.alias_pred_func(alias_pred, args.len());
+                let alias_pred_app = rty::Expr::app(func, args.clone(), None);
+                self.pred_to_fixpoint(&rty::Pred::Expr(alias_pred_app))
```
nilehmann, line 457: This is representing the associated predicate application as `rty::Expr::app` just to immediately encode it. We can go directly to fixpoint. I'd have something like this

```suggestion
                let func = self.register_const_for_alias_pred(alias_pred);
                let args = self.exprs_to_fixpoint(args);
                let pred = fixpoint::Expr::App(func, args);
                (vec![], vec![(pred, None)])
```

The trick is in `register_const_for_alias_pred` which has the following signature

```rust
fn register_const_for_alias_pred(&mut self, alias_pred: &rty::AliasPred) -> fixpoint::Var
```

This function registers a `ConstInfo` for the `AliasPred` (if there's no already one) and returns a `GlobalVar` for it. Note that the `ConstInfo` generated for an `AliasPred` should not go into the `ConstMap` because we never look it up by name (the lookup up key is the `AliasPred`).

crates/flux-middle/src/global_env.rs
```diff
@@ -216,6 +223,44 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
         self.map().const_by_name(name)
     }
 
+    pub fn sort_of_assoc_pred(&self, impl_id: DefId, name: Symbol) -> Option<List<rty::Sort>> {
```
nilehmann, line 226: The only error that's recoverable here is whether the `name` is a valid predicate, all other errors should be propagated up

```suggestion
    pub fn sort_of_assoc_pred(&self, impl_id: DefId, name: Symbol) -> QueryResult<Option<List<rty::Sort>>> {
```

If you have problems bridging between `QueryResult<T>` and `Result<T, ErrorGuaranteed>` you can use `.emit(genv.sess)` on a `QueryResult<T>` to convert it into a `Result<T, ErrorGuaranteed>`

crates/flux-middle/src/global_env.rs
```diff
@@ -216,6 +223,44 @@ impl<'sess, 'tcx> GlobalEnv<'sess, 'tcx> {
...
+            && let rty::AssocPredicateKind::Spec(sorts) = assoc_pred.kind
+        {
+            let arg_sorts: Vec<_> = args.iter().flat_map(|arg| arg.peel_out_sort()).collect();
+
+            return Some(sorts.fold_with(&mut subst::GenericSortSubst::new(&arg_sorts)));
+        };
+        None
+    }
+
+    pub fn sort_of_alias_pred(&self, alias_pred: &fhir::AliasPred) -> Option<List<rty::Sort>> {
```
nilehmann, line 242: The here, we should propagate errors up.

crates/flux-refineck/src/refine_tree.rs
```diff
@@ -690,11 +690,21 @@ mod pretty {
...
 
+    fn pred_as_expr(pred: &Pred) -> Expr {
+        match pred {
+            Pred::Expr(expr) => expr.clone(),
+            Pred::Alias(alias_pred, args) => {
+                let func = Expr::global_func(alias_pred.name, FuncKind::Asp);
+                Expr::app(func, args.clone(), None)
+            }
+        }
+    }
```
nilehmann, line 701: This is only used for printing, why not just print the `AliasPred`? We can just `preds_chain` return `(Vec<Pred>, Vec<NodePtr>)`

nilehmann (changes requested):

crates/flux-middle/src/rty/subst.rs
ranjitjhala: Done, I think, but see the note in `check_alias_pred_app`

crates/flux-middle/src/global_env.rs
ranjitjhala: done.

crates/flux-refineck/src/fixpoint_encoding.rs
```diff
@@ -565,6 +595,59 @@ where
...
+        for i in 0..arity {
+            sorts.push(rty::Sort::Var(rty::SortVar::from(i)));
+        }
+        sorts.push(rty::Sort::Bool);
+        rty::PolyFuncSort::new(arity, rty::FuncSort { inputs_and_output: List::from_vec(sorts) })
+    }
+
+    fn alias_pred_gen(&mut self, arity: usize) -> AliasPredInfo {
+        let n = self.alias_preds.len();
+        let sym = Symbol::intern(&format!("alias_pred_{n:?}"));
```
nilehmann, line 610: `sym` is only used to print a comment next to the constant declaration in the `smt2` constraint, so we could just pretty print the alias pred (or even simpler just print it's an alias pred).

crates/flux-refineck/src/fixpoint_encoding.rs
```diff
@@ -565,6 +595,59 @@ where
...
         self.genv.tcx.def_span(self.def_id)
     }
+
+    fn alias_pred_sort(arity: usize) -> rty::PolyFuncSort {
+        let mut sorts = vec![];
+        for i in 0..arity {
+            sorts.push(rty::Sort::Var(rty::SortVar::from(i)));
+        }
+        sorts.push(rty::Sort::Bool);
+        rty::PolyFuncSort::new(arity, rty::FuncSort { inputs_and_output: List::from_vec(sorts) })
```
nilehmann, line 605: hmmm, this is making the function unnecessarily polymorphic. I don't think this will cause problems because well-formedness in previous phases will ensure the function is always instantiated with the same sorts, but technically, the more right thing to do would be to get the sort from the alias pred and encode that. Maybe we can leave a comment.

nilehmann (approved):
A couple of extra comments. They are not very actionable, so feel free to merge.

crates/flux-refineck/src/refine_tree.rs
ranjitjhala: Ok I'm letting this be for now, maybe we can fix if we put the assoc-pred into the expr?

crates/flux-refineck/src/fixpoint_encoding.rs
ranjitjhala: hmm looks like its used here as an actual symbol?
https://github.com/flux-rs/flux/blob/7948b81601ea8da8795dcc0115d11a5ebc494efc/crates/flux-refineck/src/fixpoint_encoding.rs#L606

ranjitjhala: (For now, just printing the `alias_pred` as you suggest.)

crates/flux-refineck/src/fixpoint_encoding.rs
ranjitjhala: Yes, I thought about that -- but then realized its easier (if slower) to kick the can down the road and let `fixpoint` work out the sorts, by just using a polymorphic signature for the UIF. But let me add your note as a comment for now.

PR #609, opened 2024-01-25
Add support for `extern_spec` for `trait`

ranjitjhala:
For example, to let us attach associate predicates to (externally defined) traits.


crates/flux-middle/src/rty/mod.rs
```diff
@@ -233,6 +233,7 @@ pub struct AssocPredicate {
     /// [`DefId`] of the container, i.e., the impl block or trait.
     pub container_def_id: DefId,
     pub name: Symbol,
+    pub span: Span,
```
nilehmann, line 236: We should avoid having spans at `rty` because types could come for external crates for which we may not have spans.

nilehmann (changes requested):

crates/flux-middle/src/rty/mod.rs
ranjitjhala: removed.

crates/flux-driver/src/collector.rs
```diff
@@ -156,6 +163,14 @@ impl<'tcx, 'a> SpecCollector<'tcx, 'a> {
             .traits
             .insert(owner_id, surface::Trait { generics, assoc_predicates });
 
+        if attrs.extern_spec() {
+            // extern_spec dummy structs are always opaque because they contain
+            // one field: the external struct they are meant to represent.
```
nilehmann, line 168:
```suggestion
```

nilehmann (approved):

PR #640, opened 2024-06-25
Unsupported const

ranjitjhala:
Use `Constant::Opaque` instead of just falling over...


ranjitjhala:
@nilehmann -- my apologies am cutting ahead and merging this without review -- seems small enough that we can revert later if there are issues?

PR #641, opened 2024-06-25
Unsupported cast

ranjitjhala:
Don't crash on code like the below

```rust
pub fn foo(x: *const u8) -> usize {
    let bloop = x as usize;
    bloop
}
```


crates/flux-middle/src/rustc/lowering.rs
```diff
@@ -483,7 +483,11 @@ impl<'sess, 'tcx> LoweringCtxt<'_, 'sess, 'tcx> {
             rustc_mir::CastKind::PointerCoercion(ptr_coercion) => {
                 Some(CastKind::Pointer(self.lower_pointer_coercion(ptr_coercion)?))
             }
-            _ => None,
+            rustc_mir::CastKind::PointerExposeProvenance => Some(CastKind::PointerExposeProvenance),
+            _ => {
+                println!("TRACE: lower_cast_kind: {:?}", kind);
```
nilehmann, line 488:
```suggestion
```

nilehmann (approved):
Looks good.

As a side note, I don't know if we will need to track refinements on pointers and what that refinement will be (an int, an int and provenance, ...). We can always have a wrapper type over a pointer though.

PR #643, opened 2024-06-27
Don't crash when bit-shifting uints

ranjitjhala:
Add code to not crash on

```rust
// unsigned shift right
pub fn uint_shr(left: u32, right: usize) {
    let blah = left >> right;
}
```


ranjitjhala:
Am adding support for various little things as I see them...

tests/tests/neg/surface/generic_const00.rs
```diff
@@ -124,3 +124,13 @@ fn test_array_good<const N: usize>(arr: &[i32; N]) -> i32 {
...
+
+//-------------------------------------------------------------------------------
+// TODO: get this to work
+// fn test_loop<const N: usize>(arr: &[i32; N]) -> i32 {
+//     let mut x = 0;
+//     for i in 0..N {
+//         x += arr[i];
+//     }
+//     x
+// }
```
nilehmann, line 136:
```suggestion
```

crates/flux-refineck/src/sigs/mod.rs
```diff
@@ -121,7 +125,12 @@ where
...
     }
 
-    fn get(&self, op: T, btys: [BaseTy; N]) -> &Sig<N> {
-        &self.map[&(op, btys)]
+    fn get(&self, op: T, btys: [BaseTy; N]) -> Option<&Sig<N>> {
+        let _msg = format!("no sig found for {op:?} at {btys:?}", op = op, btys = btys);
+        match &self.map.get(&(op, btys)) {
+            Some(sig) => Some(sig),
+            None => None, // panic!("{}", msg),
+        }
```
nilehmann, line 133:
```suggestion
        &self.map.get(&(op, btys))
```

crates/flux-refineck/src/checker.rs
```diff
@@ -1430,7 +1458,10 @@ impl Mode for RefineMode {
         target: BasicBlock,
     ) -> Result<bool> {
         let bb_env = &ck.inherited.mode.bb_envs[&ck.def_id][&target];
-        debug_assert_eq!(&ck.snapshot_at_dominator(target).scope().unwrap(), bb_env.scope());
+        debug_assert_eq!(
+            ck.snapshot_at_dominator(target).scope().unwrap().bindings(),
+            bb_env.scope().bindings()
+        );
```
nilehmann, line 1464: was this failing? The reftgenerics should be the same throughout the entire function so the default equality implementation on `Scope` should work.

crates/flux-refineck/src/sigs/default.rs
```diff
@@ -152,6 +154,29 @@ pub(crate) fn mk_shift_ops() -> impl IntoIterator<Item = (mir::BinOp, Sig<2>)> {
...
+                (Shr, s!(fn(a: Int, b: Int) -> Int{ v: E::tt() })),
+                (Shr, s!(fn(a: Int, b: Uint) -> Int{ v: E::tt() })),
+
+                (Shl, s!(fn(a: Uint, b: Int) -> Uint{ v: E::tt() })),
+                (Shl, s!(fn(a: Uint, b: Uint) -> Uint{ v: E::tt() })),
+                (Shr, s!(fn(a: Uint, b: Int) -> Uint{ v: E::tt() })),
+                (Shr, s!(fn(a: Uint, b: Uint) -> Uint{ v: E::tt() })),
+            ]
+        })
+}
```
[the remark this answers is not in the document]
ranjitjhala: Sounds good; I'll fork it out as a separate issue.

crates/flux-refineck/src/checker.rs
ranjitjhala: Hmm good question, I have reverted it -- there was something going on with loops that was causing a crash but I can't reproduce it now!

ranjitjhala:
I think I addressed the comments above, should we merge?

nilehmann:
@ranjitjhala I think you haven't pushed any of the fixes yet. But yeah provided those are included I think this is good to merge.

nilehmann (approved):

PR #651, opened 2024-07-12
Rebase generics for trait onto impl method

ranjitjhala:


ranjitjhala:
Merging in since its a one-line change (modulo the clippy things ... which I assume are because clippy just got updated?) LMK if you want to revert...

PR #664, opened 2024-07-23
Rudimentary support for dynamic trait objects

### LINKED ISSUE #655: Supporting dynamic trait objects
As we discussed in the meeting earlier, it seems like everyone is interested in adding support for dynamic trait objects to Flux. To get the ball rolling, I put together a very simple test case (based off the Rust book dyn trait example: https://doc.rust-lang.org/rust-by-example/trait/dyn.html):
```Rust
#![allow(unused)]

struct Sheep {}
struct Cow {}


trait Animal {
    // Instance method signature
    fn noise(&self);
}

// Implement the `Animal` trait for `Sheep`.
impl Animal for Sheep {
    fn noise(&self) {
        println!("baaaaah!");
    }
}

// Implement the `Animal` trait for `Cow`.
impl Animal for Cow {
    fn noise(&self) {
        println!("moooooo!");
    }
}

fn make_two_noises(animal: &dyn Animal) {
    animal.noise();
    animal.noise();
}
```

This code currently fails to verify with the error message: 
`refinement of unsupported function` 
`this function contains unsupported features`

ranjitjhala:
Fixes #655

Totally ignores all the bound-vars stuff for `TraitObject` because I haven't understood it at all, but wanted to start iterating.


enjhnsn2:
This pr already helps a ton, although there are still some `dyn trait` related crashes when I try to apply it to Tock. So far, I've seen one main crash that I've minimized it down to:
```Rust
struct Cow {}

trait CowContainer<'a> {
    fn get_cow(&self) -> &'a Cow;
}

pub struct CowCell<'a> {
    inner: &'a dyn CowContainer<'a>,
}
```
When you run flux on this code, it errors with:
`error: internal compiler error: crates/flux-fhir-analysis/src/conv/mod.rs:444:13: unexpected! conv_poly_dyn_trait_ref`

Which points to this code: https://github.com/flux-rs/flux/blob/23c745369bec99cf6bee49dd84c8547fbeac39bc/crates/flux-fhir-analysis/src/conv/mod.rs#L438-L450

Comparing this code to other code that passes, it appears that that the problem is that the the trait being parameterized by the dyn trait has a lifetime parameter. Let me know if you need any other info

nilehmann:
I'll probably have some time to look at this tomorrow

ranjitjhala:
@enjhnsn2 are you sure about the second one, it seems to work fine for me:

```
rjhala@pozole ~/r/flux (dyn)> more tests/tests/todo/dyn02.rs 
pub trait Animal {
    fn noise(&self);
}

#[flux::trusted]
pub fn get_animal() -> &'static dyn Animal {
    unimplemented!()
}
rjhala@pozole ~/r/flux (dyn)> cargo xtask run tests/tests/todo/dyn02.rs
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/xtask run tests/tests/todo/dyn02.rs`
$ export FLUX_BUILD_SYSROOT=1
$ cargo build -p flux-rs
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
$ cargo build
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
```

ranjitjhala:
@enjhnsn2 - the above two things work now, see

* https://github.com/flux-rs/flux/blob/dyn/tests/tests/pos/surface/dyn02.rs
* https://github.com/flux-rs/flux/blob/dyn/tests/tests/pos/surface/dyn03.rs

enjhnsn2:
> pub trait Animal {
>     fn noise(&self);
> }
> 
> #[flux::trusted]
> pub fn get_animal() -> &'static dyn Animal {
>     unimplemented!()
> }

Whoops, I think I may have messed up the minimization for that one. When I was minimizing I started with tock code that pulled in the tock kernel and I think the error was in the dependency, not the minimized code. Let me try to get a better example

enjhnsn2:
Alright, I found a minimized example that triggers the error. Unfortunately its a little more complicated:
```Rust
pub trait Animal {
    fn noise(&self);
}

fn apply_closure_to_animal<F>(closure: F, animal: &'static dyn Animal)
where
    F: FnOnce(&dyn Animal),
{
    closure(animal);
}
```

ranjitjhala:
Thanks, on it!

- Ranjit.

nilehmann:
I prefer if 

> Alright, I found a minimized example that triggers the error. Unfortunately its a little more complicated:
> 
> ```rust
> pub trait Animal {
>     fn noise(&self);
> }
> 
> fn apply_closure_to_animal<F>(closure: F, animal: &'static dyn Animal)
> where
>     F: FnOnce(&dyn Animal),
> {
>     closure(animal);
> }
> ```
> 
> This bug isn't too frequent and is possible to work around, so I don't think you should block this pr on it.

This feels it'll fail for various reasons not just dynamic objects. I'd prefer if we don't fix it in this PR.

crates/flux-middle/src/rustc/ty.rs
```diff
@@ -182,6 +182,20 @@ pub enum TyKind {
     CoroutineWitness(DefId, GenericArgs),
     Alias(AliasKind, AliasTy),
     RawPtr(Ty, Mutability),
+    TraitObject(List<PolyTraitRef>, Region, TraitObjectSyntax),
```
nilehmann, line 185: This is the wrong representation at this level, instead, it should match [`TyKind::Dynamic`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/type.TyKind.html#variant.Dynamic).

crates/flux-middle/src/rty/mod.rs
```diff
@@ -951,6 +975,7 @@ pub enum BaseTy {
     Never,
     Closure(DefId, /* upvar_tys */ List<Ty>),
     Coroutine(DefId, /*resume_ty: */ Ty, /* upvar_tys: */ List<Ty>),
+    TraitObject(List<PolyTraitRef>, Region, TraitObjectSyntax),
```
nilehmann, line 978: The same here, it should match [`TyKind::Dynamic`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/type.TyKind.html#variant.Dynamic)

nilehmann (changes requested):
@ranjitjhala first round of comments. I focused only on the way dynamic objects are being represented.

crates/flux-middle/src/rustc/lowering.rs
```diff
@@ -740,6 +740,21 @@ pub(crate) fn lower_ty<'tcx>(
             let args = lower_generic_args(tcx, args)?;
             Ok(Ty::mk_generator_witness(*did, args))
         }
+        rustc_ty::Dynamic(predicates, region, kind) => {
```
nilehmann, line 743: We can emit unsupported here if we drop support for `dyn*`

crates/flux-middle/src/fhir.rs
```diff
@@ -25,6 +25,7 @@ use flux_common::{bug, span_bug};
 use flux_syntax::surface::ParamMode;
 pub use flux_syntax::surface::{BinOp, UnOp};
 use itertools::Itertools;
+use rustc_ast::TraitObjectSyntax;
```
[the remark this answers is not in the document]
ranjitjhala: Ok lets go with (1) which is simpler.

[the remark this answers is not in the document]
ranjitjhala: hmm maybe 2 is better.

crates/flux-middle/src/rustc/lowering.rs
ranjitjhala: Done in `lower_dyn_kind`

ranjitjhala:
> > This bug isn't too frequent and is possible to work around, so I don't think you should block this pr on it.
> 
> This feels it'll fail for various reasons not just dynamic objects. I'd prefer if we don't fix it in this PR.

@nilehmann  -- it turns out that for @enjhnsn2 example with `closure` and `dyn` above (see issue #666) -- just filling out the `todo!` case for `Dynamic` in the `to_rustc(...)` gets this to not crash, so I just tacked it onto this PR since its very much related to the whole conversion to/from `ExistentialPredicate/ExistentialTraitRef` that is the bulk of this PR...

crates/flux-middle/src/rustc/lowering.rs
```diff
@@ -794,6 +822,15 @@ fn lower_generic_arg<'tcx>(
...
 
+// fn lower_dyn_kind(kind: &rustc_ty::DynKind) -> Result<DynKind, UnsupportedReason> {
+//     match kind {
+//         rustc_ty::DynKind::Dyn => Ok(DynKind::Dyn),
+//         rustc_ty::DynKind::DynStar => {
+//             Err(UnsupportedReason::new(format!("unsupported dyn kind `{kind:?}`")))
+//         }
+//     }
+// }
+
```
nilehmann, line 833:
```suggestion
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -1099,6 +1133,7 @@ impl BaseTy {
             | BaseTy::Array(_, _)
             | BaseTy::Closure(_, _)
             | BaseTy::Coroutine(..)
+            | BaseTy::Dynamic(_, _)
```
nilehmann, line 1136: There's something I don't quite understand yet. We are putting `Dynamic` in `BaseTy` which means they can be indexed (by `unit` as returned by this function), however, the concrete type under the dynamic type could be indexed by another sort. I don't know if these should be somehow related or if they are completely orthogonal.

crates/flux-refineck/src/constraint_gen.rs
```diff
@@ -583,7 +583,8 @@ impl<'a, 'genv, 'tcx> InferCtxt<'a, 'genv, 'tcx> {
             (BaseTy::Bool, BaseTy::Bool)
             | (BaseTy::Str, BaseTy::Str)
             | (BaseTy::Char, BaseTy::Char)
-            | (BaseTy::RawPtr(_, _), BaseTy::RawPtr(_, _)) => Ok(()),
+            | (BaseTy::RawPtr(_, _), BaseTy::RawPtr(_, _))
+            | (BaseTy::Dynamic(_, _), BaseTy::Dynamic(_, _)) => Ok(()),
```
nilehmann, line 587: I'd add an assertion here checking both types are the same. I'm not sure if there's something odd in Rust regarding subtyping between dynamic objects so it's better to be safe (I suppose the lifetime can be different, but the existential predicate have to be the same)

nilehmann (changes requested):

crates/flux-middle/src/rty/mod.rs
ranjitjhala: I think its fine, because once its `Dynamic` you have *zero* information about the underlying concrete type, everything is abstracted to whatever is know about the `dyn`...

crates/flux-refineck/src/constraint_gen.rs
ranjitjhala: done!

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1148,20 +1182,30 @@ impl<'a, 'genv, 'tcx> ConvCtxt<'a, 'genv, 'tcx> {
...
+                    debug_assert!(has_default);
+                    let tcx = self.genv.tcx();
+                    let ty = self
+                        .genv
+                        .type_of(param.def_id)?
+                        .instantiate(tcx, into, &[])
+                        .to_ty();
+                    into.push(rty::GenericArg::Ty(ty));
+                }
+                rty::GenericParamDefKind::Lifetime => {
```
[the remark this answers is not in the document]
ranjitjhala: Ah yes, this was a wrong path that I took, that I then backtracked out of and should have deleted/reverted this code. Fixed.

ranjitjhala:
Ok I think I addressed all the issues above -- except the 

> We are putting Dynamic in BaseTy which means they can be indexed (by unit as returned by this function),

But I'm pretty certain that's (for now at least) the sensible choice - as in `dyn` you have to forget all about the underlying concrete type; I suggest we revisit this later if we have reason to (or if we think there's some use-case for indexing `Dynamic` values...)

nilehmann (approved):
Ok, I think this is good to go. we should be able to iterate from here once we have concrete examples.

re indices: more than worried about being unsound or anything I'm was wondering whether it will ever make sense to have indices on dynamic objects. We will see, but for now `unit` sounds like a safe option.

PR #668, opened 2024-07-27
Check for missing associated refinements

### LINKED ISSUE #649: ICE when assoc refinement is not defined in implementation
```rust
#[flux::assoc(fn eval(x: int) -> int)]
trait MyTrait { }

impl MyTrait for i32 { }

#[flux::sig(fn(x: i32) -> i32[<T as MyTrait>::eval(x)])]
fn test01<T: MyTrait>(x: i32) -> i32 {
    todo!()
}

fn test02() {
    test01::<i32>(0);
}
```

```
error: internal compiler error: crates/flux-fhir-analysis/src/lib.rs:197:28: assoc reft `eval` not found in impl `DefId(0:3 ~ playground[223a]::{impl#0})`
```

ranjitjhala:
* Also uses the actual sort of the associated refinement and not the poly-var + bool hack.

Unfortunately we still get the "missing error" during the trait normalization; we _want_ to use the `span` for the callsite there but its unclear how to do so, without eg adding `Span` to the `Normalizer`?

Fixes #649


nilehmann:
@ranjitjhala to report the "missing error" in the correct place I think we need to lookup the assoc refinement in [`ConvCtxt::conv_alias_reft`](https://github.com/flux-rs/flux/blob/main/crates/flux-fhir-analysis/src/conv/mod.rs/#L725) and report an error if we can't find it (this may also fix the other error in the https://github.com/flux-rs/flux/issues/649 ?)

Also, a nice to have would be a newtype over the pair `(DefId, Symbol)` that guarantees the assoc refinement exists so we only construct valid assoc refinements projection in `rty`. Doing this would probably point to the places we are doing something wrong. More generally, we could even identify  everything that doesn't have a corresponding rust item with a newtype over `(DefId, Symbol)` (e.g., qualifiers, refinement functions, uif)

ranjitjhala:
@nilehmann -- interesting thought but wouldn't that require us to basically reimplement `impl_id_of_alias_reft` (or something like it) for the `conv` phase? [ to recover the actual `impl_id` from the trait and generic-args? ]

ranjitjhala:
Silly me, we don't need to re-implement anything ... we can just use the `rty` level stuff produced _in_ the `conv` phase to do the lookup with `impl_id_of_alias_reft` ...

nilehmann:
@ranjitjhala yeah that would work. But if we have a newtype around `(DefId, Symbol)` we would know statically all the places where calling `assosc_refinement_def` and `sort_of_assoc_reft` return `Some`. This is orthogonal though, checking we only construct valid associated refinements during conv is a good first step.

ranjitjhala:
Ok, now an error is reported both in the impl AND at the callsite.

Given 

```rust
#[flux::assoc(fn eval(x: int) -> int)]
pub trait MyTrait {}

impl MyTrait for i32 {}

#[flux::trusted]
#[flux::sig(fn(x: i32) -> i32[<T as MyTrait>::eval(x)])]
fn test01<T: MyTrait>(_x: i32) -> i32 {
    todo!()
}

pub fn test02() {
    test01::<i32>(0);
}
```

Flux will complain that

```
error[E0999]: associated refinement `eval` is not defined in implementation of trait `MyTrait`
 --> tests/tests/todo/issue_649b.rs:4:1
  |
4 | impl MyTrait for i32 {}
  | ^^^^^^^^^^^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-fhir-analysis/src/compare_impl_item.rs:27:36

error[E0999]: associated refinement `eval` is not defined in this trait implementation
 --> tests/tests/todo/issue_649b.rs:4:1
  |
4 | impl MyTrait for i32 {}
  | ^^^^^^^^^^^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-fhir-analysis/src/lib.rs:201:14

TRACE: normalize-errorr Emitted(ErrorGuaranteed(()))
error[E0999]: invalid associated refinement for this function call
  --> tests/tests/todo/issue_649b.rs:13:5
   |
13 |     test01::<i32>(0);
   |     ^^^^^^^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-middle/src/queries.rs:660:29

error: aborting due to 3 previous errors
```

Ideally, we should somehow "merge" the last two errors, but at any rate, this is better than crashing so I think we should merge and move on..

crates/flux-fhir-analysis/src/compare_impl_item.rs
```diff
@@ -1,17 +1,36 @@
...
 };
-
 type Result<T = ()> = std::result::Result<T, ErrorGuaranteed>;
 
+// fn impl_span(genv: &GlobalEnv, impl_id: LocalDefId) -> Span {
+//     let hir_id = genv.tcx().local_def_id_to_hir_id(impl_id);
+//     let span = genv.tcx().hir().span(hir_id);
+//     span
+// }
+
```
nilehmann, line 15:
```suggestion
```

crates/flux-refineck/src/constraint_gen.rs
```diff
@@ -167,7 +168,16 @@ impl<'a, 'genv, 'tcx> ConstrGen<'a, 'genv, 'tcx> {
                 |br| infcx.next_bound_region_var(span, br.kind, BoundRegionConversionTime::FnCall),
                 |sort, mode| infcx.fresh_infer_var(sort, mode),
             )
-            .normalize_projections(genv, infcx.region_infcx, infcx.def_id, infcx.refparams)?;
+            .normalize_projections(genv, infcx.region_infcx, infcx.def_id, infcx.refparams);
+
+        let fn_sig = match fn_sig {
+            Ok(fn_sig) => fn_sig,
+            Err(err) => {
+                println!("TRACE: normalize-errorr {err:?}");
```
nilehmann, line 176:
```suggestion
```

nilehmann (approved):

PR #675, opened 2024-08-01
add fixpoint-passthrough for bv_shl

enjhnsn2:
I need `bv_shl` for a spec I'm writing, so I just went ahead and added it. Let me know if there is anywhere else I need to modify besides adding this line.


nilehmann (approved):
LGTM 

🚢 it

PR #676, opened 2024-08-01
extend bitvector ops passthrough api 

enjhnsn2:
I extended the fixpoint-passhtrough api with:
- bitvector size extension (bv_sign_extend, bv_zero_extend)
- a now-complete set of arithmetic operations (bv_neg, bv_udiv, bv_urem, bv_sdiv, bv_srem, bv_smod)
- and more bitwise logic (bv_or, bv_xor, bv_not)

### DIFF
--- a/crates/flux-middle/src/lib.rs
+++ b/crates/flux-middle/src/lib.rs
@@ -110,6 +110,29 @@ pub fn theory_funcs() -> &'static [TheoryFunc] {
                     rty::FuncSort::new(vec![BitVec(BvSize::Fixed(64))], Int),
                 ),
             },
+            // Bitvector extension
+            TheoryFunc {
+                name: Symbol::intern("bv_zero_extend"),
+                fixpoint_name: Symbol::intern("zero_extend"),
+                sort: rty::PolyFuncSort::new(
+                    List::from_arr([SortParamKind::BvSize]),
+                    rty::FuncSort::new(
+                        vec![rty::Sort::Int, BitVec(bv_param0)],
+                        BitVec(bv_param0),
+                    ),
+                ),
+            },
+            TheoryFunc {
+                name: Symbol::intern("bv_sign_extend"),
+                fixpoint_name: Symbol::intern("sign_extend"),
+                sort: rty::PolyFuncSort::new(
+                    List::from_arr([SortParamKind::BvSize]),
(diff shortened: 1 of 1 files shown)


crates/flux-middle/src/lib.rs
```diff
@@ -110,6 +110,23 @@ pub fn theory_funcs() -> &'static [TheoryFunc] {
...
+                ),
+            },
+            TheoryFunc {
+                name: Symbol::intern("bv_sign_extend"),
+                fixpoint_name: Symbol::intern("sign_extend"),
+                sort: rty::PolyFuncSort::new(
+                    List::from_arr([SortParamKind::BvSize]),
+                    rty::FuncSort::new(vec![rty::Sort::Int, BitVec(bv_param0)], BitVec(bv_param0)),
+                ),
+            },
```
nilehmann, line 129: @enjhnsn2 did you try these two? If I understand correctly `sign_extend` and `zero_extend` are *indexed identifiers* so we can't pass the size as an argument.

I am curious to know what `fixpoint` is doing here. @ranjitjhala?

enjhnsn2: Oh ya, fixpoint did not like that one bit. Ok, let me try to see what options are available for getting those two functions to work. First and foremost, my signature there is just wrong, it implies that the return type has the same sort as the inputs which is very wrong.

nilehmann: I suspect that fixpoint is doing something wrong here and can't express indexed identifiers (that's why we have `int_to_bv32` and `int_to_bv64`). In any case, I believe this will require some support from fixpoint.

nilehmann (approved):
There's a clippy warning, but it looks good otherwise. Feel free to merge after fixing the warning.

crates/flux-middle/src/lib.rs
enjhnsn2: Ok, the sign extension ops are pretty complicated because the sort of the output depends on the value of the argument. This works really cleanly in Z3, but I have no clue how they do it. Anyways, I've just dropped it from this commit as its not super urgent. Worst case, we could just implement fixed size extensions: 32-> 64, 16 -> 32, 16->64, 8->16, 8->32, 8->64. Its a little verbose, but it should work for most things

PR #681, opened 2024-08-06
Use fixpoint's sexp parser

### LINKED ISSUE #661: Migrate to new Fixpoint parser
Anticipating the need for user-defined sorts, I think we should move to the new s-expression fixpoint parser, i.e., stop passing `--nosmthorn`

ranjitjhala:
Fixes #661 

Depends on 

- https://github.com/ucsd-progsys/liquid-fixpoint/pull/699
- https://github.com/ucsd-progsys/liquid-fixpoint/pull/700


ranjitjhala:
@nilehmann  guess we'll have to rerun tests after merging in the fixpoint PR? or does the CI use some static `fixpoint` binary?

crates/flux-fixpoint/src/big_int.rs
```diff
@@ -27,6 +27,13 @@ impl BigInt {
     pub const ZERO: BigInt = BigInt { sign: Sign::NonNegative, val: 0 };
     pub const ONE: BigInt = BigInt { sign: Sign::NonNegative, val: 1 };
 
+    pub fn is_negative(&self) -> bool {
+        self.sign == Sign::Negative
+    }
+
+    pub fn val(&self) -> u128 {
+        self.val
+    }
```
nilehmann, line 36: I prefer if we don't expose the representation of integers. I'd add a formatting method like so

```rust
    pub fn fmt_sexp(self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self.sign {
            Sign::NonNegative => write!(f, "{}", self.val),
            Sign::Negative => write!(f, "(- {})", self.val),
        }
    }
```

crates/flux-fixpoint/src/constraint.rs
```diff
@@ -360,31 +360,31 @@ impl<T: Types> fmt::Display for Expr<T> {
                 write!(f, "({func} {})", args.iter().map(FmtParens).format(" "),)
             }
             Expr::Neg(e) => {
-                write!(f, "-{}", FmtParens(e))
+                write!(f, "(- 0 {})", FmtParens(e))
```
nilehmann, line 363: You are printing negative constants as `(- {val})`. Does that only work for literals? or can we change this to `write!("(- {})", FmtParens(e))`?

crates/flux-fixpoint/src/lib.rs
```diff
@@ -233,15 +232,7 @@ impl<T: Types> fmt::Display for Task<T> {
...
-        write!(
-            f,
-            "(var ${} ({})) // {}",
-            self.kvid,
-            self.sorts
-                .iter()
-                .format_with(" ", |sort, f| f(&format_args!("({sort})"))),
-            self.comment
-        )
+        write!(f, "(var ${} ({})) // {}", self.kvid, self.sorts.iter().join(" "), self.comment)
```
nilehmann, line 235: Unrelated to this PR, but how hard would be to change the parser to accept `;` for comments instead of `//`?

nilehmann (approved):
Some small comments

crates/flux-fixpoint/src/constraint.rs
ranjitjhala: Good catch, we don't need the `(- 0 {})` -- just `(- {})` will suffice!

crates/flux-fixpoint/src/lib.rs
ranjitjhala: You "nerdsniped" me into doing it; will just put it in this PR.

It was a boring but tedious thing to add in the [SMTParser](https://github.com/ucsd-progsys/liquid-fixpoint/pull/700)

nilehmann: That was the intention :P

PR #690, opened 2024-08-08
Support for string-literals as indices

### LINKED ISSUE #657: Support strings in refinements
We should have a string sort encoded as a string in z3/fixpoint and refine `str` by it. The following is a minimal test case we should support

```rust
#[flux::spec(
fn require_eq(x: &str[@a], &{str[@b] | a == b}))]
fn require_eq(x: &str, y: &str) {}

fn test_good() {
    require_eq("a", "a");
}

fn test_bad() {
    require_eq("a", "b");
}
```

Right now both `test_good` and `test_bad` are accepted by Flux, because strings are refined by unit.

ranjitjhala:
fixes #657


ranjitjhala:
@nilehmann how come the clippy complains in the GH conversation but no such thing when I run locally? 🤔

crates/flux-middle/src/rustc/mir.rs
```diff
@@ -281,7 +281,7 @@ pub enum Constant {
     Float(u128, FloatTy),
     Bool(bool),
     /// We only support opaque string slices, so no data stored here for now.
```
nilehmann, line 283: This comment is not longer accurate 

```suggestion
```

nilehmann (approved):

PR #691, opened 2024-08-08
Add support for bitvector extensions

### LINKED ISSUE #686: Supporting bitvector extensions
Currently Flux does not support extending bitvectors (zero-extending or sign-extending). This issue adds a test case that we would expect to pass if bitvector extension were implemented.
We would expect the following code to pass:
```Rust
#[allow(dead_code)]
#[flux::sig(fn(x: bool[true]))]
pub fn assert(_x: bool) {}


#[flux::opaque]
#[flux::refined_by(v: bitvec<32>)]
struct Register {
    inner: u32,
}

impl Register {
    #[flux::sig(fn(u32[@n]) -> Register[bv_int_to_bv32(n)])]
    #[flux::trusted]
    fn new(v: u32) -> Self {
        Register { inner: v }
    }

    #[flux::sig(fn(&Register[@n]) -> u64[bv_zext_32_64(n)])]
    #[flux::trusted]
    fn zero_extend(&self) -> u64 {
        self.inner as u64
    }

    #[flux::sig(fn(&Register[@n]) -> u64[bv_sext_32_64(n)])]
    #[flux::trusted]
    fn sign_extend(&self) -> u64 {
        self.inner as i32 as i64 as u64
    }
}

fn test_bv_extensions(){
    let r = Register::new(u32::MAX);
    assert(r.zero_extend() == u32::MAX as u64);
    assert(r.sign_extend() == u64::MAX);
}
```
Currently this code throws an error because I made up the symbols for zero/sign extension.

ranjitjhala:
fixes #686


ranjitjhala:
@nilehmann this seems straightforward enough that I'll just merge it, we can revert if you find anything egregious!

crates/flux-middle/src/lib.rs
```diff
@@ -76,6 +76,22 @@ pub static THEORY_FUNCS: LazyLock<UnordMap<Symbol, TheoryFunc>> = LazyLock::new(
...
     [
         // BitVector <-> int
+        TheoryFunc {
+            name: Symbol::intern("bv_zero_extend_32_to_64"),
+            fixpoint_name: Symbol::intern("app (_ zero_extend 32)"),
+            sort: rty::PolyFuncSort::new(
+                List::empty(),
+                rty::FuncSort::new(vec![BitVec(BvSize::Fixed(32))], BitVec(BvSize::Fixed(64))),
+            ),
+        },
```
nilehmann, line 86: @ranjitjhala I think this is wrong. `(_ zero_extend 32)` *returns* a bv of size 32. The declaration should be

```rust
        TheoryFunc {
            name: Symbol::intern("bv_zero_extend_to_64"),
            fixpoint_name: Symbol::intern("app (_ zero_extend 64)"),
            sort: rty::PolyFuncSort::new(
                List::singleton(bv_param0),
                rty::FuncSort::new(vec![BitVec(bv_param0)], BitVec(BvSize::Fixed(64))),
            ),
        }
  ```
  
The test is working because `(_ bv2int 64)`, how `bv_bv64_to_int` is encoded as in z3, seems to ignore the `64`.

**Comment by nilehmann** [comment]
don't mind my previous comment, I misunderstood it

**Comment by ranjitjhala** [comment]
Yes, I was fiddling around with that in Z3 to understand and I noted its "additive"

See https://z3prover.github.io/api/html/ml/Z3.BitVector.html

For example, running this through fixpoint

```
(constraint
    ((= (app (_ zero_extend 5) (lit "#x00000000" (BitVec Size32)))
        (lit "#x00000000" (BitVec Size32))
    ))
)
```

makes z3 complain with

```
  crash: SMTLIB2 respSat = Error "line 3 column 9562: Sorts (_ BitVec 37) and (_ BitVec 32) are incompatible"
```

nilehmann:
yeah, I noticed the same, that's why I was confused. 

If that's the behavior, to support the full thing we can't simply expose it as a theory function and we need to have dedicated sort-checking rules for it.

ranjitjhala:
Yes but the current “monomorphization” is fine I think? (As in sort
checking works for 32/64 etc.) It seems fixpoint also hardwires a fixed set
of bitvec widths.


- Ranjit.

ranjitjhala:
As in — all the more reason NOT to support the full thing till we have some
good reason to :-)

- Ranjit.

PR #704, opened 2024-08-13
Bug fixes for array code

### LINKED ISSUE #703: ICE when instantiating generic for array/iterator
The following (a simplification from https://github.com/scroll-tech/zkevm-circuits/blob/47e4cb7394b5b72be3300b9aa38733b200bf2ad5/zkevm-circuits/src/copy_circuit/copy_gadgets.rs#L62) 

```rust
pub fn gather<I: IntoIterator<Item = i32>>(inputs: I) {}

pub fn test() {
    gather([1, 2, 3])
}
```

makes `flux` yell with

```
error: internal compiler error: crates/flux-middle/src/rty/projections.rs:292:21: cannot infer substitution for param GenericParamDef { name: "N", def_id: DefId(2:4176 ~ core[5d62]::array::iter::{impl#0}::N), index: 1, pure_wrt_drop: false, kind: Const { has_default: false, is_host_effect: false, synthetic: false } }
 --> tests/tests/todo/inst00.rs:4:5
 ```

### PR DESCRIPTION
**Pull request by ranjitjhala**
- Code to convert `Array` back into `rustc_ty`
- Code to subst the `Const` part of arrays during normalization (checking candidates)

fixes #703

### CODE REVIEW

File: crates/flux-desugar/src/lib.rs [hunk 1]
```diff
@@ -105,8 +103,18 @@ pub fn desugar<'genv>(
                         fhir::Node::Item(genv.alloc(cx.desugar_impl(owner_id, impl_)?)),
                     );
                 }
+                // hir::ItemKind::OpaqueTy(_) => {
```
ranjitjhala, line 106: I didn't follow why these were two separate things, so I merged them and made it pop out an `unsupported` to allow progress, lmk if this should be reverted!

ranjitjhala:
fixes #703

crates/flux-desugar/src/lib.rs
nilehmann: Do you have a minimal example that triggers this?

This shouldn't happen because an `OpaqueTy` is desugared as part of desugaring a function, e.g., if we have 
```rust
fn foo() -> impl SomeTrait
```
the item corresponding to the opaque type generated for the `impl SomeTrait` should be desugared when we desugar `foo` so I don't see how one could be querying the `DefId` of an opaque type if the function hasn't been desugared yet.

ranjitjhala:
There was something in the ZKEVM let me dig it up…

- Ranjit.

crates/flux-middle/src/rustc/lowering.rs
```diff
@@ -50,7 +50,7 @@ pub struct UnsupportedReason {
 }
 
 impl UnsupportedReason {
-    fn new(reason: impl ToString) -> Self {
+    pub fn new(reason: impl ToString) -> Self {
```
nilehmann, line 53:
```suggestion
    fn new(reason: impl ToString) -> Self {
```

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -420,7 +420,8 @@ pub fn check_crate_wf(genv: GlobalEnv) -> Result<(), ErrorGuaranteed> {
             | DefKind::AssocFn
             | DefKind::Trait
             | DefKind::Impl { .. }
-            | DefKind::OpaqueTy => {
+            // | DefKind::OpaqueTy
```
nilehmann, line 423: Maybe we should leave a comment about why we are skipping this

nilehmann (approved):

PR #707, opened 2024-08-13
Use target type as default for unsized casts

ranjitjhala:
I added a small test that shows why we need this.

Its probably very conservative (i.e. loses information, but is it unsound?) if so we should probably add a test that shows why.


crates/flux-refineck/src/checker.rs
```diff
@@ -907,6 +907,43 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            // &T  -> & dyn U
+            if let rustc::ty::TyKind::Dynamic(_, _) = to_ty.kind() {
+                self.genv
+                    .refine_default(&self.generics, to)
+                    .with_span(self.body.span())
+            } else {
+                tracked_span_bug!("unsupported Unsize cast: from {from:?} to {to:?}")
+            }
+        } else {
+            // TODO: may be too "conservative"? Can it be unsound?
```
nilehmann, line 933: @ranjitjhala this is unsound for a `&mut [T;n] -> &mut [T][n]` coercion because one should be *invariant* on `T`. Consider

```rust
#[flux::sig(fn([u8{v: v > 0}; 4]) -> [u8{v: v > 0}; 4])]
pub fn test01(mut b: [u8; 4]) -> [u8; 4] {
    test02(&mut b);
    b
}

// This function is crashing for an unrelated reason but it should verify
#[flux::trusted] 
pub fn test02(x: &mut [u8]) {
    if x.len() > 0 {
        x[0] = 0;
    }
}
```

nilehmann: I'm working on a proper fix

ranjitjhala: So should `block_with` be used when the target is a `mut` as done when checking function params in `check_fn_call`?

nilehmann: yes, we need to convert the (strong) pointer to a mut reference by blocking the pointee. After that the cast is the same as casting a mut ref. I'll implement that, but before I'm abstracting some of the pointer to ref logic, because we reimplement it in a couple of places and it is a bit subtle.

ranjitjhala: OK so I'll let you deal with this one then?

ranjitjhala:
or should i patch in the narrow case for the original test (`b.reverse()`)

crates/flux-refineck/src/checker.rs
nilehmann: Let me deal with it.

nilehmann:
@ranjitjhala I think we can close this, right?

ranjitjhala:
Yup

- Ranjit.

PR #716, opened 2024-08-15
Convert ptr to ref before unsize cast

nilehmann:


nilehmann:
I'm going to merge this because not crashing is an improvement, but added https://github.com/flux-rs/flux/issues/717 to track the remaining issue

PR #719, opened 2024-08-15
Basic support for unevaluated constants

### LINKED ISSUE #625: Feature Request: Using compile-time constants as array lengths
Hello, it appears that using compile-time constants as array lengths. 
For example, the following code
```Rust
#[flux::constant]
const BUFLEN: usize = 100;
type FixedSizeBuffer = [i32; BUFLEN];
```
fails with the error:
```
error[FLUX]: refinement of unsupported type alias
  --> src/main.rs:36:6
   |
36 |type FixedSizeBuffer = [i32; BUFLEN];
   |      ^^^^ this type alias contains unsupported features
   |
   = note: only interger literals are supported for array lengths
```

You can workaround this by just writing it out as 
```Rust
type FixedSizeBuffer = [i32; 100];
```
directly, but thats a little ugly.

Presumably, adding support for compile-time constant lengths is pretty straight forward, would you mind either (1) implementing it, or (2) pointing me to the appropriate part of the code so that I can implement it?

Thanks

ranjitjhala:
Fixes #625


crates/flux-middle/src/rty/expr.rs
```diff
@@ -611,6 +611,10 @@ impl Expr {
                 let val = const_eval::scalar_int_to_rty_constant2(tcx, *scalar, ty).unwrap();
                 Expr::constant(val)
             }
+            ConstKind::Unevaluated(ue) => {
+                let val = const_eval::unevaluated_to_constant(tcx, ue).unwrap();
+                Expr::constant(val)
+            }
```
nilehmann, line 617: @ranjitjhala This works as a quick hack, but the proper way to do it (following what rustc does) is to evaluate constants when normalizing signature/types, i.e., the implementation of `normalize_projections` should also normalize/evaluate constants (see https://github.com/rust-lang/rust/blob/2c93fabd98d2c183bcb3afed1f7d51b2517ac5ed/compiler/rustc_trait_selection/src/traits/normalize.rs#L415)

nilehmann (changes requested):

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Sounds good - how does that work for structs? Do we normalize their types somewhere? Maybe in conv?

nilehmann: if a function signature contains a struct with const arguments we will normalize it when we normalize the signature.

ranjitjhala:
I meant in the fields of a struct?

nilehmann:
> I meant in the fields of a struct?

When we construct a struct we create a function signature from the struct definition which we will normalize before "calling" it.

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Hmm I think that -- normalizing the struct/field signatures -- is perhaps not happening (or needs to happen somewhere).

I added your suggestion -- to evaluate the const during `normalization` but the following test fails:

```rust
// #[flux::constant] Add this line if you want to refer to `BUFLEN` in *refinement* specifications
const BUFLEN: usize = 100;

struct Blob {
    data: [i32; BUFLEN],
}

pub fn test(buf: &[i32; BUFLEN]) -> i32 {
    let x0 = buf[0];
    let x1 = buf[10];
    let x2 = buf[BUFLEN - 1];
    let xbad = buf[BUFLEN]; //~ ERROR assertion might fail
    x0 + x1 + x2 + xbad
}

pub fn test_blob(blob: Blob) -> i32 {
    let x0 = blob.data[0];
    let x1 = blob.data[10];
    let x2 = blob.data[BUFLEN - 1];
    let xbad = blob.data[BUFLEN]; //~ ERROR assertion might fail
    x0 + x1 + x2 + xbad
}
```

[I assume the normalization should ensure that we have _no_ `UnevaluatedConst` left over, e.g. when actually checking the `rvalue`?]

ranjitjhala: In short, I think there is some spot where `normalization` is not happening, presumably when we lookup the field `.data` ?

ranjitjhala: Hmm maybe a `normalize` in `downcast_struct`?

nilehmann: hmm good point. Normalizing in `downcast_struct` would work (I think), but that sounds a bit wasteful because it will always normalize to the same. I wonder what rustc does in this case 🤔

nilehmann: (btw `#flux::constant` is no longer necessary after https://github.com/flux-rs/flux/pull/682 (the attribute has no effect now))

nilehmann: ah rustc doesn't have `downcast_struct`, they don't even have an analogous to `VariantSig`. I think they normalize everything when lowering into `mir` and they even annotate fields projection with a (presumably) normalized type (https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/type.PlaceElem.html#variant.Field). So yeah, we either

1. Normalize each time in `downcast_struct` (and `downcast_enum`)
2. Make this normalization into a query to cache it
3. Normalization `VariantSig` as part of `conv` 

3 sounds like the easiest and would likely be more complete. I don't know if there are going to be problems down the road for not storing the non-normalized version.

nilehmann: If we do it during `conv` we need to do it after we check structural compatibility

nilehmann: hmm I take it back. From your message in Slack I realize that to normalize we need the information at the call site so we need to do it in `downcast_struct`, so it is not that wasteful

tests/tests/neg/surface/const04.rs
```diff
@@ -0,0 +1,24 @@
+// https://github.com/flux-rs/flux/issues/625
+
+// #[flux::constant] Add this line if you want to refer to `BUFLEN` in *refinement* specifications
```
nilehmann, line 3: `flux::constant` is no longer necessary 

```suggestion
```

crates/flux-middle/src/rustc/ty.rs
```diff
@@ -234,6 +241,10 @@ impl Const {
         }
     }
 
+    pub fn from_scalar_int(scalar_int: ScalarInt, ty: Ty) -> Self {
+        Self { kind: ConstKind::Value(ty, scalar_int) }
+    }
```
nilehmann, line 246: This feels a bit unnecessary 

```suggestion
```

crates/flux-middle/src/rustc/lowering.rs
```diff
@@ -663,6 +663,10 @@ fn lower_const<'tcx>(
         rustc_type_ir::ConstKind::Value(ty, ValTree::Leaf(scalar_int)) => {
             ConstKind::Value(lower_ty(tcx, ty)?, scalar_int)
         }
+        rustc_type_ir::ConstKind::Unevaluated(c) => {
+            let args = lower_generic_args(tcx, c.args)?;
```
nilehmann, line 667: I doubt this will become a problem in the near future but we could raise unsupported if the list of args is not empty.

nilehmann (approved):

nilehmann (commented):
LGTM after fixing the merge conflict

PR #736, opened 2024-08-20
Fall back to `rustc` when normalizing

ranjitjhala:
As a temporary "stopgap" till we figure out the `ParamEnv` situation, here's a fallback that just queries `rustc` in case `flux` can't find a suitable candidate.


nilehmann (commented):
ok, let's merge this

PR #741, opened 2024-08-22
Add support for intrinsic-assume and discriminant-to-int-cast

ranjitjhala:


crates/flux-refineck/src/checker.rs
```diff
@@ -398,6 +398,11 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 // otherwise be optimized away.
             }
             StatementKind::Nop => {}
+            StatementKind::Intrinsic(NonDivergingIntrinsic::Assume(_op)) => {
+                // Currently, we only have the `assume` intrinsic, which if we're to trust rustc should be a NOP.
+                // TODO: There may be a use-case to actually "assume" the bool index associated with the operand,
+                // i.e. to strengthen the `rcx` / `env` with the assumption that the bool-index is in fact `true`...
+            }
```
nilehmann, line 405: We should at least check the operand for the side-effects

```suggestion
            StatementKind::Intrinsic(NonDivergingIntrinsic::Assume(op)) => {
                // Currently, we only have the `assume` intrinsic, which if we're to trust rustc should be a NOP.
                // TODO: There may be a use-case to actually "assume" the bool index associated with the operand,
                // i.e. to strengthen the `rcx` / `env` with the assumption that the bool-index is in fact `true`...
                let _ =  self.check_operand(rcx, env, stmt_span, op)?;
            }
```

crates/flux-refineck/src/checker.rs
```diff
@@ -944,7 +949,25 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                let mut upvar_tys = vec![];
+                for ty in &operand_tys {
+                    let ref_ty = if let TyKind::Ptr(PtrKind::Mut(re), path) = ty.kind() {
+                        env.ptr_to_ref(
+                            rcx,
+                            &mut infcx,
+                            ConstrReason::Other,
+                            *re,
+                            path,
+                            PtrToRefBound::Identity,
```
nilehmann, line 963: We probably want this to be `PtrToRefBound::Infer` because we want to infer a (possibly weaker) invariant on the location preserved by the closure.

```suggestion
                            PtrToRefBound::Infer,
```

nilehmann: Also, a stylistic suggestion (you'll need to import `flux_common::iter::IterExt`)

```rust
                let upvar_tys = operand_tys.into_iter()
                    .map(|ty| {
                        if let TyKind::Ptr(PtrKind::Mut(re), path) = ty.kind() {
                            env.ptr_to_ref(
                                rcx,
                                &mut infcx,
                                ConstrReason::Other,
                                *re,
                                path,
                                PtrToRefBound::Infer,
                            )
                        } else {
                            Ok(ty.clone())
                        }
                    })
                    .try_collect_vec()
                    .with_span(stmt_span)?;
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -340,6 +340,13 @@ impl Expr {
             .unwrap_or_else(Expr::tt)
     }
 
+    pub fn or_iter(exprs: impl IntoIterator<Item = Expr>) -> Expr {
+        exprs
+            .into_iter()
+            .reduce(|acc, e| Expr::binary_op(BinOp::Or, acc, e, None))
+            .unwrap_or_else(Expr::ff)
+    }
```
nilehmann, line 348: We already have `Expr::or_from_iter`

```suggestion
```

crates/flux-refineck/src/ghost_statements/fold_unfold.rs
```diff
@@ -341,6 +341,7 @@ impl<'a, 'genv, 'tcx, M: Mode> FoldUnfoldAnalysis<'a, 'genv, 'tcx, M> {
             | StatementKind::FakeRead(_)
             | StatementKind::AscribeUserType(_, _)
             | StatementKind::PlaceMention(_)
+            | StatementKind::Intrinsic(NonDivergingIntrinsic::Assume(_))
```
nilehmann, line 344: You should call `self.operand` on the operand

crates/flux-refineck/src/checker.rs
```diff
@@ -1139,6 +1168,14 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
         Ok(ty)
     }
 
+    fn discr_to_int_cast(adt_def: &AdtDef, bty: BaseTy) -> Ty {
+        let vals = adt_def
+            .discriminants()
+            .map(|(_, idx)| Expr::eq(Expr::nu(), Expr::from_bits(&bty, idx)))
+            .collect_vec();
```
nilehmann, line 1175: This could be a huge list. We can optimize it by finding contiguous intervals. Presumably, that could be easier for z3, but it's fine if we don't. I doubt this will cause problems now, but it sounds like a fun thing to implement!

nilehmann (approved):
A couple of suggestions and things you should fix before merging, but it looks good. Feel free to merge after fixing the issues.

crates/flux-middle/src/rty/expr.rs
ranjitjhala: oops don't know why I missed that!

crates/flux-refineck/src/checker.rs
ranjitjhala: Yes, I'll leave a TODO there for now, not clear what the best approach is... its easy and more efficient but "less precise" to make it an interval..

PR #760, opened 2024-08-29
Allow explicit lifetimes in extern specs for structs

enjhnsn2:
I added a line to the `flux_rs::extern_spec` macro to allow it to parse explicit lifetimes in extern structs. This enables extern specs like:
```Rust
#[extern_spec(std::slice)]
#[flux::refined_by(len: int)]
struct Iter<'a, T>;
```


nilehmann (approved):
Looks good. Thanks @enjhnsn2

PR #781, opened 2024-09-11
Default Assoc Reft

ranjitjhala:
Here's a whack at the default stuff, the one thing I'm puzzled by is the `TODO` in the comment below...


crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -206,10 +226,30 @@ fn assoc_refinement_def(
...
+    genv: GlobalEnv,
+    impl_id: LocalDefId,
+    name: Symbol,
+) -> QueryResult<rty::EarlyBinder<rty::Lambda>> {
+    let trait_id = genv.tcx().trait_id_of_impl(impl_id.to_def_id());
+
+    if let Some(lam) = impl_assoc_refinement_def(genv, impl_id, name)? {
+        return Ok(lam);
+    }
+    // TODO: this seems wrong, what if the trait_id is not a local id?
```
ranjitjhala, line 245: This bit bothers me... maybe instead one should use some sort of query here?

nilehmann: @ranjitjhala, definitely. This should go through the query instead of calling the local function, i.e., we should call `genv.default_assoc_refinement_def(genv, trait_id, name)`. The query will take care of mapping the right id and then redirect it to the `default_assoc_refinement_def` in this file if necessary.

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -194,11 +195,30 @@ fn assoc_refinements_of(
     Ok(rty::AssocRefinements { items: predicates })
 }
 
-fn assoc_refinement_def(
+fn default_assoc_refinement_def(
+    genv: GlobalEnv,
+    trait_id: LocalDefId,
```
nilehmann, line 200: I recently introduced [`MaybeExternId`](https://github.com/flux-rs/flux/blob/274edb4f110ee72640b896cdc05d6bb67bab7e45/crates/flux-middle/src/lib.rs#L417) to express in the type system that a local id may be for an extern item, it's not a perfect abstraction, but you at least have a reminder to be careful. Not all the code has been transitioned to use it but going forward we should start all the functions in this file implementing a query with a call to `genv.maybe_extern_id` and use the result to shadow the id passed as argument.

nilehmann (changes requested):

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -207,10 +227,30 @@ fn assoc_refinement_def(
...
+        Ok(None)
+    }
+}
+
+fn assoc_refinement_def(
+    genv: GlobalEnv,
+    impl_id: LocalDefId,
+    name: Symbol,
+) -> QueryResult<rty::EarlyBinder<rty::Lambda>> {
+    let trait_id = genv.tcx().trait_id_of_impl(impl_id.to_def_id());
```
nilehmann, line 241: re https://github.com/flux-rs/flux/pull/781/files#r1753073697: We should use the _resolved id_ here, i.e., if `impl_id` is for an extern spec we should use the def_id of the extern item. Using `MaybeExternId` makes it a bit easier to catch these errors. There's no hard rule, but in general, one should use the resolved id when calling queries either in `tcx` or `genv`.

crates/flux-middle/src/queries.rs
```diff
@@ -536,6 +542,27 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
+            let trait_id = lookup_extern(genv, trait_id).unwrap_or(trait_id);
+            if let Some(local_id) = trait_id.as_local() {
+                (self.providers.default_assoc_refinement_def)(genv, local_id, name)
+            } else if let Some(lam) = genv
+                .cstore()
+                .default_assoc_refinements_def((trait_id, name))
+            {
+                lam
+            } else {
+                bug!("TODO: implement for external crates")
```
nilehmann, line 561: Prefer `query_bug!` if it's easy to return a `QueryResult`.

```suggestion
                Err(query_bug!("TODO: implement for external crates"))
```

Also, I think this is truly unreachable rather than a TODO.

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -194,11 +195,32 @@ fn assoc_refinements_of(
     Ok(rty::AssocRefinements { items: predicates })
 }
 
-fn assoc_refinement_def(
+fn default_assoc_refinement_def(
+    genv: GlobalEnv,
+    trait_id: LocalDefId,
+    name: Symbol,
+) -> QueryResult<Option<rty::EarlyBinder<rty::Lambda>>> {
+    let impl_id = trait_id.to_def_id();
```
nilehmann, line 203: The idea is not to use `trait_id` at all before calling `maybe_extern_id`. If you _really_ need the local id converted to a `DefId` you need to do `trait_id.local_id().to_def_id()`, that being said, I think the `impl_id` in `InvalidAssocReft` should be the resolve id.

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -207,10 +229,33 @@ fn assoc_refinement_def(
...
+        Ok(None)
+    }
+}
+
+fn assoc_refinement_def(
+    genv: GlobalEnv,
+    impl_id: LocalDefId,
+    name: Symbol,
+) -> QueryResult<rty::EarlyBinder<rty::Lambda>> {
+    let err = Err(QueryErr::InvalidAssocReft { impl_id: impl_id.to_def_id(), name });
```
nilehmann, line 243: Same here. the very first thing should be `let impl_id = genv.maybe_extern_id(impl_id)`. I also thing the `impl_id` should be the resolved id.

crates/flux-fhir-analysis/src/lib.rs
nilehmann: We should also change the name of `impl_id` in the definition of `QueryErr::InvalidAssocReft` to be `container_id` or `container_def_id` now that it can refer to a trait.

ranjitjhala:
Great -- the `resolved_id()` makes various things simpler, thanks!

PR #786, opened 2024-09-12
Add example of an extern spec for std::slice::Iter and Enumerate

enjhnsn2:
Here is the enumerate example, as requested by @ranjitjhala


ranjitjhala:
Awesome thanks @enjhnsn2 !!!

PR #788, opened 2024-09-12
Pulling the iter/enum specs out into a separate file

ranjitjhala:


tests/tests/pos/surface/iter02.rs
```diff
@@ -1,115 +1,53 @@
...
+fn test_iter1(slice: &[u8]) {
     let mut iter = slice.iter();
     let next = iter.next();
     assert(next.is_some());
 }
 
-#[flux_rs::should_fail]
-#[flux_rs::sig(fn(slice: &[u8]{n: n > 0}))]
-fn test_iter1_neg(slice: &[u8]){
+#[flux::should_fail]
```
nilehmann, line 24: should this go in a `neg` test so we can annotate exactly where it fails?

nilehmann (approved):

PR #791, opened 2024-09-12
Treat `ConstantIndex` places just like `Index`

### LINKED ISSUE #725: Add support for `PlaceElem::ConstantIndex`
The next big block of errors in `zkevm` is due to code like the below

```rust
fn test() -> usize {
    let [x, y] = [0, 1];
    x
}
```

I wonder if either you should handle this @nilehmann or whether instead this may be a good intro to the whole places/tree business for me to look at?

ranjitjhala:
fixes #725


ranjitjhala:
Hmm not so fast, this test ICEs 

```rust
struct Bob {}

fn test2() -> Bob {
    let b0 = Bob {};
    let b1 = Bob {};
    let [x, y] = [b0, b1];
    x
}
```

at thread 'rustc' panicked at crates/flux-refineck/src/type_env.rs:256:13:

nilehmann (approved):
lgtm

PR #800, opened 2024-09-16
Barebones function subtyping

### LINKED ISSUE #793: ICE on FnDef
Code like the below 

```rust
#[flux::sig(fn (x:usize) -> usize[x+1])]
fn inc(x: usize) -> usize {
    x + 1
}

#[flux::sig(fn(c: Option<usize[10]>) -> Option<usize[11]>)]
pub fn test(c: Option<usize>) -> Option<usize> {
    c.map(inc)
}
```

triggers 

```
error[E0999]: unsupported terminator
 --> tests/tests/todo/fndef.rs:8:5
  |
8 |     c.map(inc)
  |     ^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-middle/src/rustc/lowering.rs:274:34
  |
  = note: unsupported type `FnDef(DefId(0:3 ~ fndef[f288]::inc), [])`
```

Seems to be an issue in both vtock and zkevm...

ranjitjhala:
fixes #793


crates/flux-infer/src/infer.rs
```diff
@@ -494,14 +510,11 @@ impl Sub {
...
-                let variances = infcx.genv.variances_of(adt_a.did());
-                for (variance, ty_a, ty_b) in izip!(variances, args_a.iter(), args_b.iter()) {
-                    self.generic_args(infcx, *variance, ty_a, ty_b)?;
-                }
-                Ok(())
+            (BaseTy::Adt(a_adt, a_args), BaseTy::Adt(b_adt, b_args)) => {
+                self.btys_def_id_args(infcx, a_adt.did(), a_args, b_adt.did(), b_args)
+            }
+            (BaseTy::FnDef(a_def_id, a_args), BaseTy::FnDef(b_def_id, b_args)) => {
+                self.btys_def_id_args(infcx, *a_def_id, a_args, *b_def_id, b_args)
```
nilehmann, line 517: This is related to how we handle generic arguments. Since they are always going to be unrefined maybe better to have a stricter assertion until we have a concrete example.

```suggestion
                debug_assert_eq!(a_def_id, b_def_id);
                assert_eq!(a_args, b_args);
                Ok(())
```

The first one is `debug_assert` because I'm confident that it will never fail. The second one we want to catch when so we have a concrete example to fix.

nilehmann (changes requested):

crates/flux-refineck/src/checker.rs
```diff
@@ -638,27 +638,141 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            .inputs()
+            .iter()
+            .map(|ty| infcx.unpack(ty))
+            .collect_vec();
+
+        // 2. Fresh names for `T_f` refine-params / Instantiate fn_def_sig and normalize it
+        infcx.push_scope();
+        let refine_args = infcx.instantiate_refine_args(*def_id).with_span(span)?;
+        let fn_def_sig = fn_def_sig
+            .instantiate(tcx, generic_args, &refine_args)
```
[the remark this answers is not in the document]
ranjitjhala: Splitting into a separate issue for now?

crates/flux-infer/src/infer.rs
```diff
@@ -482,6 +482,23 @@ impl Sub {
...
+        b_args: &[GenericArg],
+    ) -> InferResult {
+        debug_assert_eq!(a_def_id, b_def_id);
+        debug_assert_eq!(a_args.len(), b_args.len());
+        let variances = infcx.genv.variances_of(a_def_id);
+        for (variance, ty_a, ty_b) in izip!(variances, a_args.iter(), b_args.iter()) {
+            self.generic_args(infcx, *variance, ty_a, ty_b)?;
+        }
+        Ok(())
+    }
```
nilehmann, line 500: We can re-inline this.

nilehmann (approved):
OK let's merge this. We can iterate on it later.

PR #804, opened 2024-09-17
Don't keel over on trivial Fnptr subtyping

### LINKED ISSUE #803: ICE on FnPtr
[Stripped down test from `zkevm`] 

The following code 

```rust
#![allow(unused)]

type Binop = fn(i32, i32) -> i32;

struct Foo {
    f: Binop,
}

#[flux::trusted]
fn bar(b: Binop) -> i32 {
    b(3, 4)
}

fn test00(foo: Foo) -> i32 {
    let x = bar(foo.f);
    x
}
```

currently crashes with

```
error[E0999]: internal flux error: crates/flux-infer/src/infer.rs:553:22
  --> tests/tests/todo/fnptr00.rs:15:13
   |
15 |     let x = bar(foo.f);
   |             ^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-middle/src/queries.rs:794:45
   |
   = note: incompatible base types: `fn(∃b0. i32[b0], ∃b1. i32[b1]) -> ∃b2. i32[b2]` - `fn(∃b0. i32[b0], ∃b1. i32[b1]) -> ∃b2. i32[b2]`
```

i.e. at the subtyping check for `FnPtr`... is there any reason NOT to conservatively (?) check that the base types are the same in this case and then move on?

ranjitjhala:
fixes #803


nilehmann (approved):
This should do for now.

PR #805, opened 2024-09-17
Fndef poly

### LINKED ISSUE #802: Better support for `FnDef` subtyping
From #800 

I'm a bit confused about where these `generic_args` are generated and how they end up here. As it is right now I think it's sound because these args are always unrefined but eventually we would want them to be refined. For example, as implemented right now we can't prove the following correct

```rust
fn id<T>(x: T) -> T {
    x
}

#[flux::sig(fn(c: Option<usize[99]>) -> Option<usize[99]>)]
pub fn test_ok(c: Option<usize>) -> Option<usize> {
    c.map(id)
}
```

_Originally posted by @nilehmann in https://github.com/flux-rs/flux/pull/800#discussion_r1762098026_

ranjitjhala:
Fixes #802


ranjitjhala:
So merge this?

ranjitjhala:
speculatively merging -- can undo if needed its small enough!

PR #811, opened 2024-09-18
Use `ValTree` to represent constants

ranjitjhala:


crates/flux-infer/src/infer.rs
```diff
@@ -590,7 +590,11 @@ impl Sub {
             }
             (GenericArg::Lifetime(_), GenericArg::Lifetime(_)) => return Ok(()),
             (GenericArg::Const(c1), GenericArg::Const(c2)) => {
-                debug_assert_eq!(c1, c2);
+                // TODO(RJ): hack for tests/tests/pos/surface/const05.rs
+                match (&c1.kind, &c2.kind) {
+                    (rty::ConstKind::Unevaluated(_), _) | (_, rty::ConstKind::Unevaluated(_)) => (),
+                    (_, _) => debug_assert_eq!(c1, c2),
+                }
```
nilehmann, line 597: This is a bug in the normalization code. We need to update this function https://github.com/flux-rs/flux/blob/main/crates/flux-middle/src/rty/projections.rs/#L299 to evaluate the val tree. There's a `normalize` method on `rustc_middle::ty::Const` I think we can use.

nilehmann (changes requested):

crates/flux-middle/src/rty/projections.rs
```diff
@@ -298,13 +298,16 @@ impl FallibleTypeFolder for Normalizer<'_, '_, '_> {
 
     fn try_fold_const(&mut self, c: &Const) -> Result<Const, Self::Error> {
         let param_env = self.rustc_param_env();
-        let rc = c.to_rustc(self.tcx());
+        let rc = c.to_rustc(self.tcx()).normalize(self.tcx(), param_env);
```
nilehmann, line 301: I think you can simplify this to

```rust
fn try_fold_const(&mut self, c: &Const) -> Result<Const, Self::Error> {
    c.to_rustc(self.tcx())
        .normalize(self.tcx(), self.rustc_param_env()
        .lower(self.tcx())
        .map_err(|e| QueryErr::unsupported(self.def_id, e.into_err()))
|
```

ranjitjhala: yes indeed, thanks!

nilehmann (approved):
LGTM

PR #813, opened 2024-09-19
Track `GenericArgs` with Closure to support `to_rustc`

ranjitjhala:


crates/flux-middle/src/rty/mod.rs
```diff
@@ -1280,7 +1285,7 @@ pub enum BaseTy {
     Tuple(List<Ty>),
     Array(Ty, Const),
     Never,
-    Closure(DefId, /* upvar_tys */ List<Ty>),
+    Closure(DefId, /* upvar_tys */ List<Ty>, GenericArgs),
```
nilehmann, line 1288: I thought we were going to store a `flux_rustc_bridge::ty::GenericArgs` so we don't have to worry about the refinements, i.e., we store the "shape".

ranjitjhala: We need to convert back to `rustc` -- is there a `to_rustc` for the `flux_rustc_bridge::ty::GenericArgs` ?

ranjitjhala: oh wait there is...

nilehmann (approved):
:shipit:

PR #817, opened 2024-09-26
Use correct container_def_id for serializing extern-specs

### LINKED ISSUE #815: ICE: `TODO: implement for externel crates`
When I uncomment the `iter` spec in Tock ([here](https://github.com/PLSysSec/tock/blob/master/flux_support/src/extern_specs/iter.rs)), it triggers an ICE:
```
error: internal flux error: crates/flux-middle/src/queries.rs:564:21
  |
  = note: TODO: implement for external crates
```
in the `default_assoc_refinement_def` function. This is presumably just because default assoc refinements haven't been implemented for extern crates.

ranjitjhala:
fixes #815


nilehmann (approved):
LGTM

PR #824, opened 2024-09-26
Add invariant for slices

ranjitjhala:
@enjhnsn2 I added your iterator tests too see `tests/tests/surface/neg/extern_spec_iter00.rs` 

your "should be able to prove `r==n`" remark led to this PR!

### DIFF
--- a/crates/flux-common/src/bug.rs
+++ b/crates/flux-common/src/bug.rs
@@ -15,6 +15,9 @@ thread_local! {
 
 pub fn track_span<R>(span: Span, f: impl FnOnce() -> R) -> R {
     TRACKED_SPAN.with(|cell| {
+        if span.is_dummy() {
+            return f();
+        }
         let old = cell.replace(Some(span));
         let r = f();
         cell.set(old);
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -446,8 +446,12 @@ fn variants_of(
     Ok(variants)
 }
 
-fn fn_sig(genv: GlobalEnv, def_id: LocalDefId) -> QueryResult<rty::EarlyBinder<rty::PolyFnSig>> {
-    let def_id = genv.maybe_extern_id(def_id);
+fn fn_sig(
+    genv: GlobalEnv,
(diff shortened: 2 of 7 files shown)


tests/tests/pos/surface/extern_spec_trait00.rs
```diff
@@ -13,7 +13,6 @@ struct MyInt(#[field(i32[n])] i32);
 
 #[assoc(fn eq_rel(lhs: MyInt, rhs: MyInt) -> bool { lhs == rhs })]
 impl PartialEq for MyInt {
-    #[sig(fn(&Self[@lhs], &MyInt[@rhs]) -> bool[<MyInt as PartialEq<MyInt>>::eq_rel(lhs, rhs)])]
```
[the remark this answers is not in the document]
nilehmann: No, the reason it works is because for this particular test `==` is only used generically

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -446,8 +446,12 @@ fn variants_of(
     Ok(variants)
 }
 
-fn fn_sig(genv: GlobalEnv, def_id: LocalDefId) -> QueryResult<rty::EarlyBinder<rty::PolyFnSig>> {
-    let def_id = genv.maybe_extern_id(def_id);
+fn fn_sig(
+    genv: GlobalEnv,
+    orig_def_id: LocalDefId,
```
nilehmann: I like shadowing the input so we don't use it by mistake.

crates/flux-common/src/bug.rs
```diff
@@ -15,6 +15,9 @@ thread_local! {
 
 pub fn track_span<R>(span: Span, f: impl FnOnce() -> R) -> R {
     TRACKED_SPAN.with(|cell| {
+        if span.is_dummy() {
```
nilehmann, line 18: :+1:

nilehmann (approved):
:ship: it

crates/flux-fhir-analysis/src/lib.rs
ranjitjhala: Yes, I reverted.

Issue #816, opened 2024-09-25
Flux panics when ran on CHECK_OVERFLOW mode on Tock


enjhnsn2:
When I run `FLUX_CHECK_OVERFLOW=1 cargo flux` in `tock/kernel`, it fails with:
```
thread 'rustc' panicked at crates/flux-refineck/src/primops.rs:48:35:
called `Option::unwrap()` on a `None` value
```
presumably because it cannot find the overflow-enabled version of a primop being used.

I was thinking this would be a good issue for @Samir-Rashid .

To solve this we need to:
1) deduce what primop is missing
2) add it to the tables in  `crates/flux-refineck/src/primops.rs`

nilehmann:
are you interested in working on it @Samir-Rashid? 

I'd recommend changing the `unwrap` and using `tracked_span_bug` instead to know which line in `vtock` is causing the panic.

Samir-Rashid:
Yes. Thanks I'll do that.

enjhnsn2:
@828 fixes this

PR #828, opened 2024-10-01
Add missing Float UnOp

Samir-Rashid:
Please review the two commit independently. I verified #816 is solved by the first commit. Secondly, to improve Flux errors for next time, I replaced `unwrap` with the debugging version in `refineck`.

### DIFF
--- a/crates/flux-refineck/src/primops.rs
+++ b/crates/flux-refineck/src/primops.rs
@@ -173,14 +173,14 @@ fn mk_mul_rules(check_overflow: bool) -> RuleMatcher<2> {
             requires E::le(a * b, E::uint_max(uint_ty)) => ConstrReason::Overflow
             if let &BaseTy::Uint(uint_ty) = T
 
-            fn(a: T, b: T) -> T // this is underspecified
+            fn(a: T, b: T) -> T
         }
     } else {
         primop_rules!(
             fn(a: T, b: T) -> T[a * b]
             if T.is_integral()
 
-            fn(a: T, b: T) -> T // this is underspecified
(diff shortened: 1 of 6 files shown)


crates/flux-refineck/src/primops.rs
```diff
@@ -365,13 +367,16 @@ fn mk_neg_rules(check_overflow: bool) -> RuleMatcher<1> {
             fn(a: T) -> T[a.neg()]
             requires E::ne(a, E::int_min(int_ty)) => ConstrReason::Overflow
             if let &BaseTy::Int(int_ty) = T
+
+            fn(a: T) -> T[a.neg()]
+            if T.is_float()
```
nilehmann, line 372: Floats are not (currently) refined, so this generates a malformed constraint.

```suggestion
            fn(a: T) -> T
            if T.is_float()
```

crates/flux-refineck/src/primops.rs
```diff
@@ -171,14 +173,14 @@ fn mk_mul_rules(check_overflow: bool) -> RuleMatcher<2> {
             requires E::le(a * b, E::uint_max(uint_ty)) => ConstrReason::Overflow
             if let &BaseTy::Uint(uint_ty) = T
 
-            fn(a: T, b: T) -> T
+            fn(a: T, b: T) -> T // this is underspecified
```
nilehmann, line 176: Please remove these comments.

```suggestion
            fn(a: T, b: T) -> T
```

nilehmann (approved):
Awesome this looks great now!

PR #830, opened 2024-10-01
Solve `ProjectionPredicate` during normalization

### LINKED ISSUE #829: ICE: `cannot infer substitution` during normalization
(Creating a separate issue from #808 FTR)
            
@ranjitjhala This is a bit tricky. When we normalize `<MyType<...> as MyTrait<...>>::Assoc` we get back an impl of the form

```rust
impl<T1, ..., Tn> MyTrait<...> for MyType<..> { 
    type Assoc = ...
}
```
Where `T1,...,Tn` are used as args for `Trait` and `Type`. To proceed with the normalization we need to find a substitution for the `T1,...,Tn` based on the refinements we have. For example, if we have

```rust
impl<T> MyTrait for MyType<T> {
    type Assoc = T;
}
```

And we are trying to normalize `<MyType<Nat> as MyTrait>::Assoc` we get the substitution `T -> Nat` by matching `MyType<Nat>` against `MyType<T>`. This step is important because with this substitution we can infer that `<MyType<Nat> as MyTrait>::Assoc` normalizes to `Nat`.

Now, the implementation of `Iterator` for `FilterMap` looks like:

```rust
impl<B, I: Iterator, F> Iterator for FilterMap<I, F>
where
    F: FnMut(I::Item) -> Option<B>,
{
    type Item = B;
    ...
}
```
The `B` is not used in `Iterator` or in `FilterMap<I, F>` so we can't find a substitution for it when doing the matching. The `B` appears to be unconstrained but it's actually constrained by the bound `F: FnMut(I::Item) -> Option<B>`. This bounds desugars to the bounds

```rust
F: FnMut<I::Item>,
F::Output == Option<B>   
```
The second equality bound is constraining `B`

_Originally posted by @nilehmann in https://github.com/flux-rs/flux/issues/808#issuecomment-2377851856_

ranjitjhala:
Fixes #829


crates/flux-middle/src/rty/projections.rs
```diff
@@ -136,7 +137,47 @@ impl<'genv, 'tcx, 'cx> Normalizer<'genv, 'tcx, 'cx> {
...
+                    None
+                }
+            })
+            .collect();
+
+        // 3. recursively solve each of the projection predicates and unify result with the predicate's term
+        // e.g. given `i32 : Trait1<Assoc=T1>` we
+        //      - `normalize_projection_ty` on `<i32 as Trait1>::Assoc` to get `bool`
+        //      - unify `T1` with `bool`
+        for p in projection_preds.iter() {
```
nilehmann, line 172: I think this is wrong for something like:

```rust
trait Trait1 {
    type Assoc1;
}

impl Trait1 for i32 {
    type Assoc1 = bool;
}

trait Trait2 {
    type Assoc2;
}

struct S<T> {
    fld: T,
}

impl<T1, T2, T3> Trait2 for S<T1>
where
    T2: Trait1<Assoc1 = T3>,
    T1: Trait1<Assoc1 = T2>,
{
    type Assoc2 = T1;
}
```
In this example we first need to process `T1: Trait1<Assoc1 = T2>` (with the solution for `T1`) to solve `T2` before processing `T2: Trait1<Assoc1 = T3>`.

More specifically, we can only process a projection predicate `T1::Assoc = T2` if `T1` is *fully resolved* (note that normalizing something that still has generic parameters is automatically wrong because the parameter is in the context of the implementation, but we normalize in the context of the caller, i.e., the parameter will refer to the wrong thing).

I think we need a loop like so

```
while !pending.is_empty() {
    let ready = extract_fully_resolved(subst, pending);
    for clause in ready {
        // do the thing
    }
}
```

nilehmann: I guess this is exactly why applying a partial instantiation is not such a good idea. Doing `[T1 -> T1, T2 -> i32]` is technically instantiating `T1` in the context of the implementation with `T1` in the context of the caller, so one has to be very careful not to mix them.

nilehmann: actually, the substitution can genuinely instantiate one of the arguments with a parameter in the caller, so after you apply it is impossible to tell which parameters haven't been resolved and which ones are genuinely solved to a parameter in the caller.

ranjitjhala: totally not following what this caller business is about?

ranjitjhala: Ok I think I know what you mean but the _real_ issue is this: Given

* a `TVarSubst`
* a `EarlyBinder<Ty>`

How do you tell if the `TVarSubst` "resolves" the Ty?

ranjitjhala: I suppose one can walk over the `Ty` to gather the indices (debruin vars) and check if all of those are defined in `TVarSubst`?

ranjitjhala: Or! make the `instantiate` return an `Err` if that parameter is not defined...

nilehmann: Either that or to have a fallible instantiation that given a partial substitution returns `None` if the type contains parameters not in the substitution. Not much difference between the two.

(these are early params though so no debruin, but a `ParamTy`)

ranjitjhala: Perhaps we need a separate `instantiate` that takes a "partial" subst, and returns an option/result...

nilehmann: A visitor that returns the set of `ParamTy`s feels useful on its own, so I'd go with that.

ranjitjhala: In this case we need to do the actual substitution though (to get the partial instantiation that can then be recursively normalized...) so think the "fallible instantiation" is the way to go?

nilehmann: you can implement the fallible instantiation by first checking if all mentioned parameters are solved and then substituting, so it's a matter of whether to check first and then substitute, or check as you substitute. Not much difference. I don't feel strong one way or the other.

ranjitjhala: hmm. the former seems a bit redundant imo, so went with the latter.

crates/flux-middle/src/rty/projections.rs
```diff
@@ -316,11 +379,51 @@ struct TVarSubst {
...
+            Some(None) => Err(()),
+            arg => tracked_span_bug!("expected type for generic parameter, found `{arg:?}`"),
+        }
+    }
+
+    fn sort_for_param(
+        &mut self,
+        _param_ty: rustc_middle::ty::ParamTy,
+    ) -> Result<super::Sort, Self::Error> {
+        todo!()
```
nilehmann, line 397:
```suggestion
        tracked_span_bug!()
```

crates/flux-middle/src/rty/projections.rs
```diff
@@ -316,11 +379,51 @@ struct TVarSubst {
...
+    }
+
+    fn ctor_for_param(&mut self, _param_ty: rustc_middle::ty::ParamTy) -> super::SubsetTyCtor {
+        todo!()
```
nilehmann, line 401:
```suggestion
        tracked_span_bug!()
```

crates/flux-middle/src/rty/projections.rs
```diff
@@ -316,11 +379,51 @@ struct TVarSubst {
...
+    }
+
+    fn region_for_param(&mut self, _ebr: rustc_middle::ty::EarlyParamRegion) -> Region {
+        todo!()
```
nilehmann, line 405:
```suggestion
        tracked_span_bug!()
```

crates/flux-middle/src/rty/projections.rs
```diff
@@ -316,11 +379,51 @@ struct TVarSubst {
...
+    }
+
+    fn expr_for_param_const(&self, _param_const: rustc_middle::ty::ParamConst) -> Expr {
+        todo!()
```
nilehmann, line 409:
```suggestion
        tracked_span_bug!()
```

crates/flux-middle/src/rty/projections.rs
```diff
@@ -316,11 +379,51 @@ struct TVarSubst {
...
+    }
+
+    fn const_for_param(&mut self, _param: &Const) -> Const {
+        todo!()
```
nilehmann, line 413:
```suggestion
        tracked_span_bug!()
```

nilehmann (approved):
@ranjitjhala, looks good to merge (modulo small comments), but I'm not sure if it fully solves https://github.com/flux-rs/flux/issues/829. That one involves a `FnOnce::Otput` projection on a function pointer and I have the impression that the implementation of `FnOnce` for function pointers is *built-in*, so I'm not sure how that plays out. It may be ok though. That's an orthogonal issue though, just want to make sure we are not closing the #829 if it's not fully solved.

Issue #839, opened 2024-10-02
bug: The sort Str is not numeric


Samir-Rashid:
This is the first issue in what I believe are multiple Flux bugs when it looks at [this huge function](https://github.com/tock/tock/blob/65dda58c893e5b1fb49da7e4389552584758d5e1/arch/rv32i/src/pmp.rs#L2157). I will make an attempt at looking into this, but we will see if I succeed.

<summary>
<details>

```rs
error: internal compiler error: /home/mod/Documents/github/flux/crates/flux-infer/src/fixpoint_encoding.rs:384:43: fixpoint crash: CrashInfo([Array [Array [Array [Number(-1), Null], String("elaborate solver failed on:\n    a870 == a871\nwith error\n    The sort Str is not numeric\n  because\nCannot unify Str with int in expression: a870 == a871 \n  because\nElaborate fails on a870 == a871\nin environment\n    a870 := Str\n\n    a871 := int")]], String(":1:1-1:1: Error\n  elaborate solver failed on:\n      a870 == a871\n  with error\n      The sort Str is not numeric\n  because\nCannot unify Str with int in expression: a870 == a871 \n  because\nElaborate fails on a870 == a871\n  in environment\n      a870 := Str\n\n      a871 := int")])
    --> arch/rv32i/src/pmp.rs:2159:9
     |
2159 | ...   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Resul...
     |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

thread 'rustc' panicked at /home/mod/Documents/github/flux/crates/flux-infer/src/fixpoint_encoding.rs:384:43:
Box<dyn Any>
```

</details>
</summary>

I made a minimal reproducer
```rs
struct PmpCfg {
    l: bool,
}

impl PmpCfg {
    fn is_set(&self, flag: char) -> bool {
        match flag {
            'l' => self.l,
            _ => false,
        }
    }
}
```

yields this error
```rs
error: internal compiler error: crates/flux-refineck/src/checker.rs:864:41: expected discr
  --> src/main.rs:67:9
   |
67 |         match flag {
   |         ^^^^^^^^^^

thread 'rustc' panicked at crates/flux-refineck/src/checker.rs:864:41:
Box<dyn Any>
```

nilehmann:
@Samir-Rashid, this sounds like something you could tackle.

[...]

PR #840, opened 2024-10-03
fix matching with char

Samir-Rashid:
I have solved the subproblem in issue #839. Still working on the bigger error.


crates/flux-refineck/src/checker.rs
```diff
@@ -845,6 +845,13 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                         idx.clone()
                     }
                 }
+                TyKind::Indexed(BaseTy::Char, idx) => {
+                    if bits == 0 {
+                        idx.not()
+                    } else {
+                        idx.clone()
+                    }
```
Samir-Rashid, line 853: I cannot figure out what `bits` means.

nilehmann: @Samir-Rashid a `switchInt(_1)` roughly means something like

```rust
if _1 == v0 {
   goto bb0
} else if _1 == v1 {
   goto bb1
} else if _1 == v2 {
   ...
} else {
   goto bbn
}
```
i.e., test if `_1` equals a specific value and jump to a new basic block. The `bits` are the bits that make up those values. The content of the bits and what they mean depends on the type of `_1` (e.g., for `bool` the bits can be either 0 or 1)

When we check `switchInt` we try to gain control flow information from assuming the test succeeded. For example, assuming `_1` has type `i32[a]` and we giving the sequence
```rust
if _1 == 0 {
    goto bb0
} else {
   goto bb1
 }
```
we want to check `bb0` (resp. `bb1`) knowing that `a == 0` (resp. `a != 0`). This is captured by associating each successor basic block with `Guard::Pred(p)` where `p` represents the predicate we can assume when checking the basic block.

`char`'s are currently unrefined so there's no control flow information we can assume when checking successors (all successors should be associated with `Guard::None`). So for `chars` we have to ignore the bits.

Samir-Rashid:
https://github.com/flux-rs/flux/pull/840/commits/d76c1493f7b84c5cb7b44bf1b94207929faa179a mostly adds support for refining chars. Only missing thing is creating a `fixpoint::Sort::Char`. Let me know if this is not on the right track or a bad idea for some reason.

nilehmann (commented):
I only looked at it quickly, but it looks good!

I'm probably going to have time on Monday to leave more detailed comments. 

Just a quick thought about encoding: I think z3 has a "theory" of `char`'s, but I have the impression it is not exposed in fixpoint (I may be wrong). If it's not, we can encode them as ints as a first step.

nilehmann:
@Samir-Rashid I took the liberty to merge in the latest changes which were causing a conflict.

crates/flux-desugar/src/desugar.rs
```diff
@@ -1448,6 +1449,9 @@ trait DesugarCtxt<'genv, 'tcx: 'genv> {
             }
             surface::LitKind::Bool => Ok(fhir::Lit::Bool(lit.symbol == kw::True)),
             surface::LitKind::Str => Ok(fhir::Lit::Str(lit.symbol)),
+            surface::LitKind::Char => {
+                Ok(fhir::Lit::Char(u32::from(lit.symbol.as_str().parse::<char>().unwrap())))
```
nilehmann, line 1453: I'd try as much as possible to have them as `Char` so we don't lose the type invariant associated with it.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -47,6 +47,7 @@ pub enum Sort<T: Types> {
     Bool,
     Real,
     Str,
+    // Char, // TODO:
```
nilehmann, line 50: Ok, I asked around, and fixpoint doesn't support the theory of characters. It wouldn't be too hard to implement, but since we don't have a specific use case and in the interest of getting this merged, we can encode `char`s as integers for now. Does that make sense?

crates/flux-desugar/src/desugar.rs
ranjitjhala: I wonder if we can encode `Char` in fixpoint as string of size `1` ?

nilehmann: 🤔 that may be a good idea. I guess the best option depends on the theory functions for strings and how they relate to chars. For the time being, encoding char as integers is the easiest, it wouldn't be to hard to change it later.

ranjitjhala:
I agree - I wouldn’t be surprised if internally z3 encoded it as an Int too!

- Ranjit.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1118,8 +1120,9 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         e2: &rty::Expr,
         scx: &mut SortEncodingCtxt,
     ) -> QueryResult<fixpoint::Expr> {
+        // TODO: I understand zero of this
         let e = match sort {
-            rty::Sort::Int | rty::Sort::Real => {
+            rty::Sort::Int | rty::Sort::Real | rty::Sort::Char | rty::Sort::Str => {
```
Samir-Rashid, line 1124: not completely sure if `Str` should be here. our conversation about `bin_rel_to_fixpoint` has somehow completely evaporated from my mind

Samir-Rashid:
Not sure why you have this branch on 666. How can a scalar int be a char?

https://github.com/flux-rs/flux/blob/7a583efe642f4ddef90b0b00baa436cf89b108b1/crates/flux-rustc-bridge/src/lowering.rs#L649-L666

crates/flux-middle/src/fhir.rs
```diff
@@ -953,7 +953,8 @@ pub enum Lit {
     Int(i128),
     Real(i128),
     Bool(bool),
-    Str(Symbol),
+    Str(Symbol), // `rustc_span::Symbol` interns a value with the type
+    Char(u32),   // all Rust chars are u32s
```
nilehmann, line 957: Make this a `char`

```suggestion
    Char(char), 
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -904,6 +905,7 @@ pub enum Constant {
     Real(Real),
     Bool(bool),
     Str(Symbol),
+    Char(u32),
```
nilehmann, line 908: Make this a `char`
```suggestion
    Char(char),
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -478,6 +478,7 @@ impl Expr {
             }
             BaseTy::Uint(_) => ExprKind::Constant(Constant::from(bits)).intern(),
             BaseTy::Bool => ExprKind::Constant(Constant::Bool(bits != 0)).intern(),
+            BaseTy::Char => ExprKind::Constant(Constant::from(bits)).intern(),
```
nilehmann, line 481: This becomes (or something like this, haven't tested it)

```suggestion
            BaseTy::Char => {
                let c = char::from_u32(bits.try_into::<u32>().unwrap()).unwrap();
                ExprKind::Constant(Constant::Char(c)).intern(),
            }
```

crates/flux-rustc-bridge/src/mir.rs
```diff
@@ -341,8 +342,7 @@ pub enum Constant {
     Float(u128, FloatTy),
     Bool(bool),
     Str(Symbol),
-    /// We only support opaque chars, so no data stored here for now
-    Char,
+    Char(u32),
```
nilehmann, line 345: Make this a `char`

```suggestion
    Char(char),
```

crates/flux-rustc-bridge/src/mir.rs
```diff
@@ -198,6 +198,7 @@ pub enum StatementKind {
     Nop,
 }
 
+/// Corresponds to https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/mir/enum.Rvalue.html
```
nilehmann, line 201: Bare URLs are not automatically turned into clickable links

```suggestion
/// Corresponds to <https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/mir/enum.Rvalue.html>
```

crates/flux-rustc-bridge/src/mir.rs
```diff
@@ -750,7 +750,7 @@ impl fmt::Debug for Constant {
             Constant::Bool(b) => write!(f, "{b}"),
             Constant::Unit => write!(f, "()"),
             Constant::Str(s) => write!(f, "\"{s:?}\""),
-            Constant::Char => write!(f, "\"<opaque char>\""),
+            Constant::Char(c) => write!(f, "\'{}\'", unsafe { char::from_u32_unchecked(*c) }),
```
nilehmann, line 753: This becomes 

```suggestion
            Constant::Char(c) => write!(f, "\'{c}\'"),
```

A note on the way you used `unsafe` for educational purposes because it won't matter after we store a `char` instead of `u32`: The `Constant::Char` constructor is public, so anyone can call it passing an arbitrary integer. This means you *cannot* assume the integer is a valid char inside unsafe code.

crates/flux-rustc-bridge/src/lowering.rs
```diff
@@ -663,7 +663,7 @@ impl<'sess, 'tcx> MirLoweringCtxt<'_, 'sess, 'tcx> {
             TyKind::Float(float_ty) => {
                 Some(Constant::Float(scalar_to_bits(self.tcx, scalar, ty).unwrap(), *float_ty))
             }
-            TyKind::Char => Some(Constant::Char),
+            TyKind::Char => Some(Constant::Char(scalar.to_u32())),
```
nilehmann, line 666: This becomes

```suggestion
            TyKind::Char => Some(Constant::Char(scalar.try_into().unwrap())),
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -1451,6 +1456,9 @@ trait DesugarCtxt<'genv, 'tcx: 'genv> {
...
```
nilehmann, line 1460:
```suggestion
                Ok(fhir::Lit::Char(lit.symbol.as_str().parse::<char>().unwrap()))
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2022,6 +2022,7 @@ fn conv_lit(lit: fhir::Lit) -> rty::Constant {
         fhir::Lit::Real(r) => rty::Constant::Real(rty::Real(r)),
         fhir::Lit::Bool(b) => rty::Constant::from(b),
         fhir::Lit::Str(s) => rty::Constant::from(s),
+        fhir::Lit::Char(s) => rty::Constant::from(s as i128),
```
nilehmann, line 2025:
```suggestion
        fhir::Lit::Char(c) => rty::Constant::Char(c),
```

or you can define `impl From<char> for Constant`

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -596,6 +597,7 @@ fn const_to_fixpoint(cst: rty::Constant) -> fixpoint::Constant {
         rty::Constant::Int(i) => fixpoint::Constant::Numeral(i),
         rty::Constant::Real(r) => fixpoint::Constant::Decimal(r),
         rty::Constant::Bool(b) => fixpoint::Constant::Boolean(b),
+        rty::Constant::Char(c) => fixpoint::Constant::Numeral(BigInt::from(c)),
```
nilehmann, line 600:
```suggestion
        rty::Constant::Char(c) => fixpoint::Constant::Numeral(BigInt::from(u32::from(c))),
```

crates/flux-infer/src/fixpoint_encoding.rs
nilehmann: no, strings should not be here because they can't be compared using `<` or `>` in z3. There's a special function for comparing strings see https://microsoft.github.io/z3guide/docs/theories/Strings

crates/flux-middle/src/fhir.rs
```diff
@@ -1469,6 +1470,7 @@ impl fmt::Debug for Lit {
             Lit::Real(r) => write!(f, "{r}real"),
             Lit::Bool(b) => write!(f, "{b}"),
             Lit::Str(s) => write!(f, "\"{s:?}\""),
+            Lit::Char(c) => write!(f, "\'{}\'", unsafe { char::from_u32_unchecked(*c) }),
```
nilehmann, line 1473:
```suggestion
            Lit::Char(c) => write!(f, "\'{c}\'"),
```

crates/flux-refineck/src/checker.rs
```diff
@@ -1398,7 +1400,10 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 let idx = Expr::constant(rty::Constant::from(*s));
                 Ok(Ty::mk_ref(ReStatic, Ty::indexed(BaseTy::Str, idx), Mutability::Not))
             }
-            Constant::Char => Ok(Ty::char()),
+            Constant::Char(c) => {
+                let idx = Expr::constant(rty::Constant::from(*c as u128));
```
nilehmann, line 1404:
```suggestion
                let idx = Expr::constant(rty::Constant::Char(c));
```
or `rty::Constant::from(c)` if you created the impl

nilehmann (changes requested):
Looks good, but I'd store a `char` instead of an `u32` all the way down and only turn the char into an integer in fixpoint encoding.

A `ScalarInt` is just a set of bits that can represent any *scalar* value, that's it, all the primitive types with a fixed size.

nilehmann (approved):
Awesome, this looks good. Thanks @Samir-Rashid

gmax9:
> Awesome, this looks good. Thanks @Samir-Rashid



> Not sure why you have this branch on 666. How can a scalar int be a char?
> 
> https://github.com/flux-rs/flux/blob/7a583efe642f4ddef90b0b00baa436cf89b108b1/crates/flux-rustc-bridge/src/lowering.rs#L649-L666

Was this solution formally verified?

nilehmann:
Hi @gmax9, I am not sure what you mean by that in the context of this PR. Can you elaborate? While we don't have a full formal model of all Flux we have formalized the key aspects of it (see for example https://dl.acm.org/doi/10.1145/3591283)

PR #850, opened 2024-10-07
fix Constraint Type description typo in docs

nataliepopescu:
Cool project! Was reading the docs and believe this is a typo


ranjitjhala:
Thanks for spotting that @nataliepopescu !

PR #852, opened 2024-10-16
Use `tracked_span` functions in a few places 

ranjitjhala:


ranjitjhala:
Went ahead and merged as its just using the `tracked_span` functions...

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1368,8 +1368,11 @@ impl<'genv, 'tcx, P: ConvPhase> ConvCtxt<'genv, 'tcx, P> {
                     })
                 }
             }
+            fhir::Res::Def(DefKind::Union, def_id) => {
+                tracked_span_bug!("TODO: union types: {def_id:?}")
+            }
             fhir::Res::Def(..) | fhir::Res::Err => {
-                span_bug!(path.span, "unexpected resolution in conv_ty_ctor: {:?}", path.res)
+                tracked_span_bug!("unexpected resolution in conv_ty_ctor: {:?}", path.res)
```
nilehmann, line 1375: `span_bug` is strictly better. If we have access to a span we should always use it.

ranjitjhala: well it was "bypassing" the `FLUX_CATCH_BUGS=1` mode (and crashing...) ?

ranjitjhala: (or wait, am I getting it mixed up with `assert_eq` vs `tracked_span_assert_eq` ?

ranjitjhala: will revert in https://github.com/flux-rs/flux/pull/853

PR #856, opened 2024-10-18
Add sections specifying use of flux:: and flux_rs:: for files vs. crates

vrindisbacher:
@nilehmann, do we want to replace subsequent examples of refinement annotations using `flux::` with `flux_rs::`?

### DIFF
--- a/book/src/guide/run.md
+++ b/book/src/guide/run.md
@@ -47,19 +47,21 @@ To do so add the following to `Cargo.toml`:
 enabled = true
 ```
 
-### Refinement Annotations on a Package
+### Refinement Annotations on a Cargo Projects
 
-Adding refinement annotations to packages is simple. You can add `flux-rs` as a dependency in `Cargo.toml`
+Adding refinement annotations to cargo projects is simple. You can add `flux-rs` as a dependency in `Cargo.toml`
 
 ```toml
 [dependencies]
 flux-rs = { git  = "https://github.com/flux-rs/flux.git" }
 ```
 
-Then, add refinement annoations using the fully qualified path `flux_rs::`.
+Then, import attributes from `flux_rs` and add the appropriate refinement annoations.
 
 ```rust
-#[flux_rs::sig(fn(x: i32) -> i32{v: x < v)]
+use flux_rs::*
+
+#[sig(fn(x: i32) -> i32{v: x < v)]
 fn inc(x: i32) -> i32 {
     x - 1
 }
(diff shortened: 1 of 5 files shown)


book/book/404.html
```diff
@@ -90,9 +90,11 @@
...
...s</a></li><li class="chapter-item expanded "><a href="dev/develop.html"><strong aria-hidden="true">4.</strong> Develop</a></li><li><ol class="section"><li class="chapter-item expanded "><a href="dev/architecture.html"><strong aria-hidden="true">4.1.</strong> Architecture</a></li></ol></li><li class="chapter-item expanded "><li class="part-title">Examples</li><li class="chapter-item expanded "><a href="blog/01-introducing-flux.html"><strong aria-hidden="true">5.</strong> Introducing Flux</a></li><li class="chapter-item expanded "><a href="blog/02-ownership.html"><strong aria-hidden="true">6.</strong> Ownership in Flux</a></li><li class="chapter-item expanded "><a href="blog/03-vectors.html"><strong aria-hidden="true">7.</strong> Refined Vectors</a></li><li class="chapter-item expanded "><div><strong aria-hidden="true">8.</strong> Iterators</div></li><li class="chapter-item expanded "><div><strong aria-hidden="true">9.</strong> Structs</div></li><li class="chapter-item expanded "><div><strong aria-hidden="true">10.</strong> Enums</div></li><li class="chapter-item expanded affix "><li class="spacer"></li><li class="chapter-item expanded affix "><a href="about.html">About</a></li></ol>
```
nilehmann, line 93: I was super confused. We shouldn't be checking these generated files. I don't know how they got in in the first place :S . Can you remove them please and add them to .gitignore

book/src/guide/run.md
```diff
@@ -34,6 +47,24 @@ To do so add the following to `Cargo.toml`:
 enabled = true
 ```
 
+### Refinement Annotations on a Package
```
nilehmann, line 50: I was thinking that package is the wrong name for this. Maybe we should call it a "Cargo Project"

book/src/guide/run.md
```diff
@@ -34,6 +47,24 @@ To do so add the following to `Cargo.toml`:
...
+
+Adding refinement annotations to packages is simple. You can add `flux-rs` as a dependency in `Cargo.toml`
+
+```toml
+[dependencies]
+flux-rs = { git  = "https://github.com/flux-rs/flux.git" }
+```
+
+Then, add refinement annoations using the fully qualified path `flux_rs::`.
```
nilehmann, line 59: I've been tending towards importing all the attributes with `flux_rs::*` and then using them unqualified like so

```rust
use flux_rs::*;

#[sig(fn(x: i32) -> i32{v: x < v)]
fn inc(x: i32) -> i32 {
    x - 1
}
```

It is a matter of taste, but maybe we can suggest that as an alternative.

nilehmann (approved):

nilehmann:
@vrindisbacher 

Yeah, let us change all the examples to use `flux_rs` otherwise it'd be confusing, because `flux::` is not really usable. 

I sent you an invite to give you write access to the repo. Main is branch protected so everything has to go through a PR, but I typically accept PRs if I think they are good enough even with remaining comments so people can fix them and then merge at their discretion.

book/book/404.html
vrindisbacher: Ya strange. Looks like `book/` in the gitignore didn't work. I removed these rendered files.

vrindisbacher:
Great. I updated everything.

Issue #827, opened 2024-09-30
Check overflows on just 1 file


enjhnsn2:
I would like to run the equivalent of `FLUX_CHECK_OVERFLOW=1 cargo flux` but only target 1 file. The concrete use case is that I would like to check for overflows on the `ring_buffer.rs` file without having to check overflows everywhere else.

Nico assures me this would be an easy issue for @Samir-Rashid or @vrindisbacher .

nilehmann:
Yes, this would be easy to implement because the internal code is already parametric (per function) on whether to check for overflows, so it's just a matter of exposing it to the user.

vrindisbacher:
I would be open to taking this. I might need to be pointed in the right direction though.

nilehmann:
@vrindisbacher great!

The relevant points are
1. [This function](https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/lib.rs#L56) is the entry point to the code that checks a function. It takes a `CheckerConfig` that carries whether to check for overflows. Currently, the `CheckerConfig` is initialized by a global configuration and is the same for all invocations of the function.
2. [This module](https://github.com/flux-rs/flux/blob/main/crates/flux-driver/src/collector/mod.rs) *collects* all flux annotations/attributes. The [ignore and trusted](https://github.com/flux-rs/flux/blob/main/crates/flux-driver/src/collector/mod.rs#L480) annotations are the closest to what we want here.

The one tricky aspect that I don't have a definite answer for is the problem of attaching an annotation to an entire module, but we can ponder this later. This is a problem for ignore/trusted annotations as well, but what we do for them now should suffice for the ring buffer use case (I think).

vrindisbacher:
Thanks @nilehmann! @enjhnsn2, could you provide some detail on how you would expect this to work exactly, i.e. what do you do (add annotations, configs etc.) and what is the expected result?

ranjitjhala:
@vrindisbacher I would suggest something like this:

```
FLUX_CHECK_FILE=path/to/filename.rs cargo flux
```

and then you can see how `FLUX_CHECK_DEF` is implemented and adapt it to this case?

@nilehmann does that sound plausible?

nilehmann:
I think @enjhnsn2 wants something different. He wants to enable overflow checking in a single file/module and leave it disabled elsewhere. So I'm imagining an attribute similar to `#[flux::trusted]`.

[...]

PR #861, opened 2024-10-25
Add flux attribute for checking overflows

vrindisbacher:
Adds a flux attribute to check overflows, similar to `flux_rs::trusted`. Fixes (#827).


vrindisbacher:
@nilehmann, I assume I should add some tests? Is there a convention to follow when adding those?

Also, for now, since I figured we would want to keep the global configuration of `check_overflow`, I override `check_overflow` in the `CheckerConfig`. I followed a similar approach to `trusted` for applying the attr to modules.

I tested on my own file (see below) and got expected results.

```rust
#[flux::check_overflow]
mod my_mod {
    fn add(x: u32, y: u32) -> u32 {
        x + y
    }

    fn add2(x: u32) -> u32 {
        x + 2
    }
}

#[flux_rs::refined_by(inner: int)]
struct MyStruct {
    #[field(u32[inner])]
    inner: u32,
}

impl MyStruct {
    #[flux::check_overflow]
    fn add1(&self) -> u32 {
        self.inner + 1
    }

    #[flux::check_overflow]
    fn add2(&self) -> u32 {
        self.inner + 2
    }
}

#[flux::check_overflow]
impl MyStruct {
    fn add3(&self) -> u32 {
        self.inner + 3
    }

    fn add4(&self) -> u32 {
        self.inner + 4
    }
}
```

nilehmann (approved):
Looks good, but yes, please add a test. 

We have two type of tests `pos` and `neg`. A `pos` test is suppose to finish without errors while a `neg` test should finish with one. This split is mostly to separate tests with code that verifies and code that doesn't. I think this doesn't fit the type of features implemented in this PR. But anyhow, you can do something similar to `tests/neg/surface/ignore00.rs` and `tests/pos/surface/ignore00.rs`

Issue #835, opened 2024-10-02
Improve error message when a field is accessed in an opaque struct


nilehmann:
The error message could explain why the field of an opaque struct cannot be accessed and suggest using `#[flux::trusted]`.

[Example of someone confused about it](https://github.com/PLSysSec/tock/pull/3#issuecomment-2384862124)

Relevant pointers:

* https://github.com/flux-rs/flux/blob/199a8ce4324cc86ae2a1fc7d1fa90a51e5817c7d/crates/flux-refineck/src/checker.rs#L1797-L1803
* https://github.com/flux-rs/flux/blob/199a8ce4324cc86ae2a1fc7d1fa90a51e5817c7d/crates/flux-refineck/locales/en-US.ftl#L40-L41

vrindisbacher:
@nilehmann, I can take this if you'd like?

nilehmann:
@vrindisbacher that'd be great! One extra pointer. We use `rustc`'s infrastructure for reporting errors which is documented here https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-structs.html

vrindisbacher:
@nilehmann, how does this look?

File:
```rust
#![allow(unused)]
#![allow(dead_code)]

use std::collections::HashMap;

#[flux_rs::opaque]
struct MyStruct {
    inner: u32,
}

impl MyStruct {
    pub fn inner(&self) -> u32 {
        self.inner
    }
}
```

Output:
```rust
error[E0999]: cannot access fields of opaque struct `MyStruct`.
  --> ../opaque_struct.rs:13:9
   |
13 |         self.inner
   |         ^^^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-refineck/src/lib.rs:108:78
   |
help: If you'd like to use fields of `MyStruct` in methods, use `flux::trusted` instead of `flux::opaque`.
  --> ../opaque_struct.rs:7:1
   |
7  | struct MyStruct {
   | ^^^^^^^^^^^^^^^

error: aborting due to 1 previous error
```

vrindisbacher:
Ah I see. Perhaps, the best way to do this is to have a help span over the function suggesting they add `trusted` and then point them to some documentation (maybe in the flux book) about use of `opaque`?

nilehmann:
Pointing to documentation would be cool, we would have to write it though XD

So ideally, we could have something like

```rust
error[E0999]: cannot access fields of opaque struct `MyStruct`.
[...]

PR #862, opened 2024-10-26
Add additional helpful spans to opaque struct field access diagnostics

vrindisbacher:
See [here](https://github.com/flux-rs/flux/issues/835#issuecomment-2439168376)

### DIFF
--- a/crates/flux-refineck/src/lib.rs
+++ b/crates/flux-refineck/src/lib.rs
@@ -147,20 +147,19 @@ pub fn check_fn(
     dbg::check_fn_span!(genv.tcx(), local_id).in_scope(|| {
         let ghost_stmts = compute_ghost_statements(genv, local_id)
             .with_span(span)
-            .emit(&genv)?;
+            .map_err(|err| err.emit_err(&genv, def_id))?;
 
         // PHASE 1: infer shape of `TypeEnv` at the entry of join points
         let shape_result = Checker::run_in_shape_mode(genv, local_id, &ghost_stmts, config)
             // Augment the possible CheckError with the functions span so we can report
             // helpful error messages for opaque struct field accesses
-            .map_err(|x| x.with_fn_span(genv.tcx().def_span(def_id)))
-            .emit(&genv)?;
+            .map_err(|err| err.emit_err(&genv, def_id))?;
         tracing::info!("check_fn::shape");
 
         // PHASE 2: generate refinement tree constraint
         let (refine_tree, kvars) =
             Checker::run_in_refine_mode(genv, local_id, &ghost_stmts, shape_result, config)
-                .emit(&genv)?;
+                .map_err(|err| err.emit_err(&genv, def_id))?;
         tracing::info!("check_fn::refine");
 
         // PHASE 3: invoke fixpoint on the constraint
--- a/crates/flux-refineck/src/lib.rs
+++ b/crates/flux-refineck/src/lib.rs
@@ -169,7 +169,7 @@ pub fn check_fn(
 
         // PHASE 4: subtyping check for trait-method implementations
         if let Some((refine_tree, kvars)) =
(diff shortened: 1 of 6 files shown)


vrindisbacher:
Closes (#835).

vrindisbacher:
@nilehmann, is there a straight forward way to get the span for the wrapping function? I only know how to get the struct's (and it's related impls) span, since it seems like the span for the field access in the function body has no parent.

nilehmann:
no, there's no easy way right now. When we construct a `CheckerError` we only pass the `Span` of the current `Statement` being checked and I don't think you can go from that to the span of the containing function. However, I think we don't emit `CheckerError`s in too many places so something we could try is not implement `Diagnostic` for `CheckerError` this will point to all the places where we emit it and hopefully we will have the `DefId` of the function at those places, for example, here https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/lib.rs/#L154. At those places, we can call a helper function that takes a `GlobalEnv`, a  `CheckerError`, and the `DefId` of the containing function, it then constructs the `Diag` and emits it.

crates/flux-refineck/src/lib.rs
```diff
@@ -104,8 +104,11 @@ pub fn check_fn(
             .emit(&genv)?;
 
         // PHASE 1: infer shape of `TypeEnv` at the entry of join points
-        let shape_result =
-            Checker::run_in_shape_mode(genv, local_id, &ghost_stmts, config).emit(&genv)?;
+        let shape_result = Checker::run_in_shape_mode(genv, local_id, &ghost_stmts, config)
+            // Augment the possible CheckError with the functions span so we can report
+            // helpful error messages for opaque struct field accesses
+            .map_err(|x| x.with_fn_span(genv.tcx().def_span(def_id)))
```
vrindisbacher, line 110: @nilehmann, my idea was to augment `CheckerError` with the functions span, so we can use it within the `Diag` implementation for `OpaqueStruct` errors.

nilehmann: I'm not a fan of that because the span can be `None` and you have to reason about it, but in practice we always have it. Instead, we ca do:
```rust
let shape_result = Checker::run_in_shape_mode(genv, local_id, &ghost_stmts, config)
    .map_err(|err| emit_checker_err(genv, err, def_id)))?;
    
fn emit_checker_err(genv: GlobalEnv, err: CheckerError, def_id: DefId) -> ErrorGuaranteed {
   ...
}
``` 
and prove statically that we have a `def_id`.

Alternatively, we can define `emit` as a method in `CheckerError`
```rust
fn emit(self, genv: GlobalEnv, def_id: DefId) -> ErrorGuaranteed;
 ```
 
 You can get a `DiagCtxtHandle` with `genv.sess().dcx().handle()`

nilehmann (approved):

crates/flux-refineck/src/lib.rs
nilehmann: I don't feel too strongly about this though. I'll let you decide.

vrindisbacher: hmm ya. I agree that it's a bit annoying, but wouldn't we have to basically reimplement the logic in CheckerError's implementation of `Diagnostic`?

nilehmann: Yes, I suggest we remove the `Diagnostic` implementation for `CheckerError` and instead have a method/function that takes the def id and emits the error.

vrindisbacher: Ah ok - I think I've got it. The only thing I'm wondering is what the level should be for the `Query` variant of `CheckerErrKind`. My guess is that it should be `Error`? Or is there a way to get that for the `CheckerError`?

vrindisbacher:
Hmm, not sure why the CI stuff is failing. It builds fine locally, and the line it's complaining about is different than what the dialog points to...

Issue #820, opened 2024-09-26
Check that sorts are used with the correct number of generics


nilehmann:
Flux happily accepts the following

```rust
use flux_rs::*;

#[refined_by(n: int)]
struct S {
    #[field(i32[n])]
    f: i32,
}

defs! {
    fn foo(s: S<int>) -> int {
        s.n
    }
}
```

This is wrong because the sort `S` (that we automatically generate from the `refined_by` annotation) doesn't have any generics.

We should check sorts are used with the correct number of generics when converting from `fhir` to `rty` [here](https://github.com/flux-rs/flux/blob/68323a45e4da8683da779b80b1e1777d23ad66d8/crates/flux-fhir-analysis/src/conv/mod.rs#L1969)

vrindisbacher:
@nilehmann, Is this still an issue? I can give it a try.

nilehmann:
@vrindisbacher yes, it still is

PR #863, opened 2024-10-26
Add check to ensure generics for sorts are correct

vrindisbacher:
Closes (#820)

### DIFF
--- a/crates/flux-fhir-analysis/locales/en-US.ftl
+++ b/crates/flux-fhir-analysis/locales/en-US.ftl
@@ -3,7 +3,6 @@
 fhir_analysis_sort_mismatch =
     mismatched sorts
     .label = expected `{$expected}`, found `{$found}`
-
 fhir_analysis_arg_count_mismatch =
     this {$thing} takes {$expected ->
         [one] {$expected} refinement argument
@@ -201,8 +200,32 @@ fhir_analysis_too_many_generic_args =
         *[other] arguments
     }
 
+fhir_analysis_generics_on_primitive_sort = 
+    primitive sort {$name} expects {$expected -> 
+        [0] no generics
+        [one] exactly one generic argument
+        *[other] exactly {$expected} generic arguments
+    } but found {$found}
+    .label = Incorrect generics on primitive sort
+
+fhir_analysis_too_many_generics_on_sort =
+    sorts associated with this {$def_descr} should have {$max ->  
+        [0] no generic arguments
+        [one] at most one generic argument
+        *[other] at most {$max} generic arguments
+    } but {$found} generic {$found ->
+        [one] argument
+        *[other] arguments
+    } supplied
+    .label = expected {$max ->
+        [0] no generic arguments
+        [one] at most one generic argument
+        *[other] at most {$max} generic arguments
+    } on sort
+
 fhir_analysis_refined_unrefinable_type =
     type cannot be refined
+
 
 # Check impl against trait errors
 
--- a/crates/flux-fhir-analysis/locales/en-US.ftl
+++ b/crates/flux-fhir-analysis/locales/en-US.ftl
@@ -206,7 +206,7 @@ fhir_analysis_generics_on_primitive_sort =
         [one] exactly one generic argument
         *[other] exactly {$expected} generic arguments
     } but found {$found}
-    .label = Incorrect generics on primitive sort
+    .label = incorrect generics on primitive sort
 
 fhir_analysis_too_many_generics_on_sort =
     sorts associated with this {$def_descr} should have {$max ->  
@@ -222,6 +222,18 @@ fhir_analysis_too_many_generics_on_sort =
         [one] at most one generic argument
         *[other] at most {$max} generic arguments
     } on sort
+
+fhir_analysis_generics_on_type_parameter = 
+    type parameter expects no generics but found {$found}
+    .label = found generics on sort type parameter
+
+fhir_analysis_generics_on_self_alias = 
+    type alias `Self` expect not generics but found {$found}
+    .label = found generics on type `Self`
+
+fhir_analysis_generics_on_opaque_sort = 
+    user defined opaque sorts have no generics but found {$found}
+    .label = found generics on user defined opaque sort
 
 fhir_analysis_refined_unrefinable_type =
     type cannot be refined
--- a/crates/flux-fhir-analysis/locales/en-US.ftl
+++ b/crates/flux-fhir-analysis/locales/en-US.ftl
@@ -228,7 +228,7 @@ fhir_analysis_generics_on_type_parameter =
     .label = found generics on sort type parameter
 
 fhir_analysis_generics_on_self_alias = 
-    type alias `Self` expect not generics but found {$found}
+    type alias Self expects no generics but found {$found}
     .label = found generics on type `Self`
 
 fhir_analysis_generics_on_opaque_sort = 
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -1961,9 +1961,24 @@ fn conv_sort_path(
     next_infer_sort: &mut impl FnMut() -> rty::Sort,
 ) -> QueryResult<rty::Sort> {
     let ctor = match path.res {
-        fhir::SortRes::PrimSort(fhir::PrimSort::Int) => return Ok(rty::Sort::Int),
-        fhir::SortRes::PrimSort(fhir::PrimSort::Bool) => return Ok(rty::Sort::Bool),
-        fhir::SortRes::PrimSort(fhir::PrimSort::Real) => return Ok(rty::Sort::Real),
+        fhir::SortRes::PrimSort(fhir::PrimSort::Int) => {
+            if !path.args.is_empty() {
+                return Err(emit_prim_sort_generics_error(genv, path, "int", 0))?;
+            }
+            return Ok(rty::Sort::Int);
+        }
+        fhir::SortRes::PrimSort(fhir::PrimSort::Bool) => {
+            if !path.args.is_empty() {
+                return Err(emit_prim_sort_generics_error(genv, path, "bool", 0))?;
+            }
+            return Ok(rty::Sort::Bool);
+        }
+        fhir::SortRes::PrimSort(fhir::PrimSort::Real) => {
+            if !path.args.is_empty() {
+                return Err(emit_prim_sort_generics_error(genv, path, "real", 0))?;
+            }
+            return Ok(rty::Sort::Real);
+        }
         fhir::SortRes::SortParam(n) => return Ok(rty::Sort::Var(rty::ParamSort::from(n))),
         fhir::SortRes::TyParam(def_id) => {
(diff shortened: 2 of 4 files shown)


vrindisbacher:
Checking works, but I'm not sure this error is particularly helpful...

Code: 

```rust
use flux_rs::*;

#[refined_by(n: int)]
struct S {
    #[field(i32[n])]
    f: i32,
}

defs! {
    fn foo(s: S<int>) -> int {
        s.n
    }
}
```

Output:

```rust
error[E0999]: cannot instantiate base generic with opaque type or a type parameter of kind type
 --> ../sort_gen.rs:8:1
  |
8 | struct S {
  | ^^^^^^^^
-Ztrack-diagnostics: created at crates/flux-fhir-analysis/src/wf/sortck.rs:394:18

```

It would be nice if we could point to the flux def itself and say something like: 

Argument `int` supplied but definition for `S` has no generics. Either remove the generic type supplied to sort `S` or add missing generic type to `S`

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1987,9 +1991,25 @@ fn conv_sort_path(
...
 }
 
+fn path_generics_correct(path: &fhir::SortPath, sort_generics: rty::Generics) -> bool {
+    // Check that the number of generics for a path are correct
+    //
+    // Note: We include base when collecting generics (i.e. T as base) for a def id
+    // because often times there are refinements over these types (see extern spec for iter
+    // as an example)
+    //
+    if path.args.len() != sort_generics.own_default_count_with_base() {
```
nilehmann, line 2005: @vrindisbacher A sort doesn't necessarily have the same number of generics as the type it corresponds to. For example, if you write
```rust
#[flux::refined_by(keys: Set<K>)
struct Map<K, V> { ... }
```
we will generate a sort with a single generic. The mapping between type parameters and sort parameters is tracked in [AdtSortDef](https://flux-rs.github.io/flux/doc/flux_middle/rty/struct.AdtSortDef.html). Here you are basically recomputing that mapping but we could use the information in `AdtSortDef` directly.

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1979,6 +1979,10 @@ fn conv_sort_path(
         fhir::SortRes::User { name } => rty::SortCtor::User { name },
         fhir::SortRes::Adt(def_id) => {
             let sort_def = genv.adt_sort_def_of(def_id)?;
+            let generics = genv.generics_of(def_id)?;
+            if !path_generics_correct(path, generics) {
+                return Err(QueryErr::InvalidGenericArg { def_id });
```
nilehmann, line 1984: This function returns a `QueryResult`, but we should emit a specialized error. The trick is to convert a `Result<T, ErrorGuaranteed>` to a `Result<T, QueryErr>` (a.k.a `QueryResult<T>`). See for example how we report [TooFewGenericArgs](https://github.com/flux-rs/flux/blob/e150d6899cd96b9770b5e3ea3a54c0dd78a89be4/crates/flux-fhir-analysis/src/conv/mod.rs#L1458)

nilehmann (changes requested):

crates/flux-fhir-analysis/src/conv/mod.rs
vrindisbacher: @nilehmann, so if I understand correctly, we want to make sure that we are not using more generics than are defined for the type it corresponds to?

nilehmann: At this point, we can forget there's a corresponding type. We only want to look at the sort declaration.

I also just noticed that we need to do this for all sorts not just adt sorts. The syntax will for example let you write `int<bool>` which is nonsense.

* `Set` must have 1 argument
* `Map` must have 2
* For adt sorts we get the information from the `AdtSortDef`. The number of parameters is the length of https://flux-rs.github.io/flux/doc/flux_middle/rty/struct.AdtSortDefData.html#structfield.params, we must expose that value in a method.
* all other sorts must have zero arguments.

nilehmann: And the error should say something like

```
error[E0999]: sort takes 0 arguments but 1 argument was supplied
 --> ../sort_gen.rs:8:1
   |
10 |     fn foo(s: S<int>) -> int {
   |               ^
   |               |
   |               expected 0 arguments
 ```
 
For the span, I think we can take the span of the ident in the last segment of the `SortPath`

Comment (vrindisbacher) [reply crates/flux-fhir-analysis/src/conv/mod.rs]: @nilehmann, went ahead and updated this.

**Comment by vrindisbacher** [comment]
Example input:

```rust
#![allow(unused)]

use flux_rs::*;

#[refined_by(n: int)]
struct S {
    #[field(i32[n])]
    f: i32,
}

defs! {
    fn foo(s: S<int>) -> int {
        s.n
    }
}
```

Output:
```rust
error[E0999]: sorts associated with this struct should have no generic arguments but 1 generic argument supplied
  --> ../sorts.rs:12:15
   |
12 |     fn foo(s: S<int>) -> int {
   |               ^ expected no generic arguments on sort
-Ztrack-diagnostics: created at crates/flux-fhir-analysis/src/conv/mod.rs:2017:40

```

Another example:

```rust
#![allow(unused)]

use flux_rs::*;

#[refined_by(n: int)]
struct S {
    #[field(i32[n])]
    f: i32,
}

defs! {
    fn foo(x: int<bool>) -> int {
        1
    }
}
```

And output:

```rust
error[E0999]: primitive sort int expects no generics but found 1
  --> ../sorts.rs:16:15
   |
16 |     fn foo(x: int<bool>) -> int {
   |               ^^^ Incorrect generics on primitive sort
-Ztrack-diagnostics: created at crates/flux-fhir-analysis/src/conv/mod.rs:2043:17

```

crates/flux-fhir-analysis/locales/en-US.ftl
```diff
@@ -201,9 +200,33 @@ fhir_analysis_too_many_generic_args =
         *[other] arguments
     }
 
+fhir_analysis_generics_on_primitive_sort = 
+    primitive sort {$name} expects {$expected -> 
+        [0] no generics
+        [one] exactly one generic argument
+        *[other] exactly {$expected} generic arguments
+    } but found {$found}
+    .label = Incorrect generics on primitive sort
```
nilehmann, line 209: messages must start with lowercase per rustc diagnostic style guide https://rustc-dev-guide.rust-lang.org/diagnostics.html?highlight=diagnostics#diagnostic-output-style-guide

```suggestion
    .label = incorrect generics on primitive sort
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1961,9 +1961,24 @@ fn conv_sort_path(
...
+        }
+        fhir::SortRes::PrimSort(fhir::PrimSort::Real) => {
+            if !path.args.is_empty() {
+                return Err(emit_prim_sort_generics_error(genv, path, "real", 0))?;
+            }
+            return Ok(rty::Sort::Real);
+        }
         fhir::SortRes::SortParam(n) => return Ok(rty::Sort::Var(rty::ParamSort::from(n))),
         fhir::SortRes::TyParam(def_id) => {
             return Ok(rty::Sort::Param(def_id_to_param_ty(genv, def_id)))
```
nilehmann, line 1984: We need to check here sort parameters do not have arguments. An example where we should report an error:

```rust
#[flux::refined_by(f: T<int>)]
struct S<T> { 
    #[flux::field(T[f])]
    f: T 
}
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1974,11 +1989,33 @@ fn conv_sort_path(
                 .sort_of_self_ty_alias(alias_to)?
```
nilehmann, line 1989: Same here, the following should give an error

```rust
#[flux::assoc(fn foo(x: Self<int>) -> int)]
trait Trait {}

```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1974,11 +1989,33 @@ fn conv_sort_path(
...
+            rty::SortCtor::Set
+        }
+        fhir::SortRes::PrimSort(fhir::PrimSort::Map) => {
+            if path.args.len() != 2 {
+                // Has to have two arguments
+                return Err(emit_prim_sort_generics_error(genv, path, "Map", 2))?;
+            }
+            rty::SortCtor::Map
+        }
         fhir::SortRes::User { name } => rty::SortCtor::User { name },
```
nilehmann, line 2006: User defined opaque sort are a very obscure feature, but we should also give an error in a case like this:

```rust
extern crate flux_rs;

flux_rs::defs! {
    opaque sort MyOpaqueSort;
}

#[flux_rs::opaque]
#[flux_rs::refined_by(f: MyOpaqueSort<int>)]
struct S {}
```

User opaque sorts always have zero parameters

nilehmann (approved):
Cool, this looks good. I like having separate messages for primitive sorts and adt sorts.

There are a couple of extra places where we should report an error. I suppose we can have a different message for each of them but I'm fine with lumping them together in a generic "this sort doesn't have the right number of arguments" message. 

Please also add some tests to `tests/tests/neg/error_messages/conv/` 

Feel free to merge after fixing the comments or ping me if something is not clear.

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: This is akin to

```rust
fn foo<T>(x: T<i32>) {}
```

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: rust equivalent

```rust
trait Trait {
    fn foo(x: Self<i32>);
}
```

crates/flux-fhir-analysis/src/conv/mod.rs
vrindisbacher: @nilehmann, I think this would be caught by the type parameter expecting no generics?

tests/tests/neg/error_messages/conv/mismatched_generics.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[flux::assoc(fn foo(x: Self<int>) -> int)] //~ Error type alias Self expects no generics but found 1
+trait MyTrait {}
+
+flux_rs::defs! {
+    opaque sort MyOpaqueSort;
+}
+
+#[flux_rs::opaque]
+#[flux_rs::refined_by(f: MyOpaqueSort<int>)] //~ Error type parameter expects no generics but found 1
```
vrindisbacher, line 53: @nilehmann, interestingly, this is caught by the `TyParam` case, and not the user defined case?

nilehmann:
@vrindisbacher I think you put the wrong `#[diagnostic]` attributes. Both the error structs for self and opaque have

```
#[diag(fhir_analysis_generics_on_type_parameter, code = E0999)]
```

vrindisbacher:
doh! thanks... can't believe i didn't catch that...

PR #864, opened 2024-10-26
Check impl-methods are subtypes of trait-methods

### LINKED ISSUE #592: Check that `impl` satisfies `trait` specs
Currently, `flux` happily verifies the below which seems dubious:

```rust
pub trait Silly {
    #[flux::sig(fn(&Self) -> i32{v:100 < v})]
    fn bloop(&self) -> i32;
}

impl Silly for i32 {
    fn bloop(&self) -> i32 {
        0
    }
}
```

_Originally posted by @ranjitjhala in https://github.com/flux-rs/flux/pull/589#discussion_r1442110777_

ranjitjhala:
Fixes #592


ranjitjhala:
Matching PR https://github.com/PLSysSec/tock/pull/4/files

crates/flux-refineck/src/checker.rs
```diff
@@ -263,6 +307,74 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             ck.check_basic_block(infcx, env, bb)?;
         }
 
+        ck.check_trait_impl_subtyping(infcx, def_id, span)?;
+
+        Ok(())
+    }
+
+    fn has_trusted_impl(genv: &GlobalEnv, def_id: DefId) -> bool {
```
nilehmann, line 315: nit: I'd make this a method in `GlobalEnv`

crates/flux-refineck/src/checker.rs
```diff
@@ -263,6 +307,74 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            let impl_args = GenericArg::identity_for_item(genv, def_id).with_span(span)?;
+            let trait_to_impl_args = &trait_ref.args;
+            let trait_args = impl_args.rebase_onto(&tcx, impl_id, trait_to_impl_args);
+            let trait_refine_args =
+                RefineArgs::identity_for_item(&genv, trait_method_id, |i, param| {
+                    if param.mode == rty::InferMode::KVar {
+                        // See `tests/pos/surface/trait13.rs` for an example
+                        // we need to "define_var" the KVar in the trait's super-type
+                        // TODO: maybe the `trait_refine_args` business can be done inside the `check_fn_subtyping`?
+                        infcx.define_vars(&param.sort)
```
nilehmann, line 359: I find this extremely confusing and I think it's wrong in subtle ways. It'd probably help to refactor this outside of `Checker` and use a separate `InferCtxt` (there's no reason to make this part of the constraint generated for the body). Moving this out of `Checker` could also help with improving the error message

nilehmann (approved):
I'm fine with merging this, but we should refactor this out of `Checker` sooner rather than later as it's only going to get more complicated.

nilehmann (changes requested):
no wait, I think I know what's wrong with this. We are assuming the trait method and the impl method have the same number of refinement generics. This is true for type generics and that's why we can instantiate the impl method with the (rebased) identity instantiation, but requiring the same number of refinement generics won't fly. I believe we should instantiate refinement generics with inference variables.

nilehmann (commented):
No scratch that. I'm super confused

nilehmann (approved):
Well, I'm still a bit confused, but I think a lot of my confusion would be solved by using a separate `InferCtxt`. I feel that initializing the `InferCtxt` in the context of the trait would automatically do most of this right.

ranjitjhala:
By "outside" `Checker` do you mean it should be at the "level" of `fn check_fn` in `flux-refineck/src/lib.rs`?

ranjitjhala:
or at the level of `fn run_in_refine_mode`?

nilehmann:
I think this should be part of [`flux_fhir_analysis::compare_impl_item::check_impl_against_trait`](https://flux-rs.github.io/flux/doc/flux_fhir_analysis/compare_impl_item/fn.check_impl_against_trait.html), but that would require reshuffling things across crates. Putting it in [`flux_refineck::check_fn`](https://flux-rs.github.io/flux/doc/flux_refineck/fn.check_fn.html) would be a good start though.

ranjitjhala:
AFAICT `InferCtxt` can only be initialized with a `Body` and so you can't initialize it in the context of a trait-method (which may not have a `Body`?)

ranjitjhala:
@nilehmann I have done the refactoring, so the impl-trait check lives in

* `pub fn trait_impl_subtyping` (currently in checker.rs) but independent of the `Checker` itself and
* invoked as `PHASE 4` in `check_fn`

See esp. the new `args` passed into `RefineTree::new` and the subsequent `.instantiate(...)`

crates/flux-middle/src/rty/mod.rs
```diff
@@ -1536,6 +1536,37 @@ pub struct AliasTy {
...
+        let mut args = vec![];
+        for i in 0..reft_generics.count() {
+            let param = reft_generics.param_at(i, *genv)?;
+            let expr =
+                Expr::var(Var::EarlyParam(EarlyReftParam { index: i as u32, name: param.name }));
+            args.push(expr);
+        }
+        Ok(List::from_vec(args))
+    }
+}
```
nilehmann, line 1568: I believe we only use `identity_for_item_new` 

```suggestion
    fn identity_for_item(genv: &GlobalEnv, def_id: DefId) -> QueryResult<RefineArgs> {
        let reft_generics = genv.refinement_generics_of(def_id)?;
        let mut args = vec![];
        for i in 0..reft_generics.count() {
            let param = reft_generics.param_at(i, *genv)?;
            let expr =
                Expr::var(Var::EarlyParam(EarlyReftParam { index: i as u32, name: param.name }));
            args.push(expr);
        }
        Ok(List::from_vec(args))
    }
}
```

If we need them both, we should implement one in terms of the other. Or we can have a `for_item` and an `identity_for_item`, both implemented on top of `fill_item` as we do for `GenericArgs`.

crates/flux-refineck/src/checker.rs
```diff
@@ -263,6 +434,8 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             ck.check_basic_block(infcx, env, bb)?;
         }
 
+        // MICKEYMOUSE ck.check_trait_impl_subtyping(infcx, def_id, span)?;
+
```
nilehmann, line 438:
```suggestion
```

crates/flux-infer/src/infer.rs
```diff
@@ -69,25 +69,38 @@ pub struct InferCtxtRoot<'genv, 'tcx> {
...
+    //     root_id: DefId,
+    //     kvar_gen: KVarGen,
+    //     args: &GenericArgs,
+    // ) -> QueryResult<Self> {
+    //     Ok(Self {
+    //         genv,
+    //         inner: RefCell::new(InferCtxtInner::new(kvar_gen)),
+    //         refine_tree: RefineTree::new(genv, root_id, Some(args))?,
+    //     })
+    // }
```
nilehmann, line 93:
```suggestion
```

nilehmann (approved):

ranjitjhala:
ok, I'll merge if everything -- IF I get the damn tock thing to pass locally.

ranjitjhala:
merging!

PR #868, opened 2024-10-29
Update specs.md: (Possible) minor typo

t-nil:
Is it possible there is a typo there? (see code)
The sentence doesn't make sense to me otherwise, but I am not really familiar with the concepts.


nilehmann (approved):
Yes, this is indeed a typo 

Thanks!

PR #873, opened 2024-11-05
Basic support for `for i in a..b { blah }` 

ranjitjhala:
Copied over and modified mostly from @vrindisbacher 's https://github.com/vrindisbacher/flux-sandbox/blob/main/src/iterator.rs


nilehmann:
@ranjitjhala you didn't push `rrange.rs`

ranjitjhala:
Since its just tests, I'm merging -- easy enough to revert if needed.

PR #874, opened 2024-11-05
A bit of function subtyping for `ReifyFnPointer`

### LINKED ISSUE #859: Unsoundness with `ReifyFnPtr` cast
This gives me the heebie-jeebies. The function pointer doesn't have refinements but the closure or function definition you are coercing from may, so this should involve function subtyping. 

Also, I think for closures this means we never check the closure's body.

_Originally posted by @nilehmann in https://github.com/flux-rs/flux/pull/854#discussion_r1806877393_

ranjitjhala:
Fixes #859


nilehmann (changes requested):
I don't like this, `FnDef` is not technically a subtype of `FnPtr` and we are not gaining anything by making it (since fn ptrs are unrefined). I'd prefer if we do the subtyping when we check the cast (`ReifyFnPointer`). That would allow us to give a better error message. We can revisit if we ever need refined function pointers.

ranjitjhala:
> I'd prefer if we do the subtyping when we check the cast (ReifyFnPointer).

that would indeed be preferable but I don’t see how that’s possible since at that point we don’t know what the “super-type” is supposed to be?

ranjitjhala:
At best one can imagine 
1. generating a KVAR template at the cast site and creating a subtyping constraint there and 
2. then generating a subtyping between the template and the actual super type at the actual usage site 

but that seems rather overkill to me and really what the FnDef is avoiding.

ranjitjhala:
So I’d say what we’re “gaining” is avoiding this redundant indirection through Kvars, unless you see some other way to do it?

nilehmann:
We do know the type because it's unrefined. When checking a cast we have the rust target type (which in this case should be a `FnPtr`). We need the default refinement for that type and then check against it.

ranjitjhala:
Ah yes of course, we do _now_ but (hopefully soon) we'll want to refine the `FnPtr` and then we won't have that info, so why paint into a corner?

nilehmann:
I'm not convinced this is the best way to support refined function pointers. If we want the function pointer to be higher-rank we could instead put some refinement parameters (the ones not mentioned in clauses) as part of `Binder` instead of `EarlyBinder`. Also, storing a function in a variable and then applying it multiple times is rare so I'm not even convinced we need it to be higher-rank, at which point this is just an optimization (to avoid extra kvars).

ranjitjhala:
Ok @nilehmann have moved the subtyping to the cast site.

crates/flux-refineck/src/checker.rs
```diff
@@ -1637,7 +1658,7 @@ fn all_predicates_of(
 /// where the formal argument is of the form `&mut B[@n]`, e.g., the type of the first argument
 /// to `RVec::get_mut` is `&mut RVec<T>[@n]`. We should remove this after we implement opening of
 /// mutable references.
-fn infer_under_mut_ref_hack(
+pub fn infer_under_mut_ref_hack(
```
nilehmann, line 1661:
```suggestion
fn infer_under_mut_ref_hack(
```

crates/flux-refineck/src/checker.rs
```diff
@@ -235,65 +235,63 @@ fn find_trait_item(
 ///      f(x1,...,xn)
 ///  }
 /// TODO: copy rules from SLACK.
-fn check_fn_subtyping<'genv, 'tcx>(
-    genv: GlobalEnv<'genv, 'tcx>,
-    infcx: &mut InferCtxt<'_, 'genv, 'tcx>,
+pub fn check_fn_subtyping(
```
nilehmann, line 238:
```suggestion
fn check_fn_subtyping(
```

nilehmann (approved):
LGTM, just a couple of nits

Issue #870, opened 2024-10-30
Add support for functional updates


vrindisbacher:
### Motivation

Here is a minimal example to motivate this: 

```rust
#[flux_rs::refined_by(f1: int, f2: int)]
pub struct X {
    #[field(usize[f1])]
    field1: usize,
    #[field(usize[f2])]
    field2: usize,
}

impl X {

    #[flux_rs::sig( fn (self: &strg X[@old_x], usize[@val]) ensures self: X { new_x: new_x.f1 == val })]
    pub fn modify_field1(&mut self, val: usize) {
        self.field1 = val;
    }

    #[flux_rs::sig( fn (self: &strg X[@old_x], usize[@val]) ensures self: X { new_x: new_x.f2 == val })]
    pub fn modify_field2(&mut self, val: usize) {
        self.field2 = val;
    }
}

#[flux_rs::sig(
    fn (self: &strg X[@old_x])
        ensures self: X { new_x: new_x.f1 == 1 && new_x.f2 == 2 }
)]
fn modify_f1_and_f2(x: &mut X) {
    x.modify_field1(1);
    x.modify_field2(2);
}
```

The function `modify_f1_and_f2` cannot be verified: 

```rust
error[E0999]: refinement type error
  --> src/main.rs:27:32
   |
27 |   fn modify_f1_and_f2(x: &mut X) {
   |  ________________________________^
28 | |     x.modify_field1(1);
29 | |     x.modify_field2(2);
30 | | }
   | |_^ a postcondition cannot be proved
   |
note: this is the condition that cannot be proved
  --> src/main.rs:25:34
   |
25 |         ensures self: X { new_x: new_x.f1 == 1 && new_x.f2 == 2 }
   |                                  ^^^^^^^^^^^^^

warning: `flux-issue-sandbox` (bin "flux-issue-sandbox") generated 1 warning
error: could not compile `flux-issue-sandbox` (bin "flux-issue-sandbox") due to 1 previous error; 1 warning emitted
```

Specifically, we lose all information about `f1` when calling `modify_field2`. In order to make this pass, we need to add some information to the post condition of `modify_field1` and `modify_field2`. Mainly, `old_x.f1 == new_x.f1` and `old_x.f2 == new_x.f2` for `modify_field2` and `modify_field1` respectively. 

In small cases like this, this isn't too bad, but in cases where there are many fields it could get quite tedious. 

### Proposal

[...]

PR #879, opened 2024-11-07
Add functional updates

vrindisbacher:
closes (#870)


crates/flux-syntax/src/grammar.lalrpop
```diff
@@ -543,7 +543,35 @@ Level10: surface::Expr = {
             span: cx.map_span(lo, hi),
         }
     },
-    "(" <Level1> ")"
+    Constructor,
+}
+
+Constructor: surface::Expr = {
```
vrindisbacher, line 549: @nilehmann , I'm not sure how to include this into `Expr` without blowing up the Parser. Seems like the issue is coming from the rule for refinements like `new_x { new_x == .... }`. Maybe I should do as you say and collapse refinements into exprs?

nilehmann: oh, I was fearing this. Did some testing. The following works

```
Level10: surface::Expr = {
    "{" <f:Ident> ":" <arg:Level1> "}" => {
        todo!()
    },
    <lo:@L> "if" <p:Level1> "{" <e1:Level1> "}" <e2:ElseIf> <hi:@R> => {
        ...
    },
    ...
```    

but the following is ambiguous

```
Level10: surface::Expr = {
   <name:Ident>  "{" <f:Ident> ":" <arg:Level1> "}" => {
        todo!()
    },
    <lo:@L> "if" <p:Level1> "{" <e1:Level1> "}" <e2:ElseIf> <hi:@R> => {
       ...
    },
    ...
``` 

The issue is `if` expressions when a constructor is used in the condition, e.g.,

```rust
if S {f: 0} { ... } else { ... }
```    
When looking at the `{` after `S` it's not possible (without lookahead) to know whether `S` is a variable and `{` starts the `then` branch or whether `S ` is a constructor. Writing the same in rust would tell you to disambiguate with parentheses.

nilehmann: The solution in `syn` is to disallow structs in some contexts. This is not very pretty, but I don't think there is a way around it. We could something similar by parameterizing the parser with a boolean. Something like:

```
pub Expr = Level1<"true">;

Level1<AllowStruct> = NonAssoc<BinOp1, Level2<AllowStruct>>;  // <=>
Level2<AllowStruct> = LeftAssoc<BinOp2, Level3<AllowStruct>>; // =>
Level3<AllowStruct> = LeftAssoc<BinOp3, Level4<AllowStruct>>; // ||
Level4<AllowStruct> = LeftAssoc<BinOp4, Level5<AllowStruct>>; // &&
Level5<AllowStruct> = NonAssoc<BinOp5, Level6<AllowStruct>>;  // ==, !=, >=, <=
Level6<AllowStruct> = LeftAssoc<BinOp6, Level7<AllowStruct>>; // >>, <<
Level7<AllowStruct> = LeftAssoc<BinOp7, Level8<AllowStruct>>; // +, -
Level8<AllowStruct> = LeftAssoc<BinOp8, Level9<AllowStruct>>; // *, %, /
Level9<AllowStruct> = {
    <lo:@L> <op:UnOp> <e:Level10<AllowStruct>> <hi:@R> => {
        surface::Expr {
            kind: surface::ExprKind::UnaryOp(op, Box::new(e)),
            node_id: cx.next_node_id(),
            span: cx.map_span(lo, hi),
        }
    },
    <Level10<AllowStruct>>
}

Level10<AllowStruct>: surface::Expr = {
    <name:Ident> "{" <f:Ident> ":" <arg:Expr> "}" if AllowStruct == "true" => {
        todo!()
    },
    <lo:@L> "if" <p:Level1<"false">> "{" <e1:Expr> "}" <e2:ElseIf> <hi:@R> => {
        surface::Expr {
            kind: surface::ExprKind::IfThenElse(Box::new([p, e1, e2])),
            node_id: cx.next_node_id(),
            span: cx.map_span(lo, hi),
        }
    },
    ...
```

nilehmann: I'd still start by refactoring `fhir` to collapse `RefineArg` and `Expr` and submit that in a separate PR, otherwise, it'll get too big.

vrindisbacher: Ok cool - good to know

ranjitjhala:
Is this still WIP? Looks like only the formatter is grumbling?

vrindisbacher:
@nilehmann, do we want to wait until I have spreads added or do you want to take a look at this now?

nilehmann:
@vrindisbacher I'll take a look now

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -84,6 +84,39 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                }
+            }
+            Ok(())
+        } else {
+            Err(self.emit_err(errors::ArgCountMismatch::new(
+                Some(path.span),
+                String::from("type"),
+                1,
+                field_exprs.len(),
+            )))
```
nilehmann, line 116: We'll hit this with an expression like `if (S { x: 0}) { 0 } else { 1 }` or `S { x: 0} > 0`. We should report an error saying something like `expected bool, found sort constructor`. This is similar to the `UnexpectedFun` error.

tests/tests/neg/constructors/test00.rs
```diff
@@ -0,0 +1,14 @@
+#[flux::refined_by(x: int, y: int)]
+pub struct X {
+    #[flux::field(u32[x])]
+    x: u32,
+    #[flux::field(u32[y])]
+    y: u32,
+}
+
+#[flux::sig(fn (x: X[@old_x]) -> X[Y { y: 2, x: 1 }])] //~ ERROR cannot find value `Y` in this scope
```
nilehmann, line 9: We typically put tests related to error messages that are not refinement-type errors in `neg/error_messages` inside the corresponding directory of the stage (wf in this case)

crates/flux-desugar/src/resolver/refinement_resolver.rs
```diff
@@ -296,6 +296,17 @@ impl<V: ScopedVisitor> surface::visit::Visitor for ScopedVisitorWrapper<V> {
                 self.on_func(*func, expr.node_id);
             }
             surface::ExprKind::Dot(path, _) => self.on_path(path),
+            surface::ExprKind::Constructor(path, field_exprs, spread) => {
+                self.on_path(path);
+                // also need to deal with exprs in the constructor
+                // This is the case becasuse field exprs are different fields
+                for expr in field_exprs {
+                    self.visit_expr(&expr.expr);
+                }
```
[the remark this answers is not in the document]
vrindisbacher: Yep this was because I made the mistake of calling `walk_expr` in `visit_expr`, when I should have been calling `vis.visit_expr` on field expressions.

tests/tests/neg/constructors/test00.rs
vrindisbacher: Ok sounds good! I just put the tests in the right places (I think)

nilehmann (changes requested):

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -84,6 +84,34 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+        expr_span: Span,
+        field_exprs: &[fhir::FieldExpr],
+        _spread: &Option<fhir::Spread>,
+        expected: &rty::Sort,
+    ) -> Result {
+        if let rty::Sort::App(rty::SortCtor::Adt(sort_def), sort_args) = expected {
+            // these are ordered
+            let field_names = sort_def.field_names();
+            let sorts = sort_def.field_sorts(sort_args);
+            for expr in field_exprs {
```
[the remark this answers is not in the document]
vrindisbacher: So I'm wondering, do we actually need to check that there are no missing fields? 

If someone uses this syntax without specifying all fields, it is simply syntactic sugar for a normal update. That seems like it would be ok?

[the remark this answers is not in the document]
nilehmann: That's only true if you provide a spread, with no spread you have to specify all the fields, e.g., the following is wrong and should give an error


```rust
#[refined_by(x: int, y: int)]
struct S { ... }

#[sig(fn() -> S[S { x: 0 }]
fn foo() -> S { 
    ...
}
``` 


 We've been calling this functional updates, but there's no actual *update* going on here, you are always constructing a new value, by possibly copying some fields from another variable via the spread.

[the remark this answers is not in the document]
vrindisbacher: @nilehmann I went ahead and added these - I'm not sure if there's a better way to do the diagnostic for missing fields. Lmk if you have any thoughts / objections.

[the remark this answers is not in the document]
vrindisbacher: Ok - I'm going to work on spreads first and then I'll circle back to this.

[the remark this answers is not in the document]
nilehmann: I think we can simplify the code a bit if we use hashmaps and hash sets instead of vecs. 

I would expose a method in `AdtSortDef` that gives you back an `FxHashMap<Symbol, Sort>` and then

```rust
let mut sort_by_field_name = sort_def.sort_by_field_name(sort_args);
let mut used_fields = FxHashSet::default();
for expr in field_exprs {
    let Some(sort) = sort_by_field_name.get(&expr.ident.name) else {
        // report field not found
    }
    if let Some(old) = used_fields.replace(expr.ident) {
        return Err(...DuplicatefieldUsed...)
    }
    self.check_expr(&expr.expr.sort)?;
}
if let Some(_spred) = spread {
    todo!()
} else if used_fields.len() != sort_by_field_name.len() {
    // report missing fields, we can compute the difference between `sort_by_field_name` and `used_fields` here.
    // I like this approach because we don't clutter the happy path with extra logic needed to remember the 
    // remaining fields.
} else {
   ...
}
```

crates/flux-syntax/src/surface.rs
```diff
@@ -452,6 +452,27 @@ pub enum GenericArgKind {
...
+
+#[derive(Debug)]
+pub struct Spread {
+    pub expr: Expr,
+    pub span: Span,
+    pub node_id: NodeId,
+}
+
+#[derive(Debug)]
+pub enum ConstructorArgs {
```
nilehmann, line 471: I'd use singular

```suggestion
pub enum ConstructorArg {
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -112,6 +112,19 @@ impl AdtSortDef {
...
+        &self.0.field_names
+    }
+
+    pub fn sort_by_field_name(&self, args: &[Sort]) -> FxIndexMap<Symbol, Sort> {
+        let mut map = FxIndexMap::default();
+        let folded = self.0.sorts.fold_with(&mut SortSubst::new(args));
+        std::iter::zip(folded.iter(), self.0.field_names.iter()).for_each(|(sort, field_name)| {
+            map.insert(*field_name, sort.clone());
+        });
+        map
```
nilehmann, line 125: The following reads better in my opinion (or some variation of it)

```suggestion
        std::iter::zip(&self.0.field_names, &self.0.sorts.fold_with(&mut SortSubst::new(args)))
            .map(|(name, sort)| (*name, sort.clone()))
            .collect()
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -1329,11 +1329,74 @@ trait DesugarCtxt<'genv, 'tcx: 'genv> {
...
+                        fhir_id: self.next_fhir_id(),
+                    };
+                    Some(self.genv().alloc(spread))
+                } else {
+                    // Multiple spreads found - emit and error
+                    return Err(self.emit_err(errors::MultipleSpreadsInConstructor::new(
+                        spreads[1].span,
+                        spreads[0].span,
+                    )));
+                };
```
nilehmann, line 1380: maybe

```suggestion
                let spread = match &spreads[..] {
                    [] => None.
                    [s] => {
                        let spread = fhir::Spread {
                            expr: self.desugar_expr(&s.expr)?,
                            span: s.span,
                            fhir_id: self.next_fhir_id(),
                        };
                        Some(self.genv().alloc(spread))
                    }
                    [s1, s2, ..] => {
                        // Multiple spreads found - emit and error
                         return Err(self.emit_err(errors::MultipleSpreadsInConstructor::new(
                              s1.span,
                              s2.span,
                          )));
                    }
                };
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1644,10 +1646,60 @@ impl<'genv, 'tcx, P: ConvPhase> ConvCtxt<'genv, 'tcx, P> {
...
 
+    fn conv_constructor_exprs(
+        &mut self,
+        struct_def_id: DefId,
+        env: &mut Env,
+        exprs: &[fhir::FieldExpr],
+        spread: &Option<&fhir::Spread>,
+    ) -> QueryResult<List<rty::Expr>> {
+        let struct_adt = self.genv.adt_sort_def_of(struct_def_id)?;
+        let sort_by_field = FxIndexSet::from_iter(struct_adt.field_names());
```
nilehmann, line 1669: We could make `struct_adt.field_names()` return an `&FxIndexSet<Symbol>`. and even more, we can make the field `field_names` in `AdtSortDefData` an `FxIndexSet<Symbol>` instead of a `Vec<Symbol>`.

Also,

```suggestion
        let field_names = FxIndexSet::from_iter(struct_adt.field_names());
```

nilehmann: Or maybe we can invert this, map `exprs` by name, and iterate over the field names. Something like

```rust
let spread = spread
    .map(|spread| self.conv_expr(env, &spread.expr))
    .transpose()?;
let field_exprs_by_name: FxIndexMap<Symbol, &fhir::FieldExpr> =
    exprs.iter().map(|e| (e.ident.name, e)).collect();
for (idx, field_name) in struct_adt.field_names().enumerate() {
    if let Some(field_expr) = field_exprs_by_name.get(field_name) {
        assns.push(self.conv_expr(env, &field_expr.expr)?);
    } else if Some(spread) = spread {
        let proj = rty::FieldProj::Adt { def_id: struct_def_id, field: idx as u32 };
        assns.push(rty::Expr::field_proj(spread, proj));
    }
}
```

nilehmann (approved):
Ok cool, this looks great. 

I left some stylistic suggestions, but it looks good to go. There are also some clippy warnings.

crates/flux-desugar/src/desugar.rs
vrindisbacher: Oh cool I didn't know you could do that actually

crates/flux-fhir-analysis/src/conv/mod.rs
vrindisbacher: Nice I like that one.

ranjitjhala:
Yay! This is awesome thanks @vivien!


- Ranjit.

PR #882, opened 2024-11-11
Local Pointers

ranjitjhala:
This is a whack at the hack of locally converting an `&mut S` into an `&strg S` at a call-site (and then packing it back up into an `&mut S` after the call, with a subtyping obligation. See 

* [tests/neg/surface/local_ptr00.rs](tests/pos/surface/local_ptr00.rs)
* [tests/pos/surface/local_ptr00.rs](tests/neg/surface/local_ptr00.rs)


ranjitjhala:
Ok @nilehmann -- have reworked this per your suggestion, take a look!

crates/flux-refineck/src/checker.rs
```diff
@@ -703,6 +706,39 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
 
+    fn unfold_local_ptrs(
+        infcx: &mut InferCtxt<'_, 'genv, 'tcx>,
+        env: &mut TypeEnv,
+        span: Span,
+        fn_sig: &EarlyBinder<PolyFnSig>,
+        actuals: &[Ty],
+    ) -> Result<Vec<Ty>> {
+        // We *only* need to know whether each input is a &strg or not
+        let mut at = infcx.at(span);
```
nilehmann, line 717: I believe unfolding doesn't generate any constraints so we don't need the `infcx.at` (that's a nice static guarantee if we don't create an `InferCtxtAt` then we know we didn't generate any constraints).

```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -604,7 +606,7 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                             &resolved_args.lowered,
                         )
                         .with_src_info(terminator.source_info)?;
-
+                        let actuals = self.check_operands(infcx, env, terminator_span, args)?;
```
nilehmann, line 609: Is there a reason to move this here?

crates/flux-refineck/src/type_env.rs
```diff
@@ -220,6 +221,14 @@ impl<'a> TypeEnv<'a> {
         Ok(Ty::mk_ref(re, t2, Mutability::Mut))
     }
 
+    pub(crate) fn fold_local_ptrs(&mut self, infcx: &mut InferCtxtAt) -> InferResult<()> {
```
nilehmann, line 224: Maybe add a doc comment with the inference rule we wrote in the hackmd

crates/flux-refineck/src/type_env.rs
```diff
@@ -309,6 +317,18 @@ impl<'a> TypeEnv<'a> {
         Ok(())
     }
 
+    pub(crate) fn unfold_local_ptr(
```
nilehmann, line 320: Same here, we could put the inference rule in a doc comment

crates/flux-refineck/src/checker.rs
```diff
@@ -703,6 +706,39 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                Ty::ptr(PtrKind::Mut(*re), path1)
+            } else {
+                actual.clone()
+            };
+            tys.push(actual);
+        }
+        Ok(tys)
+    }
+
+    fn fold_local_ptrs(infcx: &mut InferCtxt, env: &mut TypeEnv, span: Span) -> Result {
```
nilehmann, line 737: nit: I prefer to make this a free function instead of an associated function.

nilehmann (approved):
Looks good! just a couple of small comments

crates/flux-refineck/src/checker.rs
ranjitjhala: vestige of the earlier approach with loc/place etc. (reverted!)

PR #891, opened 2024-11-19
Support `&strg` , `requires` and `ensures` in function subtyping

### LINKED ISSUE #890: Confusing error when checking trait impl against trait
When I try to check the following trait implementation ([here](https://github.com/PLSysSec/tock/blob/793c017ef05edb13c51215c7bef0622dd1735cb3/kernel/src/collections/ring_buffer.rs#L178-L184)):
```Rust
    #[flux_rs::sig(
        fn(self: &strg Self, _) -> Option<_> ensures self: Self
    )]
    fn remove_first_matching<F>(&mut self, f: F) -> Option<T>
    where
        F: Fn(&T) -> bool,
    {...}
```
against the trait spec ([here](https://github.com/PLSysSec/tock/blob/793c017ef05edb13c51215c7bef0622dd1735cb3/kernel/src/collections/queue.rs#L30-L35)):
```Rust
    #[flux_rs::sig(
        fn(self: &strg Self, _) -> Option<_> ensures self: Self
    )]
    fn remove_first_matching<F>(&mut self, f: F) -> Option<T>
    where
        F: Fn(&T) -> bool;
```
it fails with the error:
```
error[E0999]: internal flux error: crates/flux-infer/src/infer.rs:533:22
   --> kernel/src/collections/ring_buffer.rs:181:5
    |
181 | /     fn remove_first_matching<F>(&mut self, f: F) -> Option<T>
182 | |     where
183 | |         F: Fn(&T) -> bool,
    | |__________________________^
    |
    = note: incompatible types: `&'_ strg <self: RingBuffer<'_, λb0. T[b0]>[RingBuffer { a0, a1, a2 }]>` - `&'_ strg <?0e#8242: ∃b0. RingBuffer<'_, λb1. T[b1]>[b0]>`
```

I'm a little confused because the spec is token-for-token identical in both the trait and implementation, and only asserts that we need a `&strg Self`. 

I also get a similar error when there is no spec on the trait at all. I don't fully understand the implementation for trait_impl checking, so I'm not sure if this is expected or not.

 Any guidance?

ranjitjhala:
fixes #890

### DIFF
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -455,6 +455,20 @@ struct Sub {
     obligations: Vec<rty::Clause>,
 }
 
+/// [NOTE:unfold_strg_ref] We use this function to unfold a strong reference prior to a subtyping check.
+/// Normally, when checking a function body, a `StrgRef` is automatically unfolded
+/// i.e. `x:&strg T` is turned into turned into a `x:Ptr(l); l: T` where `l` is some
+/// fresh location. However, we need the below to do a similar unfolding in `check_fn_subtyping`
+/// where we just have the super-type signature that needs to be unfolded.
+/// We also add the binding to the environment so that we can:
+/// (1) UPDATE the location after the call, and
+/// (2) CHECK the relevant `ensures` clauses of the super-sig.
+/// Nico: More importantly, we are assuming functions always give back the "ownership"
+/// of the location so even though we should technically "consume" the ownership and
+/// remove the location from the environment, the type is always going to be overwritten.
+/// (there's a check for this btw, if you write an &strg we require an ensures for that
+/// location for the signature to be well-formed)
+
 impl Sub {
     fn new(reason: ConstrReason, span: Span) -> Self {
         Self { reason, span, obligations: vec![] }
@@ -487,7 +501,7 @@ impl Sub {
                 self.tys(infcx, &ty1, ty2)
             }
             (TyKind::StrgRef(_, path1, ty1), TyKind::StrgRef(_, path2, ty2)) => {
-                env.unfold_strg_ref(infcx, path1, ty1)?;
+                env.unfold_strg_ref(infcx, path1, ty1)?; // see [NOTE:unfold_strg_ref]
                 let ty1 = env.get(path1);
                 infcx.unify_exprs(&path1.to_expr(), &path2.to_expr());
                 self.tys(infcx, &ty1, ty2)
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -48,6 +48,14 @@ impl Tag {
 }
 
 #[derive(PartialEq, Eq, Clone, Copy, Hash, Debug)]
+pub enum SubtypeReason {
+    Input,
+    Output,
+    Requires,
+    Ensures,
+}
+
+#[derive(PartialEq, Eq, Clone, Copy, Hash, Debug)]
 pub enum ConstrReason {
     Call,
     Assign,
@@ -58,11 +66,8 @@ pub enum ConstrReason {
     Rem,
     Goto(BasicBlock),
     Overflow,
+    Subtype(SubtypeReason),
     Other,
-    SubtypeIn,
-    SubtypeReq,
-    SubtypeOut,
-    SubtypeEns,
 }
 
 pub struct InferCtxtRoot<'genv, 'tcx> {
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -59,6 +59,10 @@ pub enum ConstrReason {
     Goto(BasicBlock),
     Overflow,
     Other,
+    SubtypeIn,
+    SubtypeReq,
+    SubtypeOut,
+    SubtypeEns,
 }
 
 pub struct InferCtxtRoot<'genv, 'tcx> {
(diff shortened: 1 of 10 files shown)


crates/flux-infer/src/infer.rs
```diff
@@ -530,7 +542,7 @@ impl Sub {
                 }
                 Ok(())
             }
-            _ => Err(query_bug!("incompatible types: `{a:?}` - `{b:?}`"))?,
+            _ => panic!("incompatible types: `{a:?}` - `{b:?}`"), // _ => Err(query_bug!("incompatible types: `{a:?}` - `{b:?}`"))?,
```
nilehmann, line 545: Did you do this to get a backtrace? Maybe we can store the stack frame in the `QueryErr::Bug` to get the best of both worlds.

```suggestion
            _ => Err(query_bug!("incompatible types: `{a:?}` - `{b:?}`"))?,
```

crates/flux-refineck/src/type_env.rs
```diff
@@ -86,6 +87,25 @@ impl<'a> TypeEnv<'a> {
...
+        //     let ty = infcx.unpack(ty);
+        //     infcx.assume_invariants(&ty, check_overflow);
+        //     if let TyKind::StrgRef(_, path, ty) = ty.kind() {
+        //         let loc = path.to_loc().unwrap();
+        //         env.bindings.insert(loc, LocKind::Universal, ty.clone());
+        //     }
+        // }
+
+        // env.alloc(RETURN_PLACE);
+        // env
```
nilehmann, line 106:
```suggestion
```

crates/flux-infer/src/infer.rs
```diff
@@ -58,6 +59,10 @@ pub enum ConstrReason {
     Goto(BasicBlock),
     Overflow,
     Other,
+    SubtypeIn,
+    SubtypeReq,
+    SubtypeOut,
+    SubtypeEns,
```
nilehmann, line 65: Maybe we can have a separate enum as a sort of *subreason*. Not sure though.

Also, in general, I was thinking that we can improve the errors at function calls because we have access to the span of the (actual) argument

crates/flux-refineck/src/type_env.rs
```diff
@@ -329,6 +349,21 @@ impl<'a> TypeEnv<'a> {
         Ok(loc)
     }
 
+    pub(crate) fn unfold_strg_ref(
```
nilehmann, line 352: Maybe we should make a habit of writing inference rules if they are concise like in this case. I find them useful.

````suggestion
    /// ```
    /// -----------------------------------
    /// Γ ; &strg <ℓ: t> => Γ,ℓ: t ; ptr(ℓ)
    /// ```
    pub(crate) fn unfold_strg_ref(
````

File: crates/flux-infer/src/infer.rs [hunk 5]
```diff
@@ -475,6 +481,12 @@ impl Sub {
                 infcx.unify_exprs(&path1.to_expr(), &path2.to_expr());
                 self.tys(infcx, &ty1, ty2)
             }
+            (TyKind::StrgRef(_, path1, ty1), TyKind::StrgRef(_, path2, ty2)) => {
+                env.unfold_strg_ref(infcx, path1, ty1)?;
```
nilehmann, line 485: I assume not doing anything didn't work because when processing the ensures after the call we use `update_path`, so we would need something like `update_or_insert_path`. 

Maybe we can leave a comment explaining that. Or more importantly, we are assuming functions always give back the "ownership" of the location so even though we should technically "consume" the ownership and remove the location from the environment, the type is always going to be overwritten. (there's a check for this btw, if you write an &strg we require an ensures for that location for the signature to be well-formed)

crates/flux-refineck/src/checker.rs
```diff
@@ -314,26 +307,69 @@ fn check_fn_subtyping(
...
+
     let evars_sol = infcx.pop_scope().with_span(span)?;
     infcx.replace_evars(&evars_sol);
 
-    if !output.ensures.is_empty() || !oblig_output.ensures.is_empty() {
-        span_bug!(span, "Not yet handled: subtyping with ensures predicates {def_id:?}");
+    Ok(())
+}
+
+fn update_ensures(
```
nilehmann, line 327: I feel like `update_ensures` could be a method of `InferCtxt` and `check_ensures` a method in `InferCtxtAt`, but we would need to make them generic on `LocEnv`. Not sure if it's worth the trouble. Alternatively, they can both be methods of `TypeEnv`.

crates/flux-refineck/src/checker.rs
```diff
@@ -314,26 +307,69 @@ fn check_fn_subtyping(
...
+    infcx: &mut InferCtxt,
+    env: &mut TypeEnv,
+    output: &FnOutput,
+    overflow_checking: bool,
+) -> Result {
```
nilehmann, line 332: This function is infallible
```suggestion
) {
```

crates/flux-refineck/src/checker.rs
```diff
@@ -314,26 +307,69 @@ fn check_fn_subtyping(
...
+    Ok(())
+}
 
+fn check_ensures(
+    at: &mut InferCtxtAt,
+    env: &mut TypeEnv,
+    output: &FnOutput,
+    reason: ConstrReason,
+    span: Span,
+) -> Result {
```
nilehmann, line 352: Independent of making this a method in `InferCtxtAt` (or `TypeEnv`), I'd make this return an `InferResult`. Then we can avoid the extra `span` argument, and move the `with_span` to the caller.

```suggestion
) -> InferResult {
```

nilehmann (approved):
Looks good. Some stylistic comments.

crates/flux-infer/src/infer.rs
ranjitjhala: good idea -- let me spin this out into a separate issue (improving errors at calls).

PR #900, opened 2024-11-21
Push/Pop scope in function subtyping

### LINKED ISSUE #899: ICE: `unexpected evar`
Now that  #891 has added support for `&strg` in function subtyping, some of the `Queue` trait methods (as implemented by `RingBuffer`) pass, although some are still failing with the error:
```
error: internal compiler error: crates/flux-infer/src/fixpoint_encoding.rs:921:17: unexpected evar: `?0e#110`
   --> src/main.rs:855:5
    |
855 |     fn enqueue(&mut self, val: T) -> bool {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

thread 'rustc' panicked at crates/flux-infer/src/fixpoint_encoding.rs:921:17:
```

To help with debugging this issue, I made a minimal standalone Ringbuffer implementation that reproduces this issue:

```Rust

use core::slice::{Iter, SliceIndex};

pub trait Queue<T> {
    /// Returns true if the queue is full, false otherwise.
    fn is_full(&self) -> bool;

    /// If the queue isn't full, add a new element to the back of the queue.
    /// Returns whether the element was added.
    #[flux_rs::sig(fn(self: &strg Self, _) -> bool ensures self: Self)]
    fn enqueue(&mut self, val: T) -> bool;
}


#[flux_rs::extern_spec]
impl<T> [T] {
    #[flux_rs::sig(fn(&[T][@len]) -> usize[len])]
    fn len(v: &[T]) -> usize;
}


#[flux_rs::refined_by(ring_len: int, hd: int, tl: int)]
#[flux_rs::invariant(ring_len > 1)]
pub struct RingBuffer<'a, T: 'a> {
    #[field({&mut [T][ring_len] | ring_len > 1})]
    ring: &'a mut [T],
    #[field({usize[hd] | hd < ring_len})]
    head: usize,
    #[field({usize[tl] | tl < ring_len})]
    tail: usize,
}

impl<T: Copy> Queue<T> for RingBuffer<'_, T> {
    fn is_full(&self) -> bool {
        self.head == ((self.tail + 1) % self.ring.len())
    }

    #[flux_rs::sig(fn(self: &strg Self, _) -> bool ensures self: Self)]
    fn enqueue(&mut self, val: T) -> bool {
        if self.is_full() {
            // Incrementing tail will overwrite head
            false
        } else {
            self.ring[self.tail] = val;
            self.tail = (self.tail + 1) % self.ring.len();
            true
        }
    }
}
```

ranjitjhala:
fixes #899


crates/flux-refineck/src/checker.rs
```diff
@@ -315,13 +315,15 @@ fn check_fn_subtyping(
     infcx
         .subtyping(&output.ret, &super_output.ret, reason)
         .with_span(span)?;
+    let evars_sol = infcx.pop_scope().with_span(span)?;
```
ranjitjhala, line 318: These three lines are the fix - my q. is why isn't it enough to just have the stuff at the bottom? (the very last infcx.replace_evars) I note that the `update_ensures` calls `infcx.unpack` and that somehow messes stuff up and necessitates the `replace_evars` call BEFORE `update_ensures`, but why?

nilehmann: The `InferCtxt` is a cursor into a tree. Some functions like `define_var`, `assume_pred` and by extension `unpack` _advance_ the cursor. For example, defining a variable pushes a node as a child of the current node and then moves the cursor into that new node. Other functions, like `subtyping` or  `check_pred` (typically the ones defined in `InferCtxtAt`) _do not_ advance the cursor (from the caller's perspective). For example, if you call `subtyping` a subtree is pushed under the current node, and the cursor is returned to where it was. That's the purpose of `infcx.branch`: to "clone" the cursor such that the original one doesn't get modified.

`infcx.pop_scope` and `infcx.replace_evars` are supposed to be called when the `infcx` is on the same node where the `push_scope` was called, because `replace_evars` only replaces evars below that node.  What was happening here was that `pop_scope` was being called at a node further down (after calling `unpack`) so evars above that weren't being replaced.

This is very error-prone. The whole evar infrastructure needs some revamping.

nilehmann: We could accumulate solutions and apply them at the very end. The only evars we need to solve eagerly are the ones for locations because we need them to apply the ensures. Locations aside, we could even generate the constraint with evars and then solve them _in_ the constraint.

ranjitjhala: I see -- definitely quite delicate. I suggest we merge this in right now, though, (assuming that the push/pop is now done properly) and come back to clean up this push/pop/evar API later.

nilehmann (approved):
Yes, let's merge this.

PR #902, opened 2024-11-22
Refactor sort checking to deal with synthesis for constructors

### LINKED ISSUE #895: Better sort checking for constructors
Right now, the following fails:

```rust
use flux_rs::*;

#[flux::refined_by(start: T, end: T)]
struct Range<T> {
    #[flux::field(T[start])]
    pub start: T,
    #[flux::field(T[end])]
    pub end: T,
}

#[flux::sig(fn(r: Range<i32>{v: Range { start: 0, end: 1 } == v }))]
fn foo2(_r: Range<i32>) {}
```

Result: 

```rust
  --> ../example.rs:11:33
   |
11 | #[flux::sig(fn(r: Range<i32>{v: Range { start: 0, end: 1 } == v }))]
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `?1s`, found constructor
 ```
 
 A nice solution for this would be (from @nilehmann): 
 
 The nice solution for this problem is whenever we check `Range { ... }` against a sort `s` we unify it against `Range<?1s>` (for a fresh ?1s). And then check fields assuming `?1s` . If `s` is `Range<int>` we will learn `Range<?1s> = Range<int> ==> ?1s = int` and consequently check fields against int. If `s` is `?0s` we only learn `?0s = Range<?1s>` and will check fields against `?1s`.

### PR DESCRIPTION
**Pull request by vrindisbacher**
closes #895

### DIFF
--- a/crates/flux-fhir-analysis/src/wf/sortck.rs
+++ b/crates/flux-fhir-analysis/src/wf/sortck.rs
@@ -4,7 +4,7 @@ use ena::unify::InPlaceUnificationTable;
 use flux_common::{bug, iter::IterExt, result::ResultExt, span_bug, tracked_span_bug};
 use flux_errors::{ErrorGuaranteed, Errors};
 use flux_middle::{
-    fhir::{self, visit::Visitor as _, ExprRes, FhirId, FluxOwnerId, PathExpr},
+    fhir::{self, visit::Visitor as _, ExprRes, FhirId, FluxOwnerId},
     global_env::GlobalEnv,
     queries::QueryResult,
     rty::{
@@ -129,33 +129,26 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
     fn check_constructor(
         &mut self,
         expr: &fhir::Expr,
-        path: &Option<PathExpr>,
         field_exprs: &[fhir::FieldExpr],
         spread: &Option<&fhir::Spread>,
         expected: &rty::Sort,
     ) -> Result {
-        if let rty::Sort::App(rty::SortCtor::Adt(sort_def), sort_args) = expected {
-            let expected_def_id = sort_def.did();
-            if let Some(path) = path {
-                let path_def_id = match path.res {
-                    ExprRes::Struct(def_id) | ExprRes::Enum(def_id) => def_id,
-                    _ => span_bug!(expr.span, "unexpected path in constructor"),
-                };
-                if path_def_id != expected_def_id {
-                    return Err(self.emit_err(errors::SortMismatchFoundOmitted::new(
-                        path.span,
-                        expected.clone(),
-                    )));
-                }
-            } else {
-                self.wfckresults
-                    .record_ctors_mut()
-                    .insert(expr.fhir_id, sort_def.did());
-            }
-            self.check_field_exprs(expr.span, sort_def, sort_args, field_exprs, spread, expected)?;
+        let expected = self.resolve_vars_if_possible(&expected);
+        if let rty::Sort::App(rty::SortCtor::Adt(sort_def), sort_args) = &expected {
+            self.wfckresults
+                .record_ctors_mut()
+                .insert(expr.fhir_id, sort_def.did());
+            self.check_field_exprs(
+                expr.span,
+                &sort_def,
+                &sort_args,
+                field_exprs,
+                spread,
+                &expected,
+            )?;
             Ok(())
         } else {
-            Err(self.emit_err(errors::UnexpectedConstructor::new(expr.span, expected)))
+            Err(self.emit_err(errors::UnexpectedConstructor::new(expr.span, &expected)))
         }
     }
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crates/flux-fhir-analysis/src/wf/sortck.rs [hunk 1]
```diff
@@ -270,8 +263,58 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                let mut used_fields = FxHashSet::default();
+                for expr in field_exprs.iter() {
+                    // check each expression against the sort
+                    // which will unify inferred variables
+                    if let Some(sort) = sort_by_name.get(&expr.ident.name) {
+                        self.check_expr(&expr.expr, sort)?;
+                    } else {
+                        return Err(self.emit_field_not_found(&sort, expr.ident));
+                    }
+                    if let Some(old_field) = used_fields.replace(expr.ident) {
```
vrindisbacher, line 293: @nilehmann - I'm kinda annoyed that we have to do this in both checking and synthesis. It feels like this shouldn't really be here but I'm not sure there's a way around it.

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -270,8 +263,58 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                        return Err(
+                            self.emit_err(errors::DuplicateFieldUsed::new(expr.ident, old_field))
+                        );
+                    }
+                }
+                // check the spread against the sort
+                if let Some(spread) = spread {
+                    self.check_expr(&spread.expr, &sort)?;
+                } else if sort_by_name.len() != used_fields.len() {
```
vrindisbacher, line 302: Same with this

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -270,8 +256,58 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                // first get the sort based on the path - for example S { ... } => S
+                // and we should expect sort to be a struct or enum app
+                let path_def_id = match path.res {
+                    ExprRes::Struct(def_id) | ExprRes::Enum(def_id) => def_id,
+                    _ => span_bug!(expr.span, "unexpected path in constructor"),
+                };
+                let sort_def = match self.genv.adt_sort_def_of(path_def_id) {
+                    Ok(sd) => sd,
+                    Err(_) => span_bug!(path.span, "Unexpected path in constructor"),
+                };
```
nilehmann, line 270: This error needs to be propagated

```suggestion
                let sort_def = match self.genv.adt_sort_def_of(path_def_id).emit(&self.genv)?;
```

crates/flux-fhir-analysis/src/wf/sortck.rs
nilehmann: yes, I agree. We could move part of the logic to another function that only checks if all fields are present.

```rust
fn validate_constructor_fields(&self, sort_def: &AdtSortDef, exprs: &[fhir::FieldExpr], has_spread: bool);
```

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -270,8 +279,42 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                    // check each expression against the sort
+                    // which will unify inferred variables
+                    if let Some(sort) = sort_by_name.get(&expr.ident.name) {
+                        self.check_expr(&expr.expr, sort)?;
+                    }
+                }
+                // check the spread against the sort
+                if let Some(spread) = spread {
+                    self.check_expr(&spread.expr, &sort)?;
+                }
```
nilehmann, line 314: This is the same as `check_field_expr`, no? Maybe we don't need `check_constructor_shape` and could call the old `check_field_expr`.

vrindisbacher: Yea it is - good point.

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -270,8 +256,36 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+                // first get the sort based on the path - for example S { ... } => S
+                // and we should expect sort to be a struct or enum app
+                let path_def_id = match path.res {
+                    ExprRes::Struct(def_id) | ExprRes::Enum(def_id) => def_id,
+                    _ => span_bug!(expr.span, "unexpected path in constructor"),
+                };
+                let sort_def = match self.genv.adt_sort_def_of(path_def_id) {
+                    Ok(sd) => sd,
+                    Err(_) => span_bug!(path.span, "Unexpected path in constructor"),
+                };
```
nilehmann, line 270: I think this got reverted. The error should be propagated here.

nilehmann (approved):

PR #904, opened 2024-11-23
Beef up infer-mut-hack

ranjitjhala:
(To avoid it getting "confused" be extraneous `Constr`)

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -818,7 +818,7 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 .with_span(span)?;
         }
         // Replace evars
-        let evars_sol = infcx.pop_scope().unwrap(); // with_span(span)?;
+        let evars_sol = infcx.pop_scope().with_span(span)?;
         env.replace_evars(&evars_sol);
         infcx.replace_evars(&evars_sol);
 
@@ -1743,8 +1743,6 @@ fn infer_under_mut_ref_hack(
         .map(|(actual, formal)| {
             if let rty::Ref!(.., Mutability::Mut) = actual.kind()
                 && is_indexed_mut_skipping_constr(formal)
-            // && let rty::Ref!(_, ty, Mutability::Mut) = formal.kind()
-            // && let TyKind::Indexed(..) = ty.kind()
             {
                 rcx.hoister(AssumeInvariants::No)
                     .hoist_inside_mut_refs(true)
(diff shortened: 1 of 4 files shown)


crates/flux-refineck/src/checker.rs
```diff
@@ -1704,6 +1707,29 @@ fn all_predicates_of(
     .flatten()
 }
 
+struct SkipConstr;
```
nilehmann, line 1710: @ranjitjhala 

We already have [Ty::unconstr](https://flux-rs.github.io/flux/doc/flux_middle/rty/struct.Ty.html#method.unconstr) which is a more general version of this.

ranjitjhala: Interesting -- didn't know about it! 

But, not really, replacing

```rust
let ty = SkipConstr.fold_ty(ty);
```

with

```rust
 let (ty, _) = ty.unconstr();
```

Doesn't seem to work -- looks like `unconstr` is not "deep": it expects the `Constr` at the very top?

ranjitjhala: Looks like `unconstr` just peels away all the "topmost" `Constr` -- the moment it hits something else it stops. 

So it won't work with 

```rust
#[flux::sig(fn (xs: &mut { [i32][@n] | n > 1}))]
fn lib1(_xs: &mut [i32]) {}
```

nilehmann: You first need to match on the `&mut T` and then `unconstr` the `T`

```rust
 if let rty::Ref!(_, deref_ty, Mutability::Mut) = ty.kind()
        && let (deref_ty, _) = deref_ty.unconstr()
        && let TyKind::Indexed(..) = inner_ty.kind()
```

I wonder though if there's a clean way to handle this during [_canonicalization_](https://github.com/flux-rs/flux/blob/main/crates/flux-middle/src/rty/canonicalize.rs) (the thing `unpack` is implementing on top of). It is always safe to turn an `&mut {T | p}` into `{&mut T | p}` so we could get rid of it during `unpack`

ranjitjhala: This is why I have two tests -- the other one flips the order which fails your method

```rust
#[flux::sig(fn (xs: &mut { [i32][@n] | n > 1}))]
fn lib1(_xs: &mut [i32]) {}

#[flux::sig(fn (xs: { &mut  [i32][@n] | n > 1}))]
fn lib2(_xs: &mut [i32]) {}
```

But yes, if you had `canonicalized` presumably `lib1` gets turned into `lib2`?

nilehmann: ah yeah, you'd need to unconstr twice

```rust
let (ty, _) = ty.unconstr();
if let rty::Ref!(_, deref_ty, Mutability::Mut) = ty.kind() 
    && let (deref_ty, _) = deref_ty.unconstr() 
    && let TyKind::Indexed(..) = inner_ty.kind()
```

I still think there has to be a nice way in which we can canonicalize types before pattern matching on them. The canonicalize infra is almost there...

PR #911, opened 2024-11-27
Make checker trace more JSON-ish 

ranjitjhala:
RCX/ENV are still a string, but a JSON formatted one...


Cargo.toml
```diff
@@ -37,6 +37,8 @@ home = "0.5.9"
 itertools = "0.13.0"
 pad-adapter = "0.1.1"
 serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+display_json = "=0.2.1"
```
nilehmann, line 41: This dependency is using `syn1`, so I prefer to avoid it. Also, the part we are using is easy to replicate. Just add the following to `crates/flux-macros/src/lib.rs`


```rust
use quote::quote;

...

decl_derive!(
    [DebugAsJson] => debug_as_json
);

fn debug_as_json(s: synstructure::Structure<'_>) -> proc_macro2::TokenStream {
    s.gen_impl(quote! {
        gen impl std::fmt::Debug for @Self {
            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                let s = serde_json::to_string(&self).map_err(|_| std::fmt::Error::default())?;
                write!(f, "{}", s)
            }
        }
    })
}
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -776,3 +778,37 @@ mod pretty {
...
+pub struct RefineCtxtTrace {
+    bindings: Vec<RcxBind>,
+    exprs: Vec<String>,
+}
+
+#[derive(Serialize, DebugAsJson)]
+pub struct RcxBind {
+    pub name: String,
+    pub sort: String,
+}
```
nilehmann, line 793:
```suggestion
#[derive(Serialize)]
struct RcxBind {
    name: String,
    sort: String,
}

```

crates/flux-refineck/src/checker.rs
```diff
@@ -509,9 +509,15 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
             bug::track_span(span, || {
-                dbg::statement_span!("start", stmt, infcx, env, span);
+                dbg::statement!("start", stmt, infcx, env, span);
                 self.check_statement(&mut infcx, &mut env, stmt)?;
-                dbg::statement_span!("end", stmt, infcx, env, span);
+                dbg::statement!(
+                    "end",
+                    stmt,
+                    RefineCtxtTrace::new(&infcx),
+                    TypeEnvTrace::new(&env),
```
nilehmann, line 518: I'd put this inside the macro. The `Checker` doesn't care about how these are dumped.

crates/flux-refineck/src/type_env.rs
```diff
@@ -873,3 +875,32 @@ mod pretty {
...
+/// A very explicit representation of `RefineCtxt` for debugging/tracing/serialization ONLY.
+#[derive(Serialize, DebugAsJson)]
+pub struct TypeEnvTrace(Vec<TypeEnvBind>);
+
+#[derive(Serialize, DebugAsJson)]
+pub struct TypeEnvBind {
+    pub loc: String,
+    pub kind: String,
+    pub ty: String,
+}
```
nilehmann, line 888:
```suggestion
#[derive(Serialize)]
struct TypeEnvBind {
    loc: String,
    kind: String,
    ty: String,
}
```

crates/flux-refineck/src/type_env.rs
```diff
@@ -873,3 +875,32 @@ mod pretty {
         BasicBlockEnv => "basic_block_env"
     }
 }
+
+/// A very explicit representation of `RefineCtxt` for debugging/tracing/serialization ONLY.
```
nilehmann, line 879:
```suggestion
/// A very explicit representation of [`TypeEnv`] for debugging/tracing/serialization ONLY.
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -776,3 +778,37 @@ mod pretty {
         Scope,
     );
 }
+
+/// A very explicit representation of `RefineCtxt` for debugging/tracing/serialization ONLY.
```
nilehmann, line 782:
```suggestion
/// A very explicit representation of [`RefineCtxt`] for debugging/tracing/serialization ONLY.
```

nilehmann (commented):
Did you try `logreader.py` after this change? I'd like for that script to keep working.

Cargo.toml
ranjitjhala: To be clear, you mean remove the dependency on `display_json` but keep `serde_json` yes? (because we already depend on the latter in fixpoint, and also in the above snippet?)

ranjitjhala:
Oops yes, have restored the functionality of the `logreader.py` -- the "json" stuff is now emitted as an _extra_ field so as to not confuse the `logreader` (have also tested it works locally) -- which made pushing the `::new(...)` stuff inside the macro that much more essential!

nilehmann (approved):

PR #915, opened 2024-12-03
Vscode

ranjitjhala:


ranjitjhala:
@nilehmann here's the pure vscode stuff ... except i've made the `dbg` thing emit the `statement` events whenever the `dump_checker_trace` is on (not just debug-builds).

crates/flux-common/src/dbg.rs
```diff
@@ -71,9 +71,11 @@ pub use crate::_basic_block_start as basic_block_start;
 #[macro_export]
 macro_rules! _statement{
     ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr) => {{
-        let rcx_json = RefineCtxtTrace::new($rcx);
-        let env_json = TypeEnvTrace::new($env);
-        tracing::debug!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
+        if config::dump_checker_trace() {
```
nilehmann, line 74: I think we want more granularity than this because this is also going to dump the trace for the shape phase, which I don't think we ever want to expose to users. Maybe we can make `config::dump_checker_trace` return an enum with three levels (none, only check phase, all). WE would also need to pass the phase to this macro. 

We can implement this later though.

nilehmann (approved):
Looks good to me

PR #917, opened 2024-12-03
Check specified files (only)

ranjitjhala:
Adds a flag to make `flux` only check `DefId` in specified files... (to enable more interactive checking)


crates/flux-config/src/lib.rs
```diff
@@ -81,6 +85,7 @@ struct Config {
     catch_bugs: bool,
     pointer_width: PointerWidth,
     check_def: String,
+    check_files: String,
```
nilehmann, line 88: We can parse this as a list of paths (I haven't tested it)

```suggestion
    check_files: Paths,
```

and then

```rust
#[derive(Default)]
struct Paths {
    paths: Vec<PathBuf>,
}

impl<'de> Deserialize<'de> for Paths {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let paths = <&str>::deserialize(deserializer)?
            .split(',')
            .into_iter()
            .map(PathBuf::from)
            .collect();
        Ok(Paths { paths })
    }
}
```

crates/flux-config/src/lib.rs
```diff
@@ -47,6 +47,10 @@ pub fn is_cache_enabled() -> bool {
     CONFIG.cache
 }
 
+pub fn is_checked_file(file: &str) -> bool {
```
nilehmann, line 50: If we parse it as a list of files with can make this take a `Path`

```suggestion
pub fn is_checked_file(file: &Path) -> bool {
```

nilehmann (approved):
Looks good. Just one suggestion to make the configuration more structured.

crates/flux-config/src/lib.rs
nilehmann: and we should canonicalize paths (if possible) before comparing them.

crates/flux-config/src/lib.rs
nilehmann: Or more generally we could have a way to parse arbitrary lists

```rust
struct List<T> {
    values: Vec<T>,
}

impl<'de, T> Deserialize<'de> for List<T>
where
    T: Deserialize<'de>,
{
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let values = <&str>::deserialize(deserializer)?
            .split(',')
            .into_iter()
            .map(|s| T::deserialize(StrDeserializer::new(s)))
            .collect::<Result<_, _>>()?;
        Ok(List { values })
    }
}
```

ranjitjhala: Getting this weird error (when using the `PathBuf` code)?

```
rjhala@pozole ~/r/flux-demo (main) [101]> RUST_BACKTRACE=1 FLUX_CHECK_FILES=src/basics.rs cargo flux
error: process didn't exit successfully: `/Users/rjhala/.flux/flux-driver -vV` (exit status: 101)
--- stderr
thread 'main' panicked at crates/flux-config/src/lib.rs:181:13:
called `Result::unwrap()` on an `Err` value: invalid type: string "src/basics.rs", expected a borrowed string
```

nilehmann: You can try:

```rust
impl<'de> Deserialize<'de> for Paths {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let paths = String::deserialize(deserializer)?
            .split(',')
            .into_iter()
            .map(PathBuf::from)
            .collect();
        Ok(Paths { paths })
    }
}
```

if that doesn't work nvm

ranjitjhala: :-( am making the above a separate issue so I can merge this stuff in now...

ranjitjhala:
Spoke to soon - will try to decipher tomorrow

- Ranjit.

PR #919, opened 2024-12-04
Modify `vscode` extension to use `check_files`

ranjitjhala:
Modify the `vscode` extension so it passes a specific set of files to check (and generate the `FLUX_CHECKER_TRACE` for), so that we don't generate traces for ALL files which could be rather slow...


nilehmann (approved):
Didn't look at it in details but it seems fine

PR #920, opened 2024-12-04
Fixes paths thing

### LINKED ISSUE #918: Simplify `check_files`
We should represent the configuration field `check_files` as 

```suggestion
    check_files: Paths,
```

where

```rust
#[derive(Default)]
struct Paths {
    paths: Vec<PathBuf>,
}

impl<'de> Deserialize<'de> for Paths {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let paths = <&str>::deserialize(deserializer)?
            .split(',')
            .into_iter()
            .map(PathBuf::from)
            .collect();
        Ok(Paths { paths })
    }
}
```

_Originally posted by @nilehmann in https://github.com/flux-rs/flux/pull/917#discussion_r1868447480_

ranjitjhala:
Adds @nilehmann 's deserialize thing 

fixes #918

### DIFF
--- a/crates/flux-config/src/lib.rs
+++ b/crates/flux-config/src/lib.rs
@@ -48,7 +48,7 @@ pub fn is_cache_enabled() -> bool {
 }
 
 pub fn is_checked_file(file: &str) -> bool {
-    CONFIG.check_files.is_empty() || CONFIG.check_files.contains(file)
+    CONFIG.check_files.is_checked_file(file)
 }
 
 pub fn cache_path() -> PathBuf {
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Um ... ok for some odd reason the `neg` tests are all failing ? Any clue?

ranjitjhala:
@nilehmann ?

nilehmann:
calling `split(",")` on the empty springs results in a `vec![""]`, so the default config always skips everything. To account for possible whitespace:

```rust
        let paths = String::deserialize(deserializer)?
            .split(",")
            .map(str::trim)
            .filter(|s| !s.is_empty())
            .map(PathBuf::from)
            .collect();
 ```
 
 but also maybe better to make it an `Option<Paths>` to be more explicit

PR #927, opened 2024-12-05
Pretty print records using field names

ranjitjhala:
Instead of just commas or `foo.0.1.2` ... 

Screenshot 2024-12-05 at 1 56 28 PM

### DIFF
--- a/crates/flux-common/src/dbg.rs
+++ b/crates/flux-common/src/dbg.rs
@@ -68,18 +68,6 @@ macro_rules! _basic_block_start {
 }
 pub use crate::_basic_block_start as basic_block_start;
 
-// #[macro_export]
-// macro_rules! _statement{
-//     ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr) => {{
-//         if config::dump_checker_trace() {
-//             let rcx_json = RefineCtxtTrace::new($rcx);
-//             let env_json = TypeEnvTrace::new($env);
-//             tracing::info!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
-//         }
-//     }};
-// }
-// pub use crate::_statement as statement;
-
 #[macro_export]
 macro_rules! _statement{
     ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr, $genv:expr) => {{
--- a/crates/flux-common/src/dbg.rs
+++ b/crates/flux-common/src/dbg.rs
@@ -86,7 +86,7 @@ macro_rules! _statement{
         if config::dump_checker_trace() {
             let genv = $genv;
             let rcx_json = RefineCtxtTrace::new(genv, $rcx);
-            let env_json = TypeEnvTrace::new($env);
+            let env_json = TypeEnvTrace::new(genv, $env);
             tracing::info!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
         }
     }};
--- a/crates/flux-common/src/dbg.rs
+++ b/crates/flux-common/src/dbg.rs
@@ -68,11 +68,24 @@ macro_rules! _basic_block_start {
 }
 pub use crate::_basic_block_start as basic_block_start;
 
+// #[macro_export]
+// macro_rules! _statement{
+//     ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr) => {{
+//         if config::dump_checker_trace() {
+//             let rcx_json = RefineCtxtTrace::new($rcx);
+//             let env_json = TypeEnvTrace::new($env);
+//             tracing::info!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
+//         }
+//     }};
+// }
+// pub use crate::_statement as statement;
+
 #[macro_export]
 macro_rules! _statement{
-    ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr) => {{
+    ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr, $genv:expr) => {{
         if config::dump_checker_trace() {
-            let rcx_json = RefineCtxtTrace::new($rcx);
+            let genv = $genv;
+            let rcx_json = RefineCtxtTrace::new(genv, $rcx);
             let env_json = TypeEnvTrace::new($env);
             tracing::info!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
         }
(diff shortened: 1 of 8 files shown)


crates/flux-common/src/dbg.rs
```diff
@@ -68,12 +68,25 @@ macro_rules! _basic_block_start {
...
+//     ($pos:literal, $stmt:expr, $rcx:expr, $env:expr, $span:expr) => {{
+//         if config::dump_checker_trace() {
+//             let rcx_json = RefineCtxtTrace::new($rcx);
+//             let env_json = TypeEnvTrace::new($env);
+//             tracing::info!(event = concat!("statement_", $pos), stmt = ?$stmt, stmt_span = ?$span, rcx = ?$rcx, env = ?$env, rcx_json = ?rcx_json, env_json = ?env_json)
+//         }
+//     }};
+// }
+// pub use crate::_statement as statement;
+
```
nilehmann, line 82:
```suggestion
```

crates/flux-middle/src/pretty.rs
```diff
@@ -378,10 +383,14 @@ impl PrettyCx<'_> {
     pub fn hide_sorts(self, b: bool) -> Self {
         Self { hide_sorts: b, ..self }
     }
+
+    pub fn with_genv(self, genv: GlobalEnv<'genv, 'tcx>) -> Self {
+        Self { genv: Some(genv), ..self }
+    }
```
nilehmann, line 389: nit, it's a bit annoying you have to construct it `PrettyCx::default(genv.tcx()).with_genv(genv)`, so maybe we can make this a constructor

```suggestion
    pub fn default_with_genv(genv: GlobalEnv<'genv, 'tcx>) -> Self {
        Self { genv: Some(genv), Self::default(genv.tcx()) }
    }
```

or 

```rust
    pub fn with_genv(genv: GlobalEnv<'genv, 'tcx>) -> Self {
        Self { genv: Some(genv), Self::default(genv.tcx()) }
    }
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1077,11 +1078,22 @@ impl From<char> for Constant {
...
+    pub name: Symbol,
+    pub value: T,
+}
+
+impl<T: Pretty> Pretty for FieldBind<T> {
+    fn fmt(&self, cx: &PrettyCx, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        define_scoped!(cx, f);
+        w!("{}: {:?}", ^self.name, &self.value)
+    }
+}
```
nilehmann, line 1092:
```suggestion
}

```

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -277,8 +278,25 @@ impl Pretty for Ty {
             }
             TyKind::Param(param_ty) => w!("{}", ^param_ty),
             TyKind::Downcast(adt, .., variant_idx, fields) => {
-                w!("{:?}::{}", adt.did(), ^adt.variant(*variant_idx).name)?;
-                if !fields.is_empty() {
+                let is_struct = adt.is_struct();
```
nilehmann, line 281: To ponder: maybe `TyKind::Downcast` is something we should hide from users so instead of printing 
```
_1: S { x: i32[a0], y: i32[b1] }
```
we should print 

```
_1.x: i32[a0], _1.y: i32[a1]
```

But perhaps the context that it is an `S` may be relevant to users too

But also `{ x: i32[a0], y: i32[a1] }` doesn't have an equivalence in Rust syntax.

nilehmann (approved):

crates/flux-middle/src/rty/pretty.rs
ranjitjhala: Yes, a bit unclear what the right thing here is...

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1177,7 +1199,19 @@ mod pretty {
...
+                    if let Some(genv) = cx.genv
+                        && let Ok(adt_sort_def) = genv.adt_sort_def_of(def_id)
+                    {
+                        let field_binds = adt_sort_def
+                            .field_names()
+                            .iter()
+                            .zip(flds)
+                            .map(|(name, value)| FieldBind { name: *name, value: value.clone() })
+                            .collect_vec();
+                        w!("{{ {:?} }}", join!(", ", field_binds))
```
nilehmann, line 1211: wait @ranjitjhala!. Didn't notice this, I'm not sure we should skip the `def_id` blindly. So, if you write 

```rust
S { v: v = S { x: 0, y: 0 } }
```

I think we do want to keep the `S` inside

nilehmann: With this change we print `S {v: v = {x: 0, y: 0}}`

ranjitjhala: Good catch, no worries, will fix.

ranjitjhala: I'll elide the `S` (`DefId`) only when printing ... a top-level index?

nilehmann: yes, that sounds good

ranjitjhala: I used a newtype wrapper type trick to only elide the `DefId` for top-level indices, 

https://github.com/flux-rs/flux/pull/927/commits/a70354dcf924f1b9b96de10d6cbc0856dfed69b7

did you want to take a look? Or merge?

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -235,6 +236,32 @@ impl Pretty for SubsetTy {
...
+        if let ExprKind::Aggregate(AggregateKind::Adt(def_id), flds) = e.kind()
+            && let Some(genv) = cx.genv
+            && let Ok(adt_sort_def) = genv.adt_sort_def_of(def_id)
+        {
+            let field_binds = adt_sort_def
+                .field_names()
+                .iter()
+                .zip(flds)
+                .map(|(name, value)| FieldBind { name: *name, value: value.clone() })
+                .collect_vec();
```
nilehmann, line 257: We don't need to collect into a vec, `join!` takes anything iterable (I think)

```suggestion
```

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -235,6 +236,32 @@ impl Pretty for SubsetTy {
...
+        let e = &self.0;
+        let e = if cx.simplify_exprs { e.simplify() } else { e.clone() };
+        if let ExprKind::Aggregate(AggregateKind::Adt(def_id), flds) = e.kind()
+            && let Some(genv) = cx.genv
+            && let Ok(adt_sort_def) = genv.adt_sort_def_of(def_id)
+        {
+            let field_binds = adt_sort_def
+                .field_names()
+                .iter()
+                .zip(flds)
```
nilehmann, line 255: nit: I typically prefer using `zip` as a function

```suggestion
            let field_binds = iter::zip(adt_sort_def.field_names(), flds)
```

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -235,6 +236,32 @@ impl Pretty for SubsetTy {
     }
 }
 
+// This is a trick to avoid pretty printing `S [S { x: 10, y: 20}]`
+// and instead just print `S[{x: 10, y: 20}]` for struct-valued indices.
+struct TopIndexExpr(Expr);
```
nilehmann, line 241: I'd call this `IdxFmt` or `IdxPretty`

```suggestion
struct IdxFmt(Expr);
```

nilehmann (approved):
looks good

PR #930, opened 2024-12-06
Source-level names in JSON trace

ranjitjhala:
- Print `x`, `y`, `z` instead of `_1`, `_2`, `_3` for the JSON `TypeEnv` 
- Filters out places in the `TypeEnv` that do not correspond to source-level names
- Elides the printing of lifetimes

### DIFF
--- a/crates/flux-refineck/src/type_env.rs
+++ b/crates/flux-refineck/src/type_env.rs
@@ -893,7 +893,7 @@ fn loc_string(local_names: &UnordMap<Local, Symbol>, loc: &Loc) -> Option<String
         let name = local_names.get(local)?;
         return Some(format!("{}", name));
     }
-    return Some(format!("{:?}", loc));
+    Some(format!("{:?}", loc))
 }
 
 impl TypeEnvTrace {
--- a/crates/flux-refineck/src/type_env.rs
+++ b/crates/flux-refineck/src/type_env.rs
@@ -883,17 +883,31 @@ pub struct TypeEnvTrace(Vec<TypeEnvBind>);
 
 #[derive(Serialize)]
 struct TypeEnvBind {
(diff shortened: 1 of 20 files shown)


crates/flux-refineck/src/type_env.rs
```diff
@@ -886,20 +888,30 @@ struct TypeEnvBind {
...
-                let ty = WithCx::new(&cx, binding.ty.clone());
-                let ty = format!("{:?}", ty);
-                bindings.push(TypeEnvBind { loc, kind, ty });
+                // filtering out bindings that do not correspond to source-level names
+                if let Some(loc) = loc_string(local_names, loc) {
+                    let kind = format!("{:?}", binding.kind);
+                    let ty = WithCx::new(&cx, binding.ty.clone());
+                    let ty = format!("{:?}", ty);
+                    bindings.push(TypeEnvBind { loc, kind, ty });
+                }
```
nilehmann, line 914: Maybe we can store both the `loc` and the (optional) var name and decide whether to show it to the client.

crates/flux-rustc-bridge/src/lowering.rs
```diff
@@ -151,7 +151,21 @@ impl<'sess, 'tcx> MirLoweringCtxt<'_, 'sess, 'tcx> {
...
+            .iter()
+            .flat_map(|var_debug_info| {
+                if let VarDebugInfoContents::Place(place) = var_debug_info.value {
+                    let local = place.as_local()?;
+                    Some((local, var_debug_info.name))
+                } else {
+                    None
+                }
+            })
+            .collect();
```
nilehmann, line 166: I'd put this inside `Body::new`

nilehmann (approved):

PR #934, opened 2024-12-10
NestedString representation for VSCode extension

ranjitjhala:
Render trace using `NestedString` (a JSON representation) that lets the VSCode plugin display nested objects (e.g. `struct` with many fields) interactively

nested-vscode

@nilehmann any clue on how to avoid the pesky

```
warning: the feature `fmt_internals` is internal to the compiler or standard library
  --> crates/flux-middle/src/lib.rs:14:5
   |
14 |     fmt_internals
   |     ^^^^^^^^^^^^^
   |
   = note: using it is strongly discouraged
   = note: `#[warn(internal_features)]` on by default
```

which is needed in `nested_with_bound_vars`?


nilehmann:
@ranjitjhala To deal with the internal features thing we can make `fmt_bound_vars` take an `&mut impl fmt::Write` instead of `&mut Formatter`. Both `Formatter` and `String` implement `fmt::Write`.

PR #940, opened 2024-12-12
Fix parsing of angle brackets for sorts and bit vectors

### LINKED ISSUE #938: `Map<int, bitvec<32>>` does not parse
I would like to define a flux definition:
```Rust
flux_rs::defs! {
    fn bv_map_get(m: Map<int, bitvec<32>>, k:int) -> bitvec<32> { map_select(m, k) }
}
```

but it fails with the error:
```
error[E0999]: syntax error: unexpected token
  --> src/main.rs:50:37
   |
50 |     fn map_get(m: Map<int, bitvec<32>>, k:int) -> bitvec<32> { map_select(m, k) }
   |                                     ^
```

nilehmann:
Fixes #938

### DIFF
--- a/crates/flux-syntax/src/grammar.lalrpop
+++ b/crates/flux-syntax/src/grammar.lalrpop
@@ -160,14 +160,10 @@ ParamMode: surface::ParamMode = {
 }
 
 BaseSort: surface::BaseSort = {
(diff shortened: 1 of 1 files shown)


nilehmann:
ping @enjhnsn2

Issue #886, opened 2024-11-15
Add support for const promotion


ranjitjhala:
Per @nilehmann -- that is what will let us check stuff like the below

https://github.com/rust-lang/const-eval/blob/master/promotion.md

```rust
#[flux::sig(fn (a: &[&[i32][3]]))]
fn check_slice(_: &[&[i32]]) {}

fn test_slice1() {
    let x = &[1, 2, 3];
    check_slice(&[x]); // VERIFIES
}

fn test_slice2() {
    check_slice(&[&[1, 2, 3]]); // REJECTED because of const promotion?
}
```

nilehmann:
Some thoughts on this.

The mir for `test_slice2` looks like this

```rust
const test_slice2::promoted: &[&[i32]; 1] = {
    let mut _0: &[&[i32]; 1];
    let mut _1: [&[i32]; 1];
    let mut _2: &[i32];
    let mut _3: &[i32; 3];
    let mut _4: &[i32; 3];
    let mut _5: [i32; 3];

    bb0: {
        _5 = [const 1_i32, const 2_i32, const 3_i32];
        _4 = &_5;
        _3 = &(*_4);
        _2 = move _3 as &[i32] (PointerCoercion(Unsize));
        _1 = [move _2];
        _0 = &_1;
        return;
    }
}

fn test_slice2() -> () {
    let mut _0: ();
    let _1: ();
    let mut _2: &[&[i32]];
    let _3: &[&[i32]; 1];

    bb0: {
        _3 = const test_slice2::promoted;
        _2 = copy _3 as &[&[i32]] (PointerCoercion(Unsize));
        _1 = check_slice(move _2) -> [return: bb1, unwind continue];
    }

    bb1: {
        return;
    }
}
```
where `test_slice2::promoted` is an automatically generated constant with its own mir body.

I see two options to handle this. We could 
1. _Inline_ the definition of the promoted constant, that's it, jump into the promoted body while checking the main function and then back.
2. Compute the _strongest type_ for the promoted body.

ranjitjhala:
Perhaps related to #1047 ?

ranjitjhala:
I *think* this may also be related to this const-promotion issue?

```rust
use flux_rs::{assert, attrs::*};
extern crate flux_core;

#[reflect]
#[derive(PartialEq, Eq)]
pub enum State {
    On,
    Off,
}
flux_core::eq!(State); // see the `specs` PR

pub fn test_eq() {
[...]

PR #941, opened 2024-12-12
Basic support for non-integral `const` (e.g. bitvector)

ranjitjhala:
This implements STEP 1 of #935 (towards fixing #932)


nilehmann (changes requested):
I have some concerns about the way errors are reported. Right now we do "const reflection" during encoding into fixpoint. This triggers the `constant_info` query that will report an error at the constant's definition site if it's not annotated. This is problematic for a couple of reasons:

1. We can't point to the site where the user uses the constant when reporting the error. The best we can do is point at the function that generates the constraint with the constant.
2. We only report an error if the constant ends up in the constraint, which can be confusing, e.g.
    ```
    #[flux::opaque]
    #[flux::refined_by(x: int)]
    struct S {}

    const C: S = S {};
    
    #[flux::sig(fn(x: S[C]))]  // we don't produce an error if only
    fn foo(x: S) { }           // this function is defined
    
    fn baz(x: S) { // we only get an error after defining this function
        foo(x);    // and we only point you to `C`
    }
    ```
    
I think we should do the reflection during `conv` so we can point to the use site of the constant. The issue though, is that this can cause cycles if the annotation of a constant uses another constant. However, this is already the case with the implementation in this PR. It doesn't surface as cycles, because you are not expanding constants recursively, but as "weird" behavior, e.g, we can't prove the following

```rust
#[flux::opaque]
#[flux::refined_by(x: int)]
struct S {}

#[flux::constant(0)]
const A: S = S {};

#[flux::constant(A)]
const B: S = S {};

#[flux::sig(fn(x: S[B]) -> S[0])]
fn foo(x: S) -> S {
    x
}
```

vrindisbacher:
hmm @ranjitjhala, this still seems to trip up on annotations like this:

```rust
pub const START: BV32 = BV32::new(0xFFFF_FFFF);

#[sig(fn (BV32[@addr]) -> bool[addr == START])]
pub fn test_start_eq(address: BV32) -> bool {
    address == START 
}
```

```rust
   |
   |     address == START
   |                ^^^^^^^^^ a postcondition cannot be proved
   |
   |
   | #[sig(fn (BV32[@addr]) -> bool[addr == START])]
   |                                ^^^^^^^^^^^^^
```

ranjitjhala:
Looks like this new test generates `mir` like

```
fn test2(_1: BV32) -> bool {
    debug address => _1;
    let mut _0: bool;
    let mut _2: &BV32;
    let mut _3: &BV32;
    let _4: BV32;
    let mut _5: &BV32;

    bb0: {
        StorageLive(_2);
        _2 = &_1;
        StorageLive(_3);
        _5 = const test2::promoted[0];
        _3 = &(*_5);
        _0 = <BV32 as std::cmp::PartialEq>::eq(move _2, move _3) -> [return: bb1, unwind: bb2];
    }
```

which maybe means this has to do with that **promotion** business @nilehmann UGH.

https://github.com/flux-rs/flux/issues/886

Maybe this last bit will have to wait for STEP 3.

nilehmann:
Oh yeah, annoying. `==` goes through  `PartialEq` so `address == START ` is equivalent to `PartialEq::eq(&address, &START)` which causes `&START` to be const promoted. The workaround without implementing support for const promotion is to write it as

```
#[sig(fn (BV32[@addr]) -> bool[addr == START])]
pub fn test_start_eq(address: BV32) -> bool {
    let s = START;
    address == s 
}
```

vrindisbacher:
@nilehmann's tmp fix works... it was pretty annoying to do but it definitely unblocked me - checking annotations that used to hang seemingly infinitely now check in a few seconds!

ranjitjhala:
Wow. Surprising that

1. That’s why the promotion happens AND
2. One can work around it with a local AND
3. That’s it mitigates the bitvec issues!!!

- Ranjit.

vrindisbacher:
I may have been over-optimistic. It definitely helped a lot with some functions but there are still some others that take forever. I'll open an issue for these.

vrindisbacher:
Opened an issue here: #944. I don't think it's bit vector related though - it seems like this helped a ton. For example `ldmia_w` would also take forever to check until I converted everything to bitvectors.

ranjitjhala:
Ok I think this is ready to go. I'm not happy with the "HACK" which is a very special case where the `const` has generic args, but maybe that's a fight for another day.

nilehmann:
I probably going to take a look at this tomorrow

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1726,7 +1754,17 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
             fhir::ExprKind::Var(var, _) => {
                 match var.res {
                     ExprRes::Param(..) => env.lookup(var).to_expr(),
-                    ExprRes::Const(def_id) => rty::Expr::const_def_id(def_id).at(espan),
+                    ExprRes::Const(def_id) => {
+                        if P::HAS_ELABORATED_INFORMATION {
+                            let info = self.genv().constant_info(def_id)?;
+                            rty::Expr::const_def_id(def_id, info).at(espan)
```
nilehmann, line 1760: I'm not 100% sure, because both behaviors make sense and it's mostly about defaults, but  maybe we should report an error here for uninterpreted consts, e.g., in this case:


```rust
#[opaque]
#[refined_by(x: int)]
struct S { ... }

const SOME_S: S = ...

#[sig(fn() -> S[SOME_S])]
fn foo() -> S { ... }
```

```
#[sig(fn() -> S[SOME_S])]
                ^^^^^^
                sorry we can't deal with this constant
```

More generally, `ConstInfo::Uninterpreted` could be interpreted (pun intended) in two different ways

1. Sorry we can't deal with this so don't bother trying
2. Encode this constant with an uninterpreted (opaque) constant in smt

I think 1. by default is less surprising (or more discoverable) and the user should opt-in to the second behavior by annotating the const. You can pick 2. to circumvent these limitations _or_ to hide implementation details even if we support the constant.

Note that for 2. to make sense, we need to give `SOME_S` (when used in the mir) type `SOME_S[SOME_S]` and the behavior right now is the worst of both worlds because we encode `SOME_S` as an uninterpreted constant in smt but we don't give the rust const a useful type (we give it `∃v. SOME_S[v]`)

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1269,31 +1270,26 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
-                        let val = val.try_to_scalar_int()?;
-                        rty::Constant::from_scalar_int(self.genv.tcx(), val, &ty)
-                    });
-
+                // generate and insert to `const_map`
+                let val = match info {
+                    rty::ConstantInfo::Uninterpreted => None,
+                    rty::ConstantInfo::Interpreted(expr) => Some(expr.clone()),
+                };
+                let sort = self.genv.sort_of_def_id(def_id).unwrap().unwrap();
```
nilehmann, line 1288: Maybe the code would look cleaner if we only record the `DefId` here and then deal with everything in `assume_const_values` by calling `genv.conf_info` and `genv.sort_of_def_id`. But I'm conflicted, I also like the idea of having the `rty::ConstantInfo` stored inside `ExprKind::ConstDefId`.

crates/flux-middle/src/fhir.rs
```diff
@@ -155,6 +155,7 @@ pub enum Node<'fhir> {
     TraitItem(&'fhir TraitItem<'fhir>),
     ImplItem(&'fhir ImplItem<'fhir>),
     OpaqueTy(&'fhir OpaqueTy<'fhir>),
+    AnonConst(),
```
nilehmann, line 158:
```suggestion
    AnonConst,
```

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -137,6 +138,24 @@ fn adt_def(genv: GlobalEnv, def_id: LocalDefId) -> QueryResult<rty::AdtDef> {
...
+    {
+        constant
+    } else {
+        &fhir::ConstantInfo { owner, expr: None }
+    };
+    let Some(sort) = genv.sort_of_def_id(owner.def_id.to_def_id()).emit(&genv)? else {
+        return Ok(rty::ConstantInfo::Uninterpreted);
+    };
+    let wfckresults = wf::check_constant(genv, owner, constant, &sort)?;
+    conv::conv_constant(genv, local_def_id.to_def_id(), constant, &wfckresults)
```
nilehmann, line 156: I'd make this more explicit about what stuff we support (I'm assuming the expr is directly stored in `ItemKind::Const`)

```suggestion
    let def_id = genv.maybe_extern_id(local_def_id);
    let node = genv.map().node(def_id.local_id())?;
    match node {
        fhir::Node::Item { kind: ItemKind::Const(Some(expr)) } => {
            let Some(sort) = genv.sort_of_def_id(owner.def_id.to_def_id()).emit(&genv)? else {
                return Ok(rty::ConstantInfo::Uninterpreted);
            };
            let wfckresults = wf::check_constant(genv, owner, expr, &sort)?;
            conv::conv_constant(genv, local_def_id.to_def_id(), expr, &wfckresults)
        }
        fhir::Node::TraitItem { kind: TraitItemKind::Const, .. } |
        fhir::Node::ImplItem { kind: ImplItemKind::Const, .. } |
        fhir::Node::Item { kind: ItemKind::Const(None) } |
        fhir::Node::AnonConst => {
            rty::ConstantInfo::Uninterpreted
        }
```

Note that if we have the expr directly in `ItemKind::Const`, `wf::check_constant` and `conv::conv_constant` don't make much sense and we can do everything directly on the expr.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -966,6 +966,14 @@ pub enum SortArg {
     BvSize(BvSize),
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash, TyEncodable, TyDecodable)]
+pub enum ConstantInfo {
+    /// An uninterpreted constant
+    Uninterpreted,
+    /// A non-integral constant whose value is specified by the user
+    Interpreted(Expr),
```
nilehmann, line 974: It may make sense to store the `sort` here because we need it when constructing it for sort checking and then we need the sort when using it.

```suggestion
    Interpreted(Expr, Sort),
```

nilehmann (changes requested):
@ranjitjhala this look in the right direction, but left some suggestions and concerns

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: Yes, I was not clear as to which is better, so lets leave this as is for now till a clearer picture emerges..

crates/flux-desugar/src/desugar.rs
```diff
@@ -470,6 +470,53 @@ impl<'a, 'genv, 'tcx: 'genv> RustItemCtxt<'a, 'genv, 'tcx> {
...
+        &mut self,
+        def_id: LocalDefId,
+        const_info: &surface::ConstantInfo,
+    ) -> Result<fhir::TraitItem<'genv>> {
+        let constant_info = self.desugar_const_spec(def_id, const_info)?;
+        let owner_id = self.owner;
+        let generics = fhir::Generics::trivial();
+        let kind = fhir::TraitItemKind::Const(constant_info);
+        Ok(fhir::TraitItem { owner_id, generics, kind })
+    }
```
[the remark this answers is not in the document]
ranjitjhala: Const in trait seems straightforward, what is the issue? I could add tests, but I think it should "just work" ....

Impl seems hairier because of the usual resolution issues -- you probably want an "associated constant" but whats the point even of that if you can't at compile time resolve the impl --> means you can't actually know what the `const` was either?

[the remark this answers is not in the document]
ranjitjhala: For `impl` actually it makes sense to have them _if_ you only use them locally (i.e. not a `pub`) or I guess somehow at each usage site `rustc` can resolve it to one exact impl (otherwise, presumably it complains...?)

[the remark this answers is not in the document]
ranjitjhala: NVM made `None` for now!

crates/flux-fhir-analysis/src/lib.rs
ranjitjhala: Hmm, the above doesn't seem to be valid rust syntax -- the hassle is that `Item(ItemKind)` and `TraitItem(TraitItemKind)` and `ImplItem(ImplItemKind)` and I can't figure out the syntax for matching on the inner (X-Kind)'s `kind` field?

crates/flux-desugar/src/desugar.rs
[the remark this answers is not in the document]
nilehmann: Constants in traits are abstract. Implementations define them. For it to work it needs to go through normalization as we do for associated types.

Constants in inherent implementations are like item constants with the difference that they can mention generics in the impl block.

ranjitjhala:
Ok I think I did the lot, except for 

> more explicit about what stuff we support

tests/tests/todo/bitvec_const02.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[flux::trusted]
+    #[flux::sig(fn (x:u32) -> BV32[bv_int_to_bv32(x)])]
+    const fn new(val: u32) -> Self {
+        BV32(val)
+    }
+}
+
+pub trait Foo {
+    #[flux_rs::constant(bv_int_to_bv32(0x4567))]
+    const START: BV32 = BV32::new(0x4567);
```
nilehmann, line 16: My point is that the behavior you get from this annotation will be wrong if not done carefully because you can override the const in an implementation

```rust
impl Foo for Sometype {
    const START: BV32 = BV32::new(0x1234);
 }
 ```
 
 So the usual trait shenanigans apply, i.e., in the trait definition we need to treat the const abstractly and then find the concrete one by normalizing projections.
 
 I think we can make sense of the annotation by considering it only for implementations that don't override the definition (i.e., they use the default)

**Review by nilehmann — COMMENTED** [review]
@ranjitjhala the following works for me (after removing

```rust
    match node {
        fhir::Node::Item(fhir::Item { kind: fhir::ItemKind::Const(Some(expr)), .. }) => {
            todo!() // for these constants we use the expression
        }
        fhir::Node::AnonConst
        | fhir::Node::TraitItem(fhir::TraitItem { kind: fhir::TraitItemKind::Const, .. })
        | fhir::Node::ImplItem(fhir::ImplItem { kind: fhir::ImplItemKind::Const, .. }) => {
            todo!() //  these constants we try to evaluate if they are integral and return uninterpreted if we fail
        }
        _ => {
            todo!()
        }
    }
```

tests/tests/todo/bitvec_const02.rs
ranjitjhala: yup indeed.

crates/flux-fhir-analysis/src/lib.rs
ranjitjhala: odd I had tried the thing you suggested ... maybe rust-analyzer had gotten into one of its funky states...

ranjitjhala:
ok I think i took care of those changes, pls see!

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -283,17 +283,27 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+            ExprRes::Param(_, id) => Ok(self.param_sort(id)),
             ExprRes::Const(def_id) => {
-                // TODO(nilehmann) generalize const sorts
-                let ty = self.genv.tcx().type_of(def_id).no_bound_vars().unwrap();
-                assert!(ty.is_integral());
-                rty::Sort::Int
+                if let Some(sort) = self
+                    .genv
+                    .sort_of_def_id(def_id)
+                    .unwrap_or_else(|_err| span_bug!(path.span, "unknown sort for `{def_id:?}`"))
```
nilehmann, line 293: This error should be emitted

```suggestion
                    .emit(&self.genv)?
```

nilehmann (approved):

nilehmann (approved):

PR #952, opened 2024-12-17
Add bit vec ord ops

vrindisbacher:


nilehmann (approved):
Can you add a test please

PR #957, opened 2024-12-20
Pretty printing fixes

### LINKED ISSUE #955: Issues with extension and pretty printer
I found a couple of issues with the extension when trying the following (note that you need to save without running rustfmt)

```rust
#[flux::sig(fn(c: Option<bool>) -> Option<i32{v:0 <= v}>)]
pub fn test0(c: Option<bool>) -> Option<i32> {
    c.map(|b| 
        if b { 
            1 
        } else { 
            2 
        }
    )
}
```

## Type with no indices has confusing formatting

As shown in the image, `Option` is pretty printed as `Option<λb0. bool[b0]>[{..}]` . I find the ellipsis confusing because `Option` has no indices so there's nothing to expand (or rather it is indexed by a unit sort). I think we should show it as `Option<λb0. bool[b0]>` (I'm not worried about the lambda here, but we can discuss that)

image

## The extension shows trace data from the shape phase

As shown below, the extension shows a hole (i.e.,  `*`) as part of the constraints which indicates it is not filtering out the trace data coming from the shape phase (related to https://github.com/flux-rs/flux/issues/916)

image

ranjitjhala:
Fixes #955 

Screenshot 2024-12-18 at 8 55 39 PM

Screenshot 2024-12-18 at 8 55 27 PM

### DIFF
--- a/crates/flux-refineck/src/type_env.rs
+++ b/crates/flux-refineck/src/type_env.rs
@@ -929,12 +929,6 @@ impl TypeEnvTrace {
                 let local = loc_info(loc);
                 let kind = format!("{:?}", binding.kind);
                 let ty = binding.ty.nested_string(&cx);
-                // println!(
-                //     "TRACE: type-env-trace {:#?} ==> {:?} ==> {}",
-                //     binding.ty,
-                //     binding.ty.kind(),
-                //     ty
-                // );
                 let span = loc_span(genv, local_decls, loc);
                 bindings.push(TypeEnvBind { name, local, kind, ty, span });
             });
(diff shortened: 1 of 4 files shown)


crates/flux-refineck/src/type_env.rs
```diff
@@ -929,6 +929,12 @@ impl TypeEnvTrace {
                 let local = loc_info(loc);
                 let kind = format!("{:?}", binding.kind);
                 let ty = binding.ty.nested_string(&cx);
+                // println!(
+                //     "TRACE: type-env-trace {:#?} ==> {:?} ==> {}",
+                //     binding.ty,
+                //     binding.ty.kind(),
+                //     ty
+                // );
```
nilehmann, line 937:
```suggestion
```

nilehmann (approved):

PR #988, opened 2025-02-07
Add support to reflect (simple) enums

ranjitjhala:
So stuff like this works now

```rust
#[flux::reflect]
pub enum State {
    On,
    Off,
}

#[flux_rs::sig(fn () -> State[On])]
pub fn test00() -> State {
    State::On
}

#[flux_rs::sig(fn () -> State[Off])]
pub fn test01() -> State {
    State::Off
}

#[flux::sig(fn () -> State[Off])]
pub fn test02() -> State {
    State::On //~ ERROR refinement type
}

#[flux::sig(fn (State[On]) -> usize[1])]
pub fn test03(s: State) -> usize {
    match s {
        State::On => 1,
        State::Off => 0,
    }
}
```


nilehmann (changes requested):
I think at the level of `rty` we should have

crates/flux-middle/src/rty/mod.rs
```diff
@@ -67,6 +67,60 @@ use crate::{
 #[derive(Debug, Clone, Eq, PartialEq, Hash, TyEncodable, TyDecodable)]
 pub struct AdtSortDef(Interned<AdtSortDefData>);
 
+#[derive(Debug, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
+pub struct AdtSortRefined {
```
nilehmann, line 71: Maybe we can call this a `SortVariant` and make `AdtSortDef` just a list of these.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -183,6 +205,10 @@ impl<'de> Deserialize<'de> for TagIdx {
 struct SortEncodingCtxt {
     /// Set of all the tuple arities that need to be defined
     tuples: UnordSet<usize>,
+    /// Generator for fresh Reflected Datatypes
+    refl_data: IndexGen<ReflData>,
+    /// Set of all the reflected ADT (enum) definitions that need to be declared as Fixpoint data-decls
+    refl_decls: FxHashMap<DefId, ReflData>,
```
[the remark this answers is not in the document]
ranjitjhala: haha I _wanted_ to make it an `UnordMap` but couldn't for the life of me figure out how to loop (`iter`) over the `ReflData` ! :-)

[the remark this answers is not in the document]
nilehmann: Haha that's the point of `UnordMap` making it hard to expose the order. 

On second thought though, I think def ids cannot be stably sorted, so making it an UnoreMap won't work. But it just occurred to be that we can represent both the mapping and the index generator as an `FxIndexSet`. The next counter is just the next index in the se. Tgat would give is (1) a bidirectional mapping between index and DefId and (2) deterministic order (based on the order of insertion).

crates/flux-middle/src/rty/expr.rs
```diff
@@ -635,6 +661,9 @@ pub enum ExprKind {
     IfThenElse(Expr, Expr, Expr),
     KVar(KVar),
     Alias(AliasReft, List<Expr>),
+    /// A variant used in the logic to represent a variant of an ADT as a pair of the `DefId` and variant-index
+    /// TODO: extend to `Variant(DefId, List<Expr>)` to allow args...
+    Variant(DefId),
```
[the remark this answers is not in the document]
ranjitjhala: done

crates/flux-infer/src/fixpoint_encoding.rs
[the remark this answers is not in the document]
ranjitjhala: @nilehmann -- ok check if this is what you had in mind -- I'm not using a `newtype`; just a struct/wrapper around the `usize`.

ranjitjhala:
Ok I think I took care of these things *except* for the "deviates from rust behavior" remark (i.e. need to support `State::On` instead of just `On`).

Also see if the `pub struct ReflData` is what you had in mind.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -211,6 +238,16 @@ impl SortEncodingCtxt {
                 let args = args.iter().map(|s| self.sort_to_fixpoint(s)).collect_vec();
                 fixpoint::Sort::App(fixpoint::SortCtor::Map, args)
             }
+
+            rty::Sort::App(rty::SortCtor::Adt(sort_def), args) if sort_def.is_reflected() => {
```
nilehmann, line 242: I think this would be cleaner if we didn't care about reflection and only branch on whether this is a struct.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -201,14 +201,14 @@ impl Expr {
...
-        }
-    }
+    // #[track_caller]
+    // pub fn expect_adt(&self) -> (DefId, List<Expr>) {
+    //     if let ExprKind::Ctor(def_id, _, flds) = self.kind() {
+    //         (*def_id, flds.clone())
+    //     } else {
+    //         bug!("expected record, found {self:?}")
+    //     }
+    // }
```
nilehmann, line 211:
```suggestion
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -242,6 +242,10 @@ impl Expr {
         ExprKind::Constant(c).intern()
     }
 
+    // CUT pub fn variant(did: DefId) -> Expr {
+    // CUT     ExprKind::Variant(did).intern()
+    // CUT }
+
```
nilehmann, line 248:
```suggestion
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -308,7 +316,7 @@ impl Expr {
     }
 
     pub fn unit_adt(def_id: DefId) -> Expr {
```
nilehmann, line 318: Maybe rename this `unit_struct`

crates/flux-middle/src/rty/expr.rs
```diff
@@ -620,6 +631,29 @@ pub enum UnOp {
     Neg,
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
+pub enum Ctor {
+    /// for `refined_by` types
```
nilehmann, line 636: I'd prefer to avoid the distinction `refined_by` vs `reflected` and use `struct` vs `enum`.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1404,10 +1455,11 @@ pub(crate) mod pretty {
 
     pub fn aggregate_nested(
         cx: &PrettyCx,
-        def_id: DefId,
+        ctor: &Ctor, // HEREHEREHERE:reflect-adt: reflected enums with multiple fields...
```
nilehmann, line 1458:
```suggestion
        ctor: &Ctor,
```

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Hmm not following here -- we also use `refined_by` for `enum` (e.g. for measures?)

nilehmann: I want to be able to support the following

```rust
#[reflect]
struct Range<Idx> {
    pub start: Idx,
    pub end: Idx,
}
```

This should reflect the type as a struct, so it won't be true that reflected equals enum.

ranjitjhala:
Ok I think I've dealt with the various things -- except for the https://github.com/flux-rs/flux/pull/988#discussion_r1964121356

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Ok have tweaked the comment; still want to reflect (ha!) the fact that `enum` is only currently used for reflected things... [ `Struct` also means you do all those projection/packing shenanigans inside `flux` instead of leaving it to fixpoint/smt -- of course you _could_ leave it all to fixpoint for reflected-structs but that would be a bit of a change... ]

lib/flux-attrs-impl/src/lib.rs
```diff
@@ -11,6 +11,7 @@ pub const FLUX_ATTRS: &[&str] = &[
     "generics",
     "invariant",
     "opaque",
+    "reflect",
```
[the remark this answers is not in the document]
ranjitjhala: Ok I did some copy-pasta using `opaque` pls double check!

ranjitjhala:
oops sorry I think I got them all now! (but see the attrs thing...)

lib/liquid-fixpoint/src/format.rs
```diff
@@ -313,8 +313,7 @@ impl<T: Types> fmt::Display for Expr<T> {
             Expr::Atom(rel, exprs) => {
                 let [e1, e2] = &**exprs;
                 write!(f, "({rel} {e1} {e2})")
-            }
-            Expr::Variant(sort, pos) => write!(f, "{}_{pos}", sort.display()),
+            } // Expr::Variant(sort, pos) => write!(f, "{}_{pos}", sort.display()),
```
nilehmann, line 316:
```suggestion
            }
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -78,57 +126,67 @@ struct AdtSortDefData {
...
-    /// [sort variables]: Sort::Var
-    sorts: List<Sort>,
+    // The `index` is *either* the user defined indices or the "reflected" ADT
+    // kind: RefinementKind,
+    /// A vec of variants of the ADT;
+    /// - a `struct` sort -- used for types with a `refined_by` has a single variant;
+    /// - a `reflected` sort -- used for `reflected` enums have multiple variants
+    variants: Vec<AdtSortVariant>,
+    reflected: bool,
+    strukt: bool,
```
nilehmann, line 136: maybe call this `is_struct`

nilehmann (approved):
Ok looks good, a couple of small comments then ship it!

PR #989, opened 2025-02-08
add lifting for foreignitem

Zn2FePO424H2O:
I add the lift for foreign item and use it in desugar.


crates/flux-middle/src/fhir.rs
```diff
@@ -280,6 +290,22 @@ pub enum FluxItem<'fhir> {
...
+    pub span: Span,
+    pub vis_span: Span,
+}
+
+#[derive(Debug)]
+pub enum ForeignItemKind<'fhir> {
+    Fn(FnSig<'fhir>),
+    Static,
+    Type,
+}
```
nilehmann, line 307: Let's only have `Fn` as a variant here so we don't have to care about the other cases. We should also add the `Generics` part of it such that we can return it in `OwnerNode::generics`

```suggestion
pub enum ForeignItemKind<'fhir> {
    Fn(FnSig<'fhir>, Generics<'fhir>),
}
```

crates/flux-middle/src/fhir.rs
```diff
@@ -280,6 +294,22 @@ pub enum FluxItem<'fhir> {
...
+    pub owner_id: MaybeExternId<OwnerId>,
+    pub span: Span,
+    pub vis_span: Span,
+}
+
+#[derive(Debug)]
+pub enum ForeignItemKind<'fhir> {
+    Fn(FnSig<'fhir>, &'fhir Generics<'fhir>),
+    Static,
+    Type,
```
nilehmann, line 310: Let's remove these variants since we don't yet support them

```suggestion
```

crates/flux-desugar/src/lib.rs
```diff
@@ -293,10 +302,10 @@ impl CrateDesugar<'_, '_> {
         for item in specs.flux_items_by_parent.values().flatten() {
             match item {
                 surface::Item::Qualifier(qual) => {
-                    self.desugar_qualifier(qual).collect_err(&mut self.err);
+                    self.desugar_qualifier(&qual).collect_err(&mut self.err);
```
nilehmann, line 305: This is redundant `qual` already is a reference

```suggestion
                    self.desugar_qualifier(qual).collect_err(&mut self.err);
```

crates/flux-desugar/src/lib.rs
```diff
@@ -293,10 +302,10 @@ impl CrateDesugar<'_, '_> {
...
                 }
                 surface::Item::FuncDef(defn) => {
-                    self.desugar_func_defn(defn).collect_err(&mut self.err);
+                    self.desugar_func_defn(&defn).collect_err(&mut self.err);
```
nilehmann, line 308: Same here
```suggestion
                    self.desugar_func_defn(defn).collect_err(&mut self.err);
```

crates/flux-middle/src/fhir/visit.rs
```diff
@@ -281,6 +286,8 @@ pub fn walk_impl_item<'v, V: Visitor<'v>>(vis: &mut V, impl_item: &ImplItem<'v>)
     }
 }
 
+pub fn walk_foreign_item<'v, V: Visitor<'v>>(_: &mut V, _: &ForeignItem<'v>) {}
```
nilehmann, line 289: This file implements a [visitor](https://en.wikipedia.org/wiki/Visitor_pattern) over the fhir tree and clients of the visitor expect that all nodes will be visited. So this function should *walk* the structure of the foreign item the same way the other walking functions do.

nilehmann (changes requested):
Looking great @Zn2FePO424H2O! I left some comments, let me know if you have any questions.

Issue #996, opened 2025-02-20
Add basic support for extern types


nilehmann:
Add enough support to handle this

```rust
#![feature(extern_types)]

extern "C" {
    type A;

    fn create() -> *const A;
    fn take(x: *const A);
}

fn test() {
    unsafe {
        let x = create();
        take(x);
    }
}
```

nilehmann:
Some mentoring

We need to add a new type variant for foreign types. Foreign types are defined in rustc here https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/type.TyKind.html#variant.Foreign. These types are completely opaque, so no major changes should be needed other than adding them...

To implement this we should add a variant to `BaseTy::Foreign(DefId)` here https://github.com/flux-rs/flux/blob/c533702084f0d6a80edb1d039f64c31104027e0b/crates/flux-middle/src/rty/mod.rs#L153 

This will `missing variant` errors in various places and we should fill in the code in all those places.

nilehmann:
Fixed by #999

PR #999, opened 2025-02-22
Add handle to extern types 

Zn2FePO424H2O:
issue #996


crates/flux-middle/src/rty/pretty.rs
```diff
@@ -440,6 +440,9 @@ impl Pretty for BaseTy {
             BaseTy::Infer(ty_vid) => {
                 w!(cx, f, "{ty_vid:?}")
             }
+            BaseTy::Foreign(def_id) => {
+                w!(cx, f, "{def_id:?}")
```
nilehmann, line 444: When using `w!` variables cannot be inlined, otherwise we won't use the `Pretty` implementation

```suggestion
                w!(cx, f, "{:?}", def_id)
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1680,7 +1680,12 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                     )
                 }
             }
-            fhir::Res::Def(kind, def_id) => self.report_expected_type(path.span, kind, def_id)?,
+            fhir::Res::Def(kind, def_id) => {
+                match kind {
+                    DefKind::ForeignTy => flux_middle::rty::BaseTy::Foreign(def_id),
+                    _ => self.report_expected_type(path.span, kind, def_id)?,
+                }
+            }
```
nilehmann, line 1688: We need to check that the path has no generic arguments and raise an error if it does, similar to how we do for self-types. See, e.g., how we call `self.check_self_ty_generics`.

Also, you can inline the pattern on `DefKind::ForeignTy` as we do for the other cases.

```suggestion
            fhir::Res::Def(DefKind::ForeignTy, def_id) => rty::BaseTy::Foreign(def_id),
            fhir::Res::Def(kind, def_id) => self.report_expected_type(path.span, kind, def_id)?,
```

issue-996.rs
```diff
@@ -0,0 +1,15 @@
+#![feature(extern_types)]
```
nilehmann, line 1: This file should go in `tests/tests/pos/surface` so it's included in the test suite. Also, let's call it `foreign_type00.rs`

nilehmann (changes requested):

issue-996.rs
Zn2FePO424H2O: Because foreign is not fully supported, the file still raising unsupported errors.

crates/flux-middle/src/rty/pretty.rs
Zn2FePO424H2O: What is special about ty_vid such that it can be used inline?

issue-996.rs
nilehmann: What error are you getting? We should add enough support until this test passes.

Zn2FePO424H2O: error[E0999]: unsupported local declaration
  --> tests/tests/pos/surface/foreign_type00.rs:12:13
   |
12 |         let x = create();
   |             ^ this declaration has type `*const A` which is not currently supported

error[E0999]: unsupported signature
 --> tests/tests/pos/surface/foreign_type00.rs:6:5
  |
6 |     fn create() -> *const A;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: unsupported type `Foreign(DefId(0:4 ~ foreign_type00[1c71]::{extern#0}::A))`

error[E0999]: unsupported signature
 --> tests/tests/pos/surface/foreign_type00.rs:7:5
  |
7 |     fn take(x: *const A);
  |     ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: unsupported type `Foreign(DefId(0:4 ~ foreign_type00[1c71]::{extern#0}::A))`
  
  Should I support the foreign type, or let it stop reporting unsupported foreign types?

nilehmann (approved):
LGTM

PR #1000, opened 2025-02-23
Lazy `invariant` checks for `struct`

ranjitjhala:
This PR defers i.e. *lazily* does the `invariant` checks for `struct`s to places where the `struct` is *constructed* (as opposed to the checks for `enum` which are done *eagerly* at the type definition site itself.

This is because for `enum` e.g. `List[n]` the constructors themselves establish/maintain the invariant (e.g. len >= 0),
while for `struct` it really depends on the actual fields used (and its a bit awkward to duplicate the invariant onto a field
to have permit an "eager" check to pass.) 

Specifically, now the following works:

```rust
#[flux::invariant(N > 0)]
pub struct MPUGOOD<const N: usize> {
    field: i32,
}

pub fn bar<const N: usize>(x: usize, _mpu: MPUGOOD<N>) {
    let _x = x % N;
}

pub fn baz<const N: usize>() -> i32 {
    if N > 0 {
        let mpu = MPUGOOD::<N> { field: 12 };
        mpu.field
    } else {
        0
    }
}
```

@nilehmann see the `TODO(struct-inv)` note.

@vrindisbacher this should help with your demo?


vrindisbacher:
Yes absolutely helps - if we could get this merged in that would be great :)

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -539,6 +539,7 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
         enum_id: MaybeExternId,
         variant: &fhir::VariantDef,
         reflected: bool,
+        invariants: &[rty::Invariant],
```
nilehmann, line 542: This is only ever called with `&[]` so we don't need the extra parameter

```suggestion
```

vrindisbacher:
@ranjitjhala I was trying out this isolated example: 

```rust
#![allow(unused)]


#[flux_rs::invariant(MIN_REGION_SIZE > 0)]
pub struct MPU<const MIN_REGION_SIZE: usize> {
    reg1: u32,
    reg2: u32
}

impl<const MIN_REGION_SIZE: usize> MPU<MIN_REGION_SIZE> {
    fn allocate_region(
        &self,
        unallocated_memory_start: usize,
        unallocated_memory_size: usize,
        min_region_size: usize,
    ) {
        // Logical region
        let mut start = unallocated_memory_size;

        // Region start always has to align to minimum region size bytes
        if start % MIN_REGION_SIZE != 0 {
            // do stuff
        }

        // do some more stuff...
    }

}
```

but flux crashes:

```rust
    Checking flux-example v0.1.0 (/Users/vrindisbacher/Desktop/research/flux-example)
thread 'rustc' panicked at /Users/vrindisbacher/Desktop/research/flux/crates/flux-middle/src/rty/mod.rs:919:51:
invalid enum variant tag while decoding `Sort`, expected 0..15, actual 120
error: could not compile `flux-example` (bin "flux-example")

Caused by:
  process didn't exit successfully: `/Users/vrindisbacher/.flux/flux-driver --crate-name flux_example --edition=2021 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=111 --crate-type bin --emit=dep-info,metadata -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=01c720dbbe6d628c -C extra-filename=-01c720dbbe6d628c --out-dir /Users/vrindisbacher/Desktop/research/flux-example/target/flux/debug/deps -C incremental=/Users/vrindisbacher/Desktop/research/flux-example/target/flux/debug/incremental -L dependency=/Users/vrindisbacher/Desktop/research/flux-example/target/flux/debug/deps --extern flux_rs=/Users/vrindisbacher/Desktop/research/flux-example/target/flux/debug/deps/libflux_rs-fa3a4e6822fe5bb6.rmeta` (exit status: 101)
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -1195,6 +1195,7 @@ pub struct VariantSig {
     pub args: GenericArgs,
     pub fields: List<Ty>,
     pub idx: Expr,
+    pub invariants: Vec<Invariant>,
```
nilehmann, line 1198: Let's make this a list of expressions, i.e., store the invariants already applied. Also, let's call it `requires`, as in a list of predicates that need to be true to construct the variant.

```suggestion
    pub requires: List<Expr>,
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -558,11 +559,13 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
         } else {
             self.conv_expr(&mut env, &variant.ret.idx)?
         };
+        let invariants = invariants.to_vec();
```
nilehmann, line 562: We should apply the invariants here.

tests/tests/todo/struct_inv.rs
```diff
@@ -0,0 +1,2 @@
+
```
nilehmann, line 1: delete this file

crates/flux-infer/src/infer.rs
```diff
@@ -516,6 +516,8 @@ impl<'genv, 'tcx> InferCtxtAt<'_, '_, 'genv, 'tcx> {
                 this.subtyping(actual, formal, reason)?;
             }
 
+            // Check invariants
+            // TODO(struct-inv): check-invariants-here?
```
nilehmann, line 520: yes, we should also check the invariants here. Otherwise, I think we would fail to catch the following


```rust
use flux_rs::attrs::*;

#[refined_by(n: int)]
#[invariant(n > 0)])
struct A { 
    #[field(n)]
    f: i32 
}

fn baz(a: A) {
    flux_rs::assert(a.f > 0); // uh-oh
}

fn foo() {
    let mut a = A { f : 1 };
    a.f -= 1;  // this is fine, we don't need to preserve the invariant while unfolded.
    baz(a);   // force folding
}
```

nilehmann (changes requested):

nilehmann:
@vrindisbacher try `rm -rf target/`. Whenever we change a definition in `rty` that could invalidate any existing metadata file.

vrindisbacher:
@nilehmann I just realized this is definitely happening because I'm using flux in the project on `main` but this branch from the `cli`. Doh!

crates/flux-fhir-analysis/src/conv/mod.rs
ranjitjhala: Did you mean in `conv_struct_variant` (since we're just using the empty list for `enum`?)

nilehmann: ah yeah, I put the comment on the wrong code :P

ranjitjhala:
@nilehmann I think I dealt with the stuff above.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -2588,14 +2595,21 @@ impl EarlyBinder<PolyVariant> {
     pub fn to_poly_fn_sig(&self, field_idx: Option<crate::FieldIdx>) -> EarlyBinder<PolyFnSig> {
         self.as_ref().map(|poly_variant| {
             poly_variant.as_ref().map(|variant| {
+                // CUT-FIXME let ret = variant.ret();
+                // CUT-FIXME let TyKind::Indexed(_, ret_index) = ret.kind() else { bug!() };
```
nilehmann, line 2599:
```suggestion
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -2588,14 +2595,21 @@ impl EarlyBinder<PolyVariant> {
...
                     Some(i) => List::singleton(variant.fields[i.index()].clone()),
                 };
-                let requires = List::empty();
-                FnSig::new(Safety::Safe, abi::Abi::Rust, requires, inputs, output)
+
+                // CUT-FIXME let requires = variant
+                // CUT-FIXME     .requires
+                // CUT-FIXME     .iter()
+                // CUT-FIXME     .map(|inv| inv.apply(ret_index))
+                // CUT-FIXME     .collect_vec();
```
nilehmann, line 2611:
```suggestion
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -597,11 +598,20 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                     })
                     .collect(),
             );
+
+            let requires = adt_def
+                .invariants()
+                .iter()
+                .map(|inv| inv.apply(&idx))
+                .collect();
+            let requires = List::from_vec(requires);
```
nilehmann, line 607: I think you can collect directly into a `List`, haven't check if it works though.

```suggestion
            let requires = adt_def
                .invariants()
                .iter()
                .map(|inv| inv.apply(&idx))
                .collect();
```

nilehmann (approved):
LGTM

crates/flux-fhir-analysis/src/conv/mod.rs
ranjitjhala: indeed looks like that works, thanks!

ranjitjhala:
@vrindisbacher -- can you lmk if this works? does the `rm` thing solve your issue?

vrindisbacher:
It worked on this branch! I'll pull down main and give it a try later today but seems like it works.

nilehmann:
oh I just realized this is PR number #1000!

PR #1009, opened 2025-03-06
Put proc macros requirement upfront in Installation

cole-k:
This PR also fixes part of specs.md. I'm aware that it's WIP, but it is visible online.


nilehmann (approved):
LGTM

PR #1011, opened 2025-03-07
Infix bitvector operators

### LINKED ISSUE #1010: Infix bitvector operators
See 

https://github.com/flux-rs/flux/issues/1006#issuecomment-2706862621

Would be nice to support `&`, `|`, `<<` and `>>` as infix operators instead of requiring `bv_and`, `bv_or` etc.

ranjitjhala:
fixes #1010

### DIFF
--- a/crates/flux-fhir-analysis/locales/en-US.ftl
+++ b/crates/flux-fhir-analysis/locales/en-US.ftl
@@ -265,7 +265,7 @@ fhir_analysis_invalid_assoc_reft =
     associated refinement `{$name}` is not a member of trait `{$trait_}`
 
 fhir_analysis_invalid_bitvector_constant =
-    integer is not a {$sort} bit vector
+    integer is not a `{$sort}` bit vector
 
 fhir_analysis_refine_arg_mismatch =
     {$kind} takes {$expected} generic refinement {$expected ->
--- a/crates/flux-fhir-analysis/locales/en-US.ftl
+++ b/crates/flux-fhir-analysis/locales/en-US.ftl
@@ -265,7 +265,7 @@ fhir_analysis_invalid_assoc_reft =
     associated refinement `{$name}` is not a member of trait `{$trait_}`
 
 fhir_analysis_invalid_bitvector_constant =
-    integer is not a {$size} bit vector
+    integer is not a {$sort} bit vector
 
 fhir_analysis_refine_arg_mismatch =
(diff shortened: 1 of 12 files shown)


crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -394,9 +397,34 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
+            fhir::BinOp::BitAnd => "bv_and",
+            fhir::BinOp::BitOr => "bv_or",
+            fhir::BinOp::BitShl => "bv_shl",
+            fhir::BinOp::BitShr => "bv_lshr",
+            _ => span_bug!(expr.span, "unexpected operator in synth_bitvec_op"),
+        };
+        let name = crate::Symbol::intern(name);
+        let poly_fsort = flux_middle::THEORY_FUNCS.get(&name).unwrap().sort.clone();
+        let fsort = self.instantiate_func_sort(poly_fsort);
+        let args = vec![e1.clone(), e2.clone()];
```
nilehmann, line 424: Eventually, I want to move away from having bit vector operations as functions parametric on sizes and have dedicated rules for them. Smtlib is not really parametric on sizes, and we can't express the behavior of more complicated operations like `sign_extend` as function types. I'd prefer if we "inline" the function type as a rule here. But also I don't feel strongly about this.

crates/flux-fhir-analysis/locales/en-US.ftl
```diff
@@ -265,7 +265,7 @@ fhir_analysis_invalid_assoc_reft =
     associated refinement `{$name}` is not a member of trait `{$trait_}`
 
 fhir_analysis_invalid_bitvector_constant =
-    integer is not a {$size} bit vector
+    integer is not a {$sort} bit vector
```
nilehmann, line 268:
```suggestion
    integer is not a `{$sort}` bit vector
```

nilehmann (approved):
LGTM with that one caveat

crates/flux-fhir-analysis/src/wf/sortck.rs
ranjitjhala: interesting, but I think that's probably best left for a future PR. (We'd probably want to _deprecate_ the `bv_and` etc. functions too then...)

PR #1013, opened 2025-03-08
Add bitwuzla as a valid solver

vrindisbacher:
I want to add this solver to liquid fixpoint. See here: https://github.com/ucsd-progsys/liquid-fixpoint/pull/740

It would be nice if we could propogate this back to a flux option!


nilehmann:
I want some evidence on whether this actually helps before merging. Also, we already allow setting the solver per function. You can put a `flux_rs::opts(solver = "<solver>")` on a function, and it will only apply to that function.

PR #1030, opened 2025-03-16
Add support for bounded quantifiers

ranjitjhala:
That is, refinements of the form `exists i in 0 .. 5 { p(i) }` or `forall i in 0 .. 8 { q(j) }` 

So the below now works:

```rust
#![flux::defs {
    fn magic(xing:int, yonk:int) -> bool;

    fn magic_all(noob:int) -> bool {
        forall i in 0 .. 4 {
            magic(i, noob)
        }
    }

    fn magic_ex(n:int) -> bool {
        exists i in 0 .. 4 {
            i == n
        }
    }
}]

#[flux::trusted]
#[flux::sig(fn(x:i32, y:i32) ensures magic(x, y))]
pub fn do_magic(_x: i32, _y: i32) {}

// forall tests ----------------------------------------------------------------

#[flux::sig(fn({i32[@n] | magic_all(n)}) ensures magic(3, n))]
pub fn test_all_l(_x: i32) {}

#[flux::sig(fn(n:i32) ensures magic_all(n))]
pub fn test_all_r(n: i32) {
    do_magic(0, n);
    do_magic(1, n);
    do_magic(2, n);
    do_magic(3, n);
}

// exists tests ----------------------------------------------------------------
#[flux::sig(fn({i32[@n] | magic_ex(n)}) -> bool[true])]
pub fn test_exi_l(n: i32) -> bool {
    n == 0 || n == 1 || n == 2 || n == 3
}

#[flux::sig(fn(n:i32) -> bool[magic_ex(n)])]
pub fn test_exi_r(n: i32) -> bool {
    n == 0 || n == 1 || n == 2 || n == 3
}
```

@vrindisbacher can you see if this helps?


vrindisbacher:
I'll test this as soon asap! But I think this will help quite a bit.

vrindisbacher:
This works quite nicely for me!

crates/flux-syntax/src/surface.rs
```diff
@@ -507,6 +513,7 @@ pub enum ExprKind {
     Alias(AliasReft, Vec<Expr>),
     IfThenElse(Box<[Expr; 3]>),
     Constructor(Option<ExprPath>, Vec<ConstructorArg>),
+    BoundedQuant(QuantKind, RefineParam, usize, usize, Box<Expr>),
```
nilehmann, line 516: Let's use `std::ops::Range<usize>` for all these to clarify the intent. It may be slightly annoying because `Range<usize>` is not copy, so you'll have to insert some `.clone()`s but I think it'll be worth it.

```suggestion
    BoundedQuant(QuantKind, RefineParam, Range<usize>, Box<Expr>),
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -647,7 +653,13 @@ impl Ctor {
     }
 }
 
-#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, TyDecodable)]
+#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, Debug, TyDecodable)]
+pub enum QuantKind {
+    Exists,
+    Forall,
+}
```
nilehmann, line 660: Let's reuse the `fhir` one here. It's okay for `rty` to depend on `fhir` but not the other way around. You'll have to derive the necessary things on the `fhir` one. Also, derive `Encodable/Decodable` instead of `TyEncodable/TyDecodable`

```suggestion
```

crates/flux-fhir-analysis/src/wf/param_usage.rs
```diff
@@ -110,9 +110,10 @@ impl<'a, 'genv, 'tcx> ParamUsesChecker<'a, 'genv, 'tcx> {
                     self.errors.emit(InvalidParamPos::new(var.span, sort));
                 }
             }
-            fhir::ExprKind::Abs(_, body) => {
+            fhir::ExprKind::Abs(_, body) | fhir::ExprKind::BoundedQuant(_, _, _, _, body) => {
                 self.check_func_params_uses(body, true, is_top_level_var);
```
nilehmann, line 114: This would allow horn generic refinements in negative positions

```rust
use flux_rs::attrs::*;

#[sig(fn[hrn p: int -> bool](x: i32{ exists i in 0..10 { p(i) } }))]
fn foo(x: i32) { }
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1119,6 +1119,19 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
             | rty::ExprKind::ForAll(_) => {
                 span_bug!(self.def_span, "unexpected expr: `{expr:?}`")
             }
+            rty::ExprKind::BoundedQuant(kind, lo, hi, body) => {
+                let exprs = (*lo..*hi)
+                    .map(|i| {
+                        let arg = rty::Expr::constant(rty::Constant::from(i));
+                        body.replace_bound_reft(&arg)
+                    })
+                    .collect_vec();
```
nilehmann, line 1128: You don't need to collect into a vector

```suggestion
                    });
```

nilehmann (changes requested):

crates/flux-syntax/src/surface.rs
nilehmann: hmmm, then let's use `core::range::Range`. You'll have to add `#[feature(new_range_api)]`

crates/flux-fhir-analysis/src/wf/param_usage.rs
ranjitjhala: not sure I quite follow, but I replaced with the same flags as `Neg` ... ?

crates/flux-syntax/src/surface.rs
ranjitjhala: grr. `core::range::Range` does not implement `Encodable` ... ?

nilehmann: oh god, then disregard this comment...

crates/flux-fhir-analysis/src/wf/param_usage.rs
ranjitjhala: Yes, lets just stick with the latter for now!

nilehmann (approved):
LGTM

nilehmann:
I'll merge this because I have to put some changes on top

PR #1054, opened 2025-03-28
Use `define-fun` instead of inlining all `flux::def` functions

ranjitjhala:
* This is triggered with the flag `FLUX_SMT_DEFINE_FUN=1` 
* We only use this for monomorphic functions as SMTLIB does not support polymorphic functions...

(May in the future do some monomorphization in flux or fixpoint)


ranjitjhala:
@nilehmann -- I decided there was too much going on in one PR to gut the `eager-normalizing`, so I update the PR, so that

1. we still normalize, find cycles up front at the top, BUT
2. we lazily compute (per fixpoint query) the relevant definitions, by using the pre-normalized stuff.

Btw, occurs to me that its fine to do the (1) "eagerly" because it is a linear-time pass that is done once per crate (not per query).

It would be useful to memoize / cache the work currently in `fixpoint_encoding::define_funs` into a `query` that saves the transitive deps for each function, but that seems easy enough in a separate (small!) PR?

nilehmann:
> Btw, occurs to me that its fine to do the (1) "eagerly" because it is a linear-time pass that is done once per crate (not per query).

Yet Flux can't finish computing that step for veriasm. The eager inlining IS the problem. I don't have data to back this up, but with lazy inlining, there are higher chances that functions would be applied to concrete values that can be further reduced; it's more *call-by-value-ish*

ranjitjhala:
What’s the evidence that *eager* is the problem instead of just *inlining*.

my point is lazy/eager makes no difference - the *same* work is done in both cases. So if eager blows up now, then lazy will blow up but later.

At any rate how do I replicate the veriasm thing to see if this PR helps? @vrindisbacher

ranjitjhala:
(The thing you are describing is some fine-grained mix of inlining + partial evaluation / simplification to prune branches, which seems much more involved than just deferring normalization to the fixpoint-encoding…)

nilehmann:
I put my thoughts in https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/Lazy.20vs.20eager.20inlining/near/508864480

But yeah, I agree that it's a somewhat orthogonal issue and its own little project to do proper partial evaluation/inlining.

I think I'm fine with landing the approach in this PR. I'll take a closer look asap

ranjitjhala:
@nilehmann -- any idea why the tock CI is choking? It seems to work fine when I try locally?
https://github.com/flux-rs/flux/actions/runs/14207548895/job/39808456678?pr=1054

nilehmann:
Are you checking locally on latest vtock?

nilehmann:
Also, ci runs changes rebased on top of main, so make sure to merge main into your branch

vrindisbacher:
Sorry about this!

vrindisbacher:
I suppose the ci is expecting `z3` to be the solver? I can update the `flux.toml` in `vtock` but may be worth adding cvc5 to the CI?

ranjitjhala:
aha because the `flux.toml` wants cvc5 now!

vrindisbacher:
Ya exactly. I removed the `solver` from `flux.toml` and restarted the workflow.

ranjitjhala:
btw on `master` I'm now getting this error?

```

    --> kernel/src/process_standard.rs:1806:37
     |
1806 |           let breaks_and_size = match chip.mpu().allocate_app_memory_regions(
     |  _____________________________________^
1807 | |             remaining_mem_ptrs.start,
1808 | |             remaining_mem_ptrs.len,
1809 | |             min_total_memory_size,
...    |
1814 | |             &mut mpu_config,
1815 | |         ) {
     | |_________^ a precondition cannot be proved

warning: `kernel` (lib) generated 9 warnings
error: could not compile `kernel` (lib) due to 1 previous error; 9 warnings emitted
```

nilehmann:
@ranjitjhala if you merge flux master into your branch I think that error will go away

nilehmann:
> hmm wonder why that error is mysteriously _not_ showing up in the CI?

What runs in ci is not necessarily the same you run locally. In CI, changes are rebased on top of main.

PR #1061, opened 2025-04-02
Vscode interactive

ranjitjhala:
Misc improvements to the vscode-extension mode.

1. Add a toggle to make the checking "interactive" (i.e. not automatic) which is useful if a file takes a while to check!
2. Allow toggling individual bits of the display (values, constraints, types)
3. Parse the flux-checker-trace via streaming to not keel over on large files.


ranjitjhala:
@nilehmann only changes to the vscode -- i presume i can merge post CI greenlight?

PR #1062, opened 2025-04-03
Holes for generic args

### LINKED ISSUE #1056: use of const generics in `refined_by`
The following does not work:

```rust
#[flux_rs::refined_by(m: Map<int, int>)]
#[flux_rs::opaque]
pub struct S1<const N: usize> {
    arr: [usize; N]
}

const MY_N: usize = 10;

#[flux_rs::refined_by(s: S1)]
pub struct S2 {
    #[field(S1<MY_N>[s])]
    s1: S1<MY_N>
}
```

Flux reports the following: 

```rust
error[E0999]: cannot resolve `MY_N`
  --> src/lib.rs:13:16
   |
13 |     #[field(S1<MY_N>[s])]
```

Similarly, adding the const generic to `S2` doesn't help: 

```rust
#[flux_rs::refined_by(m: Map<int, int>)]
#[flux_rs::opaque]
pub struct S1<const N: usize> {
    arr: [usize; N]
}

#[flux_rs::refined_by(s: S1)]
pub struct S2<const M: usize> {
    #[field(S1<M>[s])]
    s1: S1<M>
}
```

Flux reports: 

```rust
error[E0999]: cannot resolve `M`
  --> src/lib.rs:11:16
   |
11 |     #[field(S1<M>[s])]
   |                ^
```

ranjitjhala:
fixes #1056 

attn: @nilehmann


nilehmann (changes requested):
The proper way to implement this is:

1. Add `fhir::GenericArg::Infer`
2. During desugaring, map `surface::GenericArgKind::Ty(ty)` where `ty.is_hole()` to `fhir::GenericArg::Infer`.
3. During conv, since we have the `rty::GenericParam`, we know whether is a const or ty, so we map it to `rty::GenericArg::Ty(TyKind::Infer(..))` or `rty::GenericArg::Const(Const::Infer(..))` accordingly

Nothing needs to happen in `struct_compat` because *const holes* are already being filled.

ranjitjhala:
Ok closing this one; instead look at #1063 which is indeed considerably simpler!

PR #1065, opened 2025-04-04
Use `NormalizeInfo` to Inline during Constraint Simplification

ranjitjhala:
So we don't expand out the "top"-level applications that appear in constraints.


ranjitjhala:
@nilehmann am taking the liberty of merging this one-line PR, lmk if there is an issue!

PR #1066, opened 2025-04-04
`hide` and `reveal` flux-definitions

ranjitjhala:
So you can treat function definitions as uninterpreted by default (`hide`) and 
locally `reveal` them only where needed to check particular functions.

See the  `hide00.rs` tests.

@nilehmann -- I'll try to wrangle the parser to use `#[hide]` as suggested, but that seems orthogonal to the rest of the PR.


crates/flux-syntax/src/lexer.rs
```diff
@@ -64,6 +64,7 @@ pub enum Token {
     Sort,
     Opaque,
     Local,
+    Hide,
```
nilehmann, line 67: I don't want to add unnecessary keywords

```suggestion
```

crates/flux-syntax/src/lexer.rs
```diff
@@ -136,6 +137,7 @@ impl Token {
             Token::Sort => "sort",
             Token::Opaque => "opaque",
             Token::Local => "local",
+            Token::Hide => "hide",
```
nilehmann, line 140:
```suggestion
```

crates/flux-syntax/src/lexer.rs
```diff
@@ -171,6 +173,7 @@ struct Symbols {
     sort: Symbol,
     opaque: Symbol,
     local: Symbol,
+    hide: Symbol,
```
nilehmann, line 176:
```suggestion
```

crates/flux-syntax/src/lexer.rs
```diff
@@ -197,6 +200,7 @@ impl<'t> Cursor<'t> {
                 bitvec: Symbol::intern("bitvec"),
                 opaque: Symbol::intern("opaque"),
                 local: Symbol::intern("local"),
+                hide: Symbol::intern("hide"),
```
nilehmann, line 203:
```suggestion
```

crates/flux-syntax/src/lexer.rs
```diff
@@ -281,6 +285,7 @@ impl<'t> Cursor<'t> {
             TokenKind::Ident(symb, _) if symb == self.symbs.sort => Token::Sort,
             TokenKind::Ident(symb, _) if symb == self.symbs.opaque => Token::Opaque,
             TokenKind::Ident(symb, _) if symb == self.symbs.local => Token::Local,
+            TokenKind::Ident(symb, _) if symb == self.symbs.hide => Token::Hide,
```
nilehmann, line 288:
```suggestion
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -93,7 +99,23 @@ fn parse_flux_item(cx: &mut ParseCtxt) -> ParseResult<Item> {
...
+}
+
+fn advance_if_pragma(cx: &mut ParseCtxt, tok: crate::lexer::Token) -> ParseResult<bool> {
+    let toks =
+        [Tok::Pound, Tok::OpenDelim(Delimiter::Bracket), tok, Tok::CloseDelim(Delimiter::Bracket)];
+    advance_if_many(cx, &toks)
+}
+
 fn parse_reft_func(cx: &mut ParseCtxt) -> ParseResult<SpecFunc> {
+    let hide = advance_if_pragma(cx, Tok::Hide)?;
```
nilehmann, line 118: Since we don't yet have other attributes, I'd keep this specialized

```suggestion
fn parse_hide_attr(cx: &mut ParseCtxt) -> ParseResult<bool> {
    if !cx.advance_if(Tok::Pound) {
        return Ok(false);
    }
    cx.expect(Tok::OpenDelim(Bracket))?;
    cx.expect("hide")?;
    cx.expect(Tok::CloseDelim(Bracket))?;
    return Ok(true)
}

fn parse_reft_func(cx: &mut ParseCtxt) -> ParseResult<SpecFunc> {
    let hide = parse_hide_attr(cx);
```

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -153,9 +153,11 @@ pub struct ConstDecl<T: Types> {
 #[derive_where(Hash)]
 pub struct FunDef<T: Types> {
     pub name: T::Var,
+    pub arity: usize,
     pub args: Vec<(T::Var, Sort<T>)>,
     pub out: Sort<T>,
-    pub body: Expr<T>,
+    // Optional body for uninterpreted ("hide") functions
+    pub body: Option<Expr<T>>,
```
nilehmann, line 160: We already have `ConstDecl` for `uif`. We should decide whether to emit a const or a fun def in `fixpoint_encoding`

crates/flux-middle/src/rty/normalize.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
 #[derive(TyEncodable, TyDecodable)]
 pub struct NormalizedDefns {
     krate: CrateNum,
-    defns: UnordMap<FluxId<DefIndex>, NormalizeInfo>,
+    pub defns: UnordMap<FluxId<DefIndex>, NormalizeInfo>,
```
nilehmann, line 23: I think you are not accessing this outside
```suggestion
    defns: UnordMap<FluxId<DefIndex>, NormalizeInfo>,
```

crates/flux-middle/src/global_env.rs
```diff
@@ -591,6 +596,14 @@ impl<'genv, 'tcx> Map<'genv, 'tcx> {
         }
     }
 
+    pub fn fn_reveals_for(self, def_id: LocalDefId) -> QueryResult<&'genv [FluxDefId]> {
```
nilehmann, line 599: You are not accessing this function outside.

```suggestion
    fn fn_reveals_for(self, def_id: LocalDefId) -> QueryResult<&'genv [FluxDefId]> {
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1567,33 +1552,44 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                    (0, fixpoint::Sort::Int)
                 });
+
             let info = self.genv.normalized_info(did);
             for dep in rty::normalize::local_deps(&info.body) {
                 wkl.push(dep.to_def_id());
             }
             let (args, body) = self.body_to_fixpoint(&info.body, scx)?;
+            let revealed = reveals.contains(&did);
+            let body = if info.hide && !revealed { None } else { Some(body) };
```
nilehmann, line 1584: We should decide here whether to emit a `FunDef` or a `ConstDecl`

nilehmann (changes requested):
I don't like the word hide for this, but I also don't have better suggestions 🤷🏽‍♂️

crates/flux-middle/src/global_env.rs
ranjitjhala: It is used in `define_funs` inside `fixpoint_encoding` (to see if a function should be `define` or `constant` ...)

lib/liquid-fixpoint/src/lib.rs
ranjitjhala: Yes, good point! But note that this will require beefing up the `DataSort` currently used by `ConstDecl` to include `func` sorts.

nilehmann: But it already does, no? We are already using it for uifs. See `register_uif`

ranjitjhala: NVM ignore above

ranjitjhala:
Ok @nilehmann -- have addressed all issues except the `pub fn fn_reveals_for` which seems to to require the `pub`...

tests/tests/neg/error_messages/reveal00.rs
```diff
@@ -0,0 +1,23 @@
+#![flux::defs {
```
nilehmann, line 1: This file should go in `tests/tests/neg/error_messages/desugar/`

nilehmann (approved):

PR #1070, opened 2025-04-07
add cvc5 to VTock Workflow

vrindisbacher:


vrindisbacher:
Looks like it installed properly in the `test` workflow run

.github/workflows/ci.yml
```diff
@@ -101,6 +101,15 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Install CVC5
+        run: |
+          wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.0/cvc5-Linux-x86_64-static.zip
```
nilehmann, line 106: I think the latest is `1.2.1` btw, but I'm not sure if there's any reason to have the latest version.

```suggestion
          wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-x86_64-static.zip
```

nilehmann (approved):

.github/workflows/ci.yml
vrindisbacher: Probably a good idea. Mine is newer than 1.2.0 as well.

PR #1077, opened 2025-04-13
Work around silly fixpoint parser bug

ranjitjhala:
Add a `f$` in front of function names to work around

https://github.com/ucsd-progsys/liquid-fixpoint/issues/750


ranjitjhala:
merging this trivial bugfix lmk if there are any issues @nilehmann

PR #1081, opened 2025-04-20
Report span of super-type when incompatible

ranjitjhala:


crates/flux-middle/src/queries.rs
```diff
@@ -77,6 +77,11 @@ pub enum QueryErr {
         trait_id: DefId,
         name: Symbol,
     },
+    IncompatibleSubtyping {
+        msg: String,
+        span: Span,
+        super_span: Option<Span>,
+    },
```
nilehmann, line 84: This shouldn't be here. `QueryErr` is for errors that happen deep inside a query but need to be reported somewhere else. If anything, this should be an `InferErr`, but see comment below.

tests/tests/neg/error_messages/mismatched_mutability_2.rs
```diff
@@ -0,0 +1,16 @@
...
+
+impl MyTrait for () {
+    #[flux::sig(fn(x: &strg i32[@n]) ensures x: i32[n+1])]
+    fn foo1(x: &mut i32) {
+        *x += 1;
+    }
+}
+
+impl MyTrait for usize {
+    fn foo1(_x: &mut i32) { //~ ERROR incompatible subtyping
```
nilehmann, line 14: This smells like a bug, not an error we should report. Are we not using `ptr_to_ref` on the strong reference like we do for function calls?

nilehmann (changes requested):

ranjitjhala:
Hmm good point re ptr-to-ref I’m guessing not!

- Ranjit.

ranjitjhala:
Subsumed by #1085

PR #1086, opened 2025-04-23
Normalize sorts before wf checking

ranjitjhala:
To get this `extern_spec` to work we need to allow the index `[0, n]` 
which is the valid index for the `Iter` struct; to do so we need to 
normalize the sort of the `bty`, which is what this PR does.

```rust
#[extern_spec]
impl<'a, T, A: Allocator> IntoIterator for &'a Vec<T, A> {
    #[flux::sig(fn (&Vec<T, A>[@n]) -> <&Vec<T, A> as IntoIterator>::IntoIter[0,n])]
    fn into_iter(v: &'a Vec<T, A>) -> <&'a Vec<T, A> as IntoIterator>::IntoIter;
}

#[flux::sig(fn(slice: &Vec<u8>[@n]) -> usize[n])]
fn test_iter_for_loop_vec(vec: &Vec<u8>) -> usize {
    let n = vec.len();
    let mut count = 0;
    for v in vec {
        count += 1;
    }
    count
}
```

### DIFF
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -35,6 +35,7 @@ use rustc_errors::Diagnostic;
 use rustc_hash::FxHashSet;
 use rustc_hir::{OwnerId, Safety, def::DefKind, def_id::DefId};
 use rustc_index::IndexVec;
+use rustc_infer::infer::TyCtxtInferExt;
 use rustc_middle::{
     middle::resolve_bound_vars::ResolvedArg,
     ty::{self, AssocItem, AssocKind, BoundVar, TyCtxt},
@@ -46,6 +47,8 @@ use rustc_span::{
 use rustc_target::spec::abi;
 use rustc_trait_selection::traits;
 use rustc_type_ir::DebruijnIndex;
+
+use crate::conv::ty::TypingMode;
 
 /// Wrapper over a type implementing [`ConvPhase`]. We have this to implement most functionality as
 /// inherent methods instead of defining them as default implementation in the trait definition.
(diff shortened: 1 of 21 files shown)


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1158,6 +1162,21 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
...
+            .infer_ctxt()
+            .build(TypingMode::non_body_analysis());
+        if let Some(def_id) = self.owner().def_id()
+            && let Ok(normalized_sort) = sort.normalize_sorts(def_id.into(), genv, &infcx)
+        {
+            normalized_sort
+        } else {
+            sort
+        }
+    }
```
nilehmann, line 1178: This is the function I was talking about. We should normalize there. Note that `normalize_sorts` already normalizes weak aliases.

https://github.com/flux-rs/flux/blob/main/crates/flux-fhir-analysis/src/wf/sortck.rs#L513

tests/tests/neg/surface/vec02.rs
```diff
@@ -2,9 +2,18 @@
 
 use std::ops::Index;
 
+#[path = "../../lib/option.rs"]
+mod option;
+
+#[path = "../../lib/slice.rs"]
+mod slice;
+
```
nilehmann, line 10: Are these extra imports needed?

I hate the tests that import other files, because if something fails on those files, a bunch of tests fail, and it's hard to see at a glance what's wrong. We should keep tests as self-contained as possible.

ranjitjhala: The `option` is needed for the `Iterator` which uses the indexed-option.

I've split the test up into those for `iter_slice` and `iter_vec` where each only requires the corresponding import.

crates/flux-fhir-analysis/src/conv/mod.rs
ranjitjhala: Ok I moved the call to `normalize_sort` to this place, please take a look!

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -507,15 +512,36 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
-    // we do that maybe we can also stop expanding aliases in `ConvCtxt::conv_sort`.
-    pub(crate) fn normalize_weak_alias_sorts(&mut self) -> QueryResult {
+    // for sort checking, including normalizing projections. [RJ: this is done now]
+    // Maybe we can do lazy normalization. Once we do that maybe we can also stop
+    // expanding aliases in `ConvCtxt::conv_sort`.
+    pub(crate) fn normalize_sorts(&mut self) -> QueryResult {
+        let genv = self.genv;
         for sort in self.sort_of_bty.values_mut() {
-            *sort = self.genv.deep_normalize_weak_alias_sorts(sort)?;
+            let sort1 = genv.deep_normalize_weak_alias_sorts(sort)?;
```
nilehmann, line 540: We don't need to call `deep_normalize_weak_alias_sort` because `normalize_sorts` already normalizes weak aliases (I think)

nilehmann (approved):
I think we don't need to call `deep_normalize_weak_alias_sorts` anymore, bit otherwise looks good.

crates/flux-fhir-analysis/src/wf/sortck.rs
ranjitjhala: Good point, I removed the call.

PR #1088, opened 2025-04-26
Parse `&mut` + `ensures` as `&strg`

ranjitjhala:
This implements Niko Matsakis' suggestion to let us just write the update clause with an `ensures` instead of necessarily having to spell out `&strg` e.g. the below works (and you get an error if the relevant location is _not_ a `&mut`.

```rust
#[flux::sig(fn (x: &mut i32[@n]) ensures x: i32[n+1])]
pub fn incr(x: &mut i32) {
    *x += 1;
}

#[flux::sig(fn () -> i32[11])]
pub fn client_safe() -> i32 {
    let mut p = 10;
    incr(&mut p);
    p
}
```

### DIFF
--- a/crates/flux-driver/src/collector/mod.rs
+++ b/crates/flux-driver/src/collector/mod.rs
@@ -935,6 +935,9 @@ mod errors {
                         }
                     }
                     ParseErrorKind::CannotBeChained => "operator cannot be chained".to_string(),
+                    ParseErrorKind::InvalidBinding => {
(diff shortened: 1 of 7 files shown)


ranjitjhala:
ping @nilehmann

nilehmann (approved):

ranjitjhala:
was wondering why this wasn't working -- looks like I forgot to merge!

PR #1093, opened 2025-05-06
Add support for closure specifications

ranjitjhala:
So stuff like the below works

```rust
#[flux::sig(fn (n:i32, frog: F) -> i32{v:n<=v} where F: FnOnce(i32{v:n <= v}) -> i32{v:n <= v})]
pub fn test001<F>(n: i32, frog: F) -> i32
where
    F: FnOnce(i32) -> i32,
{
    let a = n + 1;
    frog(a)
}

#[flux::sig(fn () -> i32{v:10<=v})]
pub fn test001_client() -> i32 {
    test001(10, |x| x + 1)
}


#[flux::sig(fn (n:i32, frog: F) -> i32{v:n<=v} where F: FnOnce(i32{v:n <= v}) -> i32{v:n <= v})]
pub fn test001_err<F>(n: i32, frog: F) -> i32
where
    F: FnOnce(i32) -> i32,
{
    frog(99) //~ ERROR refinement type
}

#[flux::sig(fn () -> i32{v:20<=v})]
pub fn test001_client_err() -> i32 {
    test001(20, |x| x - 1) //~ ERROR refinement type
}
```

But still todo: allowing binders in the closure spec e.g. `F: FnOnce(i32[@n]) -> i32[n+1]`.


ranjitjhala:
@nilehmann -- I still need to do the dependent/binder signatures, but maybe its a good idea to first get feedback / merge this in since its not very small...

crates/flux-infer/src/projections.rs
```diff
@@ -262,24 +265,63 @@ impl<'infcx, 'genv, 'tcx> Normalizer<'infcx, 'genv, 'tcx> {
...
+        candidates: &mut Vec<Candidate>,
+    ) -> InferResult {
+        let tcx = self.tcx();
+        let rustc_obligation = obligation.to_rustc(tcx);
+        for predicate in predicates {
+            if let ClauseKind::Projection(pred) = predicate.kind_skipping_binder() {
+                let pred_projection_ty = pred.projection_ty.to_rustc(tcx);
+                if pred_projection_ty == rustc_obligation {
+                    let parent_def_id = pred_projection_ty.trait_ref(tcx).def_id;
+                    self.subtype_aliasty(parent_def_id, &pred.projection_ty, obligation)?;
```
nilehmann, line 298: I'd check this when we confirm the candidate

crates/flux-refineck/src/type_env/place_ty.rs
```diff
@@ -118,9 +126,14 @@ impl LookupMode for NoUnfold {
 }
 
 impl PlacesTree {
-    pub(crate) fn unfold(&mut self, infcx: &mut InferCtxt, key: &impl LookupKey) -> InferResult {
+    pub(crate) fn unfold(
+        &mut self,
+        infcx: &mut InferCtxt,
+        key: &impl LookupKey,
+        span: Span,
+    ) -> InferResult {
```
nilehmann, line 134: Not sure how well this would work, but instead of adding a `Span` to all these functions, we could try passing an `InferCtxtAt` directly

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -630,8 +630,47 @@ fn parse_indices(cx: &mut ParseCtxt) -> ParseResult<Indices> {
...
+    let args = vec![in_arg, out_arg]; // inputs and output
+    let segment = PathSegment { ident, args, node_id: cx.next_node_id() };
+    let hi = cx.hi();
+    Ok(Path {
+        segments: vec![segment],
+        refine: vec![],
+        node_id: cx.next_node_id(),
+        span: cx.mk_span(lo, hi),
+    })
+}
```
nilehmann, line 670: Maybe in another PR, but I think it would make things easier if the AST is closer to the syntax, like in rustc, where they separate between parenthesized args and angle-bracketed args (https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.GenericArgs.html). In particular, I think it would help with the "scopes" in `refinement_resolver.rs`

nilehmann (changes requested):

ranjitjhala:
I tried the InferCtxtAt but after going down a rathole (it has an extra
lifetime parameter) was stymied by one of those mysterious “lifetime
outlives” errors…

- Ranjit.

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: Ok I'll pull into a separate issue.

crates/flux-refineck/src/type_env/place_ty.rs
nilehmann: Ok, but there's some simpler cleanup we can do. Instead of passing a `Span` to `lookup_inner` and then threading it down, we can put the `Span` in the `Unfold` struct. Then we don't need to pass a `Span` when there's no `InferCtxt`.

ranjitjhala:
Uff. let's see what's breaking on `vtock` ...

crates/flux-refineck/src/checker.rs
```diff
@@ -229,20 +231,70 @@ impl<'ck, 'genv, 'tcx> Checker<'ck, 'genv, 'tcx, RefineMode> {
...
     }
 }
 
+/// The function `hoist_with_binders` uses the `LocalHoister` machinery to convert
+/// a (closure) function template _without_ binders, e.g. `fn ({v.i32 | *}) -> {v.i32|*})`
+/// into one _with_ input binders, e.g. `forall <a:int>. fn ({i32[a]|*}) -> {v.i32|*}`
+/// after which the hole-filling machinery can be used to fill in the holes.
+/// This lets us get "dependent signatures" where the output can refer to the input.
+/// e.g. see `tests/pos/surface/closure09.rs`
+fn hoist_input_binders(poly_sig: &PolyFnSig) -> PolyFnSig {
```
nilehmann, line 255: Cole needed the same so he implemented it here https://github.com/flux-rs/flux/blob/12d42d997ac645495eb0e44774f3dddf96116519/crates/flux-middle/src/rty/canonicalize.rs#L229. We should reuse that. You'll need to modify it to collapse holes.

crates/flux-refineck/src/checker.rs
```diff
@@ -229,20 +231,70 @@ impl<'ck, 'genv, 'tcx> Checker<'ck, 'genv, 'tcx, RefineMode> {
...
+        .map(|ty| hoister.hoist(ty))
+        .collect_vec();
+
+    delegate.bind(|_vars, preds| {
+        let (mut holes, mut preds): (Vec<_>, Vec<_>) = preds
+            .into_iter()
+            .partition(|pred| matches!(pred.kind(), ExprKind::Hole(HoleKind::Pred)));
+        if let Some(hole) = holes.pop() {
+            preds.push(hole);
+        }
```
nilehmann, line 274: maybe
```suggestion
    delegate.bind(|_vars, mut preds| {
        let mut seen_hole = false;
        preds.retain(|pred| {
          if let ExprKind::Hole(HoleKind::Pred) = pred.kind() {
              if seen_hole {
                  false
              } else {
                  seen_hole = true;
                  true
              }
          } else {
              true
          }
        });
            
```

It's technically more efficient because it doesn't have to allocate extra vectors, but I dunno, looks kind of ugly.

nilehmann: This one is cute.

```rust
    delegate.bind(|_vars, mut preds| {
        let mut keep_hole = true;
        preds.retain(|pred| {
          if let ExprKind::Hole(HoleKind::Pred) = pred.kind() {
              std::mem::replace(&mut keep_hole, false)
          } else {
              true
          }
        });
```

crates/flux-refineck/src/checker.rs
```diff
@@ -897,56 +944,96 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
-        &mut self,
-        infcx: &mut InferCtxt<'_, 'genv, 'tcx>,
-        clauses: &[Binder<FnTraitPredicate>],
-        span: Span,
-    ) -> Result {
-        for clause in clauses {
-            // FIXME(nilehmann) we shouldn't be skipping this binder
-            let clause = clause.skip_binder_ref();
-            self.check_fn_trait_clause(infcx, clause, span)?;
+            Some(self_ty) => {
```
nilehmann, line 1023: Maybe we can specialize this for when `self_ty` is `BaseTy::Param`

crates/flux-refineck/src/type_env/place_ty.rs
ranjitjhala: done

crates/flux-desugar/src/resolver/refinement_resolver.rs
```diff
@@ -158,12 +161,35 @@ impl<V: ScopedVisitor> surface::visit::Visitor for ScopedVisitorWrapper<V> {
         });
     }
 
+    fn visit_trait_ref(&mut self, trait_ref: &surface::TraitRef) {
+        match trait_ref.as_fn_trait_ref() {
+            Some((in_arg, out_arg)) => {
+                self.with_scope(ScopeKind::FnTraitInput, |this| {
+                    this.on_fn_trait_input(in_arg, trait_ref.node_id);
+                    surface::visit::walk_generic_arg(this, in_arg);
+                    surface::visit::walk_generic_arg(this, out_arg);
```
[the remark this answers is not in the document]
ranjitjhala: added a test too.

crates/flux-desugar/src/resolver/refinement_resolver.rs
```diff
@@ -741,6 +768,18 @@ impl ScopedVisitor for RefinementResolver<'_, '_, '_> {
         self.scopes.pop();
     }
 
+    fn on_fn_trait_input(&mut self, in_arg: &surface::GenericArg, trait_node_id: NodeId) {
+        let params = ImplicitParamCollector::new(
+            self.resolver.genv.tcx(),
+            &self.resolver.output.path_res_map,
+            ScopeKind::FnInput,
```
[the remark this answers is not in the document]
ranjitjhala: whoops! good catch thanks!

crates/flux-syntax/src/surface/visit.rs
```diff
@@ -317,21 +317,26 @@ pub fn walk_variant<V: Visitor>(vis: &mut V, variant: &VariantDef) {
...
     vis.visit_async(&fn_sig.asyncness);
-    vis.visit_generics(&fn_sig.generics);
     walk_list!(vis, visit_refine_param, &fn_sig.params);
     for requires in &fn_sig.requires {
         walk_list!(vis, visit_refine_param, &requires.params);
         vis.visit_expr(&requires.pred);
     }
     walk_list!(vis, visit_fn_input, &fn_sig.inputs);
     vis.visit_fn_output(&fn_sig.output);
+    vis.visit_generics(&fn_sig.generics);
```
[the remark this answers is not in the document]
ranjitjhala: hmm I moved it to the top and that seems to work fine too; is that ok?

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -4,12 +4,23 @@ use derive_where::derive_where;
...
-pub(crate) static DEFAULT_QUALIFIERS: LazyLock<[Qualifier<DefaultTypes>; 11]> =
+pub(crate) static DEFAULT_QUALIFIERS: LazyLock<[Qualifier<DefaultTypes>; 13]> =
     LazyLock::new(|| {
         // -----
         // UNARY
         // -----
-
+        // (qualif EqTrue ((v bool)) (v))
+        let eqtrue = Qualifier {
+            args: vec![("v", Sort::<DefaultTypes>::Bool)],
```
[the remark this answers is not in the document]
ranjitjhala: Weird! I only put it in because I was getting an error otherwise, but seems to be working fine now, just as `Sort::Bool`!

crates/flux-middle/src/global_env.rs
```diff
@@ -413,11 +413,24 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
         }
     }
 
-    pub fn is_fn_once_output(&self, def_id: DefId) -> bool {
-        self.tcx()
-            .require_lang_item(rustc_hir::LangItem::FnOnceOutput, None)
-            == def_id
-    }
+    pub fn is_fn_output(&self, def_id: DefId) -> Option<ClosureKind> {
```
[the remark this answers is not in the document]
ranjitjhala: I think its because I want the `ClosureKind` (see `fn find_self_ty_fn_sig`); the tests all pass right now with `FnOnce` and hence returning `Some(ClosureKind::FnOnce)` as the `ClosureKind` but that seems like its shady because perhaps at some point we will need the _actual_ `ClosureKind`?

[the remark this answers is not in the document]
ranjitjhala: Hmm OTOH I wonder now if maybe this code always just returns `ClosureKind::FnOnce` since that 
is where the `Output` is defined?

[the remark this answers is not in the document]
ranjitjhala: hmm. ok reverting to the old `FnOnce` thing!

crates/flux-refineck/src/checker.rs
ranjitjhala: done!

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -707,6 +707,8 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
         let tcx = self.genv().tcx();
         let predicates = tcx.predicates_of(def_id);
         let unrefined_clauses = predicates.predicates;
+        // println!("TRACE: match_clauses {def_id:?} (unrefined) => {unrefined_clauses:#?}");
+        // println!("TRACE: match_clauses {def_id:?} (refined)   => {refined_clauses:#?}");
```
nilehmann:
```suggestion
```

crates/flux-infer/src/infer.rs
```diff
@@ -386,6 +398,10 @@ impl<'infcx, 'genv, 'tcx> InferCtxt<'infcx, 'genv, 'tcx> {
     pub fn cursor(&self) -> &Cursor<'infcx> {
         &self.cursor
     }
+
+    // pub fn scope(&self) -> Option<Scope> {
+    //     self.cursor.marker().scope()
+    // }
```
nilehmann:
```suggestion
```

crates/flux-infer/src/projections.rs
```diff
@@ -310,6 +418,10 @@ impl<'infcx, 'genv, 'tcx> Normalizer<'infcx, 'genv, 'tcx> {
             self.rustc_param_env(),
             obligation.to_rustc(self.tcx()).trait_ref(self.tcx()),
         );
+        // println!(
+        //     "TRACE: assemble_candidates_from_impls (0) [{:?}] => {trait_pred:?}",
+        //     self.def_id()
+        // );
```
nilehmann:
```suggestion

```

crates/flux-middle/src/global_env.rs
```diff
@@ -413,10 +413,25 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
...
+    /// and `Fn`/`FnMut` inherit it, so this should suffice to check if the `def_id`
+    /// corresponds to `LangItem::FnOnceOutput`. However, we may need to do some extra
+    /// work to get the correct `ClosureKind`?
+    pub fn is_fn_output(&self, def_id: DefId) -> Option<ClosureKind> {
+        // let tcx = self.tcx();
+        // let assoc_item = tcx.opt_associated_item(def_id)?;
+        // let trait_def_id = assoc_item.trait_container(tcx)?;
+        // let res = self.tcx().fn_trait_kind_from_def_id(trait_def_id);
+        // println!("TRACE: is_fn_output ==> {res:?}");
+        // res
```
nilehmann:
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -229,20 +230,58 @@ impl<'ck, 'genv, 'tcx> Checker<'ck, 'genv, 'tcx, RefineMode> {
...
+//         rty::FnSig::new(
+//             fn_sig.safety,
+//             fn_sig.abi,
+//             preds.into(),
+//             inputs.into(),
+//             fn_sig.output().clone(),
+//         )
+//     })
+// }
+
```
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -897,56 +931,93 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                    .skip_binder_ref()
+                    .to_ty()
+                    .to_rustc(tcx)
+                    == self_ty
+            {
+                return Ok(proj_clause.map(|proj_pred| {
+                    let self_ty = proj_pred.self_ty().to_ty();
+                    let tupled_args = proj_pred.projection_ty.args[1].expect_base().to_ty();
+                    let output = proj_pred.term.to_ty();
+                    FnTraitPredicate { kind, self_ty, tupled_args, output }.fndef_sig()
```
nilehmann: I don't remember what we do with this `kind`, but this looks wrong. We need the kind of the corresponding trait predicate, but this always returns `FnOnce` (I think it was also returning `FnOnce` before you changed `is_fn_output`)

If I understand correctly, a bound like `F: Fn(i32) -> i32` will produce two clauses `F: Fn<(i32,)>` and `<F as FnOnce>::Output == i32`. This is finding the `<F as FnOnce>::Output == i32` clause, which doesn't contain the kind, for the kind we need to find the corresponding `F: Fn<(i32,)>`. One way to do this would be to call `split_of_fn_trait_clauses` on the predicates and match against the `FnTraitPredicate` directly

crates/flux-refineck/src/compare_impl_item.rs
```diff
@@ -91,13 +91,15 @@ fn check_assoc_reft(
...
-        .normalize_projections(infcx)?;
+        .normalize_projections(&mut infcx.at(impl_span))?;
+
+    // println!("TRACE: impl_sort = {impl_sort:?}");
+    let trait_sort = infcx.genv.sort_of_assoc_reft(trait_assoc_id)?;
+    // println!("TRACE: trait_sort (0) = {trait_sort:?}");
+    let trait_sort = trait_sort.instantiate(infcx.tcx(), &impl_trait_ref.args, &[]);
+    // println!("TRACE: trait_sort (1) = {trait_sort:?}");
+    let trait_sort = trait_sort.normalize_projections(&mut infcx.at(impl_span))?;
+    // println!("TRACE: trait_sort (2) = {trait_sort:?}");
```
nilehmann:
```suggestion
    let trait_sort = infcx.genv.sort_of_assoc_reft(trait_assoc_id)?;
    let trait_sort = trait_sort.instantiate(infcx.tcx(), &impl_trait_ref.args, &[]);
    let trait_sort = trait_sort.normalize_projections(&mut infcx.at(impl_span))?;
```

nilehmann (changes requested):
One major comment https://github.com/flux-rs/flux/pull/1093#discussion_r2122322083

crates/flux-refineck/src/checker.rs
ranjitjhala: Yup -- that `kind` was wrong! Much cleaner with the `split_off_fn_trait_clauses` thanks, have made that change!

crates/flux-middle/src/global_env.rs
```diff
@@ -413,10 +413,19 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
         }
     }
 
-    pub fn is_fn_once_output(&self, def_id: DefId) -> bool {
-        self.tcx()
+    /// Nico notes that that the `Output` associated type is defined in `FnOnce`,
+    /// and `Fn`/`FnMut` inherit it, so this should suffice to check if the `def_id`
+    /// corresponds to `LangItem::FnOnceOutput`. However, we may need to do some extra
+    /// work to get the correct `ClosureKind`?
+    pub fn is_fn_output(&self, def_id: DefId) -> Option<ClosureKind> {
```
nilehmann: I think we can make this return a boolean again

```suggestion
    pub fn is_fn_output(&self, def_id: DefId) -> bool {
```

nilehmann (approved):
One last small comment, otherwise lgtm

crates/flux-middle/src/global_env.rs
ranjitjhala: ah of course, I made a note to do this and forgot thanks for catching!

PR #1100, opened 2025-05-28
Add flag for trusted and ignore

Zn2FePO424H2O:
I add two flags for trusted and ignore. When these flags are set to true, Flux will trust/ignore everything by default.

`flux -Ftrusted=on -Fignore=on test.rs`


nilehmann (approved):
Looks good, thanks @Zn2FePO424H2O!

PR #1119, opened 2025-06-09
Loop `self-ty` search to parent

### LINKED ISSUE #1116: ICE: `cannot find self_ty_fn_sig`
Triggered all over `core`; minimal-ish test

```rust
pub struct Split<'a, T: 'a, P>
where
    P: FnMut(&T) -> bool,
{
    pub v: &'a [T],
    pub pred: P,
}

impl<'a, T: 'a, P: FnMut(&T) -> bool> Split<'a, T, P> {
    pub fn new(slice: &'a [T], pred: P) -> Self {
        Self { v: slice, pred }
    }
}
```

ranjitjhala:
Instead of just failing.

fixes #1116

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -900,18 +900,27 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
         span: Span,
     ) -> Result<PolyFnSig> {
         let tcx = self.genv.tcx();
-        let predicates = self
-            .genv
-            .predicates_of(self.def_id)
-            .with_span(span)?
-            .instantiate_identity()
-            .predicates;
+        let mut def_id = Some(self.def_id);
+        while let Some(did) = def_id {
+            let generic_predicates = self
+                .genv
+                .predicates_of(did)
+                .with_span(span)?
+                .instantiate_identity();
+            let predicates = generic_predicates.predicates;
 
-        for poly_fn_trait_pred in Clause::split_off_fn_trait_clauses(self.genv, &predicates).1 {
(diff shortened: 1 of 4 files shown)


crates/flux-refineck/src/checker.rs
```diff
@@ -900,18 +900,32 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
         span: Span,
     ) -> Result<PolyFnSig> {
         let tcx = self.genv.tcx();
-        let predicates = self
-            .genv
-            .predicates_of(self.def_id)
-            .with_span(span)?
-            .instantiate_identity()
-            .predicates;
+        let mut def_id = Some(self.def_id);
```
nilehmann: You should start with a `DefId` here

```suggestion
        let mut def_id = Some(self.def_id.to_def_id());
```

nilehmann (changes requested):

crates/flux-refineck/src/checker.rs
```diff
@@ -900,18 +900,32 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            // Continue to the parent if we didn't find a match
+            def_id = match generic_predicates.parent {
+                Some(parent) => {
+                    self.genv
+                        .resolve_id(parent)
+                        .as_maybe_extern()
+                        .map(|z| z.local_id().to_def_id())
+                }
+                None => None,
+            };
```
nilehmann: No need to resolve now that `def_id` is a `DefId`

```suggestion
            def_id = generic_predicates.parent;
```

nilehmann (approved):
One small comment

crates/flux-refineck/src/checker.rs
ranjitjhala: nice!

PR #1120, opened 2025-06-10
Add support for enum `Ctor`

ranjitjhala:
Allows stuff like 

```rust

pub enum Blah {
    MkBlah(i32),
}

pub fn test00<T, F>(f: F) -> T
where
    F: FnOnce(i32) -> T,
{
    f(0)
}

pub fn test01() -> Blah {
    test00(Blah::MkBlah)
}
```

### DIFF
--- a/tests/tests/neg/error_messages/extern_specs/invalid_enum.rs
+++ b/tests/tests/neg/error_messages/extern_specs/invalid_enum.rs
@@ -1,0 +1,7 @@
+use flux_rs::attrs::*;
+
+#[extern_spec]
+enum Option<T> {
+    //~^ ERROR invalid extern_spec
+    Some(T),
+}


crates/flux-driver/src/collector/extern_specs.rs
```diff
@@ -108,6 +108,18 @@ impl<'a, 'sess, 'tcx> ExternSpecCollector<'a, 'sess, 'tcx> {
...
+        let extern_enum_def = self.tcx().adt_def(extern_id);
+
+        // Collect all constructor DefIds from variants
+        for (extern_variant, variant) in extern_enum_def.variants().iter().zip(enum_def.variants) {
+            if let Some(extern_ctor) = extern_variant.ctor_def_id()
+                && let Some(ctor) = variant.data.ctor_def_id()
+            {
+                self.insert_extern_id(ctor, extern_ctor)?;
+            }
+        }
```
nilehmann: We should report an error here in case of a mismatch; otherwise, the mapping may not make sense, i.e., 

1. If the number of variants is different
2. If two variants are not simultaneously a ctor

Additionally, we could check if variants are the same kind of `Ctor` (`Unit` or `Fn`/`Tuple`). I think this is not strictly necessary, but it'd be good to be strict to avoid confusion.

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -504,13 +525,49 @@ fn variants_of(
...
+            Some(fn_sig)
+        }
+        fhir::Node::Ctor(ctor_def_id) => {
+            let (def_id, variant_idx) = get_enum_info_from_ctor(&genv, ctor_def_id).unwrap();
+            match genv.variant_sig(def_id, variant_idx)? {
+                rty::Opaqueness::Opaque => Err(query_bug!(def_id, "expected transparent enum"))?,
+                rty::Opaqueness::Transparent(variant_sig) => {
+                    return Ok(variant_sig.to_poly_fn_sig(None));
+                }
+            }
```
```suggestion
            let tcx = genv.tcx();
            let variant_id = tcx.parent(def_id.resolved_id());
            let enum_id = tcx.parent(variant_id);
            let variant_idx = tcx.adt_def(enum_id).variant_index_with_id(variant_id);
            let sig = genv
                .variant_sig(enum_id, variant_idx)?
                .map(|sig| sig.to_poly_fn_sig(None))
                .ok_or_else(|| query_bug!(ctor_id, "expected transparent enum"))?;


```

crates/flux-middle/src/fhir.rs
```diff
@@ -121,6 +121,7 @@ pub enum Node<'fhir> {
     ImplItem(&'fhir ImplItem<'fhir>),
     OpaqueTy(&'fhir OpaqueTy<'fhir>),
     ForeignItem(&'fhir ForeignItem<'fhir>),
+    Ctor(LocalDefId),
```
nilehmann: Storing `LocalDefId` is dangerous because we don't know if they refer to extern specs. But we don't need this anymore with the change I suggested in `fn fn_sig`

```suggestion
    Ctor,
```

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -357,6 +359,25 @@ fn generics_of(genv: GlobalEnv, def_id: LocalDefId) -> QueryResult<rty::Generics
...
+                    })?;
+                    conv::conv_generics(genv, generics, def_id, false)
+                }
+                MaybeExternId::Extern(_, def_id) => {
+                    // for an extern, get generics from rustc
+                    let rustc_generics = genv.lower_generics_of(def_id);
+                    refining::refine_generics(genv, def_id, &rustc_generics)
+                }
+            }
+        }
```
nilehmann: I think we can merge this into the arm below because the ctor is represented as a child of the `Enum` from the perspective of generics so the `rty::Generics` will always be empty and point to the parent, i.e., it'll look like

```
Generics { parent: Some(DefId(0:3 ~ playground[f6cf]::Foo)), parent_count: 1, own_params: [], param_def_id_to_index: {}, has_self: false, has_late_bound_regions: None }
```

nilehmann (changes requested):

crates/flux-fhir-analysis/src/lib.rs
ranjitjhala: good catch!

ranjitjhala:
should probably merge this one before `bump` ?

crates/flux-driver/src/collector/extern_specs.rs
```diff
@@ -108,6 +108,25 @@ impl<'a, 'sess, 'tcx> ExternSpecCollector<'a, 'sess, 'tcx> {
...
         self.inner.collect_enum_def(enum_id, attrs, enum_def)?;
 
+        // Add stuff about Ctor
+        // Get the AdtDef for the enum
+        let extern_enum_def = self.tcx().adt_def(extern_id);
+
+        // Collect all constructor DefIds from variants
+        let extern_variants = extern_enum_def.variants();
+        let enum_variants = enum_def.variants;
+        debug_assert_eq!(extern_variants.len(), enum_variants.len());
```
nilehmann: this error should be reported, not an assert. It can happen with stuff like

```rust
#[extern_spec]
enum Option<T> {
    Some(T),
}
```

ranjitjhala:
Ok, added that

PR #1122, opened 2025-06-11
Bump toolchain to nightly-2025-06-02

ranjitjhala:


crates/flux-metadata/src/encoder.rs
```diff
@@ -129,10 +128,10 @@ impl SpanEncoder for EncodeContext<'_, '_> {
     }
 }
 
-impl<'tcx> TyEncoder for EncodeContext<'_, 'tcx> {
+impl<'tcx> TyEncoder<'tcx> for EncodeContext<'_, 'tcx> {
     const CLEAR_CROSS_CRATE: bool = true;
 
-    type I = TyCtxt<'tcx>;
+    // CUT type I = TyCtxt<'tcx>;
```
nilehmann:
```suggestion
```

crates/flux-syntax/src/lexer.rs
```diff
@@ -49,8 +49,14 @@ pub enum Token {
...
-    OpenDelim(Delimiter),
-    CloseDelim(Delimiter),
+    OpenParen,
+    CloseParen,
+    OpenBrace,
+    CloseBrace,
+    OpenBracket,
+    CloseBracket,
+    OpenInvisible(InvisibleOrigin),
+    CloseInvisible(InvisibleOrigin),
```
nilehmann, line 59: I'm confused, why did you split this?

crates/flux-middle/src/fhir.rs
```diff
@@ -230,7 +230,7 @@ impl<'fhir> Item<'fhir> {
 pub enum ItemKind<'fhir> {
     Enum(EnumDef<'fhir>),
     Struct(StructDef<'fhir>),
-    TyAlias(TyAlias<'fhir>),
+    TyAlias(Box<TyAlias<'fhir>>),
```
nilehmann: You should allocate this in the arena

```suggestion
    TyAlias(&'fhir TyAlias<'fhir>),
```

crates/flux-middle/src/global_env.rs
```diff
@@ -59,9 +59,9 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
         self.inner.tcx
     }
 
-    pub fn hir(&self) -> rustc_middle::hir::map::Map<'tcx> {
-        self.tcx().hir()
-    }
+    // TODO(BUMP) pub fn hir(&self) -> rustc_middle::hir::map::Map<'tcx> {
+    // TODO(BUMP)     self.tcx().hir()
+    // TODO(BUMP) }
```
nilehmann:
```suggestion
```

crates/flux-rustc-bridge/src/lowering.rs
```diff
@@ -789,14 +789,14 @@ impl<'tcx> Lower<'tcx> for rustc_ty::ValTree<'tcx> {
 
     fn lower(self, _tcx: TyCtxt<'tcx>) -> Self::R {
         if let Some(scalar_int) = self.try_to_scalar_int() {
-            return crate::ty::ValTree::Leaf(scalar_int);
+            crate::ty::ValTree::Leaf(scalar_int)
```
nilehmann: Wait you didn't change this from the previous PR

nilehmann (changes requested):

crates/flux-syntax/src/lexer.rs
ranjitjhala: because the rust people did so ... https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/token/enum.TokenKind.html#variant.OpenParen ...

crates/flux-rustc-bridge/src/lowering.rs
ranjitjhala: oops -- good catch! looks like it was in 2 places! this and `lower_constant`...

crates/flux-common/src/mir_storage.rs
```diff
@@ -44,15 +44,22 @@ pub unsafe fn store_mir_body<'tcx>(
 /// # Safety
 ///
 /// See the module level comment.
-#[expect(clippy::needless_lifetimes, reason = "we want to be very explicit about lifetimes here")]
+// #[expect(clippy::needless_lifetimes, reason = "we want to be very explicit about lifetimes here")]
```
nilehmann:
```suggestion
```

crates/flux-common/src/mir_storage.rs
```diff
@@ -44,15 +44,22 @@ pub unsafe fn store_mir_body<'tcx>(
...
     _tcx: TyCtxt<'tcx>,
     def_id: LocalDefId,
 ) -> BodyWithBorrowckFacts<'tcx> {
     let body_with_facts: BodyWithBorrowckFacts<'static> = SHARED_STATE.with(|state| {
-        match state.borrow_mut().remove(&def_id) {
+        let mut map = state.borrow_mut();
+        let mut keys = String::new();
+        for k in map.keys() {
+            keys.push_str(format!("{k:?}").as_str());
+        }
```
nilehmann: This is dead code. Funny that Clippy doesn't realize that, but I suppose it depends on whether allocation should be an observable behavior.

```suggestion
```

crates/flux-desugar/src/resolver.rs
```diff
@@ -134,7 +134,7 @@ impl<'genv, 'tcx> CrateResolver<'genv, 'tcx> {
             let def_kind = match item.kind {
                 ItemKind::Use(path, kind) => {
                     match kind {
-                        hir::UseKind::Single => {
+                        hir::UseKind::Single(_) => {
```
nilehmann: If I'm reading the documentation of [Single](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.UseKind.html#variant.Single) correctly, we can get the name from the field in this variant instead of getting the last segment of the path. It may also be the correct thing to do to handle `use path as name` imports. I believe we wren't handling this case.

crates/flux-driver/src/callbacks.rs
```diff
@@ -263,6 +257,21 @@ fn mir_borrowck<'tcx>(
     unsafe {
         flux_common::mir_storage::store_mir_body(tcx, def_id, body_with_facts);
     }
+}
+
+// #[expect(clippy::needless_lifetimes, reason = "we want to be explicit about lifetimes here")]
```
nilehmann:
```suggestion
```

crates/flux-macros/src/diagnostics/subdiagnostic.rs
```diff
@@ -88,25 +86,122 @@ impl SubdiagnosticDerive {
...
                     #implementation
                 }
             }
         });
+
         ret
     }
 }
 
+// impl SubdiagnosticDerive {
```
nilehmann: Remove this code

crates/flux-macros/src/diagnostics/subdiagnostic.rs
```diff
@@ -384,11 +479,11 @@ impl<'parent, 'a> SubdiagnosticDeriveVariantBuilder<'parent, 'a> {
                 Ok(quote! {})
             }
             "subdiagnostic" => {
-                let f = &self.parent.f;
+                // let f = &self.parent.f;
```
nilehmann:
```suggestion
```

crates/flux-macros/src/diagnostics/subdiagnostic.rs
```diff
@@ -535,12 +630,12 @@ impl<'parent, 'a> SubdiagnosticDeriveVariantBuilder<'parent, 'a> {
         let span_field = self.span_field.value_ref();
 
         let diag = &self.parent.diag;
-        let f = &self.parent.f;
+        // let f = &self.parent.f;
```
nilehmann:
```suggestion
```

crates/flux-middle/src/rty/fold.rs
```diff
@@ -180,7 +180,7 @@ pub trait TypeVisitable: Sized {
 
             fn visit_binder<T: TypeVisitable>(&mut self, t: &Binder<T>) -> ControlFlow<()> {
                 self.outer_index.shift_in(1);
-                t.super_visit_with(self);
+                let _ = t.super_visit_with(self);
```
nilehmann: This is wrong. This function should break if this call breaks because the value is used to signal if an escaping variable is found (we break and stop the traversal when finding one)

```suggestion
                t.super_visit_with(self)?;
```

nilehmann: I believe this was already broken.

crates/flux-refineck/src/checker.rs
```diff
@@ -1898,7 +1896,7 @@ fn infer_under_mut_ref_hack(rcx: &mut InferCtxt, actuals: &[Ty], fn_sig: &PolyFn
 }
 
 impl Mode for ShapeMode {
-    const NAME: &str = "shape";
+    const _NAME: &str = "shape";
```
nilehmann: Let's remove the `_` and put a `#[expect(...)]`. Sometimes I use this name for debugging to print only something in one of the modes.

```suggestion
    const NAME: &str = "shape";
```

crates/flux-refineck/src/checker.rs
```diff
@@ -1951,7 +1949,7 @@ impl Mode for ShapeMode {
 }
 
 impl Mode for RefineMode {
-    const NAME: &str = "refine";
+    const _NAME: &str = "refine";
```
nilehmann: same here

crates/flux-refineck/src/ghost_statements/fold_unfold.rs
```diff
@@ -129,7 +129,7 @@ struct FoldUnfoldAnalysis<'a, 'genv, 'tcx, M> {
 }
 
 trait Mode: Sized {
-    const NAME: &'static str;
+    const _NAME: &'static str;
```
nilehmann, line 132: same in this file

crates/flux-syntax/src/lexer.rs
```diff
@@ -300,31 +312,35 @@ impl<'t> Cursor<'t> {
...
                     .push_back((span.lo() + BytePos(1), Token::Gt, span.hi()));
                 return;
             }
-            TokenKind::Not => Token::Not,
+            TokenKind::Bang => Token::Bang,
             TokenKind::PathSep => Token::PathSep,
             TokenKind::DotDot => Token::DotDot,
-            _ => Token::Invalid,
+            tok => {
+                println!("Invalid token: {tok:?}");
```
nilehmann:
```suggestion
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -101,9 +102,9 @@ fn parse_hide_attr(cx: &mut ParseCtxt) -> ParseResult<bool> {
     if !cx.advance_if(Tok::Pound) {
         return Ok(false);
     }
-    cx.expect(Tok::OpenDelim(Bracket))?;
+    cx.expect(open_delim(Bracket))?;
```
nilehmann: Since we have a dedicated token now we should use it directly `Tok::OpenBracket` (the same for all the other cases)

crates/flux-refineck/src/ghost_statements/points_to.rs
```diff
@@ -250,10 +251,15 @@ struct CollectPointerToBorrows<'a> {
     tracked_places: FxHashMap<PlaceIndex, flux_rustc_bridge::mir::Place>,
     stmts: &'a mut GhostStatements,
     before_state: Vec<(PlaceIndex, FlatSet<Loc>)>,
+    results: IndexVec<BasicBlock, State>,
```
nilehmann: I believe we don't need to mutate it, so we can make this slice and avoid the clone.

```suggestion
    results: &'a IndexSlice<BasicBlock, State>,
```

crates/flux-refineck/src/ghost_statements/points_to.rs
```diff
@@ -49,17 +50,17 @@ pub(crate) fn add_ghost_statements<'tcx>(
...
+    let mut analysis = fixpoint.analysis;
+    let results = fixpoint.results;
+    let mut visitor = CollectPointerToBorrows::new(&map, stmts, results.clone());
+    visit_reachable_results(body, &mut analysis, &results, &mut visitor);
 
     Ok(())
 }
 
-type Results<'a, 'tcx> = rustc_mir_dataflow::Results<'tcx, PointsToAnalysis<'a>>;
+// CUT type Results<'a> = rustc_mir_dataflow::Results<PointsToAnalysis<'a>>;
```
nilehmann:
```suggestion
```

nilehmann (approved):
Mostly small changes

crates/flux-syntax/src/lexer.rs
ranjitjhala: did you want to unsplit?

crates/flux-middle/src/rty/fold.rs
ranjitjhala: good catch!

crates/flux-syntax/src/lexer.rs
nilehmann: No, it's fine like this

PR #1124, opened 2025-06-14
Better simplification

cole-k:
# Changes
Two big changes to simplification in this PR:

1. Simplification was removing trivial implications by doing equality comparison on `Expr`s. This almost never works because `Expr`s have spans in them which are often different. I fixed that by bringing over a function I wrote in another branch to erase spans. (we may want an architectural change to handle this if it keeps coming up)
2. We can do slightly better for removing trivial implications. We were previously checking to see if we find anything like `foo => foo` to `true`, but this does not simplify constraints such as `foo => bar => foo`. Additionally, if we have `foo /\ bar => foo`, we wouldn't simplify this expression either. I collect all of the assumed predicates and use them to perform a deeper simplification.

**Care should be taken to make sure I'm not accidentally introducing unsoundness.**

# Impact

## On a simple test

I tried this on the `kmeans.rs` tests.

With the better simplification
```
$ find log-simp -name '*.simp.fluxc' -exec wc {} \;
      18     122     719 log-simp/fluxtests.kmeans-kmeans.simp.fluxc
      19     120     708 log-simp/fluxtests.kmeans-init_centers.simp.fluxc
       5      34     176 log-simp/fluxtests.kmeans-add.simp.fluxc
      49     298    2456 log-simp/fluxtests.kmeans-kmeans_step.simp.fluxc
      20     131     819 log-simp/fluxtests.kmeans-nearest.simp.fluxc
       4      27     134 log-simp/fluxtests.rvec-{impl#3}-index.simp.fluxc
       5      36     188 log-simp/fluxtests.rvec-{impl#4}-index_mut.simp.fluxc
      17     111     653 log-simp/fluxtests.kmeans-normalize_centers.simp.fluxc
       5      34     176 log-simp/fluxtests.kmeans-dist.simp.fluxc
       6      41     214 log-simp/fluxtests.kmeans-normal.simp.fluxc
```

Without the better simplification
```
$ find log-without-simp -name '*.simp.fluxc' -exec wc {} \;
      25     198    1098 log-without-simp/fluxtests.kmeans-kmeans.simp.fluxc
      19     120     708 log-without-simp/fluxtests.kmeans-init_centers.simp.fluxc
       9      62     326 log-without-simp/fluxtests.kmeans-add.simp.fluxc
      64     452    3442 log-without-simp/fluxtests.kmeans-kmeans_step.simp.fluxc
      32     199    1335 log-without-simp/fluxtests.kmeans-nearest.simp.fluxc
       5      35     175 log-without-simp/fluxtests.rvec-{impl#3}-index.simp.fluxc
       6      44     229 log-without-simp/fluxtests.rvec-{impl#4}-index_mut.simp.fluxc
      28     215    1207 log-without-simp/fluxtests.kmeans-normalize_centers.simp.fluxc
       8      54     284 log-without-simp/fluxtests.kmeans-dist.simp.fluxc
       9      57     314 log-without-simp/fluxtests.kmeans-normal.simp.fluxc
```

Here's a specific file to show the difference:

With the better simplification (`kmeans-add`)

```
∀ n: RVec.  
  (0 ≤ n.0 ∧ n.0 ≥ 0) =>  
    $k0(0) ~ Goto(bb2) at 24:9: 24:10
    ∀ a0: int.  
      ($k0(a0) ∧ a0 < n.0) => $k0(a0 + 1) ~ Goto(bb2) at 25:17: 30:6
```

Without the better simplification (`kmeans-add`)
```
∀ n: RVec.  
  (0 ≤ n.0 ∧ 0 ≤ n.0 ∧ n.0 ≥ 0) =>  
    $k0(0) ~ Goto(bb2) at 24:9: 24:10
    ∀ a0: int.  
      ($k0(a0) ∧ a0 < n.0) =>  
        (a0 < n.0) ~ Call at 26:19: 26:22
        (a0 < n.0) ~ Call at 27:19: 27:22
        (a0 < n.0) ~ Call at 28:10: 28:13
        $k0(a0 + 1) ~ Goto(bb2) at 25:17: 30:6
```

## On a real-world test

I timed the current simplification versus the better simplification on `rv32i` in tock. With the better simplification it took about 24s and without it took about 67s.

It also shows a modest decrease in size (only on the smt2 files though)

```
$ find log-simp -type f -regex '.*.smt2' -print0 | xargs -0 du -ch | tail -n 1
5.2M	total
```

```
$ find log-without-simp -type f -regex '.*.smt2' -print0 | xargs -0 du -ch | tail -n 1
6.1M	total
```

### DIFF
--- a/crates/flux-infer/src/refine_tree.rs
+++ b/crates/flux-infer/src/refine_tree.rs
@@ -16,6 +16,7 @@ use flux_middle::{
     },
 };
 use itertools::Itertools;
+use rustc_hash::FxHashSet;
 use rustc_middle::ty::TyCtxt;
 use serde::Serialize;
 
@@ -55,7 +56,7 @@ impl RefineTree {
     }
 
     pub(crate) fn simplify(&mut self, genv: GlobalEnv) {
-        self.root.borrow_mut().simplify(genv);
+        self.root.borrow_mut().simplify(genv, FxHashSet::default());
     }
 
     pub(crate) fn into_fixpoint(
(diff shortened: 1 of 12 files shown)


crates/flux-infer/src/refine_tree.rs
```diff
@@ -353,31 +354,41 @@ impl std::ops::Deref for NodePtr {
...
-            NodeKind::Trace(_) | NodeKind::Root(_) | NodeKind::ForAll(..) => {
+            _ => {}
+        }
+
+        // Then simplify the children
+        // (the order matters here because we need to collect assumed preds first)
+        for child in &self.children {
+            // Need to clone the assumed_preds here because it may (and probably
+            // will) diverge across branches
+            child.borrow_mut().simplify(genv, assumed_preds.clone());
```
nilehmann: You can implement this with a [`SnapshotMap`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_data_structures/snapshot_map/struct.SnapshotMap.html) to avoid the `clone` and quadratic behavior.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -578,6 +586,18 @@ impl Expr {
...
+    pub fn erase_spans(&self) -> Expr {
+        struct SpanEraser {}
+        impl TypeFolder for SpanEraser {
+            fn fold_expr(&mut self, e: &Expr) -> Expr {
+                let e = e.super_fold_with(self);
+                Expr { kind: e.kind, espan: None }
+            }
+        }
+        let mut eraser = SpanEraser {};
+        eraser.fold_expr(self)
```
nilehmann:
```suggestion
        struct SpanEraser;
        impl TypeFolder for SpanEraser {
            fn fold_expr(&mut self, e: &Expr) -> Expr {
                e.super_fold_with(self).at_opt(None)
            }
        }
        self.fold_with(&mut SpanEraser)
```

crates/flux-infer/src/refine_tree.rs
cole-k: Haskell: "Look what they have to do to mimic a fraction of our power"

nilehmann (approved):

crates/flux-infer/src/refine_tree.rs
nilehmann: We could use [immutable data structures](https://docs.rs/im/latest/im/struct.HashMap.html), but why would you consciously choose to do something more inefficient?

Issue #1114, opened 2025-06-06
Unfold associated refinements with default definitions


vrindisbacher:
The following should check, but results in `postcondition cannot be proved`.

```rust
use flux_rs::*;
#[assoc(fn f1() -> bool)]
#[assoc(fn f2() -> bool { <Self as Foo>::f1() })]
trait Foo {

    #[spec(fn () -> bool[<Self as Foo>::f1()])]
    fn get_bool() -> bool;
}

struct Bar<F: Foo> {
    inner: F
}

impl<F: Foo> Bar<F> {
   #[spec(fn () -> bool[<F as Foo>::f1()])]
    fn implies() -> bool {
        F::get_bool()
    }

    #[spec(fn () -> bool[<F as Foo>::f2()])]
    fn foo_bar_baz() -> bool {
        Self::implies()
    } 
}
```

Looks like flux is not unfolding the definition of `f2` despite having a default impl. Here's the constraint generated: 

```rust
∀ . (<(λb0. F[b0]) as Foo>::f1() = <(λb1. F[b1]) as Foo>::f2()) ~ Ret at 211:9: 211:24
```

nilehmann:
@vrindisbacher this is the correct behavior because an implementation could redefine the associated refinement so there's no guarantee on what the body is when used generically. I suppose this means it doesn't solve the original issue you were having though

vrindisbacher:
Ahhh... I see. Hmmm - can we define a different primitive that does unfold the definition?

vrindisbacher:
Perhaps something that can only be added at the trait definition site?

nilehmann:
I was thinking the same. We could have something like that and it would be a subset of the feature we actually want which is `defs!` functions with trait bounds. This would be probably a lot easier to implement.

vrindisbacher:
Yea exactly. This sounds good.

vrindisbacher:
Hey @nilehmann any update here? Is this something that's being worked on?

vrindisbacher:
Is this something I could do you think? Or would it be complicated?

nilehmann:
@vrindisbacher, I think you can take a crack at it. If you want we can discuss how to implement it. Or I can post some pointers here.

nilehmann:
Some pointers

[...]

PR #1128, opened 2025-06-18
Add final associated reft

vrindisbacher:
Adds final associated refinements: #1114


tests/tests/pos/surface/final_assoc_reft01.rs
```diff
@@ -0,0 +1,20 @@
+#[flux::assoc(fn f1() -> bool)]
+trait Baz {
+    #[flux::spec(fn (s: Self) -> bool[<Self as Baz>::f1()])]
+    fn into_bool(self) -> bool;
+}
+
+#[flux::final_assoc(fn f2() -> bool { <T as Baz>::f1() })]
```
nilehmann: we could make this `#[flux::assoc(final fn f2() -> bool { <T as Baz>::f1() })`. There is a proposal to do the same for trait methods https://internals.rust-lang.org/t/pre-rfc-final-trait-methods/18407, so we could be consistent with that. But I don't have strong opinions.

tests/tests/neg/surface/final_assoc_reft00.rs
```diff
@@ -0,0 +1,2 @@
+#[flux::final_assoc(fn foo() -> bool)] //~ERROR final associated refinements must have a body
+pub trait MyTrait {}
```
nilehmann: This file should go in `tests/neg/error_messages/desugar`

crates/flux-syntax/src/surface.rs
```diff
@@ -238,6 +238,7 @@ pub struct TraitAssocReft {
     pub output: BaseSort,
     pub body: Option<Expr>,
     pub span: Span,
+    pub r#final: bool,
```
nilehmann: We typically end keywords with `_`, I think we do it a couple of times for `trait_` and `impl_`, so maybe better to be consistent with that.

```suggestion
    pub final_: bool,
```

crates/flux-middle/src/rty/mod.rs
```diff
@@ -751,9 +751,29 @@ pub struct CoroutineObligPredicate {
     pub output: Ty,
 }
 
+#[derive(Copy, Clone, Encodable, Decodable, Hash, PartialEq, Eq)]
+pub struct AssocRefinementMeta {
```
nilehmann: I would call this `AssocRefinement` or `AssocReft`

crates/flux-middle/src/rty/mod.rs
```diff
@@ -751,9 +751,29 @@ pub struct CoroutineObligPredicate {
...
+    pub def_id: FluxDefId,
+    pub r#final: bool,
```
nilehmann: It'd be good to leave a comment here saying that this field only applies for traits, because we also use the same struct to refer to associated refinements in impls.

crates/flux-infer/src/projections.rs
```diff
@@ -97,12 +97,31 @@ impl<'a, 'infcx, 'genv, 'tcx> Normalizer<'a, 'infcx, 'genv, 'tcx> {
...
     }
 
+    fn alias_reft_is_final(&mut self, alias_reft: &AliasReft) -> QueryResult<bool> {
+        let reft_id = alias_reft.assoc_id;
+        let assoc_refinements = self.genv().assoc_refinements_of(reft_id.parent())?;
+        Ok(assoc_refinements
+            .items
+            .iter()
+            .find(|meta| meta.def_id == reft_id && meta.r#final)
+            .is_some())
```
nilehmann: This is duplicating the logic in `AssocRefinements::find`. I would make that method return a `&AssocRefinementMeta` and then we can extract the necessary information (either the ID or whether it's final) from there.

crates/flux-refineck/src/compare_impl_item.rs
```diff
@@ -21,13 +21,14 @@ pub fn check_impl_against_trait(genv: GlobalEnv, impl_id: MaybeExternId) -> Quer
...
 
-    for trait_assoc_id in &trait_assoc_refts.items {
+    for trait_assoc_meta in &trait_assoc_refts.items {
+        let trait_assoc_def_id = trait_assoc_meta.def_id();
         let has_default = genv
-            .default_assoc_refinement_body(*trait_assoc_id)?
+            .default_assoc_refinement_body(trait_assoc_def_id)?
             .is_some();
-        if !impl_names.contains(&trait_assoc_id.name()) && !has_default {
+        if !impl_names.contains(&trait_assoc_meta.name()) && !has_default {
```
nilehmann: We should report an error here if the impl tries to redefine a final assoc refinement. So that would be the case where

`impl_names.contains(&trait_assoc_meta.name()) && is_final`

nilehmann (changes requested):
Looks good, mostly stylistic comments, except for one where I think we are missing an extra check when an impl tries to redefine a final associated refinement. I think in that case we will ignore it, but better to give an error to avoid confusion.

crates/flux-infer/src/projections.rs
nilehmann: Actually, scratch that. I think we can do better and expose a method like this

```rust
fn get(&self, assoc_id: FluxDefId) -> AssocRefinementMeta
```

Because if we have a `FluxDefId` and the `AssocRefinements` associated with the parent of that `FluxDefId`, then it's guaranteed that the associated refinement will exist in the list.

vrindisbacher:
@nilehmann all of these should be addresed.

crates/flux-metadata/src/lib.rs
```diff
@@ -314,14 +314,15 @@ fn encode_def_ids<K: Eq + Hash + Copy>(
                     .insert(key, genv.refinement_generics_of(def_id));
                 let assocs = genv.assoc_refinements_of(def_id);
                 if let Ok(assocs) = &assocs {
-                    for assoc_id in &assocs.items {
-                        let key = assoc_id_to_key(*assoc_id);
+                    for assoc_meta in &assocs.items {
```
nilehmann: We should probably not called these variables `*_meta` anymore.

```suggestion
                    for assoc_reft in &assocs.items {
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -150,9 +150,12 @@ pub(crate) fn parse_trait_assoc_refts(cx: &mut ParseCtxt) -> ParseResult<Vec<Tra
 /// ```text
 /// ⟨trait_assoc_reft⟩ := fn ⟨ident⟩ ( ⟨refine_param⟩,* ) -> ⟨base_sort⟩ ;?
 ///                     | fn ⟨ident⟩ ( ⟨refine_param⟩,* ) -> ⟨base_sort⟩ ⟨block⟩
+///                     | final fn ⟨ident⟩ ( ⟨refine_param⟩,* ) -> ⟨base_sort⟩ ⟨block⟩
 /// ```
 fn parse_trait_assoc_reft(cx: &mut ParseCtxt) -> ParseResult<TraitAssocReft> {
     let lo = cx.lo();
+    // let final_ = cx.advance_if(Tok::Final);
```
```suggestion
```

vrindisbacher:
Cool fixed!

PR #1132, opened 2025-06-19
Use function param names to generate better bound variables

nilehmann:
Use the name of parameters in functions to generate a more meaningful `BoundReftKind::Named` for existentials at the "top level"

### DIFF
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -19,7 +19,7 @@ use flux_middle::{
     queries::{QueryErr, QueryResult},
     query_bug,
     rty::{
-        self, ESpan, INNERMOST, List, RefineArgsExt, WfckResults,
+        self, BoundReftKind, ESpan, INNERMOST, List, RefineArgsExt, WfckResults,
         fold::TypeFoldable,
         refining::{self, Refine, Refiner},
     },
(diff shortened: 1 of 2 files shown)


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1696,7 +1723,15 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
         };
         let sort = bty.sort();
         let bty = bty.shift_in_escaping(1);
-        let ctor = rty::Binder::bind_with_sort(rty::Ty::indexed(bty, rty::Expr::nu()), sort);
+        let kind = match name {
+            Some(name) => BoundReftKind::Named(name),
+            None => BoundReftKind::Annon,
```
cole-k, line 1728: While you're at it this should probably be named `BoundReftKind::Anon`.

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1143,12 +1155,25 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
             .find_by_ident_and_kind(self.tcx(), assoc_name, assoc_tag, trait_def_id)
     }
 
+    fn conv_ty_with_name(
+        &mut self,
+        env: &mut Env,
+        ty: &fhir::Ty,
+        name: Symbol,
+    ) -> QueryResult<rty::Ty> {
+        if let fhir::TyKind::BaseTy(bty) = &ty.kind {
```
cole-k, line 1164: Why is this only done for `BaseTy`s?

cole-k (commented):
Code LGTM, but one thing I'm confused about: will this also work for un-annotated code?

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: The signatures generated for unannotated functions use `BaseTy` and we generate the existential for them during `conv`, so this is the case that matters. It won't however use a name for the index behind references which I'm not sure wether we should do.

PR #1133, opened 2025-06-20
Misc failure things

ranjitjhala:
changing `unwrap` to `tracked_span_bug!` type things...

### DIFF
--- a/book/src/dev/develop.md
+++ b/book/src/dev/develop.md
@@ -1,5 +1,13 @@
 # Developer's Guide
 
+## Backtraces
+
+You can use the usual `RUST_BACKTRACE=1` environment variable to enable backtraces, if you build and install `flux` as:
+
+```sh
+$ cargo x install --profile dev
+```
+
 ## Regression Tests
 
 You can run the various regression tests in the `tests/pos` and `tests/neg` directories using
--- a/crates/flux-desugar/src/lib.rs
+++ b/crates/flux-desugar/src/lib.rs
@@ -242,11 +242,14 @@ pub fn desugar<'genv>(
         rustc_hir::Node::Ctor(rustc_hir::VariantData::Tuple(_, _, _)) => {
             nodes.insert(def_id, fhir::Node::Ctor);
         }
+        rustc_hir::Node::Synthetic => {
+            nodes.insert(def_id, fhir::Node::Synthetic);
+        }
         node => {
             if let Some(ident) = node.ident() {
-                span_bug!(ident.span, "unsupported node: {node:?}");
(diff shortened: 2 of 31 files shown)


book/src/dev/develop.md
```diff
@@ -1,5 +1,13 @@
 # Developer's Guide
 
+## Backtraces
+
+You can use the usual `RUST_BACKTRACE=1` environment variable to enable backtraces, if you build and install `flux` as:
```
nilehmann: You also get backtraces for `release`, so this is more like to get better backtraces use `dev`

nilehmann (approved):

crates/flux-infer/src/projections.rs
```diff
@@ -280,12 +280,16 @@ impl<'a, 'infcx, 'genv, 'tcx> Normalizer<'a, 'infcx, 'genv, 'tcx> {
...
                     .associated_items(impl_def_id)
                     .in_definition_order()
                     .find(|item| item.trait_item_def_id == Some(obligation.def_id))
                     .map(|item| item.def_id)
-                    .unwrap();
+                else {
+                    return Err(query_bug!(
+                        "no associated type for {obligation:?} in impl {impl_def_id:?}"
+                    ));
+                };
```
nilehmann: You can do `ok_or_else(|| query_bug!(...))?`

crates/flux-desugar/src/lib.rs
```diff
@@ -242,11 +242,14 @@ pub fn desugar<'genv>(
         rustc_hir::Node::Ctor(rustc_hir::VariantData::Tuple(_, _, _)) => {
             nodes.insert(def_id, fhir::Node::Ctor);
         }
+        rustc_hir::Node::Synthetic => {
+            nodes.insert(def_id, fhir::Node::Synthetic);
+        }
```
nilehmann: Does the example that triggered this still crash? If not, and this is the fix, maybe we can add a test.

ranjitjhala: hmm yes, I wanted to add a test but it was tickled by some sort of strange async/trait business inside core... let me see if its still needed...

ranjitjhala: Would you prefer we removed this code if 

1. The error is no longer triggered in `core` and
2. I can't replicate it as a test?

book/src/dev/develop.md
ranjitjhala: good point will update comment

crates/flux-desugar/src/lib.rs
ranjitjhala: nuked it; easy enough to add back if we want!

PR #1136, opened 2025-06-21
Support conversion from char-to-int

ranjitjhala:


ranjitjhala:
@nilehmann here's the int/char PR -- I didn't realize that thanks to @Samir-Rashid we already supported char literals, so the whole thing works quite nicely! (thanks @Samir-Rashid !!!)

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -356,8 +360,17 @@ pub enum ThyFunc {
 }
 
 impl ThyFunc {
-    pub const ALL: [ThyFunc; 35] = [
+    pub fn is_erased_in_encoding(&self) -> bool {
+        match self {
+            ThyFunc::CharToInt | ThyFunc::IntToChar => true,
+            _ => false,
+        }
```
nilehmann: I'm not a fan. `ThyFunc` is supposed to be the list of functions supported in fixpoint. This won't work when/if we switch to the Rust fixpoint rewrite. But I guess this is fine for now, considering the alternative is to create a new type of function or duplicate the enum.

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -356,8 +360,17 @@ pub enum ThyFunc {
...
+    }
```
nilehmann:
```suggestion
        matches!(self, ThyFunc::CharToInt | ThyFunc::IntToChar)
    }
```

lib/liquid-fixpoint/src/lib.rs
ranjitjhala: Yes, I agree - and realized the same working on #1137 - maybe better, in a subsequent PR, to encode these as `SpecFunc` (which _is_ a new type of function?)

PR #1137, opened 2025-06-24
Extensible primops

ranjitjhala:
You can add extra properties for primitive operations like `>>` or `<<` etc. which work directly on `int`-indexed values like `usize`, `u32` etc. by

```rust
defs! {
    property ShiftByTwo[<<](x:int, y:int) -> int {
       y == 2 => [<<](x, y) == 4*x
    }
}

pub fn test2() {
    let x = 1;
    let x = x << 2;
    let x = x << 2;
    assert(x == 16) //~ ERROR refinement type error
}
```

### DIFF
--- a/book/src/guide/use.md
+++ b/book/src/guide/use.md
@@ -672,6 +672,22 @@ You can switch on overflow checking
 You can provide _properties_ to be used when doing computations with
 primitive operations like `<<` or `>>`.
 
+Given a primop `op` with signature `(t1,...,tn) -> t` we define
+a refined type for `op` expressed as a [`RuleMatcher`]
+
+```rust
+op :: (x1: t1, ..., xn: tn) -> { t[op_val[op](x1,...,xn)] | op_rel[x1,...,xn] }
+```
+
+that is, using two _uninterpreted functions_ `op_val` and `op_rel` that respectively denote
+
+1. The _value_ of the primop, and
+2. Some invariant _relation_ that holds for the primop.
+
+The latter can be extended by the user via a `property` definition,
+which allows us to customize primops like `<<` with extra "facts"
+or lemmas. See `tests/tests/pos/surface/primops00.rs` for an example.
+
 ```rust
 {{#include ../../../tests/tests/pos/surface/primops00.rs}}
 ```


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -57,6 +57,7 @@ pub(crate) struct AfterSortck<'a, 'genv, 'tcx> {
     next_type_index: u32,
     next_region_index: u32,
     next_const_index: u32,
+    prim_app_sort: FxHashMap<fhir::BinOp, rty::Sort>,
```
ranjitjhala: Everything around this feels a bit janky to me ... ( the business of figuring out what the output sort of the primitive UIF should be.) Maybe its better to have some "global" knowledge that associates the sort directly with the `BinOp` ?

One thing I dislike is you may want the `property` to talk about multiple values e.g. for sake of example `[>>](x, y) == [<<](x, y)` -- and you cannot now...

crates/flux-middle/src/rty/expr.rs
```diff
@@ -674,13 +682,36 @@ impl Ctor {
...
+
 #[derive(Clone, PartialEq, Eq, Hash, TyEncodable, Debug, TyDecodable)]
 pub enum ExprKind {
     Var(Var),
     Local(Local),
     Constant(Constant),
     ConstDefId(DefId),
     BinaryOp(BinOp, Expr, Expr),
+    /// An UIF application representing a primitive operation
+    PrimApp(PrimFunc, Expr, Expr),
```
ranjitjhala: Or is it better to make `PrimFunc` a variant of `SpecFuncKind` ? I think so now...

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -292,6 +293,24 @@ impl<'genv> fhir::visit::Visitor<'genv> for Wf<'_, 'genv, '_> {
...
+                .as_conv_ctxt()
+                .conv_sort(&prim_prop.output)
+                .emit(&self.errors)
+        {
+            self.infcx.prim_app_sort.insert(prim_prop.op, output);
+            self.infcx
+                .check_expr(&prim_prop.body, &rty::Sort::Bool)
+                .collect_err(&mut self.errors);
+        } else {
+            panic!("FIXME: prim_prop args should always have two elements");
```
ranjitjhala: TODO(RJ): fix this!

crates/flux-middle/src/rty/expr.rs
ranjitjhala: This is now split out. 

In `rty` we just use plain `App` everywhere _but_ have separate `GlobalFunc` (stuff that goes to fixpoint) and `InternalFunc` (stuff that does not); the latter also includes `Char/Int` etc. which are pulled out of `ThyFunc`.

In `fhir` we have a separate `PrimApp`, that uses the `BinOp` (to avoid shenanigans mangling the `BinOp` with the `Path` stuff needed in regular `App`).

ranjitjhala:
@nilehmann -- I think I did the stuff we discussed yesterday, so this is ready for a look!

crates/flux-syntax/src/surface.rs
```diff
@@ -69,6 +71,21 @@ pub struct SpecFunc {
     pub hide: bool,
 }
 
+/// A (currently global) *primop property*; see tests/tests/pos/surface/
+#[derive(Debug)]
+pub struct PrimProp {
```
nilehmann: I prefer to have `op` in the name of these things. So `PrimoOpProp` or `PrimopProp`

```suggestion
pub struct PrimOpProp {
```

crates/flux-fhir-analysis/src/wf/mod.rs
```diff
@@ -292,6 +296,30 @@ impl<'genv> fhir::visit::Visitor<'genv> for Wf<'_, 'genv, '_> {
...
+                Some(prim_prop.span),
+                String::from("primop"),
+                sorts.len(),
+                prim_prop.args.len(),
+            ));
+            return;
+        }
+        for (arg, sort) in prim_prop.args.iter().zip(sorts) {
+            self.infcx.declare_param(*arg, sort);
+        }
```
nilehmann: This is a bit weird because it overrides the declaration created in `declare_params_for_flux_item`. I would either (1) put this in `declare_params_for_flux_item` and change the behavior for primop props or (2) implement this as unifying the sort to the inference variables already declared. Probably (1) is easier.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1043,6 +1043,32 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                let expr = if let Some(prim_rel) = self.genv.prim_rel_for(op)? {
+                    prim_rel.body.replace_bound_refts(args)
+                } else {
+                    rty::Expr::tt()
+                };
+                self.expr_to_fixpoint(&expr, scx)
+            }
+            InternalFuncKind::CharToInt | InternalFuncKind::IntToChar => {
+                self.expr_to_fixpoint(&args[0], scx)
+            }
```
nilehmann: Maybe we can leave a comment about why this is erased

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1043,6 +1043,32 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                let args = self.exprs_to_fixpoint(args, scx)?;
+                Ok(fixpoint::Expr::App(Box::new(func), args))
+            }
+            InternalFuncKind::Rel(op) => {
+                let expr = if let Some(prim_rel) = self.genv.prim_rel_for(op)? {
+                    prim_rel.body.replace_bound_refts(args)
+                } else {
+                    rty::Expr::tt()
+                };
+                self.expr_to_fixpoint(&expr, scx)
```
nilehmann, line 1064: Maybe

```suggestion
                if let Some(prim_rel) = self.genv.prim_rel_for(op)? {
                    let expr = prim_rel.body.replace_bound_refts(args)
                    self.expr_to_fixpoint(&expr, scx)
                } else {
                    Ok(fixpoint::Expr::Constant(Constant::Boolean(true)))
                }
```

Or we can define a constant and use `fixpoint::Expr::TRUE`

```rust
impl<T: Types> Expr<T> {
    pub const TRUE: Self = fixpoint::Expr::Constant(Constant::Boolean(true));
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1413,6 +1437,37 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         })
     }
 
+    fn prim_op_sort(op: &rty::BinOp, span: Span) -> rty::PolyFuncSort {
```
nilehmann, line 1450: hmm, this logic is a bit duplicated in `sortck.rs`. They are not exactly the same because this is on `rty` and the other one on `fhir`. For now, maybe we can leave a comment saying that they have to be in sync.

But more generally, we are abusing these bin ops, because we are using them for interpreted operations on bit vectors and uninterpreted functions on integers. So when `Binop::BitShr` appears in a `ExprKind::BinOp`, it means bit vectors, but inside `ExprKind::InternalFunc` it means `int`. I suppose it'll fine for now...

crates/flux-macros/src/primops.rs
```diff
@@ -308,7 +313,14 @@ impl Parse for Output {
...
             let content;
             bracketed!(content in input);
-            Ok(Output::Indexed(bty, content.parse()?))
+            let idx = content.parse()?;
+            if input.peek(token::Bracket) {
+                let pred;
+                bracketed!(pred in input);
+                Ok(Output::Constr(bty, idx, pred.parse()?))
+            } else {
+                Ok(Output::Indexed(bty, idx))
```
nilehmann: We could parse here something closer to the surface syntax like `{ T | p }`

```rust
impl Parse for Output {
    fn parse(input: ParseStream) -> syn::Result<Self> {
        if input.peek(token::Brace) {
            let content;
            braced!(content in input);
            let bty = content.parse()?;
            let idx = parse_index(&content)?;
            let _: Token![|] = content.parse()?;
            Ok(Output::Constr(bty, idx, content.parse()?))
        } else {
            let bty: syn::Ident = input.parse()?;
            if input.peek(token::Bracket) {
                Ok(Output::Indexed(bty, parse_index(&input)?))
            } else if input.peek(token::Brace) {
                let content;
                braced!(content in input);
                let _: syn::Ident = content.parse()?;
                let _: Token![:] = content.parse()?;
                Ok(Output::Exists(bty, content.parse()?))
            } else {
                Ok(Output::Base(bty))
            }
        }
    }
}

fn parse_index(input: ParseStream) -> syn::Result<TokenStream> {
    let content;
    bracketed!(content in input);
    content.parse()
}
```

crates/flux-macros/src/primops.rs
```diff
@@ -400,4 +412,5 @@ where
 
 mod kw {
     syn::custom_keyword!(requires);
+    syn::custom_keyword!(ensures);
```
nilehmann:
```suggestion
```

crates/flux-fhir-analysis/src/wf/mod.rs
ranjitjhala: Good catch -- this code is not needed, so I can just delete it.

ranjitjhala: ok I changed the `declare_params_for_flux_item` so it special cases the primop by grabbing the `Sort` from the actual `>>` ...

crates/flux-macros/src/primops.rs
ranjitjhala: Much better, thanks!

crates/flux-fhir-analysis/src/wf/mod.rs
ranjitjhala: See this https://github.com/flux-rs/flux/blob/ee876b87ab458e4c6762f08bb135f34c2f54a942/crates/flux-fhir-analysis/src/wf/mod.rs#L164

ranjitjhala:
Ok @nilehmann I think I did the above, but you probably want to take a look at the `declare_params_for_flux_item` to see if that's what you had in mind.

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -143,6 +145,40 @@ fn parse_sort_decl(cx: &mut ParseCtxt) -> ParseResult<SortDecl> {
...
+    let op = parse_binop(cx)?;
+    cx.expect(Token::CloseBracket)?;
+
+    // Parse the args
+    let params = parens(cx, Comma, |cx| parse_refine_param(cx, RequireSort::No))?;
+    // CUT for param in &params {
+    // CUT     if !matches!(param.sort, Sort::Infer) {
+    // CUT         return Err(ParseError { kind: crate::ParseErrorKind::InvalidSort, span: param.span });
+    // CUT     }
+    // CUT }
```
```suggestion
```

crates/flux-macros/src/primops.rs
```diff
@@ -400,4 +418,5 @@ where
...
```
nilehmann:
```suggestion
```

ranjitjhala: strange, I thought I already did the "commit suggestion" last time around, hmm...

PR #1139, opened 2025-06-26
Check files cargo

ranjitjhala:
Use cargo manifest to specify `checked_files` instead of the command line. So now, we can write stuff like

```toml
[package.metadata.flux]
enabled = true
check_files = [ "src/ascii/ascii_char.rs" ]
```

in the `Cargo.toml` and then `flux` will only check the file matching that suffix.

### DIFF
--- a/crates/flux-config/src/flags.rs
+++ b/crates/flux-config/src/flags.rs
@@ -139,10 +139,16 @@ pub struct Paths {
 }
 
 impl Paths {
+    fn file_matches(p: &PathBuf, file: &str) -> bool {
+        let res = p.to_str().map_or(false, |p| p == file);
+        println!("TRACE: file_matches[{p:?}] {file}) ==> {res}");
+        res
+    }
+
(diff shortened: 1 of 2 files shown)


crates/flux-config/src/flags.rs
```diff
@@ -139,10 +139,16 @@ pub struct Paths {
 }
 
 impl Paths {
+    fn file_matches(p: &PathBuf, file: &str) -> bool {
+        let res = p.to_str().map_or(false, |p| p == file);
```
nilehmann: I'm confused. How is this different than before? It is still checking for equality, no?

ranjitjhala:
Ugh sorry git branch foolishness let me fix


- Ranjit.

ranjitjhala:
have updated -- checks if suffix matches; I think maybe I'll change it so if the path ends with `*` it allows any *substring* match so we specify the whole module?

PR #1140, opened 2025-06-27
Implement `-Finclude` flag with glob pattern

nilehmann:


crates/flux-config/src/lib.rs
```diff
@@ -57,8 +57,8 @@ pub fn ignore_default() -> bool {
     FLAGS.ignore_default
 }
 
-pub fn is_checked_file(file: &str) -> bool {
-    FLAGS.check_files.is_checked_file(file)
+pub fn is_checked_file(file: &Path) -> bool {
```
ranjitjhala, line 60: Does this mean that the old `check-files` stuff is no longer used?

PR #1151, opened 2025-06-28
Reorder docs

ranjitjhala:
I've reordered the docs so that

1. The "tutorial"-y bits come at the top 1,2,3,4,...
2. The "internals"-y bits are at the end like an appendix
3. The "tutorial"-y bits are interactive


book/src/guide/specifications.md
```diff
@@ -155,7 +228,59 @@ with `const` generics
...
+    inner: std::collections::HashMap<K, V>,
+}
+```
+
+**Note that opaque structs **can not** have refined fields.**
+
+Now, we can define `get` for our refined map as follows:
+
+```rust
+#[generics(K as base, V as base)]
```
nilehmann: `generics` is a no-op now. We should probably remove it

```suggestion
```

book/src/guide/specifications.md
```diff
@@ -155,7 +228,59 @@ with `const` generics
...
+    {
+        self.inner.get(k)
+    }
+
+}
+```
+
+Note that if we do not mark these methods as `trusted`, we will get an error that looks like...
+
+```rust
```
nilehmann: This is not rust

```suggestion
```text
```

book/src/guide/specifications.md
```diff
@@ -155,7 +228,59 @@ with `const` generics
 
 ### Opaque Structs
 
-See the [specifications guide](specs.md) for more about the `opaque` annotation.
+Flux offers an attribute `opaque` which can be used on structs. A module defining an opaque struct should define a trusted API, and clients of the API should not access struct fields directly. This is particularly useful in cases where users need to define a type indexed by a different type than the structs fields. For example, `RMap` (see below) defines a refined HashMap, indexed by a `Map` - a primitive sort defined by flux.
+
+```rust
```
nilehmann, line 233: I think by default, `rust` blocks have a "play" button (which sends the query to goto). I don't know if we want some of these blocks to have the button.

book/src/guide/specifications.md
```diff
@@ -532,9 +694,71 @@ TODO
...
+`std::mem` into a module path and then transforms the function into
+
+```rust
+#[extern_spec]
+#[spec(fn(x: &mut T[@vx], y: &mut T[@vy]) ensures x: T[vy], y: T[vx])]
+#[allow(unused, dead_code)]
+fn __flux_extern_spec_swap<T>(x: &mut T, y: &mut T) {
+    std::mem::swap(x, y)
+}
+```
```
nilehmann: We can probably get rid of this.

```suggestion
You shouldn't need to know the details, but if you're curious,
here's how the macro works. In the code above, flux parses the
`std::mem` into a module path and then transforms the function into

```rust
#[extern_spec]
#[spec(fn(x: &mut T[@vx], y: &mut T[@vy]) ensures x: T[vy], y: T[vx])]
#[allow(unused, dead_code)]
fn __flux_extern_spec_swap<T>(x: &mut T, y: &mut T) {
    std::mem::swap(x, y)
}
```
```

book/flux_theme/favicon.png
nilehmann, line 1: maybe we can make this image smaller

nilehmann (approved):

book/src/guide/specifications.md
ranjitjhala: Right, let me leave that to "future work" -- I mashed together the old `specs.md` and `use.md` here, but yes, I think its a good idea to also make these parts clickable.

book/src/guide/specifications.md
ranjitjhala: So to clarify: delete the whole discussion? The "suggestion" on github is  no-op (as in the suggestion seems to be the original text itself...?)

book/src/guide/specifications.md
nilehmann: I meant the opposite. To make some of them not clickable. By default they all get the play button (I think) and you have to annotate them with `noplayground` for it to just be text. Or did your javascript changes completely change the way blocks work?

book/src/guide/specifications.md
nilehmann: ups, yeah, I think we can remove this, no need to "stabilize" the expansion.

PR #1168, opened 2025-07-07
remove `lib/option.rs`

ranjitjhala:
instead use `flux_core`


ranjitjhala:
@nilehmann here's a first PR that removes the `option.rs` and `range.rs` ...

crates/flux-metadata/src/lib.rs
```diff
@@ -388,6 +388,9 @@ fn encode_def_ids<K: Eq + Hash + Copy>(
                     .refinement_generics_of
                     .insert(key, genv.refinement_generics_of(def_id));
             }
+            DefKind::Ctor(CtorOf::Variant, _) => {
+                tables.fn_sig.insert(key, genv.fn_sig(def_id));
+            }
```
nilehmann: I'd put this together with `DefKind::Fn` and `DefKind::AssocFn`. Also, I think what we want is `DefKind::Ctor(_, CtorKind::Fn)`

ranjitjhala: Note that we do use the `CtorOf::Variant` when constructing the `fn_sig` -- should that be changed?

nilehmann: I recently added support for tuple structs, so we now support `CtorOf::Variant` and `CtorOf::Struct`. The defining feature here is that they are functions (instead of constants, which are generated for unit structs and variants)

PR #1170, opened 2025-07-08
Remove "lib/step.rs"

ranjitjhala:


lib/flux-core/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![no_std]
+#![feature(step_trait)]
```
nilehmann: We have to put this under a cfg, otherwise it won't work for non-Flux builds.

```suggestion
#![cfg_attr(flux, feature(step_trait))]
```

lib/flux-core/src/iter/range.rs
```diff
@@ -1,15 +1,8 @@
...
-
-extern crate flux_core;
-
-use std::iter::Step;
+use flux_attrs::*;
 
 // TODO(RJ): use default spec `true` for `can_step_forward` and `can_step_backward`
 
-#[extern_spec(core::ops)]
+#[extern_spec(core::iter)]
```
nilehmann, line 19: I think we should document these associated refinements. It's not clear to me what each one is supposed to mean. Like, why can a `usize` always step forward? Also, I'm a bit hesitant to have associated refinements that don't have default bodies because that will break any implementation that we have not yet refined.

Maybe we should also use these associated refinements to refine the methods in the trait to sanity check them.

nilehmann (changes requested):

lib/flux-core/src/iter/range.rs
ranjitjhala: Lets separate some concerns here: 

1. breakage-from-missing-defaults, 
2. docs, 
3. why-this-spec.

I think (1) is a big one because it may just prevent using `flux-core` at all. 

(2) and (3) seem orthogonal to (1) and likely iteratable on. 

I'm not sure what you mean by "refine the methods to sanity check"?

re: (1) maybe we should have a way to generate a default UIF?

nilehmann: The `Step` trait has a couple of methods, and the associated refinements are not used to specify them. I understand that the end goal is to specify iterators, but the implementation of `Iterator` for `Range` ultimately uses the methods in `Step`, so one should be able to specify them with the associated refinements.

The comment about documentation is more about me not being able to say if these associated refinements make sense. But perhaps this is more about this PR not standing on its own, because I need to see how it interacts with `Iterator`.

The UIF is an interesting idea; it also crossed my mind. But depending on how one specify things it may also break users, but in the sense that you won't be able to use the methods, because you don't have enough information.

ranjitjhala: Was just looking at the _actual_ `Step` in core, btw, they have a lot of specs explicitly spelled out.

ranjitjhala: Ok I think I dealt with (1), (2) and (3), using UIFs, adding comment and tweaking the specs.

Ultimately, would be cool to actually validate against the real implementations in `core`!

ranjitjhala: Yes, maybe I should lump `Iterator` in as well (since we can't really check the assoc-reft in isolation, without actually verifying them against their real implementations...)

tests/tests/lib/iterator.rs
```diff
@@ -29,8 +26,8 @@ struct Zip<A, B>;
 #[flux_rs::assoc(fn step(self: Range<A>, other: Range<A>) -> bool { <A as Step>::can_step_forward(self.start, 1) => other.start == <A as Step>::step_forward(self.start, 1) } )]
 impl<A: Step> Iterator for Range<A> {
     #[flux_rs::sig(
-        fn(self: &strg Range<A>[@old_range]) -> Option<A[old_range.start]>[old_range.start < old_range.end]
-            ensures self: Range<A>{r: (<A as Step>::can_step_forward(old_range.start, 1) && old_range.start < old_range.end)=> (r.start == <A as Step>::step_forward(old_range.start, 1) && r.end == old_range.end) }
+        fn(self: &mut Range<A>[@old]) -> Option<A[old.start]>[old.start < old.end]
+            ensures self: Range<A>{r: (old.start < old.end => r.start == <A as Step>::step_forward(old.start, 1)) && r.end == old.end }
```
nilehmann: Sounds like we don't need `can_step_forward` for specifying iterators? Maybe we should get rid of it.

ranjitjhala:
Yup was thinking the same

- Ranjit.

ranjitjhala:
@nilehmann ok -- have moved all of the `iter/ator` stuff together with `step`... [ will leave `lib/vec.rs` for a separate PR]

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,33 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[assoc(fn valid_item(self: Self, item: Self::Item) -> bool { true })]
+#[assoc(fn size(self: Self) -> int)]
+#[assoc(fn done(self: Self) -> bool)]
+#[assoc(fn step(self: Self, other: Self) -> bool { true })]
+trait Iterator {
+    #[flux::sig(fn(self: &strg Self[@curr_s]) -> Option<Self::Item>[!<Self as Iterator>::done(curr_s)] ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)})]
```
nilehmann: I found this hard to read in a single line. I would format it like so. 

```suggestion
    #[flux::sig(
        fn(self: &strg Self[@curr_s]) -> Option<Self::Item>[!<Self as Iterator>::done(curr_s)] 
        ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)}
    )]
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn next(&mut self) -> Option<Self::Item>;
+
+    #[flux::sig(fn(Self[@s]) -> Enumerate<Self>[0, s])]
+    fn enumerate(self) -> Enumerate<Self>
+    where
+        Self: Sized;
+
+    #[flux::sig(fn(Self[@s], f: F) -> Map<Self, F>[s] where F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> B)]
```
nilehmann: Same here and other places

```suggestion
    #[flux::sig(
        fn(Self[@s], f: F) -> Map<Self, F>[s]
        where 
            F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> B
    )]
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,33 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[assoc(fn valid_item(self: Self, item: Self::Item) -> bool { true })]
+#[assoc(fn size(self: Self) -> int)]
+#[assoc(fn done(self: Self) -> bool)]
+#[assoc(fn step(self: Self, other: Self) -> bool { true })]
```
nilehmann: You can put these in a single block btw. I find it a bit more readable. 

Also, we are using `assoc` in some places and `reft` in others; they both mean the same, but we should probably pick one. I don't have strong opinions

```suggestion
#[assoc(
    fn valid_item(self: Self, item: Self::Item) -> bool { true }
    fn size(self: Self) -> int);
    fn done(self: Self) -> bool);
    fn step(self: Self, other: Self) -> bool { true }
)]
```

lib/flux-core/src/slice/iter.rs
```diff
@@ -10,6 +10,15 @@ impl<'a, T> Iter<'a, T> {
     fn as_slice(&self) -> &'a [T];
 }
 
+#[extern_spec(core::slice)]
+#[flux::assoc(fn size(x: Iter) -> int { x.len - x.idx })]
+#[flux::assoc(fn done(x: Iter) -> bool { x.idx >= x.len })]
+#[flux::assoc(fn step(x: Iter, y: Iter) -> bool { x.idx + 1 == y.idx && x.len == y.len})]
+impl<'a, T> Iterator for Iter<'a, T> {
+    #[flux::sig(fn(self: &strg Iter<T>[@curr_s]) -> Option<_>[curr_s.idx < curr_s.len] ensures self: Iter<T>{next_s: curr_s.idx + 1 == next_s.idx && curr_s.len == next_s.len})]
```
nilehmann:
```suggestion
    #[flux::sig(fn(self: &mut Iter<T>[@curr_s]) -> Option<_>[curr_s.idx < curr_s.len] ensures self: Iter<T>{next_s: curr_s.idx + 1 == next_s.idx && curr_s.len == next_s.len})]
```

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,60 @@
+use core::{iter::Step, ops};
```
nilehmann, line 1: We still need to fix this such that it works for non-flux builds. We should either put this import under a `cfg` or the entire module.

lib/flux-core/src/iter/adapters/skip.rs
```diff
@@ -0,0 +1,14 @@
+use flux_attrs::*;
```
nilehmann, line 1: I don't think we are using this in any test; we should either add a test or remove it.

And talking about tests, I added support for writing tests directly in the modules where we add the extern specs (see `slice/mod.rs` for example). My rationale was that it would be better to have them close to the specs. But it is a bit more complicated to write tests like that so we could move them to the normal test infrastructure.

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,33 @@
...
+    where
+        Self: Sized;
+
+    #[flux::sig(fn(Self[@s], f: F) -> Map<Self, F>[s] where F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> B)]
+    fn map<B, F>(self, f: F) -> Map<Self, F>
+    where
+        Self: Sized,
+        F: FnMut(Self::Item) -> B;
+
+    #[flux::sig(fn(Self[@s], f: F) where F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> () )]
```
nilehmann: We should be consistent with the use of spec

```suggestion
    #[spec(fn(Self[@s], f: F) where F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> () )]
```

This should work, but it'll be better to also add `spec` to the list in `flux_attrs_impl::FLUX_ATTRS`.

nilehmann (changes requested):

lib/flux-core/src/iter/range.rs
ranjitjhala: I put the module under a `cfg(flux)`

lib/flux-core/src/iter/adapters/skip.rs
ranjitjhala: Well, plus we may as well reuse these specs for our regular tests...

ranjitjhala: I simplified the spec for `Skip` and added a test for it.

lib/flux-core/src/iter/adapters/enumerate.rs
```diff
@@ -0,0 +1,15 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[refined_by(idx: int, inner: I)]
+struct Enumerate<I>;
+
+#[extern_spec(core::iter)]
+#[assoc(fn size(x: Enumerate<I>) -> int { <I as Iterator>::size(x.inner) })]
+#[assoc(fn done(x: Enumerate<I>) -> bool { <I as Iterator>::done(x.inner)})]
+#[assoc(fn step(x: Enumerate<I>, y: Enumerate<I>) -> bool { <I as Iterator>::step(x.inner, y.inner)})]
```
nilehmann: note that I changed the definition of `step`, which can be made more precise.

```suggestion
#[assoc(
    fn size(x: Enumerate<I>) -> int { <I as Iterator>::size(x.inner) }
    fn done(x: Enumerate<I>) -> bool { <I as Iterator>::done(x.inner) }
    fn step(x: Enumerate<I>, y: Enumerate<I>) -> bool { 
        x.idx + 1 == y.idx && <I as Iterator>::step(x.inner, y.inner)
    }
)]
```

lib/flux-core/src/iter/adapters/map.rs
```diff
@@ -0,0 +1,11 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[refined_by(inner: I)]
+struct Map<I, F>;
+
+#[extern_spec(core::iter)]
+#[assoc(fn size(x: Map<I>) -> int { <I as Iterator>::size(x.inner) })]
+#[assoc(fn done(x: Map<I>) -> bool { <I as Iterator>::done(x.inner)})]
+#[assoc(fn step(x: Map<I>, y: Map<I>) -> bool { <I as Iterator>::step(x.inner, y.inner)})]
```
nilehmann:
```suggestion
#[assoc(
    fn size(x: Map<I>) -> int { <I as Iterator>::size(x.inner) }
    fn done(x: Map<I>) -> bool { <I as Iterator>::done(x.inner)}
    fn step(x: Map<I>, y: Map<I>) -> bool { <I as Iterator>::step(x.inner, y.inner)}
)]
```

lib/flux-core/src/iter/adapters/skip.rs
```diff
@@ -0,0 +1,10 @@
...
+
+#[extern_spec(core::iter)]
+#[refined_by(size: int)]
+struct Skip<I>;
+
+#[extern_spec(core::iter)]
+#[assoc(fn size(r: Skip) -> int { r.size } )]
+#[assoc(fn step(self: Skip, other: Skip) -> bool { other.size == self.size - 1 } )]
```
nilehmann:
```suggestion
#[assoc(
    fn size(r: Skip) -> int { r.size }
    fn step(self: Skip, other: Skip) -> bool { other.size == self.size - 1 }
)]
```

lib/flux-core/src/iter/adapters/skip.rs
```diff
@@ -0,0 +1,10 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[refined_by(size: int)]
+struct Skip<I>;
+
+#[extern_spec(core::iter)]
+#[assoc(fn size(r: Skip) -> int { r.size } )]
```
nilehmann: This is morally wrong; we should store the number of skipped elements in the refined_by and compute the size based on the inner iterator here.

lib/flux-core/src/iter/adapters/zip.rs
```diff
@@ -0,0 +1,15 @@
+use flux_attrs::*;
+
+#[extern_spec(core::iter)]
+#[refined_by(a: A, b: B, idx: int, len: int, a_len: int)]
+struct Zip<A, B>;
```
nilehmann: There's no corresponding method in `Iterator` for constructing a `Zip`, so this is currently useless. It'd be a bit complicated to specify the zip method because it involves `IntoIterator`, so I'd just remove it as it's not being used in any tests.

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,60 @@
...
+        fn step_forward(start: int, count: int) -> int { start + count }
+        fn size(lo: int, hi: int) -> int { hi - lo }
+    )]
+    //
+}
+
+#[extern_spec(core::ops)]
+#[assoc(fn valid_item(self: Range<A>, item: A) -> bool { self.start <= item && item < self.end })]
+#[assoc(fn size(self: Range<A>) -> int { <A as Step>::size(self.start, self.end) })]
+#[assoc(fn done(self: Range<A>) -> bool { <A as Step>::size(self.start, self.end) <= 0})]
```
nilehmann:
```suggestion
#[assoc(
    fn valid_item(self: Range<A>, item: A) -> bool { self.start <= item && item < self.end }
    fn size(self: Range<A>) -> int { <A as Step>::size(self.start, self.end) }
    fn done(self: Range<A>) -> bool { <A as Step>::size(self.start, self.end) <= 0 }
)]
```

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,60 @@
...
+impl<A: Step> Iterator for ops::Range<A> {
+    #[spec(
+        fn(self: &mut Range<A>[@old]) -> Option<A[old.start]>[old.start < old.end]
+            ensures self: Range<A>{r: (old.start < old.end => r.start == <A as Step>::step_forward(old.start, 1)) && r.end == old.end }
```
nilehmann:
```suggestion
        ensures self: Range<A>{r: (old.start < old.end => r.start == <A as Step>::step_forward(old.start, 1)) && r.end == old.end }
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[extern_spec(core::iter)]
+#[assoc(
+    fn valid_item(self: Self, item: Self::Item) -> bool { true }
+    fn size(self: Self) -> int { default_iterator_size(self) }
+    fn done(self: Self) -> bool { default_iterator_done(self) }
+    fn step(self: Self, other: Self) -> bool { true }
+)]
+trait Iterator {
+    #[flux::sig(
```
nilehmann:
```suggestion
    #[spec(
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,53 @@
...
+        fn(self: &mut Self[@curr_s]) -> Option<Self::Item>[!<Self as Iterator>::done(curr_s)]
+        ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)}
+    )]
+    fn next(&mut self) -> Option<Self::Item>;
+
+    #[flux::sig(fn(Self[@s]) -> Enumerate<Self>[0, s])]
```
nilehmann:
```suggestion
    #[spec(fn(Self[@s]) -> Enumerate<Self>[0, s])]
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn enumerate(self) -> Enumerate<Self>
+    where
+        Self: Sized;
+
+    #[flux::sig(
```
nilehmann:
```suggestion
    #[spec(
```

lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -0,0 +1,53 @@
...
+    where
+        Self: Sized;
+
+    #[spec(fn(Self[@s], f: F) where F: FnMut(Self::Item{item: <Self as Iterator>::valid_item(s, item)}) -> () )]
+    fn for_each<F>(self, f: F)
+    where
+        Self: Sized,
+        F: FnMut(Self::Item);
+
+    #[flux::sig(fn (Self[@s]) -> B{v: <B as FromIterator<Self::Item>>::with_size(v, <Self as Iterator>::size(s))})]
```
nilehmann:
```suggestion
    #[spec(fn (Self[@s]) -> B{v: <B as FromIterator<Self::Item>>::with_size(v, <Self as Iterator>::size(s))})]
```

lib/flux-core/src/slice/iter.rs
```diff
@@ -10,6 +10,15 @@ impl<'a, T> Iter<'a, T> {
     fn as_slice(&self) -> &'a [T];
 }
 
+#[extern_spec(core::slice)]
+#[flux::assoc(fn size(x: Iter) -> int { x.len - x.idx })]
+#[flux::assoc(fn done(x: Iter) -> bool { x.idx >= x.len })]
+#[flux::assoc(fn step(x: Iter, y: Iter) -> bool { x.idx + 1 == y.idx && x.len == y.len})]
```
nilehmann:
```suggestion
#[assoc(
    fn size(x: Iter) -> int { x.len - x.idx }
    fn done(x: Iter) -> bool { x.idx >= x.len }
    fn step(x: Iter, y: Iter) -> bool { x.idx + 1 == y.idx && x.len == y.len}
)]
```

lib/flux-core/src/slice/iter.rs
```diff
@@ -10,6 +10,15 @@ impl<'a, T> Iter<'a, T> {
...
+impl<'a, T> Iterator for Iter<'a, T> {
+    #[flux::sig(fn(self: &mut Iter<T>[@curr_s]) -> Option<_>[curr_s.idx < curr_s.len] ensures self: Iter<T>{next_s: curr_s.idx + 1 == next_s.idx && curr_s.len == next_s.len})]
```
nilehmann:
```suggestion
    #[spec(fn(self: &mut Iter<T>[@curr_s]) -> Option<_>[curr_s.idx < curr_s.len] ensures self: Iter<T>{next_s: curr_s.idx + 1 == next_s.idx && curr_s.len == next_s.len})]
```

lib/flux-core/src/iter/adapters/map.rs
```diff
@@ -0,0 +1,11 @@
...
+impl<B, I: Iterator, F: FnMut(I::Item) -> B> Iterator for Map<I, F> {}
```
nilehmann: I think there's unsoundness here. This shouldn't work without annotating `next`. Without annotation, we will use the default signature, which shouldn't be a subtype of the signature in the trait. So there's a missing check.

lib/flux-core/src/iter/adapters/skip.rs
```diff
@@ -0,0 +1,10 @@
...
+#[assoc(fn step(self: Skip, other: Skip) -> bool { other.size == self.size - 1 } )]
+impl<I: Iterator> Iterator for Skip<I> {}
```
nilehmann: Same as `Map`, this shouldn't work without annotating `next`

nilehmann (changes requested):

lib/flux-core/src/iter/adapters/skip.rs
ranjitjhala: Unfortunately you can’t quite do that because (if i understand correctly) that requires “stepping” the inner iterator N times (when you construct the Skip). 

thats why instead we’re just erasing the inner iterator more or less and tracking the count. 

one way could be to “generalize” step to take a “count” but I’m not sure that’s strictly preferable?

nilehmann: hmmm, it's implemented by calling `inner.nth` the first time, setting `n` to 0, and then `inner.next` on subsequent calls. So yeah, one would need to be able to talk about taking n steps. I suppose we would need that if we want to specify `nth`.

I don't think we should specify everything, but I wish we had a more principled way to only track some aspects of the `Iterator`. I suppose the way it's currently specified is fine for now.

ranjitjhala: I want to fiddle around with generalizing step; lets see what happens...

lib/flux-core/src/iter/adapters/map.rs
ranjitjhala: Hmm bit unfortunate that we don't automatically "specialize" the trait-method `spec` to the `impl`...

lib/flux-core/src/iter/adapters/skip.rs
ranjitjhala: done

lib/flux-core/src/iter/adapters/skip.rs
nilehmann: Also, the current spec can be problematic if you use interior mutability to change the size of the inner iterator. Consider:


```rust
use std::cell::RefCell;

struct Wrapper<'a, I> {
    inner: &'a RefCell<I>
}

impl<'a, I> Wrapper<'a, I> {
    fn new(inner: &'a RefCell<I>) -> Self {
        Wrapper { inner }
    }
}

impl<I: Iterator> Iterator for Wrapper<'_, I> {
    type Item = I::Item;
    
    fn next(&mut self) -> Option<Self::Item> {
        self.inner.borrow_mut().next()
    }
}

fn foo(x: &[i32]) {
    let iter1 = RefCell::new(x.iter());
    let iter2 = Wrapper::new(&iter1);
    
    let mut iter3 = iter2.skip(1);
    iter1.borrow_mut().next(); // this is changing the size of the inner iterator
    for x in iter3 {
        println!("{x}");
    }
}

fn main() {
    foo(&[1, 2, 3, 4, 5]);
}
```

Whether this is unsound depends on how you specify `Wrapper`, but in general, I'm unsure about the interaction between this and other weird iterators.

lib/flux-core/src/iter/adapters/map.rs
ranjitjhala: (have added `next` for now)

lib/flux-core/src/iter/adapters/skip.rs
nilehmann: hmmm, I guess interior mutability is fine because it's the responsibility of the interiorly mutable iterator not to promise a size it can't uphold.

lib/flux-core/src/iter/adapters/zip.rs
ranjitjhala: ok removed.

lib/flux-core/src/iter/adapters/skip.rs
ranjitjhala: letting this be.

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,62 @@
...
+    #![reft(
+        fn step_forward(start: Self, count: int) -> Self {
+            default_step_step_forward(start, count)
+        }
+        fn size(lo: Self, hi: Self) -> int {
+            default_step_size(lo, hi)
+        }
+    )]
+    //
+}
```
nilehmann: Let's be consistent; we should either use `assoc` everywhere or `reft` everywhere.

```suggestion
#[assoc(
    fn step_forward(start: Self, count: int) -> Self {
        default_step_step_forward(start, count)
    }
    fn size(lo: Self, hi: Self) -> int {
        default_step_size(lo, hi)
    }
)]
trait Step { }
```

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[extern_spec(core::iter)]
+impl Step for usize {
+    #![reft(
+        fn step_forward(start: int, count: int) -> int { start + count }
+        fn size(lo: int, hi: int) -> int { hi - lo }
+    )]
+    //
+}
```
nilehmann:
```suggestion
#[assoc(
    fn step_forward(start: int, count: int) -> int { start + count }
    fn size(lo: int, hi: int) -> int { hi - lo }
)]
impl Step for usize { }
```

lib/flux-core/src/iter/range.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[extern_spec(core::iter)]
+impl Step for i32 {
+    #![reft(
+        fn step_forward(start: int, count: int) -> int { start + count }
+        fn size(lo: int, hi: int) -> int { hi - lo }
+    )]
+    //
+}
```
nilehmann:
```suggestion
#![assoc(
    fn step_forward(start: int, count: int) -> int { start + count }
    fn size(lo: int, hi: int) -> int { hi - lo }
)]
impl Step for i32 { }
```

lib/flux-core/src/iter/adapters/map.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#[extern_spec(core::iter)]
+#[assoc(
+    fn size(x: Map<I>) -> int { <I as Iterator>::size(x.inner) }
+    fn done(x: Map<I>) -> bool { <I as Iterator>::done(x.inner)}
+    fn step(x: Map<I>, y: Map<I>) -> bool { <I as Iterator>::step(x.inner, y.inner)}
+)]
+impl<B, I: Iterator, F: FnMut(I::Item) -> B> Iterator for Map<I, F> {
+    #[spec(fn(self: &mut Self[@curr_s]) -> Option<B>[!<Self as Iterator>::done(curr_s)]
+             ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)})]
```
nilehmann:
```suggestion
           ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)})]
```

lib/flux-core/src/iter/adapters/map.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#[extern_spec(core::iter)]
+#[refined_by(inner: I)]
+struct Map<I, F>;
+
+#[extern_spec(core::iter)]
+#[assoc(
+    fn size(x: Map<I>) -> int { <I as Iterator>::size(x.inner) }
+    fn done(x: Map<I>) -> bool { <I as Iterator>::done(x.inner)}
+    fn step(x: Map<I>, y: Map<I>) -> bool { <I as Iterator>::step(x.inner, y.inner)}
```
nilehmann:
```suggestion
    fn done(x: Map<I>) -> bool { <I as Iterator>::done(x.inner) }
    fn step(x: Map<I>, y: Map<I>) -> bool { <I as Iterator>::step(x.inner, y.inner) }
```

nilehmann (approved):
A couple of more formatting comments, but otherwise LGTM

lib/flux-core/src/iter/adapters/skip.rs
ranjitjhala: oops my "letting this be" precedes seeing your remarks.

Yes, I think with interior mutability the whole thing is a bit YOLO... :-)

ranjitjhala:
I think we should use `assoc` and just drop `reft` from the parsing?

PR #1171, opened 2025-07-09
Remove "lib/vec.rs"

ranjitjhala:
This moves the stuff in `tests/lib/vec.rs` into a new crate `lib/flux-alloc` 

depends on first merging #1170


ranjitjhala:
@nilehmann here's the separate PR for `vec.rs`

lib/flux-alloc/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// #![no_std]
+#![cfg_attr(flux, feature(allocator_api))]
+
+#[cfg(flux)]
+pub mod vec;
+
+// TODO(RJ): I get an "unused extern crate" warning here,
```
ranjitjhala, line 7: @nilehmann any idea how to deal with this?

nilehmann: For now, we can silence it with a `#[expect(...)]`. In the long term, I think we need to rethink how all these (extern) extern specs are implemented.

PR #1182, opened 2025-07-14
Elaborate function application sorts in `rty`

ranjitjhala:
1. Changes the `rty::Expr::App` to explicitly track the `List<SortArg>` that were inferred during the `sortck`
2. Generalizes `char_to_int` to "polymorphic" ~~`to_int`~~ `cast`
3. Uses the explicit `SortArg` during fixpoint-conversion for ~~`to_int`~~ `cast`.

Allows stuff like

```rust
use flux_rs::attrs::*;

#[spec(fn (x:u8) -> u32[cast(x)])]
pub fn test_int_to_int(x: u8) -> u32 {
    x as u32
}

#[spec(fn (x:char) -> u32[cast(x)])]
pub fn test_char_to_int(x: char) -> u32 {
    x as u32
}

#[spec(fn (x:bool) -> u32[cast(x)])]
pub fn test_bool_to_int(x: bool) -> u32 {
    x as u32
}

#[spec(fn (x:bool) -> u32[cast(x)])]
pub fn test_bool_to_int_with_if(x: bool) -> u32 {
    if x { 1 } else { 0 }
}

pub struct S {}

#[spec(fn(x:S, n:usize[to_int(x)]) -> usize[cast(x)])]
pub fn foo(_x: S, n: usize) -> usize {
    n
}
```

### DIFF
--- a/book/src/guide/specifications.md
+++ b/book/src/guide/specifications.md
@@ -957,3 +957,11 @@ or lemmas. See `tests/tests/pos/surface/primops00.rs` for an example.
 ```rust,noplayground
 {{#include ../../../tests/tests/pos/surface/primops00.rs}}
 ```
+
+## Casting Sorts to Int
+
+You can convert refinements of different sorts to `int` using the `to_int` internal function.
+
+```rust,noplayground
+{{#include ../../../tests/tests/pos/surface/to_int00.rs}}
+```
--- a/book/src/guide/specifications.md
+++ b/book/src/guide/specifications.md
@@ -958,9 +958,10 @@ or lemmas. See `tests/tests/pos/surface/primops00.rs` for an example.
 {{#include ../../../tests/tests/pos/surface/primops00.rs}}
 ```
 
-## Casting Sorts to Int
+## Casting Sorts
 
-You can convert refinements of different sorts to `int` using the `to_int` internal function.
+You can convert refinements of different sorts -- e.g. `int` to `char` or `int` to `bool` --
+using the `cast` internal function.
 
 ```rust,noplayground
 {{#include ../../../tests/tests/pos/surface/to_int00.rs}}
--- a/crates/flux-fhir-analysis/src/wf/sortck.rs
+++ b/crates/flux-fhir-analysis/src/wf/sortck.rs
@@ -37,6 +37,7 @@ pub(super) struct InferCtxt<'genv, 'tcx> {
     sort_of_bty: FxHashMap<FhirId, rty::Sort>,
     sort_of_literal: FxHashMap<FhirId, (rty::Sort, Span)>,
     sort_of_bin_op: FxHashMap<FhirId, (rty::Sort, Span)>,
+    sorts_of_fn_app: FxHashMap<FhirId, (List<rty::SortArg>, Span)>,
     path_args: UnordMap<FhirId, rty::GenericArgs>,
     sort_of_alias_reft: FxHashMap<FhirId, rty::FuncSort>,
 }
@@ -68,6 +69,7 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
             sort_of_bin_op: Default::default(),
             path_args: Default::default(),
             sort_of_alias_reft: Default::default(),
+            sorts_of_fn_app: Default::default(),
         }
     }
 
@@ -291,7 +293,9 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
                 let Some(poly_fsort) = self.is_coercible_to_func(&sort, callee.fhir_id) else {
                     return Err(self.emit_err(errors::ExpectedFun::new(callee.span, &sort)));
                 };
-                let fsort = self.instantiate_func_sort(poly_fsort);
+                let fhir_id = expr.fhir_id;
+                let span = expr.span;
+                let fsort = self.instantiate_func_sort(poly_fsort, fhir_id, span);
                 self.synth_app(fsort, args, expr.span)
             }
             fhir::ExprKind::BoundedQuant(.., body) => {
@@ -399,7 +403,7 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
             ExprRes::GlobalFunc(SpecFuncKind::Thy(itf)) => {
                 Ok(rty::Sort::Func(THEORY_FUNCS.get(&itf).unwrap().sort.clone()))
             }
-            ExprRes::GlobalFunc(SpecFuncKind::CharToInt) => {
+            ExprRes::GlobalFunc(SpecFuncKind::ToInt) => {
                 Ok(rty::Sort::Func(rty::PolyFuncSort::new(
                     List::empty(),
                     rty::FuncSort::new(vec![rty::Sort::Char], rty::Sort::Int),
@@ -506,7 +510,12 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
         Ok(fsort.output().clone())
     }
 
-    fn instantiate_func_sort(&mut self, fsort: rty::PolyFuncSort) -> rty::FuncSort {
+    fn instantiate_func_sort(
+        &mut self,
+        fsort: rty::PolyFuncSort,
+        fhir_id: FhirId,
+        span: Span,
+    ) -> rty::FuncSort {
         let args = fsort
             .params()
             .map(|kind| {
@@ -516,6 +525,8 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
                 }
             })
             .collect_vec();
+        self.sorts_of_fn_app
+            .insert(fhir_id, (List::from_vec(args.clone()), span));
         fsort.instantiate(&args)
     }
 
@@ -796,6 +807,25 @@ impl<'genv> InferCtxt<'genv, '_> {
                 .fully_resolve(&sort)
                 .map_err(|_| self.emit_err(errors::CannotInferSort::new(span)))?;
             self.wfckresults.bin_op_sorts_mut().insert(fhir_id, sort);
+        }
+
+        // Make sure that function applications are fully resolved
+        for (fhir_id, (sort_args, span)) in std::mem::take(&mut self.sorts_of_fn_app) {
+            let mut res = vec![];
+            for sort_arg in &sort_args {
+                let sort_arg = if let rty::SortArg::Sort(sort) = sort_arg {
+                    let sort = self
+                        .fully_resolve(&sort)
+                        .map_err(|_| self.emit_err(errors::CannotInferSort::new(span)))?;
+                    rty::SortArg::Sort(sort)
+                } else {
+                    sort_arg.clone()
+                };
+                res.push(sort_arg);
+            }
+            self.wfckresults
+                .fn_app_sorts_mut()
+                .insert(fhir_id, res.into());
         }
 
         // Make sure all parameters are fully resolved
--- a/crates/flux-fhir-analysis/src/wf/sortck.rs
+++ b/crates/flux-fhir-analysis/src/wf/sortck.rs
@@ -5,6 +5,7 @@ use flux_common::{bug, iter::IterExt, result::ResultExt, span_bug, tracked_span_
 use flux_errors::{ErrorGuaranteed, Errors};
 use flux_infer::projections::NormalizeExt;
 use flux_middle::{
+    THEORY_FUNCS,
     fhir::{self, ExprRes, FhirId, FluxOwnerId, visit::Visitor as _},
     global_env::GlobalEnv,
     queries::QueryResult,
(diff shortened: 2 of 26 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1064,10 +1072,27 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                    [rty::SortArg::Sort(from), rty::SortArg::Sort(to)] if is_id_cast(from, to) => {
+                        self.expr_to_fixpoint(&args[0], scx)
+                    }
+                    [rty::SortArg::Sort(rty::Sort::Bool), rty::SortArg::Sort(rty::Sort::Int)] => {
+                        let expr = rty::Expr::ite(
+                            args[0].clone(),
+                            rty::Expr::constant(rty::Constant::from(1)),
+                            rty::Expr::constant(rty::Constant::from(0)),
+                        );
+                        self.expr_to_fixpoint(&expr, scx)
```
```suggestion
                        Ok(fixpoint::Expr::IfThenElse(Box::new([
                            self.expr_to_fixpoint(&expr, scx)?,
                            fixpoint::Expr::int(1),
                            fixpoint::Expr::int(0),
                        ])));
                        
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1459,6 +1484,30 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         }
     }
 
+    fn define_const_for_to_int(
```
```suggestion
    fn define_const_for_cast(
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -377,6 +377,7 @@ enum ConstKey<'tcx> {
     Alias(FluxDefId, rustc_middle::ty::GenericArgsRef<'tcx>),
     Lambda(Lambda),
     PrimOp(rty::BinOp),
+    ToInt(List<rty::SortArg>),
```
```suggestion
    Cast(List<rty::SortArg>),
```

we could also make this `Cast(rty::Sort, rty::Sort)` to be very explicit.

crates/flux-middle/src/fhir.rs
```diff
@@ -1193,7 +1193,7 @@ pub enum SpecFuncKind {
     /// User-defined functions with a body definition
     Def(FluxDefId),
     /// Char-Int Conversions
-    CharToInt,
+    Cast,
```
nilehmann, line 1196: Adjust documentation

crates/flux-middle/src/sort_of.rs
```diff
@@ -41,6 +43,22 @@ impl GlobalEnv<'_, '_> {
...
+                    List::from_arr([SortParamKind::Sort, SortParamKind::Sort]),
+                    rty::FuncSort::new(
+                        vec![rty::Sort::Var(rty::ParamSort::from(0_usize))],
+                        rty::Sort::Var(rty::ParamSort::from(1_usize)),
+                    ),
+                )
+            }
+        }
+    }
+
```
nilehmann: Please revert this change. At some point, I want to stop treating theory functions as "functions" and have dedicated rules for them, so we can get rid of parametric bit vector sizes. This change goes in the wrong direction.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1459,6 +1484,30 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+        let key = ConstKey::ToInt(sort_args.into());
+        self.const_map
+            .entry(key)
+            .or_insert_with(|| {
+                let fsort = self
+                    .genv
+                    .sort_of_spec_func(&flux_middle::fhir::SpecFuncKind::Cast)
+                    .instantiate(sort_args);
+                let fsort = rty::PolyFuncSort::new(List::empty(), fsort);
+                let sort = scx.func_sort_to_fixpoint(&fsort);
```
nilehmann: Since I suggested getting rid of `sort_of_spec_func`, we can inline the function sort here. We are already relying on `cast` having exactly one argument for the encoding, so this is already a leaky abstraction, and we don't lose much with the extra "coordination".

book/src/guide/specifications.md
```diff
@@ -957,3 +957,11 @@ or lemmas. See `tests/tests/pos/surface/primops00.rs` for an example.
...
 {{#include ../../../tests/tests/pos/surface/primops00.rs}}
 ```
+
+## Casting Sorts to Int
+
+You can convert refinements of different sorts to `int` using the `to_int` internal function.
+
+```rust,noplayground
+{{#include ../../../tests/tests/pos/surface/to_int00.rs}}
+```
```
nilehmann: Update documentation to use `cast`

tests/tests/pos/surface/to_int00.rs
```diff
@@ -1,12 +1,16 @@
 #![flux::defs {
+    fn as_int(x:int) -> int {
+        x
+    }
+
     fn is_ascii_digit(c:char) -> bool {
-        let i = char_to_int(c);
-        48 <= i && i <= 57
+        let i = cast(c);
+        as_int(48) <= i && i <= as_int(57)
```
nilehmann: Hmm, I thought we could use an `int` postfix to force an integer, but even though we are parsing it correctly, it is not working as intended.

```
        48int <= i && i <= 57int
```
We could also implement a fallback behavior for unconstrained num vars so they become int 🤔

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1064,10 +1072,27 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                    }
+                    [rty::SortArg::Sort(rty::Sort::Bool), rty::SortArg::Sort(rty::Sort::Int)] => {
+                        let expr = rty::Expr::ite(
+                            args[0].clone(),
+                            rty::Expr::constant(rty::Constant::from(1)),
+                            rty::Expr::constant(rty::Constant::from(0)),
+                        );
+                        self.expr_to_fixpoint(&expr, scx)
+                    }
+                    // Treat all other to_int conversions as uninterpreted
```
nilehmann: I'm not a fan of casts being silently "ignored" as uninterpreted functions. That's a recipe for "why doesn't my code verify?"

But on the plus side, if everything is uniformly castable, then you can have, for example, a function generic on a sort that is castable to `int` without having any form of bounds.

Maybe we can have the cake and eat it too, if we reject statically casts that don't make sense when both sides are concrete.

crates/flux-infer/src/fixpoint_encoding.rs
nilehmann: Also, it just occurred to me that we can define a single generic uif for `cast`, but maybe we want to move away from emitting those?

ranjitjhala: I changed the Cast to use (Sort, Sort) but I can't follow your second comment ("maybe we want to move away from emitting those"); do you mean we should report those as errors?

nilehmann: Instead of monomorphizing cast in flux we could emit a single generic UIF in fixpoint. But maybe we want to move away from generating generic UIFs in fixpoint.

crates/flux-middle/src/sort_of.rs
ranjitjhala: ok have inlined this back at the usage site inside `sortck`

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: done!

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: Well sometimes you want it _even_ if the types are concrete, e.g. in core they do `u32 -> f64` casts. 

Now you might argue, that you can actually use the (inlined) `|_| ()` as the cast function in that case... (i.e. just replace the unit index?) 

Anyways I'm not super wedded to the UIF business, and I do agree that UIF can cause puzzles downstream!

tests/tests/pos/surface/to_int00.rs
ranjitjhala: should I try to implement this *fallback* in this PR or separately?

ranjitjhala:
@nilehmann I addressed the comments, but there are two remainders:

1. Remove the UIF for non-supported casts.
2. Implement the default-int-literal.

Thoughts?

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: Ok got it. I think its fine to generate generic UIFs in fixpoint too... I don't think this matters too much one way or another. (The bigger issue is whether to use UIFs _at all_ in this case, or just disallow...)

nilehmann: I was thinking that if we stop emitting generic UIFs that would make the transition to the fixpoint Rust reimplementation easier, but that's a long way ahead...

tests/tests/pos/surface/to_int00.rs
ranjitjhala: Hmm its a bit more complicated as we have to also treat `48 <= i` as an `int` operation.

Why is `48` given a `NumVar` in the first place (instead of just being `int` ?

crates/flux-infer/src/fixpoint_encoding.rs
nilehmann: Yes, there are genuine use cases for the uif. The problem is when it's unintended.

This could be implemented as a warn-by-default lint that can be silenced. But we don't have any other examples of that so I don't know how hard that would be.

tests/tests/pos/surface/to_int00.rs
nilehmann: Int literals can be treated as bit vectors, ints, or reals dependening on their inferred use.

The way the fallback would work is that we record the places where we generate a NumVar for an int literal. Then, at the end of sort checking we check if they are solved and if they are not we unify them with int. That should fix the sort of binary relations as well.

ranjitjhala: Right now it seems that we only generate `NumVar` for int-literals. When you do what you're suggesting -- the `fully_resolve` check inside `into_results` ? You can turn the literals to `int` but the `sort_of_bin_op` remains unresolved...? So somehow you'd have to "re-propagate" the int-ness there too, no?

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: How about we have error-by-default but you instead can have a flag to `allow_uninterpreted_cast` ?

nilehmann: Yes, that would have the intended behavior. We could even make it per item if we add it as an option to `InferOpts`.

nilehmann: Also, there are two ways in which one could fail. 

1. During sort checking
2. During fixpoint encoding

I'm thinking about 1.

ranjitjhala: for sure!

tests/tests/pos/surface/to_int00.rs
ranjitjhala: I think best to leave this for a separate PR. 

In this case, it is easy enough to just write a monomorphic wrapper like so

```rust
fn cast_to_int(c:char) -> int {
  cast(c)
}

fn is_ascii_digit(c:char) -> bool {
  let i = cast_to_int(c);
  48 <= i && i <= 57
}
```

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: ah, yes I think that's a ways off! Still maybe we can leave it in here -- with the monomorphization-in-rust -- to plan ahead...

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -798,6 +820,41 @@ impl<'genv> InferCtxt<'genv, '_> {
             self.wfckresults.bin_op_sorts_mut().insert(fhir_id, sort);
         }
 
+        let allow_uninterpreted_cast = if let Some(def_id) = self.owner.def_id() {
+            self.genv.infer_opts(def_id).allow_uninterpreted_cast
+        } else {
+            false
+        };
```
nilehmann: maybe

```suggestion
        let allow_uninterpreted_cast = self
            .owner
            .def_id()
            .map_or(false, |def_id| self.genv.infer_opts(def_id).allow_uninterpreted_cast);
```

ranjitjhala:
@nilehmann lmk if there is more to be done here..

Issue #1188, opened 2025-07-16
Arithmetic Ops with "lazy" overflow checking


ranjitjhala:
From discussion [on zulip]([#verify-std > uint -&gt; uint downcast @ 💬](https://flux-rs.zulipchat.com/#narrow/channel/486369-verify-std/topic/uint.20-.3E.20uint.20downcast/near/528961968))

We want give arith ops, e.g. (+) (say) the following types

- **no-check** : u32[n] -> u32[m] -> u32[n + m]
- **strict** : u32[n] -> u32[m] -> u32[n + m] requires n + m <= u32::max
- **lazy** : u32[n] -> u32[m] -> u32{v: let k = n + m in k <= u32::max => v == k}

Currently the **default** is `no-check` which is unsound; ideally we'd switch to **lazy**.

ranjitjhala:
closed by #1215

PR #1189, opened 2025-07-16
"lazy" casts to smaller widths

ranjitjhala:
When casting from `uint_width1` to `uint_width2` where `width2` is LESS THAN `width1`, 
we can use a **lazy** type of the form

`$uint_width2{v: idx <= uint_max(width2) => v == idx}`

meaning that IF the `idx` is provably within bounds, then it is preserved by the cast.

Allows stuff like

```rust
#[spec(fn(n:u64{n < 100}) -> u32{v: v < 100})]
pub fn uint_uint_lossy(n: u64) -> u32 {
    n as u32
}
```

Will be generalized to address #1188 

**NOTE** No noticeable perf difference in `vtock` (just with cast) ~ 24s on `main` and the new PR branch.

### DIFF
--- a/tests/tests/neg/casts/int_to_int.rs
+++ b/tests/tests/neg/casts/int_to_int.rs
@@ -25,5 +25,5 @@ pub fn uint_int_lossless() -> i32 {
 
 #[flux::sig(fn() -> usize[42])]
 pub fn unsigned_to_usize() -> usize {
-    42u128 as usize
+    43u128 as usize //~ ERROR refinement type
 }


tests/tests/neg/casts/int_to_int.rs
```diff
@@ -25,5 +25,5 @@ pub fn uint_int_lossless() -> i32 {
 
 #[flux::sig(fn() -> usize[42])]
 pub fn unsigned_to_usize() -> usize {
-    42u128 as usize //~ ERROR refinement type
+    42u128 as usize
```
nilehmann: maybe change the output to `usize[43]` like the other cases?

PR #1193, opened 2025-07-17
Basic Support for Detached Specifications

ranjitjhala:
This works now

```rust
#![flux::specs {

   fn inc(n:i32) -> i32{v: n < v};

   fn id(n:i32) -> i32[n];

}]

pub fn inc(n: i32) -> i32 {
    n - 1 //~ ERROR refinement type
}

pub fn id(n: i32) -> i32 {
    n
}
```


crates/flux-syntax/src/parser/mod.rs
```diff
@@ -389,7 +432,7 @@ fn mut_as_strg(inputs: Vec<FnInput>, ensures: &[Ensures]) -> ParseResult<Vec<FnI
 ///             ⟨-> ⟨ty⟩⟩?
 ///             ⟨requires⟩ ⟨ensures⟩ ⟨where⟩
 /// ```
-pub(crate) fn parse_fn_sig(cx: &mut ParseCtxt) -> ParseResult<FnSig> {
+pub(crate) fn parse_fn_sig(cx: &mut ParseCtxt, eof: bool) -> ParseResult<FnSig> {
```
nilehmann: I would use an arbitrary `end` token and pass `token::Semi` for detached specs, making it more consistent with Rust's syntax.

```suggestion
pub(crate) fn parse_fn_sig<T: PeekExpected>(cx: &mut ParseCtxt, end: T) -> ParseResult<FnSig> {
```

crates/flux-driver/src/collector/mod.rs
```diff
@@ -536,6 +542,102 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
             self.specs.infer_opts.insert(def_id, infer_opts);
         }
     }
+
+    fn collect_detached_specs(&mut self, attrs: &mut FluxAttrs, def_id: LocalDefId) -> Result {
```
nilehmann: I'd put the stuff related to detached specs in a separate module (like we do for extern specs). This will probably become complicated soon.

crates/flux-driver/src/collector/mod.rs
```diff
@@ -536,6 +542,102 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
...
+        }
+    }
+}
+
+struct DetachedIdentResolver<'a> {
+    items: &'a mut HashMap<Ident, (surface::DetachedItem, Option<LocalDefId>)>,
+}
+
+impl<'tcx> hir::intravisit::Visitor<'tcx> for DetachedIdentResolver<'_> {
+    fn visit_item(&mut self, item: &'tcx Item<'tcx>) {
```
nilehmann: Oh I thought we were going to use [`TyCtxt::module_children_local`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.module_children_local) like we do [in name resolution](https://github.com/flux-rs/flux/blob/8ff117f14ed85215256bb4f60961abc06ccbbe49/crates/flux-desugar/src/resolver.rs#L693). This api is, in my opinion, slightly nicer to use because you don't need the visitor. It's also a bit lower level so we could eventually reuse it for writing extern specs.

crates/flux-syntax/src/surface.rs
```diff
@@ -110,6 +109,27 @@ pub struct TyAlias {
     pub span: Span,
 }
 
+#[derive(Debug)]
+pub struct DetachedSpecs {
+    pub items: Vec<DetachedItem>,
+}
+
+#[derive(Debug)]
+pub enum DetachedItem {
```
nilehmann: I'm hoping we can reuse this definition for normal specs so we can make the code in name resolution nicer, so I prefer to call this simply `Item` (and rename the current `Item` to `FluxItem`). But we can always rename this later.

Right now we have separate functions for different type of items, but they all behave uniformly, e.g.
https://github.com/flux-rs/flux/blob/8ff117f14ed85215256bb4f60961abc06ccbbe49/crates/flux-desugar/src/resolver.rs#L314)
https://github.com/flux-rs/flux/blob/8ff117f14ed85215256bb4f60961abc06ccbbe49/crates/flux-desugar/src/resolver.rs#L322
If we had a node for items we could just visit the ast.

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: done.

crates/flux-syntax/src/surface.rs
ranjitjhala: done!

crates/flux-driver/src/collector/mod.rs
ranjitjhala: Did we decide to make this change to module-children or keep the hir-visitor?

nilehmann: module children

ranjitjhala: Hmm there is something odd. 

Note that we want to traverse the `ModuleChildren` recursively (for nested `mod`).

The odd thing is that `ModChild` only has the `DefId` of the inner definitions, not the `LocalDefId`.

However to (recursively) get the children, we need the `LocalDefId`? 

I suppose we could use [`as_local`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/def_id/struct.DefId.html#method.as_local)

Is that what you had in mind?

ranjitjhala: ok, have switched to the `module_children_local` (should be easy enough to revert if we want...)

ranjitjhala:
Have addressed the above.

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+}
+
+///```text
+/// <field>  ::= refined_by(<refined_by>)?
+///              invariant(<expr>)?
+/// ```
+fn parse_detached_refined_by_and_invariants(
+    cx: &mut ParseCtxt,
+) -> ParseResult<(Option<Vec<RefineParam>>, Vec<Expr>)> {
+    let tok_refined_by = token::Ident(Symbol::intern("refined_by"));
```
nilehmann: You have to add this as a pre-interned keyword or symbol to [`flux-syntax::symbols`](https://github.com/flux-rs/flux/blob/main/crates/flux-syntax/src/symbols.rs)

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    let tok_invariant = token::Ident(Symbol::intern("invariant"));
+    let refined_by = if cx.peek(tok_refined_by) {
+        cx.expect(tok_refined_by)?;
+        Some(parens(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?)
+    } else if cx.peek(token::OpenBracket) {
+        Some(brackets(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?)
+    } else {
+        None
+    };
```
nilehmann: We said we were going to pick just one

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+/// ```
+fn parse_detached_field(cx: &mut ParseCtxt) -> ParseResult<(Ident, Ty)> {
+    let ident = parse_ident(cx)?;
+    cx.expect(token::Colon)?;
+    let ty = parse_type(cx)?;
+    Ok((ident, ty))
+}
+
+///```text
+/// <field>  ::= refined_by(<refined_by>)?
```
nilehmann: This is not a field.

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
     } else if lookahead.peek(kw::Property) {
-        parse_primop_property(cx).map(Item::PrimOpProp)
+        parse_primop_property(cx).map(FluxItem::PrimOpProp)
     } else {
         Err(lookahead.into_error())
     }
 }
 
+///```text
+/// <specs> ::= <spec>*
```
nilehmann: Use `⟨` and `⟩` for all these comments

```suggestion
/// ⟨specs⟩ ::= ⟨spec⟩*
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    let refined_by = if cx.peek(tok_refined_by) {
+        cx.expect(tok_refined_by)?;
+        Some(parens(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?)
+    } else if cx.peek(token::OpenBracket) {
+        Some(brackets(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?)
+    } else {
+        None
+    };
+    let invariants = if cx.peek(tok_invariant) {
+        cx.expect(tok_invariant)?;
```
nilehmann: Use `advance_if`

```suggestion
    let invariants = if cx.advance_if(tok_invariant) {
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+        vec![delimited(cx, Parenthesis, |cx| parse_expr(cx, true))?]
+    } else {
+        vec![]
+    };
+    Ok((refined_by, invariants))
+}
+
+///```text
+/// <field>  ::= <Ident> : <Ty>
```
nilehmann: This comment is wrong

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,21 +94,147 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+/// ```
+fn parse_detached_variant(cx: &mut ParseCtxt) -> ParseResult<(Ident, VariantDef)> {
+    let ident = parse_ident(cx)?;
+    let variant = parse_variant(cx, true)?;
```
nilehmann: Maybe this could be cleaner if we make the `ident` optional inside `parse_variant`, similar to how the ident is optional for function specs.

crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -0,0 +1,193 @@
...
+        &mut self,
+        ident: Ident,
+        owner_id: OwnerId,
+        struct_def: surface::StructDef,
+    ) -> Result {
+        let entry = self
+            .inner
+            .specs
+            .structs
+            .entry(owner_id)
```
nilehmann: This is a bit confusing. I'd instead either

1. Unconditionally insert `struct_def`. This will return `Some` if there was an element before, so you can then match on that.
2. Match on the entry and handle the occupied and vacant cases independently.

I think 2 will eventually be better because you have access to the old and new values to report the error.

crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+fn resolve_idents_in_scope(
+    tcx: TyCtxt,
+    scope: LocalDefId,
+    items: &mut HashMap<Ident, (surface::Item, Option<DefId>)>,
+) {
+    for child in tcx.module_children_local(scope) {
+        let ident = child.ident;
+        if let Res::Def(kind, def_id) = child.res {
```
nilehmann: Maybe we can start like this to avoid some nesting and repetition.

```rust
        let Some(val) = items.get_mut(&ident) else { continue };
        let Res::Def(kind, def_id) = child.res else { continue };
        ...
```

crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -0,0 +1,193 @@
...
+            if let DefKind::Fn = kind
+                && let Some(val) = items.get_mut(&ident)
+                && matches!(val.0, surface::Item::FnSig(_, _))
+                && val.1.is_none()
```
nilehmann: I'm not sure if this can happen in practice, but if this is `Some`, we should probably report an ambiguity error. 

Also, I'm not sure if we should find a match _if_ the `DefKind` matches the `surface::Item` or if we should _expect_ them to match.

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: I couldn't remember which, do you recall?

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: good idea!

ranjitjhala:
@nilehmann I think this is ready for your perusal -- I've added all the features we need for the next round of `verify-std`, namely `fn`, `struct`, `enum`, `impl`, `impl-for`, `trait` and `reft`.

I might still have to fiddle around a bit to get the above to work nicely with the `mod` ...

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+///```text
+/// ⟨specs⟩ ::= ⟨fn-spec⟩
+///           | ⟨struct-spec⟩
+///           | ⟨enum-spec⟩
+///           | mod  <PATH> { ⟨specs⟩ }
+///           | impl <PATH> { ⟨impl-spec⟩ }
+/// ```
+pub(crate) fn parse_detached_item(cx: &mut ParseCtxt) -> ParseResult<Item> {
+    let mut lookahead = cx.lookahead1();
+    if lookahead.peek(kw::Fn) || lookahead.peek(token::Pound) {
```
nilehmann: I think a more elegant way to implement this is to parse a list of attributes unconditionally before the lookahead (and collect them into a vec). This, however, adds quite a lot more failures to consider that should be reported after parsing (duplicated attributes, attributes used on the wrong items, ...).

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    cx.expect(token::Colon)?;
+    let ty = parse_type(cx)?;
+    Ok((ident, ty))
+}
+
+///```text
+/// ⟨refine_info⟩ ::= [(Ident : ⟨sort⟩)*]?
+///                   invariant(<expr>)?
+/// ```
+fn parse_detached_refine_info(
```
nilehmann: I am now leaning towards parsing `refined_by` and `invariant` as attributes. I don't think you gain much readability from the "inline" syntax, and probably the less we deviate from the attached syntax, the better. relates to https://github.com/flux-rs/flux/pull/1193#discussion_r2237461161 if we decide to implement it as attributes

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+        parse_detached_impl(cx)
+    } else if lookahead.peek(kw::Trait) {
+        parse_detached_trait(cx)
+    } else {
+        Err(lookahead.into_error())
+    }
+}
+
+///```text
+/// ⟨field⟩ ::= Ident : ⟨Ty⟩
```
```suggestion
/// ⟨field⟩ ::= ⟨ident⟩ : ⟨type⟩
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    let invariants = if cx.advance_if(kw::Invariant) {
+        vec![delimited(cx, Parenthesis, |cx| parse_expr(cx, true))?]
+    } else {
+        vec![]
+    };
+    Ok((refined_by, invariants))
+}
+
+///```text
+/// ⟨enum⟩ := enum Ident ⟨refine_info⟩ { ⟨variant⟩* }
```
nilehmann, line 164:
```suggestion
/// ⟨enum⟩ := enum ⟨ident⟩ ⟨refine_info⟩ { ⟨variant⟩* }
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    let span = fn_sig.span;
+    let ident = fn_sig
+        .ident
+        .ok_or(ParseError { kind: crate::ParseErrorKind::InvalidDetachedSpec, span })?;
+    let fn_spec = FnSpec { fn_sig: Some(fn_sig), qual_names: None, reveal_names: None, trusted };
+    Ok(Item { ident, kind: fn_spec })
+}
+
+///```text
+/// ⟨mod⟩ ::= mod Ident { ⟨specs⟩ }
```
```suggestion
/// ⟨mod⟩ ::= mod ⟨ident⟩ { ⟨specs⟩ }
```

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -94,29 +97,232 @@ pub(crate) fn parse_flux_items(cx: &mut ParseCtxt) -> ParseResult<Vec<Item>> {
...
+    let items = until(cx, token::Eof, parse_detached_item)?;
+    Ok(surface::DetachedSpecs { items })
+}
+
+///```text
+/// ⟨specs⟩ ::= ⟨fn-spec⟩
+///           | ⟨struct-spec⟩
+///           | ⟨enum-spec⟩
+///           | mod  <PATH> { ⟨specs⟩ }
+///           | impl <PATH> { ⟨impl-spec⟩ }
```
```suggestion
///           | ⟨mod⟩
///           | ⟨impl⟩
```

crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -0,0 +1,420 @@
...
+                surface::ItemKind::TraitImpl(detached_impl) => {
+                    let key = (ImplKey(item.ident.name), detached_impl.trait_.name);
+                    let span = item.ident.span;
+                    if let std::collections::hash_map::Entry::Vacant(e) = trait_impls.entry(key) {
+                        e.insert((detached_impl, None, span));
+                    } else {
+                        return Err(collector.errors.emit(errors::AttrMapErr {
+                            span,
+                            message: format!("multiple impls for `{}`", item.ident.name),
+                        }));
```
nilehmann: You shouldn't reuse `AttrMapErr` which is intended for errors when parsing attributes of the form `#[name(key1=val1,...)]`. Also, if you use this, the wording will be quite confusing because it says `invalid attribute: multiple impls for ident`, an talking about the detached spec as an attribute doesn't make much sense. This applies to all the places in this module where you use `AttrMapErr`

nilehmann (commented):
I haven't looked too much at the implementation in `detached_specs.rs`, but that's well isolated so I'm not too worried because we can change it later. 

I'm more worried about the concrete syntax now.

crates/flux-driver/src/collector/detached_specs.rs
nilehmann: @ranjitjhala you haven't addressed this comment

crates/flux-driver/src/collector/detached_specs.rs
nilehmann: @ranjitjhala you haven't addressed this comment

crates/flux-driver/src/collector/detached_specs.rs
nilehmann: @ranjitjhala you haven't addressed this comment

crates/flux-driver/src/collector/detached_specs.rs
ranjitjhala: Oops sorry I missed this! That approach -- match on occ/variant is a bit gross because it turns out we seem to add some default "trivial" for structs, and so you have to duplicate the actual "attaching" code in the `Occupied` and `Vacant` cases...

ranjitjhala: meh its not so bad, have replaced with the explicit entry match.

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: have reverted to `refined_by`

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: Yes, also the specific attributes currently supported -- `hide`, `assoc` and `trusted` -- are always "disjoint", i.e. I can't think of any situation where you want more than one...

crates/flux-syntax/src/parser/mod.rs
nilehmann: By attribute I mean 

```rust
#[refined_by(x: int)]
#[invariant(...)]
struct Struct {
   ...
}
```

crates/flux-driver/src/collector/detached_specs.rs
ranjitjhala: oops, good catch!

crates/flux-driver/src/collector/detached_specs.rs
ranjitjhala: re: expect -- can a `fn` name (`Ident`) match that of a `mod` ?

ranjitjhala: I suppose if that happens its bad.. better to "expect".

crates/flux-driver/src/collector/detached_specs.rs
ranjitjhala: indeed, have split into separate errors.

crates/flux-syntax/src/parser/mod.rs
nilehmann: Functions can have multiple attributes `#[trusted]`, `#[qualifiers(...)]`, `#[opts(...)]` and if we implement https://github.com/flux-rs/flux/pull/1193#discussion_r2237472458 then structs and enums could have multiple ones too.

crates/flux-driver/src/collector/detached_specs.rs
ranjitjhala: have made it an "expect"

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: ok, have switched to use lists-of-attrs for `invariant` and `refined_by` ... (am deferring fancier error reporting to later though...)

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: done.

ranjitjhala:
The "AI" seems to think

> The job failed because the script tried to download a binary from https://github.com/ucsd-progsys/liquid-fixpoint/releases/download/nightly/fixpoint-aarch64-apple-darwin, but the server returned error 503 (Service Unavailable). This caused curl to exit with code 56, which failed the job.

but that URL itself seems to be working fine?

nilehmann:
@ranjitjhala seems GitHub is having some downtime and it's randomly failing. Rerunning the job seems to have fixed it. I'll add retry logic to the download step to account for flukes like this.

ranjitjhala:
@nilehmann sorry I rewrote this code because it was bugging me, and to fix an issue with core. 

The syntax is still the same, but I think this follows your suggested style, do you want to take a look?

nilehmann (approved):
LGTM

PR #1223, opened 2025-08-07
Strict underflow checking

ranjitjhala:
Add a `"strict-under"` mode which is like `"lazy"` but where we strictly check that unsigned operators preserve non-negativity.

### DIFF
--- a/crates/flux-refineck/src/primops.rs
+++ b/crates/flux-refineck/src/primops.rs
@@ -58,6 +58,7 @@ pub(crate) fn match_bin_op(
         OverflowMode::Strict => &OVERFLOW_STRICT_BIN_OPS,
         OverflowMode::Lazy => &OVERFLOW_LAZY_BIN_OPS,
         OverflowMode::None => &OVERFLOW_NONE_BIN_OPS,
+        OverflowMode::StrictUnder => &OVERFLOW_STRICT_UNDER_BIN_OPS,
     };
     table.match_inputs(&op, [(bty1.clone(), idx1.clone()), (bty2.clone(), idx2.clone())])
 }
@@ -72,6 +73,7 @@ pub(crate) fn match_un_op(
         OverflowMode::Strict => &OVERFLOW_STRICT_UN_OPS,
         OverflowMode::None => &OVERFLOW_NONE_UN_OPS,
         OverflowMode::Lazy => &OVERFLOW_LAZY_UN_OPS,
+        OverflowMode::StrictUnder => &OVERFLOW_STRICT_UNDER_UN_OPS,
     };
     table.match_inputs(&op, [(bty.clone(), idx.clone())])
 }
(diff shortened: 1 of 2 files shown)


crates/flux-refineck/src/primops.rs
```diff
@@ -235,6 +280,20 @@ fn mk_add_rules(overflow_mode: OverflowMode) -> RuleMatcher<2> {
             }
         }
 
+        // like Lazy, but we also check for underflow on unsigned addition
+        OverflowMode::StrictUnder => {
```
nilehmann: It doesn't make sense to check for underflow for addition, only for substraction.

crates/flux-refineck/src/primops.rs
```diff
@@ -197,6 +229,15 @@ static OVERFLOW_LAZY_UN_OPS: LazyLock<RuleTable<mir::UnOp, 1>> = LazyLock::new(|
     }
 });
 
+static OVERFLOW_STRICT_UNDER_UN_OPS: LazyLock<RuleTable<mir::UnOp, 1>> = LazyLock::new(|| {
```
nilehmann: You can't negate an unsigned integer so unary operators are not affected by the new mode.

nilehmann (changes requested):

crates/flux-refineck/src/primops.rs
```diff
@@ -316,6 +389,20 @@ fn mk_sub_rules(overflow_mode: OverflowMode) -> RuleMatcher<2> {
...
         }
 
+        // like Lazy, but we also check for underflow on unsigned subtraction
+        OverflowMode::StrictUnder => {
+            primop_rules! {
+                fn(a: T, b: T) -> T{v: E::implies(valid_int(a - b, int_ty), E::eq(v, a - b)) }
+                if let &BaseTy::Int(int_ty) = T
+
+                fn(a: T, b: T) -> T{v: E::implies(valid_uint(a - b, uint_ty), E::eq(v, a - b)) }
+                requires nonnegative(a - b) => ConstrReason::Overflow // check for underflow
```
nilehmann: Since we are at this, maybe we can have a dedicated reason, i.e., an error message. The one you get right now talks about overflow, which can be confusing.

crates/flux-refineck/src/primops.rs
nilehmann: hmm, now that I think about it, the whole table thing is a bit weird since some operations are not affected. maybe we should push the branching on the mode into the rule itself.

crates/flux-refineck/src/primops.rs
ranjitjhala: duh! thanks!

crates/flux-refineck/src/primops.rs
ranjitjhala: I agree, there is also too much code duplication IMO.

crates/flux-refineck/src/primops.rs
```diff
@@ -316,6 +361,20 @@ fn mk_sub_rules(overflow_mode: OverflowMode) -> RuleMatcher<2> {
             }
         }
 
+        // like Lazy, but we also check for underflow on unsigned subtraction
```
nilehmann, line 364: The same here, there can't be underflow when multiplying unsigned integers

crates/flux-refineck/src/primops.rs
```diff
@@ -275,6 +306,20 @@ fn mk_mul_rules(overflow_mode: OverflowMode) -> RuleMatcher<2> {
             }
         }
 
+        // like Lazy, but we also check for underflow on unsigned multiplication
```
nilehmann: There can't be underflow when multiplying

nilehmann (approved):

PR #1224, opened 2025-08-07
Improve how types are pretty printed (by way of many hacks)

cole-k:
The gist of this PR is converting flux's type pretty printing to match its own syntax.

<details>
<summary> More details </summary>
We have to do all sorts of hacks to accomplish this because the desugaring process does not track the original syntax. I made some efforts to ensure that the outputs are not only valid syntax, but close to the syntax that would have been written (e.g. if we desugar `u32` into `{v. u32[v]}`, I have hacks to remove the existential that is unnecessary in the frontend).
</details>

This doesn't affect anything user-facing. There are some todos (`// TODO`).

<details>
<summary> What needs to be done? </summary>
Anything related to `PrettyNested`.
Additionally, I didn't incorporate redundant bound var elimination (see previous section) for `Lambdas`, nor did I change their pretty printed syntax. Functions in the `Ty` pretty printing itself should convert lambdas into existentials.
</details>

### DIFF
--- a/crates/flux-middle/src/rty/canonicalize.rs
+++ b/crates/flux-middle/src/rty/canonicalize.rs
@@ -435,7 +435,7 @@ mod pretty {
                 CanonicalTy::Constr(constr) => w!(cx, f, "{:?}", constr),
                 CanonicalTy::Exists(poly_constr) => {
                     let redundant_bvars = poly_constr.skip_binder_ref().redundant_bvars().into_iter().collect();
-                    cx.with_bound_vars_removable(poly_constr.vars(), redundant_bvars, |f_body| {
+                    cx.with_bound_vars_removable(poly_constr.vars(), redundant_bvars, None, |f_body| {
                         let constr = poly_constr.skip_binder_ref();
                         if constr.pred().is_trivially_true() {
                             w!(cx, f_body, "{:?}", &constr.ty)
--- a/crates/flux-middle/src/rty/canonicalize.rs
+++ b/crates/flux-middle/src/rty/canonicalize.rs
@@ -30,8 +30,7 @@
 //!
 //! [existentials]: TyKind::Exists
 //! [constraint predicates]: TyKind::Constr
-use std::ops::ControlFlow;
-use std::fmt::Write;
+use std::{fmt::Write, ops::ControlFlow};
 
 use flux_arc_interner::List;
 use flux_macros::{TypeFoldable, TypeVisitable};
@@ -40,7 +39,9 @@ use rustc_ast::Mutability;
 use rustc_type_ir::{BoundVar, INNERMOST};
 
 use super::{
-    fold::{TypeFoldable, TypeFolder, TypeSuperFoldable, TypeVisitor, TypeVisitable}, BaseTy, Binder, BoundVariableKind, Expr, FnSig, GenericArg, GenericArgsExt, PolyFnSig, SubsetTy, Ty, TyCtor, TyKind, TyOrBase
+    BaseTy, Binder, BoundVariableKind, Expr, FnSig, GenericArg, GenericArgsExt, PolyFnSig,
+    SubsetTy, Ty, TyCtor, TyKind, TyOrBase,
+    fold::{TypeFoldable, TypeFolder, TypeSuperFoldable, TypeVisitable, TypeVisitor},
 };
 use crate::rty::{ExprKind, HoleKind};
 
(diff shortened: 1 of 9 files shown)


cole-k:
@ranjitjhala can you rebase your branch that needs these changes on top and tell me if the sigs look OK? I can probably make some small changes if you can give me specific examples of things not rendering correctly.

crates/flux-middle/src/rty/canonicalize.rs
```diff
@@ -352,6 +354,15 @@ pub enum CanonicalTy {
...
 }
 
+impl TypeVisitable for CanonicalTy {
+    fn visit_with<V: TypeVisitor>(&self, visitor: &mut V) -> ControlFlow<V::BreakTy> {
+        match self {
+            Self::Constr(constr_ty) => constr_ty.visit_with(visitor),
+            Self::Exists(binder) => binder.visit_with(visitor),
+        }
+    }
+}
```
nilehmann: You can `#[derive(TypeVisitable)]`

crates/flux-middle/src/rty/fold.rs
```diff
@@ -221,6 +222,130 @@ pub trait TypeVisitable: Sized {
...
+    /// versus the (unlikely) edge case
+    ///
+    ///     exists b0. RMat<f32>[{rows: b0, cols: b0}]
+    ///
+    /// for which the existential is now necessary.
+    ///
+    /// NOTE: this only applies to refinement bvars.
+    fn redundant_bvars(&self) -> Vec<BoundVar> {
+        struct RedundantBVarFinder {
+            debruijn: DebruijnIndex,
```
nilehmann: We typically call this `current_index`

crates/flux-middle/src/rty/fold.rs
```diff
@@ -221,6 +222,130 @@ pub trait TypeVisitable: Sized {
...
+            bvar_occurrences: FxHashMap<BoundVar, usize>,
+            bvar_index_occurrences: FxHashMap<BoundVar, usize>,
```
nilehmann: I think we can make this an `FxHashSet<BoundVar>`, no?

crates/flux-middle/src/rty/fold.rs
```diff
@@ -221,6 +222,130 @@ pub trait TypeVisitable: Sized {
...
+            bvar_occurrences: FxHashMap::default(),
+            bvar_index_occurrences: FxHashMap::default(),
+        };
+        let _ = self.visit_with(&mut finder);
+
+        finder
+            .bvar_index_occurrences
+            .keys()
+            .copied()
+            .filter(|var_index| *finder.bvar_occurrences.get(var_index).unwrap_or(&0) == 1)
```
nilehmann: Maybe

```suggestion
            .filter(|var_index| *finder.bvar_occurrences.get(var_index) == Some(&1))
```

crates/flux-middle/src/rty/fold.rs
```diff
@@ -221,6 +222,130 @@ pub trait TypeVisitable: Sized {
...
+                {
+                    self.bvar_occurrences
+                        .entry(*var)
+                        .and_modify(|count| {
+                            *count += 1;
+                        })
+                        .or_insert(1);
+                }
+                e.super_visit_with(self)
+            }
```
nilehmann, line 287:
```suggestion
            }
            
```

crates/flux-middle/src/rty/fold.rs
```diff
@@ -221,6 +222,130 @@ pub trait TypeVisitable: Sized {
...
+    ///
+    /// which may be rewritten as `RMat<f32>`,
+    /// versus the (unlikely) edge case
+    ///
+    ///     exists b0. RMat<f32>[{rows: b0, cols: b0}]
+    ///
+    /// for which the existential is now necessary.
+    ///
+    /// NOTE: this only applies to refinement bvars.
+    fn redundant_bvars(&self) -> Vec<BoundVar> {
```
nilehmann: Every time you call this function, you immediately turn it into a set, maybe we can return a set right away.

```suggestion
    fn redundant_bvars(&self) -> FxHashSet<BoundVar> {
```

nilehmann (commented):
A couple of small comments. There are various parts I haven't looked at yet.

crates/flux-middle/src/rty/fold.rs
cole-k: I don't think so. We're looking for bvars that occur a single time. We could accomplish this maybe using two sets or an enum instead of a `usize`.

cole-k:
I fixed the nits. I want to emphasize that this is (1) untested and (2) doesn't have `PrettyNested` implemented, but I'm pretty sure I grabbed all the commits that make the types pretty in my other branch. We can always roll it back if it doesn't do anything.

crates/flux-middle/src/rty/fold.rs
nilehmann: You are using this field to collect the variables that appear in index position, not to count them. `bvar_occurrences` is used to count.

cole-k: Ah, fair point. I'll rename the fields too.

nilehmann (approved):

PR #1225, opened 2025-08-08
`hyperlink` events in trace to support jump to def in vscode

ranjitjhala:


crates/flux-middle/src/def_id.rs
```diff
@@ -22,18 +22,36 @@ use rustc_span::Symbol;
 ///   when item existence is guaranteed
 /// * The type is parametric over the parent `Id` type to support various id types (e.g., [`DefId`],
 ///   [`MaybeExternId`])
-#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, Encodable, Decodable)]
+#[derive(Debug, Copy, Clone, Encodable, Decodable)]
 pub struct FluxId<Id> {
     parent: Id,
     name: Symbol,
+    span: Span,
```
nilehmann: I really don't like this. Please create a query `flux_def_ident_span` (similar to `TyCtxt::def_ident_span`) and query the span when you need it.

crates/flux-middle/src/fhir.rs
```diff
@@ -1069,6 +1081,21 @@ pub struct PathExpr<'fhir> {
...
+    pub fn new(
+        tcx: TyCtxt,
+        res: ExprRes,
+        segments: &'fhir [Ident],
+        fhir_id: FhirId,
+        span: Span,
+    ) -> Self {
+        if let Some(dst_span) = res.def_span(tcx) {
+            dbg::hyperlink!(tcx, span, dst_span);
+        }
```
nilehmann: I don't like having the side effect here. All the places we create a `PathExpr` are in desugar, so we can have a method in the context.

nilehmann (changes requested):

crates/flux-middle/src/def_id.rs
ranjitjhala: Can't quite do that, because the `DefId` alone is the wrong one -- its some parent etc. -- so we have to actually track the span ourselves. Nevertheless, this felt a bit icky and triggered a bug with the `Eq` instance, so I'll track the `Span` elsewhere.

nilehmann: That's what I mean by create a query. The query should record the span that we have in the FluxItem associated to a particular FluxId. A query is required if you want links to definitions in external crates, otherwise we can grab the span for local items from the fhir::FluxItem

ranjitjhala: ok have made it a query, you can take a look to see if that's what you had in mind?

crates/flux-middle/src/fhir.rs
ranjitjhala: have moved this whole thing to `conv`, please take a look

ranjitjhala:
have made the above edits, I think -- please take a look!

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -86,6 +88,10 @@ fn func_sort(genv: GlobalEnv, def_id: FluxId<MaybeExternId>) -> rty::PolyFuncSor
     }
 }
 
+fn func_span(genv: GlobalEnv, def_id: FluxId<MaybeExternId>) -> Span {
+    genv.fhir_spec_func_body(def_id.local_id()).unwrap().span
+}
```
nilehmann: We can make this generic for any flux definition, not just functions. Let's call it `flux_def_ident_span`. We must be careful because a `FluxId` can point either to a `FluxItem` or to an `AssocRefinement`. Also we probably want the span of the identifier not of the entire definition.

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2083,11 +2088,13 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
     fn conv_expr(&mut self, env: &mut Env, expr: &fhir::Expr) -> QueryResult<rty::Expr> {
         let fhir_id = expr.fhir_id;
         let espan = ESpan::new(expr.span);
+        let tcx = self.tcx();
         let expr = match expr.kind {
             fhir::ExprKind::Var(var, _) => {
                 match var.res {
                     ExprRes::Param(..) => env.lookup(&var).to_expr(),
                     ExprRes::Const(def_id) => {
+                        dbg::hyperlink!(tcx, var.span, tcx.def_span(def_id));
```
nilehmann:
```suggestion
                        dbg::hyperlink!(tcx, var.span, tcx.def_ident_span(def_id));
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2098,11 +2105,13 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                         }
                     }
                     ExprRes::Variant(variant_def_id) => {
+                        dbg::hyperlink!(tcx, var.span, tcx.def_span(variant_def_id));
```
nilehmann:
```suggestion
                        dbg::hyperlink!(tcx, var.span, tcx.def_ident_span(variant_def_id));
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2098,11 +2105,13 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
...
                         let enum_def_id = self.tcx().parent(variant_def_id);
                         let idx = variant_idx(self.tcx(), variant_def_id);
                         rty::Expr::ctor_enum(enum_def_id, idx)
                     }
                     ExprRes::ConstGeneric(def_id) => {
+                        dbg::hyperlink!(tcx, var.span, tcx.def_span(def_id));
```
nilehmann:
```suggestion
                        dbg::hyperlink!(tcx, var.span, tcx.def_ident_span(def_id));
```

crates/flux-desugar/src/desugar.rs
```diff
@@ -54,6 +54,16 @@ pub(crate) fn desugar_qualifier<'genv>(
     })
 }
 
+fn path_expr<'genv>(
```
nilehmann: I don't think any of the changes in this file are needed. Please revert them

crates/flux-desugar/src/lib.rs
```diff
@@ -309,7 +309,8 @@ impl CrateDesugar<'_, '_> {
     fn desugar_flux_items(&mut self, specs: &Specs) {
         for (parent, items) in &specs.flux_items_by_parent {
             for item in items {
-                let def_id = FluxLocalDefId::new(parent.def_id, item.name().name);
+                let ident = item.name();
```
nilehmann: Same here, these changes are not needed.

crates/flux-desugar/src/resolver.rs
```diff
@@ -100,12 +100,14 @@ impl<'genv, 'tcx> CrateResolver<'genv, 'tcx> {
             for item in items {
                 match item {
                     surface::FluxItem::Qualifier(qual) => {
-                        let def_id = FluxLocalDefId::new(parent.def_id, qual.name.name);
-                        self.qualifiers.insert(qual.name.name, def_id);
+                        let name = qual.name.name;
```
nilehmann: Same here

nilehmann (changes requested):

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2302,11 +2311,21 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
...
         let expr = match func.res {
             ExprRes::Param(..) => env.lookup(func).to_expr(),
             ExprRes::GlobalFunc(kind) => {
                 match Self::conv_spec_func(&kind) {
-                    Ok(func) => rty::Expr::global_func(func),
+                    Ok(func) => {
+                        match func {
+                            rty::SpecFuncKind::Uif(flux_id) | rty::SpecFuncKind::Def(flux_id) => {
+                                let dst_span = self.genv().func_span(flux_id);
+                                dbg::hyperlink!(self.genv().tcx(), span, dst_span);
```
nilehmann: Conversion happens twice, so you are emitting all hyperlinks twice.

ranjitjhala: yikes, well not the worst I suppose, one could add the MODE?

crates/flux-desugar/src/desugar.rs
ranjitjhala: In this case the `.span` is indeed just that of the identifier -- have renamed the field to `ident_span` to clarify.

crates/flux-fhir-analysis/src/conv/mod.rs
ranjitjhala: I added the. mode.

PR #1232, opened 2025-08-13
Refine iterator spec

vrindisbacher:
I believe that the iterator should only increase it's index if the iterator is not done.

@ranjitjhala @nilehmann do you agree?

### DIFF
--- a/lib/flux-core/src/iter/traits/iterator.rs
+++ b/lib/flux-core/src/iter/traits/iterator.rs
@@ -16,7 +16,10 @@ defs! {
 trait Iterator {
     #[spec(
         fn(self: &mut Self[@curr_s]) -> Option<Self::Item>[!<Self as Iterator>::done(curr_s)]
-        ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)}
+        ensures self: Self{next_s: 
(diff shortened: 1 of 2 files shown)


lib/flux-core/src/iter/traits/iterator.rs
```diff
@@ -16,7 +16,10 @@ defs! {
 trait Iterator {
     #[spec(
         fn(self: &mut Self[@curr_s]) -> Option<Self::Item>[!<Self as Iterator>::done(curr_s)]
-        ensures self: Self{next_s: <Self as Iterator>::step(curr_s, next_s)}
+        ensures self: Self{next_s:
```
ranjitjhala, line 19: @vrindisbacher why do you need this change? 

Technically, the `step` is a _relation_ and for each `impl` you can define it so that IF the `done(curr_s)`  is true then `step(curr_s, next_s)` only holds if `next_s == curr_s`; so we shouldn't really need to change the spec at all (or if we do, it should suffice to just modify the `next` for the appropriate `impl`?

vrindisbacher: That's a fine approach too. I'm not sure what's best, but I would imagine that in general the iterator doesn't `step` if it is `done`? That's why I put it here.

PR #1233, opened 2025-08-13
Bot elimination

ranjitjhala:
Simplify the fixpoint constraints to identify KVars that can be set to

1. `False` i.e. (`bot`) because they never appear in the HEAD position of a constraint;
2. `True` i.e. (`top`) because they never appear in the ASSM position of a constraint.

This gives substantial savings, e.g. in `vtock`, the time for check `update_regions` in 
the `lazy-overflow` branch by

```
$ FLUXFLAGS="-Finclude=def:update_regions" cargo flux -p cortexm
```

Goes from 500s (8m+)  to about 10s.

### DIFF
--- a/crates/flux-infer/src/refine_tree.rs
+++ b/crates/flux-infer/src/refine_tree.rs
@@ -834,19 +834,13 @@ impl Node {
 struct ConstraintDeps {
     /// description of each constraint
     edges: IndexVec<ClauseId, ClauseInfo>,
-    /// set of edges where kvid appears as ASSM
-    kv_lhs: FxHashMap<KVid, FxHashSet<ClauseId>>,
     /// set of edges where kvid appears as HEAD
     kv_rhs: FxHashMap<KVid, FxHashSet<ClauseId>>,
 }
 
 impl ConstraintDeps {
     fn new(node: &Node) -> Self {
-        let mut graph = Self {
-            edges: IndexVec::default(),
-            kv_lhs: FxHashMap::default(),
-            kv_rhs: FxHashMap::default(),
-        };
+        let mut graph = Self { edges: IndexVec::default(), kv_rhs: FxHashMap::default() };
         graph.build(node, &mut vec![]);
 
         graph
@@ -855,15 +849,6 @@ impl ConstraintDeps {
     fn head(&self, cid: ClauseId) -> VertexId {
         let info = &self.edges[cid];
         info.rhs
-    }
-
-    fn cstrs_with_kvar_as_assumption(&self, kvid: &KVid) -> Vec<ClauseId> {
-        self.kv_lhs
-            .get(kvid)
-            .unwrap_or(&FxHashSet::default())
-            .iter()
-            .copied()
-            .collect()
     }
 
     fn build(&mut self, node: &Node, ctx: &mut Vec<KVid>) {
@@ -891,15 +876,25 @@ impl ConstraintDeps {
         ctx.truncate(n); // restore ctx
     }
 
+    fn kv_lhs(&self) -> FxHashMap<KVid, FxHashSet<ClauseId>> {
+        let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
+        for (edge_id, edge_info) in self.edges.iter_enumerated() {
+            for kvid in &edge_info.lhs {
+                res.entry(*kvid).or_default().insert(edge_id);
+            }
+        }
+        res
+    }
+
     fn insert_edge(&mut self, lhs: &[KVid], rhs: VertexId) {
         // Create and insert edge
         let edge_info = ClauseInfo { lhs: lhs.iter().copied().collect(), rhs };
         let edge_id = self.edges.push(edge_info);
 
         // Add edge_id to kv_lhs for each kvar in lhs
-        for kvid in lhs {
-            self.kv_lhs.entry(*kvid).or_default().insert(edge_id);
-        }
+        // for kvid in lhs {
+        //     self.kv_lhs.entry(*kvid).or_default().insert(edge_id);
+        // }
         // Add edge_id to kv_rhs for rhs kvar
         if let VertexId::KVar(kvid) = rhs {
             self.kv_rhs.entry(kvid).or_default().insert(edge_id);
@@ -911,6 +906,9 @@ impl ConstraintDeps {
     fn bot_kvars(self) -> Assignment {
         // set of BOT kvars (initially, all)
         let mut assignment = Assignment::new(Label::Bot);
+
+        // set of edges where kvid appears as ASSM
+        let kv_lhs: FxHashMap<KVid, FxHashSet<ClauseId>> = self.kv_lhs();
 
         // set of BOT kvars in LHS of each constraint with KVar HEAD
         let mut bot_assms: IndexVec<ClauseId, Option<FxHashSet<KVid>>> =
@@ -945,15 +943,15 @@ impl ConstraintDeps {
                 // un-BOT the head kvar
                 assignment.remove(kvid);
                 // remove the head kvar from all (bot) assumptions where it currently occurs
-                for cid in self.cstrs_with_kvar_as_assumption(&kvid) {
+                for cid in kv_lhs.get(&kvid).unwrap_or(&FxHashSet::default()).iter() {
                     // if cid HEAD is a kvar
-                    if let VertexId::KVar(rhs_kvid) = self.head(cid) {
-                        let Some(ref mut assms) = bot_assms[cid] else {
+                    if let VertexId::KVar(rhs_kvid) = self.head(*cid) {
+                        let Some(ref mut assms) = bot_assms[*cid] else {
                             tracked_span_bug!("missing constraint info for {:?}", cid);
                         };
                         assms.remove(&kvid);
                         if assignment.has_label(rhs_kvid) && assms.is_empty() {
-                            candidates.push(cid);
+                            candidates.push(*cid);
                         }
                     };
                 }
--- a/crates/flux-infer/src/refine_tree.rs
+++ b/crates/flux-infer/src/refine_tree.rs
@@ -833,24 +833,25 @@ impl Node {
 #[derive(Debug)]
 struct ConstraintDeps {
     /// description of each constraint
-    edges: IndexVec<ClauseId, ClauseInfo>,
+    assumptions: IndexVec<ClauseId, FxHashSet<KVid>>,
+    /// head of each constraint
+    heads: IndexVec<ClauseId, VertexId>,
 }
 
 impl ConstraintDeps {
     fn new(node: &Node) -> Self {
-        let mut graph = Self { edges: IndexVec::default() };
+        let mut graph = Self { assumptions: IndexVec::default(), heads: IndexVec::default() };
         graph.build(node, &mut vec![]);
-
         graph
     }
 
-    fn head(&self, cid: ClauseId) -> VertexId {
-        let info = &self.edges[cid];
-        info.rhs
+    fn insert_clause(&mut self, assumptions: &[KVid], head: VertexId) {
+        self.assumptions.push(assumptions.iter().copied().collect());
+        self.heads.push(head);
     }
 
-    fn build(&mut self, node: &Node, ctx: &mut Vec<KVid>) {
-        let n = ctx.len();
+    fn build(&mut self, node: &Node, assumptions: &mut Vec<KVid>) {
+        let n = assumptions.len();
         match &node.kind {
             NodeKind::Head(expr, _) => {
                 expr.visit_conj(&mut |e: &Expr| {
@@ -859,15 +860,15 @@ impl ConstraintDeps {
                         ExprKind::KVar(kvar) => Some(VertexId::KVar(kvar.kvid)),
                         _ => Some(VertexId::Conc),
                     };
-                    if let Some(rhs) = vertex {
-                        self.edges.push(ClauseInfo::new(ctx, rhs));
+                    if let Some(head) = vertex {
+                        self.insert_clause(assumptions, head);
                     }
                 });
             }
             NodeKind::Assumption(expr) => {
                 expr.visit_conj(&mut |e: &Expr| {
                     if let ExprKind::KVar(kvar) = e.kind() {
-                        ctx.push(kvar.kvid);
+                        assumptions.push(kvar.kvid);
                     }
                 });
             }
@@ -875,17 +876,17 @@ impl ConstraintDeps {
         };
 
         for child in &node.children {
-            self.build(&child.borrow(), ctx);
+            self.build(&child.borrow(), assumptions);
         }
 
-        ctx.truncate(n); // restore ctx
+        assumptions.truncate(n); // restore ctx
     }
 
     /// set of edges where kvid appears as ASSM
     fn kv_lhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
         let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
-        for (edge_id, edge_info) in self.edges.iter_enumerated() {
-            for kvid in &edge_info.lhs {
+        for (edge_id, kvids) in self.assumptions.iter_enumerated() {
+            for kvid in kvids {
                 res.entry(*kvid).or_default().insert(edge_id);
             }
         }
@@ -897,9 +898,9 @@ impl ConstraintDeps {
     /// set of edges where kvid appears as HEAD
     fn kv_rhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
         let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
-        for (edge_id, edge_info) in self.edges.iter_enumerated() {
-            if let VertexId::KVar(kvid) = edge_info.rhs {
-                res.entry(kvid).or_default().insert(edge_id);
+        for (edge_id, head) in self.heads.iter_enumerated() {
+            if let VertexId::KVar(kvid) = head {
+                res.entry(*kvid).or_default().insert(edge_id);
             }
         }
         res.into_iter()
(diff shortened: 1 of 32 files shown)


ranjitjhala:
@nilehmann here's the updated PR with the bot/top phases split out

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+        }
+
+        assignment
+    }
+}
+
+#[derive(Debug)]
+struct Graph {
+    /// description of each constraint
+    edges: FxHashMap<CstrId, CstrInfo>,
```
nilehmann: Make `CstrId` a `newtype_index!` and store this as an `IndexVec`.

Also, we've been calling them constraints, but a more accurate term is (horn) clause. I don't feel strongly about this.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+            .keys()
+            .chain(self.kv_rhs.keys())
+            .copied()
+            .collect::<FxHashSet<_>>()
+            .into_iter()
+    }
+}
+
+/// helper function to split an `Expr` into its constituent `VertexId`s
+fn expr_vertices(expr: &Expr, vertices: &mut Vec<VertexId>) {
```
nilehmann: We implement something similar inside `Expr::flatten_conj`. I would factor that out and expose it. Something like

```rust
        pub fn visit_conj(&self, f: &mut impl FnMut(&Expr)) {
            if let ExprKind::BinaryOp(BinOp::And, e1, e2) = self.kind() {
                e1.visit_conj(f);
                e2.visit_conj(f);
            } else {
                f(self)
            }
        }
```

Then we can use this method in the two places where you are calling `expr_vertices` and avoid the intermediate allocation.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+}
+
+#[derive(Debug)]
+struct Graph {
+    /// description of each constraint
+    edges: FxHashMap<CstrId, CstrInfo>,
+    /// set of edges where kvid appears as ASSM
+    kv_lhs: FxHashMap<KVid, FxHashSet<CstrId>>,
+    /// set of edges where kvid appears as HEAD
+    kv_rhs: FxHashMap<KVid, FxHashSet<CstrId>>,
```
nilehmann: I think these two can be `Vec` because we are always inserting a fresh `CstrId`

```suggestion
    kv_lhs: FxHashMap<KVid, Vec<CstrId>>,
    /// set of edges where kvid appears as HEAD
    kv_rhs: FxHashMap<KVid, Vec<CstrId>>,
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+    }
+
+    fn cstrs_with_kvar_as_assumption(&self, kvid: &KVid) -> Vec<CstrId> {
+        self.kv_lhs
+            .get(kvid)
+            .unwrap_or(&FxHashSet::default())
+            .iter()
+            .copied()
+            .collect()
+    }
```
nilehmann: If we are making the values in `kv_lhs` vectors, we can avoid the allocation here with

```suggestion
    fn cstrs_with_kvar_as_assumption(&self, kvid: &KVid) -> &[CstrId] {
        self.kv_lhs.get(kvid).map_or(&[], Vec::as_slice) 
    }
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+
+    /// Computes the set of all kvars that can be assigned to Bot (False),
+    /// because they are not (transitively) reachable from any concrete ASSUMPTION.
+    fn bot_kvars(&self, graph: Graph) -> Assignment {
+        // set of BOT kvars (initially, all)
+        let mut assignment = Assignment::new(graph.kvars(), Label::Bot);
+
+        // set of BOT kvars in LHS of each constraint with KVar HEAD
+        let mut bot_assms: FxHashMap<CstrId, FxHashSet<KVid>> = graph
```
nilehmann: Can we modify `graph.edges` in-place instead of doing this clone?

If we do that, you'll have lifetime issues when calling `cstrs_with_kvar_as_assumption` in combination with the change I suggested in that function, but we can inline the function to fix it.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+                }
+            }
+        }
+
+        assignment
+    }
+}
+
+#[derive(Debug)]
+struct Graph {
```
nilehmann: I think we can hardly call this a graph now. It's more like a condensed representation of the set of constraints/clauses.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+            _ => {}
+        }
+        for child in &self.children {
+            child.borrow_mut().simplify_with_assignment(assignment);
+        }
+    }
+
+    /// Computes the set of all kvars that can be assigned to Bot (False),
+    /// because they are not (transitively) reachable from any concrete ASSUMPTION.
+    fn bot_kvars(&self, graph: Graph) -> Assignment {
```
nilehmann: This is not using `&self`, and it could be a method in `Graph`.

crates/flux-infer/src/refine_tree.rs
nilehmann: We only use `kv_lhs` for `bot` and `kv_rhs` for `top`. Maybe we can compute them from the edges in each case. We would need to compute the set of kvars some other way, though. Or maybe we don't need to compute the set of kvars at all if we store the complement of the assigned set in `Assignment`, i.e., if it's not in the set, it's considered as marked with the label.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +791,328 @@ impl RefineCtxtTrace {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Label {
+    /// Kvar can be solved to false
+    Bot,
+    /// Kvar can be solved to true
+    Top,
+}
+
+struct Assignment {
+    inner: FxHashMap<KVid, Label>,
```
nilehmann: We always assign everything to the same label, so we can make this a set and store the `Label` as a separate field.

crates/flux-infer/src/refine_tree.rs
ranjitjhala: Which intermediate allocation? We'd still have to make the `vec![]` to push the vertices into, yes?

nilehmann: You are collecting into a vec and then immediately iterating over that vec and discarding it.

ranjitjhala: I rejigged this, take a look to see if its what you meant?

crates/flux-infer/src/refine_tree.rs
ranjitjhala: Not quite, if the same KVar appears twice in an LHS then you'd end up inserting the CstrId twice...

crates/flux-infer/src/refine_tree.rs
ranjitjhala: renamed to `ConstraintDeps` ...

crates/flux-infer/src/refine_tree.rs
ranjitjhala: oh yes, forgot to move them!

crates/flux-infer/src/refine_tree.rs
ranjitjhala: see above re: duplication.

crates/flux-infer/src/refine_tree.rs
ranjitjhala: great idea, done!

nilehmann: Good point, but you are already computing the `lhs` as a set just before pushing, so you can iterate over the deduplicated set of kvars.

crates/flux-infer/src/refine_tree.rs
ranjitjhala: nvm got it!

crates/flux-infer/src/refine_tree.rs
ranjitjhala: OTOH if you made the `ctx` a set then can make this a `Vec` ...

ranjitjhala: but that makes it tedious to "pop" stuff off the context... leaving this for now.

ranjitjhala: did as you suggest.

crates/flux-infer/src/refine_tree.rs
ranjitjhala: I did something different (to avoid lifetime issues!) -- I split the `ClauseInfo` into separate maps:

```rust
#[derive(Debug)]
struct ConstraintDeps {
    /// description of each constraint
    assumptions: IndexVec<ClauseId, FxHashSet<KVid>>,
    /// head of each constraint
    heads: IndexVec<ClauseId, VertexId>,
}
```

This lets us straightforwardly "consume" the `assumptions" to do the in-place updates in `bot_kvars`.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+        assignment
+    }
+}
+
+newtype_index! {
+    struct ClauseId {}
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum VertexId {
```
nilehmann:
```suggestion
enum Head {
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+            self.build(&child.borrow(), assumptions);
+        }
+
+        assumptions.truncate(n); // restore ctx
+    }
+
+    /// set of edges where kvid appears as ASSM
+    fn kv_lhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
+        let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
+        for (edge_id, kvids) in self.assumptions.iter_enumerated() {
```
nilehmann:
```suggestion
        for (clause_id, kvids) in self.assumptions.iter_enumerated() {
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+        }
+        res.into_iter()
+            .map(|(k, v)| (k, v.into_iter().collect()))
+            .collect()
+    }
+
+    /// set of edges where kvid appears as HEAD
+    fn kv_rhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
+        let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
+        for (edge_id, head) in self.heads.iter_enumerated() {
```
nilehmann:
```suggestion
        for (clause_id, head) in self.heads.iter_enumerated() {
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+            NodeKind::Head(expr, _) => {
+                expr.visit_conj(&mut |e: &Expr| {
+                    let vertex = match e.kind() {
+                        ExprKind::BinaryOp(BinOp::And, _, _) => None,
+                        ExprKind::KVar(kvar) => Some(VertexId::KVar(kvar.kvid)),
+                        _ => Some(VertexId::Conc),
+                    };
+                    if let Some(head) = vertex {
+                        self.insert_clause(assumptions, head);
+                    }
```
nilehmann:
```suggestion
                    if let ExprKind::KVar(kvar) = e.kind() {
                        self.insert_clause(assumptions, VertexId::KVar(kvar.kvid))
                    } else {
                        self.insert_clause(assumptions, VertexId::Conc)
                    };
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+        assumptions.truncate(n); // restore ctx
+    }
+
+    /// set of edges where kvid appears as ASSM
+    fn kv_lhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
+        let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
+        for (edge_id, kvids) in self.assumptions.iter_enumerated() {
+            for kvid in kvids {
+                res.entry(*kvid).or_default().insert(edge_id);
+            }
```
nilehmann: I still don't see why you have to insert into a set and can't push directly into a vec. Since `kvids` is a set, you will never push the same `edge_id` into the same `kvid`.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+            .map(|(k, v)| (k, v.into_iter().collect()))
+            .collect()
+    }
+
+    /// set of edges where kvid appears as HEAD
+    fn kv_rhs(&self) -> FxHashMap<KVid, Vec<ClauseId>> {
+        let mut res: FxHashMap<KVid, FxHashSet<ClauseId>> = FxHashMap::default();
+        for (edge_id, head) in self.heads.iter_enumerated() {
+            if let VertexId::KVar(kvid) = head {
+                res.entry(*kvid).or_default().insert(edge_id);
```
nilehmann: Same here. I don't see why we can't push into a vec directly.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+            .filter_map(|(cid, lhs)| if lhs.is_empty() { Some(cid) } else { None })
+            .collect();
+
+        // while there is a candidate constraint, that has NO BOT kvars in lhs
+        while let Some(cid) = candidates.pop() {
+            if let VertexId::KVar(kvid) = self.heads[cid] {
+                // un-BOT the head kvar
+                assignment.remove(kvid);
+                // remove the head kvar from all (bot) assumptions where it currently occurs
+                for cid in kv_lhs.get(&kvid).unwrap_or(&vec![]).iter() {
```
nilehmann:
```suggestion
                for cid in kv_lhs.get(&kvid).map_or(&[], Vec::as_slice) {
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
     }
 }
+
+impl Node {
+    /// replace bot-kvars with false
+    fn simplify_bot(&mut self, genv: GlobalEnv, assumed_preds: &mut SnapshotMap<Expr, ()>) {
+        let graph = ConstraintDeps::new(self);
+        let bots = graph.bot_kvars();
+        self.simplify_with_assignment(&bots);
+        self.simplify(genv, assumed_preds);
```
nilehmann: Is this needed just to remove the nodes from the tree? I don't love that we have to do full simplification again. Maybe we can factor out from `Node::simplify` the part that alters the shape of the tree.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -780,3 +792,254 @@ impl RefineCtxtTrace {
...
+        for child in &self.children {
+            child.borrow_mut().simplify_with_assignment(assignment);
+        }
+    }
+}
+
+#[derive(Debug)]
+struct ConstraintDeps {
+    /// assumptions for each clause
+    assumptions: IndexVec<ClauseId, FxHashSet<KVid>>,
```
nilehmann, line 854: Data-oriented programming. I like it!

crates/flux-infer/src/refine_tree.rs
ranjitjhala: duh. sorry about that!

crates/flux-infer/src/refine_tree.rs
ranjitjhala: Yes, its to do the pruning.

Well, the thing is that once you replace a HEAD with `true` you may end up "propagating" that all the way up to the top, so its not _just_ the matter of pruning the constraints where you have a BOT/False assumption...

ranjitjhala: In short, its not really clear how to factor it out. 

I added a `SimplifyPhase` parameter: 

- when its `Full` we do the `.normalize(..).simplify(..)` on each inner `Expr`;
- when its `Partial` we skip the above, and only propagate the TOP/BOT stuff.

crates/flux-infer/src/refine_tree.rs
ranjitjhala: Sadly, your change throws some odd error (mismatch between `&[ClauseId; 0]` and `&[ClauseId]` ... ???

```rust
error[E0271]: expected `as_slice` to return `&[_; 0]`, but it returns `&[ClauseId]`
    --> crates/flux-infer/src/refine_tree.rs:944:58
     |
 944 |                 for cid in kv_lhs.get(&kvid).map_or(&[], Vec::as_slice) {
     |                                              ------      ^^^^^^^^^^^^^ expected `&[_; 0]`, found `&[ClauseId]`
     |                                              |
     |                                              required by a bound introduced by this call
     |
     = note: expected reference `&[_; 0]`
                found reference `&[ClauseId]`
note: required by a bound in `std::option::Option::<T>::map_or`
    --> /Users/rjhala/.rustup/toolchains/nightly-2025-08-07-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:1219:32
     |
1217 |     pub const fn map_or<U, F>(self, default: U, f: F) -> U
     |                  ------ required by a bound in this associated function
1218 |     where
1219 |         F: ~const FnOnce(T) -> U + ~const Destruct,
     |                                ^ required by this bound in `Option::<T>::map_or`


```

nilehmann (approved):

PR #1256, opened 2025-08-22
Hack to check bitfields

nilehmann:


nilehmann:
Ah, it was a hack to get some number for the paper. We can close it.

PR #1266, opened 2025-09-09
WIP Implement Fusion for Solving Horn Constraints

petros-marko:
### Parsing

First pass parses string into s-expression type. Second pass converts s-expression to Constraint. It definitely allows some things that it shouldn't (e.g. a list of arguments nested in more parens than needed). The error messages aren't very detailed and there are some kinds of expressions that don't work yet(e.g. function declarations). 

### Solving

Eliminates acyclic kvars and then applies predicate abstraction. Lots of cloning things around that could be improved. Only supports Z3 as the solver. The build fails if `Z3_SYS_Z3_HEADER` is not in the environment.

### DIFF
--- a/lib/liquid-fixpoint/Cargo.toml
+++ b/lib/liquid-fixpoint/Cargo.toml
@@ -15,7 +15,7 @@ z3 = { version = "0.12.1", optional = true }
 [features]
 default = []
 # Enable Z3 SMT solver integration for constraint satisfiability checking (optional)
-z3-solver = ["dep:z3"]
+rust-fixpoint = ["dep:z3"]
 nightly = []
 
 [lints]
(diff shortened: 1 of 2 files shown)


nilehmann (changes requested):
@petros-marko as we discussed the other day, can you make this a different crate instead of modifying `lib/liquid-fixpoint`?

nilehmann:
> I want to use the code you wrote to go between SMT and Flux types while still using the original liquid-fixpoint backend. If we merge this into main, is this going to be possible?
> 
> If not, can we move this to a separate crate so we can merge it into main?

It depends on what you want to use from this. If you want to reuse the code that goes from `fixpoint::Expr` to `z3`, it has to be in the same crate; otherwise, you'd need to transform between `rty` and the representation in the new crate.

But also, I realize that the code that transforms into `z3` is not general enough to be used with the types we generate from Flux, so that would have to be adapted. If you don't want to re-use that, I don't think there's much on this PR that you could use, except as an example of how to use z3 api.

PR #1275, opened 2025-09-12
Make solving generic

petros-marko:
First pass at making the solving work for any `T: Types`.

The core change is adding a trait bound to `Types::Var` to say we should be able to create a `Var` from a `KVar` and a number. Because of the orphan rule I had to create a new trait instead of using `From`. I'm pretty sure we should be able to do away with this because this is only used to create variables that will later be substituted, but it works as a sort of hack...

The Z3 bindings require that variable names be convertible into String or usize. Since `Identifier` was already part of the trait bound for `Types::Var`, I was able to convert variables to strings without adding more traits. If that's somehow inappropriate, we can be more explicit.


lib/liquid-fixpoint/src/lib.rs
```diff
@@ -79,17 +83,6 @@ pub trait Identifier: Sized {
...
-
-impl Types for DefaultTypes {
-    type Sort = &'static str;
-    type KVar = &'static str;
-    type Var = &'static str;
-    type Tag = String;
-    type Decimal = u32;
-    type String = String;
-}
-
```
petros-marko: I'm guessing we don't actually want to delete this; the `'static` lifetime really messed with my approach (since there's no way to create `&'static str` instances at runtime) and so I did this temporarily to make sure I could get it to work. How attached are we to this @nilehmann?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -135,6 +135,13 @@ pub mod fixpoint {
         }
     }
 
+    impl FromPair<KVid, i32> for Var {
+        fn from(p: (KVid, i32)) -> Self {
+            let uniq = (p.0.as_u32() << 8) & (p.1 as u32);
+            Var::Local(LocalVar::from(uniq))
```
petros-marko: VERY HACKY! I'm not familiar with what these variable forms are. As I mentioned in the PR description, I don't think it really matters what we create here because it'll later be substituted, it just has to come from the KVar type and a number (and ideally be bijective, which this isn't)

cole-k: I don't get it: why are you not instead converting a kvar to the KVar in the fixpoint types? You could for example change `type KVar = (KVid, i32)`.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -1,115 +1,8 @@
-use std::{hash::Hash, sync::LazyLock, vec};
+use std::hash::Hash;
 
 use derive_where::derive_where;
 
-use crate::{DefaultTypes, Types};
-
-pub(crate) static DEFAULT_QUALIFIERS: LazyLock<[Qualifier<DefaultTypes>; 13]> =
```
cole-k: Where did these go? Isn't the plan to add these qualifiers and their definitions lazily?

petros-marko: @nilehmann will have to correct me if I'm wrong; I don't think these were used. They were using `DefaultTypes` another implementer of the `Types` trait that is not what `flux` uses. I (temporarily) removed `DefaultTypes` and had to also remove this as a result.

lib/liquid-fixpoint/src/lib.rs
cole-k: Could you just use `String` here?

lib/liquid-fixpoint/src/constraint_solving.rs
```diff
@@ -239,13 +184,13 @@ impl Constraint<ParsingTypes> {
...
             Constraint::Pred(Pred::KVar(kvid, args), _tag) if var.eq(kvid) => {
-                let args_eq: Vec<Expr<ParsingTypes>> = (0..)
+                let args_eq: Vec<Expr<T>> = (0..)
                     .zip(args.iter())
                     .map(|(idx, arg)| {
                         Expr::Atom(
                             BinRel::Eq,
                             Box::new([
-                                Expr::Var(format!("karg${}#{}", kvid, idx).to_string()),
+                                Expr::Var(FromPair::from((kvid.clone(), idx))),
```
nilehmann: I don't understand all the details, but it seems that this is generating a list `[#0 = x, #1 = y, #2 = z, ...]`. Then all the `#n` are substituted with concrete expressions to yield `[e0 = x, e1 = y, e2 = z, ...]`. So it seems to me that we don't need to generate the intermediate variables at all

petros-marko: Yeah, that's what I was saying in the description. I think we can do it in one step and avoid creating the intermediate value. I'll do that as soon as I get the tests to pass again. @cole-k figured out that the DEFAULT_QUALIFIERS were indeed needed.

lib/liquid-fixpoint/src/constraint_solving.rs
```diff
@@ -239,38 +184,23 @@ impl Constraint<ParsingTypes> {
...
-                            BinRel::Eq,
-                            Box::new([
-                                Expr::Var(format!("karg${}#{}", kvid, idx).to_string()),
-                                Expr::Var(arg.clone()),
-                            ]),
-                        )
-                    })
-                    .collect();
-                vec![(vec![], args_eq)]
+                let arg_vars = args.iter().map(|arg| Expr::Var(arg.clone())).collect();
```
nilehmann, line 192: nit: maybe we can put this in a wrapper type to mark this represents a *solution*

petros-marko: @nilehmann When you say a wrapper type you mean something like `struct Solution(T::Var)`? So instead of wrapping it in an `Expr` I'd wrap it in that and then in `do_elim` unwrap and rewrap?

nilehmann: I meant more like

```rust
struct Solution<T: Types> {
    args: Vec<Expr<T>>
}
```

to document what the list of expressions is supposed to mean. 

Or maybe 

```rust
struct Solution<T: Types> {
    binders: Vec<Bind<T>>,
    args: Vec<Expr<T>>,
}
```

petros-marko: I'll do that and merge :)

PR #1276, opened 2025-09-17
Allow passing in features when running tests

petros-marko:
The title says it all. It'll be useful to run tests with the `liquid-fixpoint/rust-fixpoint` feature enabled. Maybe I could add the same support for the `xtask run` subcommand.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -19,6 +19,8 @@ xflags::xflags! {
     cmd xtask {
         /// If true, run all cargo commands with `--offline`
         optional --offline
+        /// If true, run cargo build commands with --features liquid-fixpoint/rust-fixpiont
+        optional --rust-fixpoint
 
         /// Run regression tests
         cmd test {
@@ -26,8 +28,6 @@ xflags::xflags! {
             optional filter: String
             /// Do not check tests in Flux libs.
             optional --no-lib-tests
-            /// Build binaries with features enabled
-            optional --features features: String
         }
         /// Run the `flux` binary on the given input file.
         cmd run {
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -130,7 +132,7 @@ fn test(args: Test) -> anyhow::Result<()> {
         dst: local_sysroot_dir()?,
         build_libs: BuildLibs { force: false, tests: !args.no_lib_tests, libs: FluxLib::ALL },
     };
-    let flux = build_binary("flux", config.profile)?;
+    let flux = build_binary("flux", config.profile, args.features)?;
```
nilehmann: I don't think this will have any effect. This only builds the `flux` binary with the features which doesn't have `liquid-fixpoint` as a dependency. The feature needs to be enabled when building `flux-driver`

petros-marko: Sorry for the premature PR; I realized this problem very shortly after I did the PR and I'm working on it.

nilehmann: I see a couple of options.

1. Have the following in the workspace's `Cargo.toml` and uncomment `rust-fixpoint` if you are working on that
   ```toml
   liquid-fixpoint = { path = "./lib/liquid-fixpoint", version = "0.1.0", features = [
     "nightly",
     # "rust-fixpoint"
   ] }
   ```
2. Transitively add features to crates such that it can be enabled for `flux-driver`
3. Patch the dependency somehow (https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html)

petros-marko: I mean, it works if we pass `--features liquid-fixpoint/rust-fixpoint` when building `flux-driver`. I just updated the PR to make it do that. I'm not sure if something I'm doing is bad practice, so I'm open to your Ideas here.

nilehmann: What happens in the editor after calling `cargo x test --features liquid-fixpoint/rutst-fixpoint`? I think `rust-analyzer` will call `cargo check` without enabling the feature, which will probably trigger a re-check of more crates than necessary.

nilehmann: But maybe you already have the feature enabled in rust-analyzer.

nilehmann: I don't think there's a silver bullet, but you'll be the one using this the most, so I'm fine with whatever you think will make your life the easiest.

petros-marko: I've enabled the feature in rust-analyzer. When I type the command before making these changes I just get the `xtask` help blurb because it doesn't recognize `--features <features>` as anything. After my changes, it builds and runs as expected.

PR #1277, opened 2025-09-17
Plugs the rust fixpoint implementation into flux (behind the rust-fixpoint feature)

petros-marko:
Makes the rust solver return `FixpointResult` instead of bool. The stats are definitely wrong but the Safe/Unsafe outcome should be correct.
- adds support for constants in the environment
- stops dropping the tags associated with constraints
- adds support for function sorts (partially)
- resolves an infinite loop bug by simplifying the constraint before handing it off to the solver

Lots of flux tests passing! Some failing :(. A few falling into infinite loops.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -442,6 +442,9 @@ where
         let mut constants = self.ecx.const_map.into_values().collect_vec();
         constants.extend(define_constants);
 
+        // The rust fixpoint implementation does not yet support polymorphic functions.
+        // For now we avoid including these by default so that cases where they are not needed can work.
+        // Should be removed when support is added.
         #[cfg(not(feature = "rust-fixpoint"))]
         for rel in fixpoint::BinRel::INEQUALITIES {
             // ∀a. a -> a -> bool
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -442,6 +442,7 @@ where
         let mut constants = self.ecx.const_map.into_values().collect_vec();
         constants.extend(define_constants);
 
+        #[cfg(not(feature = "rust-fixpoint"))]
         for rel in fixpoint::BinRel::INEQUALITIES {
             // ∀a. a -> a -> bool
             let sort = fixpoint::Sort::mk_func(
--- a/lib/liquid-fixpoint/src/constraint.rs
+++ b/lib/liquid-fixpoint/src/constraint.rs
@@ -247,6 +247,21 @@ impl<T: Types> Pred<T> {
             Pred::Expr(_) => true,
         }
     }
+
+    pub fn simplify(&mut self) {
+        match self {
+            Pred::And(conjuncts) => {
+                if conjuncts.len() == 0 {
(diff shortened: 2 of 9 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -442,6 +442,7 @@ where
         let mut constants = self.ecx.const_map.into_values().collect_vec();
         constants.extend(define_constants);
 
+        #[cfg(not(feature = "rust-fixpoint"))]
```
nilehmann, line 448: Please add a comment explaining why this is excluded and that we should eventually add support for it.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -247,6 +247,21 @@ impl<T: Types> Pred<T> {
             Pred::Expr(_) => true,
         }
     }
+
+    pub fn simplify(&mut self) {
```
nilehmann: Try to reduce the visibility as much as possible; otherwise, it's hard to tell where something is being used, and rustc won't tell us when something is not used anymore. This is in general, you have many functions marked as `pub` that are not used outside the crate.

```suggestion
    pub(crate) fn simplify(&mut self) {
```

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -210,6 +217,32 @@ pub enum FixpointResult<Tag> {
     Crash(CrashInfo),
 }
 
+impl<Tag> FixpointResult<Tag> {
+    pub fn is_safe(&self) -> bool {
+        matches!(self, FixpointResult::Safe(_))
+    }
+
+    pub fn merge(self, other: FixpointResult<Tag>) -> Self {
```
nilehmann, line 228: If we are faking the stats and the result, then I'd return the fake data in `Task::run` and avoid all this `merge` business.

lib/liquid-fixpoint/src/cstr2smt2.rs
```diff
@@ -1,326 +1,417 @@
...
-        if lhs.sort_kind() != rhs.sort_kind() {
-            panic!("binary operands must have the same sort");
+fn atom_to_z3<'ctx, T: Types>(
+    ctx: &'ctx Context,
+    bin_rel: &BinRel,
+    operands: &Box<[Expr<T>; 2]>,
+    env: &mut Env<'ctx, T>,
+) -> ast::Dynamic<'ctx> {
+    let loperand = expr_to_z3(ctx, &operands[0], env);
+    let roperand = expr_to_z3(ctx, &operands[1], env);
```
nilehmann: nit: you call these `lhs` and `rhs` in other places, so maybe we can be consistent with that.

lib/liquid-fixpoint/src/lib.rs
petros-marko: Not entirely fake; some tests were failing because in `Unsafe` case I was dropping the associated tag. If you have an idea of how to keep the merging for errors but make it explicit that the rest is fake, I'm open to change it.

nilehmann: I see. We can iterate over this later, but it feels to me that it will be better to store this in some state that we update for each constraint.

PR #1278, opened 2025-09-18
Add the default qualifiers when using rust-fixpoint

petros-marko:
`DEFAULT_QUALIFIERS` were added during the serialization, which doesn't happen when using `rust-fixpoint`. This PR creates another copy of these qualifiers named `FIXPOINT_QUALIFIERS` (because they're of type `Qualifier<FixpointTypes>`) which are added explicitly during `Task` creation. 

These changes fix a bunch of tests, eg `pos/array/array02.rs`


nilehmann:
Hmmm. To avoid the repetition I'd remove the qualifiers from the `liquid-fixpoint` crate and use this in both cases. I think it makes sense to let the user specify the default qualifiers instead of being harcoded by the fixpoint crate.

crates/flux-infer/src/fixpoint_qualifiers.rs
```diff
@@ -0,0 +1,152 @@
+use std::sync::LazyLock;
+
+use liquid_fixpoint::{BinOp, BinRel, Sort};
+
+use crate::fixpoint_encoding::fixpoint::{
+    LocalVar, Var,
+    fixpoint_generated::{Expr, Qualifier},
+};
+pub(crate) static FIXPOINT_QUALIFIERS: LazyLock<[Qualifier; 13]> = LazyLock::new(|| {
```
nilehmann, line 9: I'd call this (and the module) `DEFAULT_QUALIFIERS`

PR #1281, opened 2025-09-18
Glob prefix fix

ranjitjhala:
Padding an *empty* `glob_prefix` as a prefix of the pattern made the glob patterns like `src/ch03_structs.rs` become `/src/ch03_structs.rs` which of course doesn't match the file, thwarting the `include`...


ranjitjhala:
@nilehmann this is a one-liner but you probably want to take a look in case I've muddle something up...

PR #1284, opened 2025-09-19
Make rust fixpoint faster by preserving solver and environment

petros-marko:
Use the same solver and constant bindings across constraints. On my machine this sped up `cargo x --rust-fixpoint test` by almost 2x.

### DIFF
--- a/lib/liquid-fixpoint/src/constraint_with_env.rs
+++ b/lib/liquid-fixpoint/src/constraint_with_env.rs
@@ -28,9 +28,7 @@ use crate::Assignments;
 
 #[cfg(feature = "rust-fixpoint")]
 impl<T: Types> ConstraintWithEnv<T> {
-    pub fn compute_initial_assignments(
-        &self,
-    ) -> Assignments<'_, T> {
+    pub fn compute_initial_assignments(&self) -> Assignments<'_, T> {
         let mut assignments = HashMap::new();
 
         for decl in &self.kvar_decls {
--- a/lib/liquid-fixpoint/src/constraint_with_env.rs
+++ b/lib/liquid-fixpoint/src/constraint_with_env.rs
@@ -3,7 +3,7 @@ use derive_where::derive_where;
 use {
     crate::{
         FixpointResult,
-        cstr2smt2::{Env, is_constraint_satisfiable_inner, new_binding},
+        cstr2smt2::{Env, is_constraint_satisfiable, new_binding},
     },
     itertools::Itertools,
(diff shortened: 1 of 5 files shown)


lib/liquid-fixpoint/src/constraint_with_env.rs
```diff
@@ -117,11 +122,19 @@ impl<T: Types> ConstraintWithEnv<T> {
     }
 
     pub fn solve_by_predicate_abstraction(&mut self) -> FixpointResult<T::Tag> {
-        let mut qualifiers: Vec<Qualifier<T>> = vec![];
-        qualifiers.extend(self.qualifiers.to_vec().into_iter());
-        let kvar_assignment = self.solve_for_kvars(&qualifiers);
+        let cfg = Config::new();
+        let ctx = Context::new(&cfg);
+        let solver = Solver::new(&ctx);
```
ranjitjhala, line 126: Why not make `solver` itself  a field of the `ctx`?

petros-marko: These are coming from the z3 bindings crate so I can't change them. I tried to package them up in a convenient struct but it seems impossible to deal with the reference lifetimes correctly (AI suggested some unsafe rust stuff that I guess we want to avoid). Maybe I'm misunderstanding what you meant though.

ranjitjhala: Ah yes, I meant call `Solver::new(...) inside `Context::new` but yes, lifetime shenanigans could be a pain here!

lib/liquid-fixpoint/src/cstr2smt2.rs
```diff
@@ -361,7 +355,7 @@ fn z3_sort<'ctx, T: Types>(ctx: &'ctx Context, s: &Sort<T>) -> z3::Sort<'ctx> {
     }
 }
 
-fn is_constraint_satisfiable_inner<'ctx, T: Types>(
+pub(crate) fn is_constraint_satisfiable_inner<'ctx, T: Types>(
```
nilehmann: maybe?

```suggestion
pub(crate) fn is_constraint_satisfiable<'ctx, T: Types>(
```

lib/liquid-fixpoint/src/constraint_with_env.rs
nilehmann: What we could do is to put `Solver` inside `ConstraintWithEnv`, but meh

PR #1286, opened 2025-09-22
Add support for most theory functions

petros-marko:
- Bump z3 bindings version up (is there a reason not to?) because it does away with the context stuff that made the code weird in #1284 and it makes strlen easier to implement
- Add support for most theory functions

15 more tests passing with these changes

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -157,6 +157,14 @@ pub mod fixpoint {
     #[derive(Hash, Clone, Debug)]
     pub struct SymStr(pub Symbol);
 
+    #[cfg(feature = "rust-fixpoint")]
+    impl FixpointFmt for SymStr {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "{}", self.0)
+        }
+    }
+
+    #[cfg(not(feature = "rust-fixpoint"))]
     impl FixpointFmt for SymStr {
         fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
             write!(f, "\"{}\"", self.0)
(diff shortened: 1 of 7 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1175,7 +1183,7 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
                 fixpoint::Expr::Let(vars[0].into(), Box::new([init, body]))
             }
             rty::ExprKind::GlobalFunc(SpecFuncKind::Thy(itf)) => {
-                fixpoint::Expr::Var(fixpoint::Var::Itf(*itf))
+                fixpoint::Expr::ThyFunc(*itf)
```
nilehmann: We should remove `fixpoint::Var::Itf`.

lib/liquid-fixpoint/src/cstr2smt2.rs
```diff
@@ -197,79 +194,256 @@ fn binop_to_z3<'ctx, T: Types>(
...
+            let elem = expr_to_z3(&args[0], env);
+            let set = expr_to_z3(&args[1], env).as_set().unwrap();
+            set.member(&elem).into()
+        }
+        ThyFunc::SetEmpty | ThyFunc::SetSng => {
+            let arg = expr_to_z3(&args[0], env);
+            let arg_sort = arg.get_sort();
+            let empty = ast::Set::empty(&arg_sort);
+            empty.add(&arg).into()
+        }
```
petros-marko: This is obviously wrong for `SetEmpty`. I was looking at a constraint dump when I started implementing this and it was shown as `set_empty 0` so I assumed we take an argument to add to the empty set. This seems wrong (thanks @nilehmann for the thorough tests); CI not failing since this is behind a feature, so I just won't merge without fixing.

lib/liquid-fixpoint/src/cstr2smt2.rs
```diff
@@ -197,79 +194,259 @@ fn binop_to_z3<'ctx, T: Types>(
...
-fn expr_to_z3<'ctx, T: Types>(
-    ctx: &'ctx Context,
-    expr: &Expr<T>,
-    env: &mut Env<'ctx, T>,
-) -> ast::Dynamic<'ctx> {
+fn thy_func_application_to_z3<T: Types>(
+    func: &ThyFunc,
+    args: &Vec<Expr<T>>,
+    env: &mut Env<T>,
+) -> ast::Dynamic {
```
nilehmann, line 201:
```suggestion
fn thy_func_application_to_z3<T: Types>(
    func: ThyFunc,
    args: &[Expr<T>],
    env: &mut Env<T>,
) -> ast::Dynamic {
```

PR #1293, opened 2025-09-30
Encode `flux` "struct" using `fixpoint` / SMT `datatype`

ranjitjhala:
cc @cole-k @petros-marko

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -80,6 +80,7 @@ pub mod fixpoint {
         DataCtor(AdtId, VariantIdx),
         TupleCtor { arity: usize },
         TupleProj { arity: usize, field: u32 },
+        DataProj { adt_id: AdtId, /* variant_idx: VariantIdx, */ field: u32 },
         UIFRel(BinRel),
         Param(EarlyReftParam),
         ConstGeneric(ParamConst),
@@ -108,6 +109,7 @@ pub mod fixpoint {
                 }
                 Var::TupleCtor { arity } => write!(f, "mktuple{arity}"),
                 Var::TupleProj { arity, field } => write!(f, "tuple{arity}${field}"),
+                Var::DataProj { adt_id, field } => write!(f, "fld{}${field}", adt_id.as_u32()),
                 Var::UIFRel(BinRel::Gt) => write!(f, "gt"),
                 Var::UIFRel(BinRel::Ge) => write!(f, "ge"),
                 Var::UIFRel(BinRel::Lt) => write!(f, "lt"),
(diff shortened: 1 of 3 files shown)


petros-marko (approved):
Looks good to me. should this happen only with rust-fixpoint enabled or is it ok to happen always?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -310,11 +313,19 @@ impl SortEncodingCtxt {
...
+                            .field_sorts_id()
+                            .iter()
+                            .enumerate()
+                            .map(|(i, sort)| {
+                                fixpoint::DataField {
+                                    name: fixpoint::Var::DataProj { adt_id, field: i as u32 },
+                                    sort: self.sort_to_fixpoint(sort),
+                                }
+                            })
+                            .collect::<Vec<_>>();
```
```suggestion
                            .collect_vec();
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -296,10 +298,11 @@ impl SortEncodingCtxt {
     }
 
     fn append_adt_decls(
+        &mut self,
         genv: GlobalEnv,
-        adt_sorts: FxIndexSet<DefId>,
         decls: &mut Vec<fixpoint::DataDecl>,
     ) -> QueryResult {
+        let adt_sorts = self.adt_sorts.clone();
```
nilehmann: Calling `self.sort_to_fixpoint` can insert more ids in `self.adt_sorts`, so you have to iterate until there are no more new ids (like we do in `define_funs`)

All praise to the Borrow Checker. Do not trust the LLMs.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -119,6 +119,10 @@ impl AdtSortVariant {
         self.sorts.fold_with(&mut SortSubst::new(args))
     }
 
+    pub fn field_sorts_id(&self) -> List<Sort> {
```
nilehmann: maybe

```suggestion
    pub fn field_sorts_instantiate_identity(&self) -> List<Sort> {
        self.sorts.clone()
    }
```

or `field_sorts_identity`

nilehmann (changes requested):

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: hmm interesting. i suppose they must also be serialized per dependencies, as otherwise z3 will yell.

though i wonder if `fixpoint` takes care of the order somehow...

ranjitjhala: hallelujah! https://github.com/ucsd-progsys/liquid-fixpoint/blob/e766a7ea6e59dfc853ee23bce25dbf61b7714dde/src/Language/Fixpoint/Types/Utils.hs#L115-L116

ranjitjhala: ok I think I managed to "fixpoint" this -- but can you take a look?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -296,28 +299,45 @@ impl SortEncodingCtxt {
...
+        // We iterate until we have processed all adt sorts because processing one adt sort may
+        // discover new adt sorts to process (e.g., if an adt field has an adt sort).
+        let mut done = FxHashSet::default();
+        while done.len() < self.adt_sorts.len() {
+            let adt_sorts = self.adt_sorts.clone();
+            for (idx, adt_def_id) in adt_sorts.iter().enumerate().skip(done.len()) {
+                if !done.insert(*adt_def_id) {
+                    continue;
+                }
+                let adt_id = AdtId::from_usize(idx);
```
nilehmann: `adt_sort` is an `FxIndexSet` which is a set, but also preserves insertion order and can be indexed by position, so this can be simplified to

```rust
        let mut idx = 0;
        while let Some(adt_def_id) = self.adt_sort.get_index(idx) {
            let adt_id = AdtId::from_usize(idx);
            .
            .
            .
            idx += 1;
        }
```

nilehmann: Although I wonder if there's a smarter way to traverse this, as they almost come out in reverse dependency order except for the initial set...

ranjitjhala: Your `idx` method is much nicer, thanks! have implemented. Yes, its not quite reverse dependency; anyways, fixpoint deals with that ...

PR #1301, opened 2025-10-02
Fix `<...>` in vscode extension

ranjitjhala:


tools/vscode/src/extension.ts
```diff
@@ -307,6 +315,7 @@ async function runCargoFlux(workspacePath: string, file: string, trace: boolean,
   statusBarItem.show();
 
   let fluxFlags = `-Finclude=${includeValue}`;
+  fluxFlags += ` -Fcache=1`;
```
nilehmann, line 318: `-Fcache` takes a path. Also, this should be enabled in `Cargo.toml`

ranjitjhala: Quite so! Odd the the plugin was not crashing with this? I suppose it was just using a file called `1` (!)

fixed in #1304

ranjitjhala: I'm confused. I have

```
rjhala@pozole ~/r/flux-demo (main)> more flux.toml
cache = true
scrape_quals = true
```

where _is_ the cache saved

ranjitjhala: (I should say, I have had this for a _long_ time...)

ranjitjhala: As suggested here: https://flux-rs.github.io/flux/guide/install.html?highlight=cache#cargo-projects

nilehmann: `cargo flux` picks a suitable path for you. There's a concurrency bug though...

`cargo flux` uses the same path for all crates but they are checked in parallel so there could be race conditions.

PR #1302, opened 2025-10-02
Parse flags in cargo-flux

nilehmann:
This is so we can more selectively forward them to cargo.

Fixes the issue discussed in https://flux-rs.zulipchat.com/#narrow/channel/486098-general/topic/Flux.20as.20.22dev-dependency.22/with/542782348

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,10 +40,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.20"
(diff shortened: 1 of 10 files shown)


ranjitjhala:
This seems like a "user-facing" change but I didn't see any changes in the `.md` ?

nilehmann:
Everything that's documented is still supported. In a sense, this is being honest about what we support. Now, when you run `cargo flux --help`, it displays the options we know will work.

PR #1307, opened 2025-10-03
Do enough generation for tock theorems

petros-marko:
- Include refinement function definitions in fixpoint translation when a function is marked as `proven_externally`
- Stop panicking at the first externally proven function that doesn't typecheck so that one run can generate multiple theorems
- Support more expressions and (bitvec related) theory functions

This adds enough so that all `trusted` functions in [this part of tock](https://github.com/PLSysSec/tock/blob/master/arch/cortex-m/src/tcb/theorems.rs) can instead be marked with `proven_externally` and have theorems be generated for them. I can try to prove the theorems 🤷🏽

I made the choice to represent bitvectors in lean as `BitVec`s. Maybe that's not always optimal.

### DIFF
--- a/crates/flux-infer/src/lean_format.rs
+++ b/crates/flux-infer/src/lean_format.rs
@@ -1,7 +1,7 @@
 use core::fmt;
 
 use itertools::Itertools;
-use liquid_fixpoint::{FixpointFmt, Identifier};
+use liquid_fixpoint::{FixpointFmt, Identifier, ThyFunc};
 
 use crate::{
     fixpoint_encoding::fixpoint::{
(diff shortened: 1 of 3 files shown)


crates/flux-infer/src/lean_format.rs
```diff
@@ -98,7 +136,37 @@ impl<'a> fmt::Display for LeanExpr<'a> {
...
             }
-            _ => todo!(),
+            Expr::Neg(inner) => {
+                write!(f, "(-{})", LeanExpr(inner.as_ref()))
+            }
+            Expr::IfThenElse(ite) => {
+                let [condition, if_true, if_false] = ite.as_ref();
+                write!(
+                    f,
+                    "(ite {} {} {})",
```
nilehmann: Is there a reason to use `ite` here instead of `if e then e else e`?

petros-marko: Seemed like a more direct translation, but I can change it 🤷

nilehmann (approved):

PR #1313, opened 2025-10-07
Detached hacks

ranjitjhala:
Fixes @1311 and perhaps also @enjhnsn2 's issue but I can't seem to find a link to that.

Apologies in advance @nilehmann  -- the more I worked on the correct solution (using the `resolver`) the more difficult it seemed to get, and I really want to just get this working, finish the book chapters, and get back to `verify-std`!

This modest patch

1. Uses the names that are in the current scope of where the detached-spec is, and
2. Handles `TraitImpl` somewhat better (again, using the current scope).

### DIFF
--- a/crates/flux-driver/src/collector/detached_specs.rs
+++ b/crates/flux-driver/src/collector/detached_specs.rs
@@ -16,26 +16,36 @@ use rustc_span::{Symbol, def_id::DefId};
 use crate::collector::{FluxAttrs, SpecCollector, errors};
 type Result<T = ()> = std::result::Result<T, ErrorGuaranteed>;
 
-#[derive(PartialEq, Eq, Debug, Hash)]
-struct ImplKey(Symbol);
+#[derive(PartialEq, Eq, Debug, Hash, Clone, Copy)]
+// struct ImplKey(Symbol);
+enum LookupRes {
+    DefId(DefId),
+    Name(Symbol),
+}
 
-impl ImplKey {
-    fn new(tcx: TyCtxt, ty: &Ty) -> Self {
-        let name = match ty.kind() {
+impl LookupRes {
+    fn from_name<T: std::fmt::Debug>(thing: &T) -> Self {
+        let str = format!("{thing:?}");
+        LookupRes::Name(Symbol::intern(&str))
+        // LookupRes::Name(Symbol::intern(name))
+    }
+
+    fn new(ty: &Ty) -> Self {
+        match ty.kind() {
             rustc_middle::ty::TyKind::Adt(adt_def, _) => {
-                let def_id = adt_def.did();
-                tcx.def_path_str(def_id)
+                LookupRes::DefId(adt_def.did())
+                // let def_id = adt_def.did();
+                // tcx.def_path_str(def_id)
             }
-            _ => format!("{ty:?}"),
-        };
-        ImplKey(Symbol::intern(&name))
+            _ => Self::from_name(ty),
+        }
     }
 }
 
 #[derive(PartialEq, Eq, Debug, Hash)]
 struct TraitImplKey {
     trait_: Symbol,
-    self_ty: ImplKey,
+    self_ty: LookupRes,
 }
 
 fn path_to_symbol(path: &surface::ExprPath) -> Symbol {
@@ -61,8 +71,9 @@ fn item_def_kind(kind: &surface::ItemKind) -> Vec<DefKind> {
     }
 }
 
+#[derive(Debug)]
 struct ScopeResolver {
-    items: HashMap<(Symbol, DefKind), DefId>,
+    items: HashMap<(Symbol, DefKind), LookupRes>,
 }
 
 impl ScopeResolver {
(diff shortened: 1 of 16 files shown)


crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -8,34 +8,43 @@ use rustc_errors::ErrorGuaranteed;
...
+enum LookupRes {
+    DefId(DefId),
+    Name(Symbol),
+}
+
+impl LookupRes {
+    fn from_name<T: std::fmt::Debug>(thing: &T) -> Self {
+        let str = format!("{thing:?}");
+        LookupRes::Name(Symbol::intern(&str))
+        // LookupRes::Name(Symbol::intern(name))
```
nilehmann:
```suggestion
```

crates/flux-driver/src/collector/detached_specs.rs
```diff
@@ -8,34 +8,43 @@ use rustc_errors::ErrorGuaranteed;
...
-    fn new(tcx: TyCtxt, ty: &Ty) -> Self {
-        let name = match ty.kind() {
+    fn new(ty: &Ty) -> Self {
+        match ty.kind() {
             rustc_middle::ty::TyKind::Adt(adt_def, _) => {
-                let def_id = adt_def.did();
-                tcx.def_path_str(def_id)
+                LookupRes::DefId(adt_def.did())
+                // let def_id = adt_def.did();
+                // tcx.def_path_str(def_id)
```
nilehmann:
```suggestion
```

ranjitjhala:
I wanted to test this out on @enjhnsn2 's RingBuffer stuff; I think its a strict improvement, for example 

https://github.com/PLSysSec/tock/compare/05451dbb1fca9f5fed7bada5e7500628e847ab33..evan-ringbuffer

So a bunch of those nested `mod` are gone and replaced with just `use ...`.

lib/flux-core/src/cmp.rs
```diff
@@ -17,8 +17,8 @@ trait PartialEq<Rhs: PointeeSized = Self>: PointeeSized {
 
 #[macro_export]
 macro_rules! eq {
-    ($type_name:ident) => {
-        #[specs {
+    ($type_name:path) => {
+        #[flux_rs::specs {
```
nilehmann: We have to gate this for it to work in normal builds.

```suggestion
        #[cfg_attr(flux, flux::specs {
```

tests/tests/pos/detached/detach_impl06.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn test() -> usize {
+        crate::x::y::Thing::new(true).gromp()
+    }
+}
+
+// -------------------------------------------------------------------
+#[flux_rs::specs {
+    mod x {
+        mod y {
+            impl MyTrait for Thing<T> {
```
nilehmann: I don't think this one should work. `MyTrait` is not in any reasonable sense in scope here.

nilehmann (changes requested):

lib/flux-core/src/cmp.rs
ranjitjhala: should one gate this, isn't it better to gate the `eq!` invocation itself?

tests/tests/pos/detached/detach_impl06.rs
ranjitjhala: Haha. Yes, I agree. In fact I had to scratch my head to figure out why it was working in the same place (TLDR; string hacks...) let me think about how to make it break...

lib/flux-core/src/cmp.rs
nilehmann: If you are using "attached" specs, I think it'd be nicer for it to work out of the box like this

```rust
#[derive(PartialEq)]
struct S {...}
flux_core::eq!(S);
```

nilehmann: It'd be even nicer to have :)

```rust
#[derive(LawfulPartialEq)]
struct S { ... }
```

(or whatever name we bikeshed)

tests/tests/pos/detached/detach_impl06.rs
ranjitjhala: Ok, I've made it more robust: the `trait` names must all be _fully qualified_ now.
For some reason, you can say `From` instead of `std::convert::From` and yet `From` doesn't show up in the `module_local_children`... odd.

ranjitjhala: So the example above you have to now say `a::b::MyTrait`...

ranjitjhala: Any idea why the `module_children_local` doesn't pick up the traits? [ I tried using `module_children` as in `resolver.rs` but that crashes... ]

lib/flux-core/src/cmp.rs
ranjitjhala: ah so you mean cleaner to sweep "the gate" under the macro? fair point!

tests/tests/pos/detached/detach_impl06.rs
nilehmann: If the id is local, you have to call `module_children_local` and otherwise call `module_children`. I don't know exactly what you mean by it doesn't pick up the traits. `From` is in the prelude, I don't know if that's important.

nilehmann:
> By “doesn’t pick up” I mean isn’t returned in the iterator returned by
module_children_local

But what trait is not returned in which module? I don't think that's the issue. We can resolve traits just fine using `module_children_local` in `resolver.rs`

PR #1317, opened 2025-10-08
Allow opaque sorts to take type arguments

petros-marko:
Allows something like
```rust
use flux_rs::*;

defs! {
    opaque sort Seq<T>;
    fn nil<T>() -> Seq<T>;
    fn cons<T>(v: T, tail: Seq<T>) -> Seq<T>;
}

/// The elements of the inner vec in reverse order
#[opaque]
#[refined_by(elems: Seq<T>)]
struct Foo<T> {
    inner: Vec<T>
}

impl<T> Foo<T> {
    #[trusted]
    #[sig(fn() -> Self[nil()])]
    pub fn new() -> Self {
        Self {
            inner: Vec::new()
        }
    }

    #[trusted]
    #[sig(fn(self: &mut Self[@elems], elem: T[@v]) ensures self : Self[cons(v, elems)])]
    pub fn push(&mut self, v: T) {
        self.inner.push(v)
    }
}

#[sig(fn() -> Foo<i32>[nil()])]
fn test_empty() -> Foo<i32> {
    Foo::new()
}

#[sig(fn() -> Foo<bool>[cons(false, nil())])]
fn test_cons() -> Foo<bool> {
    let mut list = Foo::new();
    list.push(false);
    list
}
```
to be written

### DIFF
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -20,7 +20,7 @@ use flux_common::{
 };
 use flux_middle::{
     THEORY_FUNCS,
-    def_id::{FluxLocalDefId, MaybeExternId},
+    def_id::{FluxDefId, MaybeExternId},
     fhir::{self, FhirId, FluxOwnerId, QPathExpr},
     global_env::GlobalEnv,
     queries::{QueryErr, QueryResult},
@@ -953,9 +953,9 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
     fn check_user_defined_sort_param_count(
         &mut self,
         path: &fhir::SortPath<'_>,
-        def_id: FluxLocalDefId,
+        def_id: FluxDefId,
     ) -> QueryResult {
-        let expected_param_count = self.genv().sort_decl_param_count(def_id)?;
+        let expected_param_count = self.genv().sort_decl_param_count(def_id);
         if path.args.len() != expected_param_count {
             let err = errors::IncorrectGenericsOnUserDefinedOpaqueSort::new(
                 path.segments.last().unwrap().span,
--- a/crates/flux-middle/src/fhir.rs
+++ b/crates/flux-middle/src/fhir.rs
@@ -898,7 +898,7 @@ pub enum SortRes {
     /// A primitive sort.
     PrimSort(PrimSort),
     /// A user declared sort.
-    User(FluxLocalDefId),
+    User(FluxDefId),
     /// A sort parameter inside a polymorphic function or data sort.
     SortParam(usize),
     /// The sort associated to a (generic) type parameter
--- a/crates/flux-middle/src/queries.rs
+++ b/crates/flux-middle/src/queries.rs
@@ -23,7 +23,7 @@ use rustc_macros::{Decodable, Encodable};
 use rustc_span::{Span, Symbol};
 
 use crate::{
-    def_id::{FluxDefId, FluxId, FluxLocalDefId, MaybeExternId, ResolvedDefId},
+    def_id::{FluxDefId, FluxId, MaybeExternId, ResolvedDefId},
     fhir,
     global_env::GlobalEnv,
     rty::{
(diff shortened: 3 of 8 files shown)


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -926,8 +926,8 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                 self.check_prim_sort_generics(path, fhir::PrimSort::Map)?;
                 rty::SortCtor::Map
             }
-            fhir::SortRes::User { name } => {
-                if !path.args.is_empty() {
+            fhir::SortRes::User { name, params } => {
+                if path.args.len() != params {
```
petros-marko: Should I create a new error type for this or can we repurpose the existing one (and write a new correct message for it)?

crates/flux-middle/src/fhir.rs
```diff
@@ -910,7 +911,7 @@ pub enum SortRes {
     /// A primitive sort.
     PrimSort(PrimSort),
     /// A user declared sort.
-    User { name: Symbol },
+    User { name: Symbol, params: usize },
```
nilehmann: A `SortRes` should have only enough information to recover the sort. If we need the number of params for a sort, we should query it from the sort declaration. Also, since we are going to revive this feature we should assign sort declarations a `FluxDefId` like we do for other flux defs.

```suggestion
    User(FluxDefId),
```

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: I'd repurpose this one (change the name maybe). Also, since we are at it, i'd move this to a function like we do with `check_prim_sort_generics`. I'm trying to move away from having `new` methods in the error and instead construct the error inside a method in whatever context we are working on.

crates/flux-middle/src/fhir.rs
petros-marko: What does "querying it from the sort declaration" look like? Is [the handling of SortRes::Adt](https://github.com/flux-rs/flux/blob/f4d733370cd77f6f01ba656fe606ce74abdd6565/crates/flux-fhir-analysis/src/conv/mod.rs#L939-L952) a good example to follow?

nilehmann: Yes, I think we want something similar to `AdtSortDef` but for user-declared (opaque) sorts

crates/flux-middle/src/queries.rs
```diff
@@ -522,6 +526,16 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
 
+    pub(crate) fn sort_decl_param_count(
+        &self,
+        genv: GlobalEnv,
+        def_id: FluxLocalDefId,
+    ) -> QueryResult<usize> {
+        run_with_cache(&self.sort_decl_param_count, def_id, || {
+            (self.providers.sort_decl_param_count)(genv, def_id)
+        })
+    }
```
nilehmann: We don't yet have support for importing Flux definitions from other crates, so a sort name would never resolve to a definition in an external crate. However, extern sorts can show up as part of extern signatures and other items. Thus, we should take a `FluxDefId` here and grab the result from the `cstore` if the id is not local. This will require:

1. Save the result of the query in the crate metadata like we do for `func_sort` here https://github.com/flux-rs/flux/blob/main/crates/flux-metadata/src/lib.rs#L345-L357
2. Add a method to the `CrateStore` trait here https://github.com/flux-rs/flux/blob/main/crates/flux-middle/src/cstore.rs#L10
3. Implement the new method like we do for `func_sort` here https://github.com/flux-rs/flux/blob/main/crates/flux-metadata/src/lib.rs#L311-L313
4. Modify this query to dispatch the id to either the local provider or to the `cstore` like we do for `func_sort`.

crates/flux-middle/src/fhir.rs
```diff
@@ -895,7 +898,7 @@ pub enum SortRes {
     /// A primitive sort.
     PrimSort(PrimSort),
     /// A user declared sort.
-    User { name: Symbol },
+    User(FluxLocalDefId),
```
nilehmann: As I mentioned, we don't yet have support for importing Flux definitions, but a sort could, in theory, come from an external crate. I think we can prepare for it and make this a `FluxDefId` without too many problems (like we do for `SpecFuncKind`).

```suggestion
    User(FluxDefId),
```

PR #1323, opened 2025-10-10
Strengthen `otherwise` cases with `remaining` cases

ranjitjhala:
@nilehmann you probably want to take a look at this, esp. the change to `Discr(..)` 

Lets us prove stuff like the below which somewhat frustratingly, we could not before...

```rust
#[reflect]
pub enum Bob {
    A,
    B,
    C,
}

#[spec(fn (&Bob[@b1], &Bob[@b2]) -> bool[b1 == b2])]
fn is_eq(b1: &Bob, b2: &Bob) -> bool {
    match (b1, b2) {
        (Bob::A, Bob::A) => true,
        (Bob::B, Bob::B) => true,
        (Bob::C, Bob::C) => true,
        _ => false,
    }
}
```

### DIFF
--- a/crates/flux-middle/src/rty/expr.rs
+++ b/crates/flux-middle/src/rty/expr.rs
@@ -1388,7 +1388,7 @@ pub(crate) mod pretty {
                     }
                 }
                 ExprKind::IsCtor(def_id, variant_idx, idx) => {
-                    w!(cx, f, "is::{:?}::{:?}( {:?} )", def_id, ^variant_idx, idx)
+                    w!(cx, f, "({:?} is {:?}::{:?})", idx, def_id, ^variant_idx)
                 }
                 ExprKind::Ctor(ctor, flds) => {
                     let def_id = ctor.def_id();


crates/flux-refineck/src/checker.rs
```diff
@@ -1284,7 +1295,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                     .as_bty_skipping_existentials()
                     .unwrap_or_else(|| tracked_span_bug!())
                     .expect_adt();
-                Ok(Ty::discr(adt_def.clone(), place.clone()))
+                let idx =
+                    if let TyKind::Indexed(_, idx) = ty.kind() { Some(idx.clone()) } else { None };
+                Ok(Ty::discr(adt_def.clone(), place.clone(), idx))
```
nilehmann: We don't need to store the `idx` here, we can look up the type again from the place when we need it.

crates/flux-refineck/src/checker.rs
```diff
@@ -1052,16 +1052,27 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
-
-        if remaining.len() == 1 {
+        let guard = if remaining.len() == 1 {
+            // If there's only one variant left, we know for sure that this is the one, so can force an unfold
             let (_, variant_idx) = remaining
                 .into_iter()
                 .next()
                 .unwrap_or_else(|| tracked_span_bug!());
-            successors.push((targets.otherwise(), Guard::Match(place.clone(), variant_idx)));
+            Guard::Match(place.clone(), variant_idx)
```
nilehmann, line 1074: I'm not sure if it'll be cleaner, but another way to implement this would be to generalize `Guard::Match` to take a list of `VariantIdx`s as in "the place was found to be of one of those variants", and then decide what to do when jumping to the successor.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -786,6 +796,8 @@ pub enum ExprKind {
     /// about the scope).
     Hole(HoleKind),
     ForAll(Binder<Expr>),
+    /// Is the expression constructed from constructor of the given Ctor (should be Enum)
+    IsCtor(Ctor),
```
nilehmann: Maybe we should add the expression being tested here instead of going through an application.

```suggestion
    /// Is the expression constructed from constructor of the given Ctor (should be Enum)
    IsCtor(Ctor, Expr),
```

Also, does it make sense to have this for `Ctor::Struct` or should we require `Ctor::Enum`?, i.e., 

```suggestion
    /// Is the expression constructed from constructor of the given Ctor (should be Enum)
    IsCtor(DefId, VariantIdx, Expr),
```

crates/flux-refineck/src/checker.rs
ranjitjhala: done

ranjitjhala: I don't think that's cleaner -- I considered it, and the "semantics" of the `Guard::Match` is that you then force an "unfold" at that index which we're not doing here, so I thought best to just leave that as the semantics of `Guard::Match` and not muddy the waters, and instead use the pure `Guard::Pred` instead.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -81,6 +81,7 @@ pub mod fixpoint {
         Global(GlobalVar, Option<Symbol>),
         Local(LocalVar),
         DataCtor(AdtId, VariantIdx),
+        IsDataCtor(AdtId, VariantIdx),
```
nilehmann: I think we should make a dedicated expr for this in `fixpoint`, such that we can translate to Z3 in `fixpoint-rs` (similar to how we have a special expression for theory functions)

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Yes, I think this turned out simpler, thanks!

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: You mean move the variant `IsDataCtor` from `Var` to `Expr`, yes?

nilehmann: Yes

ranjitjhala: Well then you'd need `DefId` and `VariantIdx` in there which seems to add some strange dependency?
 
 For now, I've 
 
 1. Left the `Var::DataCtor` as before,
 2. Removed the `Var::IsDataCtor`,
 2. Added an `IsCtor(T::Var, Box<Self>)` variant to the `enum Expr<T: Types>`.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1381,6 +1387,9 @@ pub(crate) mod pretty {
                         w!(cx, f, "({:?})", join!(", ", flds))
                     }
                 }
+                ExprKind::IsCtor(def_id, variant_idx, idx) => {
```
nilehmann, line 1390: Maybe we can print this as `e is Enun::Variant`, bit I don't feel strongly.

PR #1327, opened 2025-10-14
Make more tests pass with fixpoint-rs

nilehmann:
Implement a couple of missing features. Each commit implements a different fix.

### DIFF
--- a/lib/liquid-fixpoint/src/cstr2smt2.rs
+++ b/lib/liquid-fixpoint/src/cstr2smt2.rs
@@ -233,6 +233,21 @@ fn thy_func_application_to_z3<T: Types>(
             let arg2 = expr_to_z3(&args[1], env).as_set().unwrap();
             ast::Set::set_union(&[&arg1, &arg2]).into()
         }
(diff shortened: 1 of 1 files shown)


crates/flux-infer/src/fixpoint_qualifiers.rs
```diff
@@ -16,7 +16,7 @@ pub(crate) static FIXPOINT_QUALIFIERS: LazyLock<[Qualifier; 13]> = LazyLock::new
     let eqfalse = Qualifier {
         name: String::from("EqFalse"),
         args: vec![(Var::Local(LocalVar::from(0u32)), Sort::Bool)],
-        body: Expr::Neg(Box::new(Expr::Var(Var::Local(LocalVar::from(0u32))))),
+        body: Expr::Not(Box::new(Expr::Var(Var::Local(LocalVar::from(0u32))))),
```
nilehmann, line 19: This has always been wrong, but it looks like fixpoint-hs is more lenient and accepts `(- true)`

PR #1334, opened 2025-10-15
Generate defs independent of any constraint

petros-marko:
First pass at doing this (very hacky)

Pros:
- it works (see below for when it doesn't work)
- adding support for opaque functions and sorts seems straightforward

Cons:
- probably wrecks query caching
- seems brittle

I tried to run this on tock like I did before. tock uses some structs refined by tuples unrelated to the theorems we want to prove in lean, but because this approach looks to generate lean definitions for all flux defs regardless of use in externally proven theorems, it crashes. Adding support for Adt sorts in the lean encoder would fix this.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -419,17 +419,13 @@ impl<'genv, 'tcx, Tag> FixpointCtxt<'genv, 'tcx, Tag>
 where
     Tag: std::hash::Hash + Eq + Copy,
 {
-    pub fn new(
-        genv: GlobalEnv<'genv, 'tcx>,
-        def_id: Option<MaybeExternId>,
-        kvars: KVarGen,
-    ) -> Self {
+    pub fn new(genv: GlobalEnv<'genv, 'tcx>, def_id: MaybeExternId, kvars: KVarGen) -> Self {
         Self {
             comments: vec![],
             kvars,
             scx: SortEncodingCtxt::default(),
             genv,
-            ecx: ExprEncodingCtxt::new(genv, def_id),
+            ecx: ExprEncodingCtxt::new(genv, Some(def_id)),
             kcx: Default::default(),
             tags: IndexVec::new(),
             tags_inv: Default::default(),
@@ -439,7 +435,7 @@ where
     pub fn check(
         mut self,
         cache: &mut FixQueryCache,
-        def_id: Option<MaybeExternId>,
+        def_id: MaybeExternId,
         constraint: fixpoint::Constraint,
         kind: FixpointQueryKind,
         scrape_quals: bool,
@@ -451,18 +447,13 @@ where
             return Ok(vec![]);
         }
         let def_span = self.ecx.def_span();
-        let (define_funs, define_constants, qualifiers) = if let Some(def_id) = def_id {
-            let (define_funs, define_constants) = self.ecx.define_funs(def_id, &mut self.scx)?;
-            let qualifiers = self
-                .ecx
-                .qualifiers_for(def_id.local_id(), &mut self.scx)?
-                .into_iter()
-                .chain(FIXPOINT_QUALIFIERS.iter().cloned())
-                .collect();
-            (define_funs, define_constants, qualifiers)
-        } else {
-            (vec![], vec![], FIXPOINT_QUALIFIERS.iter().cloned().collect())
-        };
+        let (define_funs, define_constants) = self.ecx.define_funs(def_id, &mut self.scx)?;
+        let qualifiers = self
+            .ecx
+            .qualifiers_for(def_id.local_id(), &mut self.scx)?
+            .into_iter()
+            .chain(FIXPOINT_QUALIFIERS.iter().cloned())
+            .collect();
 
         let kvars = self.kcx.into_fixpoint();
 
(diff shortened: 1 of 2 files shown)


crates/flux-driver/src/callbacks.rs
```diff
@@ -148,7 +151,32 @@ struct CrateChecker<'genv, 'tcx> {
 
 impl<'genv, 'tcx> CrateChecker<'genv, 'tcx> {
     fn new(genv: GlobalEnv<'genv, 'tcx>) -> Self {
-        CrateChecker { genv, cache: QueryCache::load() }
+        let checker = CrateChecker { genv, cache: QueryCache::load() };
+        checker.encode_flux_items_in_lean();
+        checker
+    }
+
+    fn encode_flux_items_in_lean(&self) {
```
nilehmann: We cannot emit lean unconditionally (failing tests are a testament to that). We should do this if (1) there's at least one function marked as `proven_externally` or (2) you opted into lean with some global flag.

crates/flux-driver/src/callbacks.rs
```diff
@@ -148,7 +151,32 @@ struct CrateChecker<'genv, 'tcx> {
...
+        let mut fcx: FixpointCtxt<'_, '_, i32> = FixpointCtxt::new(self.genv, None, None);
```
nilehmann: Going through the code, I realize now that we don't need a `FixpointCtxt`, we only need an `ExprEncodingCtxt` and a `SortEncodingCtxt`. That way, we don't need to make the `KvarGen` optional.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -107,7 +108,7 @@ pub mod fixpoint {
         fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
             match self {
                 Var::Global(v, None) => write!(f, "c{}", v.as_u32()),
-                Var::Global(v, Some(sym)) => write!(f, "f${}${}", sym, v.as_u32()),
+                Var::Global(v, Some(sym)) => write!(f, "f${}${}", sym.name(), v.as_u32()),
```
```suggestion
                Var::Global(v, Some(did)) => write!(f, "f${}${}", did.name(), v.as_u32()),
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -446,17 +451,21 @@ where
         }
         let def_span = self.ecx.def_span();
         let def_id = self.ecx.def_id;
+        let (define_funs, define_constants, qualifiers) = if let Some(def_id) = def_id {
```
nilehmann: I think we should modify this method to take the `def_id` as a parameter instead of grabbing it from the `ExprEncodingCtxt`.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1071,7 +1099,7 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
     }
 
     fn def_span(&self) -> Span {
-        self.genv.tcx().def_span(self.def_id)
+        self.genv.tcx().def_span(self.def_id.unwrap())
```
nilehmann: The idea of this span is to give a better message when panicking, if we `unwrap` here that completely defeats the purpose.

```suggestion
        self.def_id
            .map_or(DUMMY_SP, |def_id| self.genv.tcx().def_span(def_id))
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1812,6 +1841,24 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         Ok((defs, consts))
     }
 
+    pub fn to_fun_def(
```
nilehmann: For consistency

```suggestion
    pub fn fun_def_to_fixpoint(
```

nilehmann (changes requested):

crates/flux-driver/src/callbacks.rs
petros-marko: In the changes I made, I just added a flag to the config to trigger the generation of defs, meaning that if you have `proven_externally` annotations, you also need to enable the flag. Is that OK or should I get it to emit the defs regardless of the flag if some function is proven externally?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -434,6 +439,7 @@ where
     pub fn check(
         mut self,
         cache: &mut FixQueryCache,
+        def_id: Option<MaybeExternId>,
```
nilehmann: I meant that we don't need this to be optional if we make it a parameter of the function

```suggestion
        def_id: MaybeExternId,
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -418,7 +419,11 @@ impl<'genv, 'tcx, Tag> FixpointCtxt<'genv, 'tcx, Tag>
 where
     Tag: std::hash::Hash + Eq + Copy,
 {
-    pub fn new(genv: GlobalEnv<'genv, 'tcx>, def_id: MaybeExternId, kvars: KVarGen) -> Self {
+    pub fn new(
+        genv: GlobalEnv<'genv, 'tcx>,
+        def_id: Option<MaybeExternId>,
```
nilehmann: Same here, I think we don't need this to be optional now that we are only using `ExprEncodingCtxt`.

```suggestion
        def_id: MaybeExternId,
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -513,29 +532,34 @@ where
         }
     }
 
+    pub fn to_fun_def(&mut self, def_id: FluxDefId) -> QueryResult<fixpoint::FunDef> {
+        self.ecx.fun_def_to_fixpoint(def_id, &mut self.scx)
+    }
+
```
nilehmann: This function is not used

```suggestion
```

PR #1340, opened 2025-10-18
Track imm args for decoding from fixpoint

cole-k:
We still don't have a good story for tracking the local vars that we generate from bound variables, but hopefully those are rare enough that we don't need to worry about them just yet.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -830,8 +830,8 @@ struct LocalVarEnv {
     layers: Vec<Vec<fixpoint::LocalVar>>,
     /// While it might seem like the signature should be
     /// [`UnordMap<fixpoint::LocalVar, rty::Var>`], we encode the arguments to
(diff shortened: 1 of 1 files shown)


cole-k:
If you look at the commit history you will see that `rustfmt` is not idempotent (farewell CI :money_with_wings:)

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -828,7 +828,7 @@ struct LocalVarEnv {
     fvars: UnordMap<rty::Name, fixpoint::LocalVar>,
     /// Layers of late bound variables
     layers: Vec<Vec<fixpoint::LocalVar>>,
-    reverse_map: UnordMap<fixpoint::LocalVar, rty::Var>,
+    reverse_map: UnordMap<fixpoint::LocalVar, rty::Expr>,
```
nilehmann, line 835: Can you please add a comment here explaining that this is an `rty::Expr` because of A-normalization

PR #1343, opened 2025-10-20
Generate Typeclass for Opaque Defs

petros-marko:
```rust
flux_rs::defs! {
    opaque sort FSeq<T>;
    fn Fempty<T>() -> FSeq<T>;
    fn Fcons<T>(v: T, s: FSeq<T>) -> FSeq<T>;
    fn FmapInc(s: FSeq<int>) -> FSeq<int>;
    fn Fsingleton(v: int) -> FSeq<int> {
        Fcons(v, Fempty())
    }
}
#[flux::proven_externally]
#[flux::sig(fn() ensures FmapInc(Fsingleton(1)) == Fsingleton(2))]
pub fn test02() {}

#[flux::proven_externally]
#[flux::sig(fn() ensures FmapInc(Fcons(2, Fcons(1, Fsingleton(0)))) == Fcons(3, Fcons(2, Fsingleton(1))))]
pub fn test03() {}
```
can now be proven (see the [lean_proofs.zip](https://github.com/user-attachments/files/23006463/lean_proofs.zip) for details).
In the lean proofs:
- `{Defs, InferredInstance, OpaqueFluxDefs, Test02, Test03}.lean` are generated
- `{Instance, Test02Proof, Test03Proof}.lean` only generate a skeleton if the file does not exist, the user has to fill them in
- `InstTheorems.lean` is just hand written. Maybe we could automate the generation though...

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -161,9 +161,8 @@ pub mod fixpoint {
                     write!(f, "Adt{}", adt_id.as_u32())
                 }
                 // There's no way to declare opaque sorts in the fixpoint horn syntax so we encode user
-                // declared opaque sorts, type parameter sorts, and (unormalizable) type alias sorts as
-                // integers. Well-formedness should ensure values of these sorts are used "opaquely",
-                // i.e., the only values of these sorts are variables.
+                // declared opaque sorts as integers. Well-formedness should ensure values of these
+                // sorts are used "opaquely", i.e., the only values of these sorts are variables.
                 DataSort::User(..) => {
                     write!(f, "int")
                 }
@@ -260,6 +259,10 @@ impl SortEncodingCtxt {
                     args,
                 )
             }
+            // There's no way to declare opaque sorts in the fixpoint horn syntax so we encode type
+            // parameter sorts and (unormalizable) type alias sorts as integers. Well-formedness
+            // should ensure values of these sorts are used "opaquely", i.e., the only values of
+            // these sorts are variables.
             rty::Sort::Param(_)
             | rty::Sort::Alias(rty::AliasKind::Opaque | rty::AliasKind::Projection, ..) => {
                 fixpoint::Sort::Int
--- a/crates/flux-infer/src/lean_encoding.rs
+++ b/crates/flux-infer/src/lean_encoding.rs
@@ -10,13 +10,10 @@ use flux_middle::{
     global_env::GlobalEnv,
     queries::{QueryErr, QueryResult},
 };
-use itertools::Itertools;
 
 use crate::{
     fixpoint_encoding::fixpoint,
-    lean_format::{
-        self, LeanConstDecl, LeanDataField, LeanSort, LeanSortDecl, LeanSortVar, LeanVar,
-    },
+    lean_format::{self, LeanConstDecl, LeanSortDecl, LeanSortVar, LeanVar},
 };
 
 pub struct LeanEncoder<'genv, 'tcx, 'a> {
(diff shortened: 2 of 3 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -138,6 +138,7 @@ pub mod fixpoint {
     pub enum DataSort {
         Tuple(usize),
         Adt(AdtId),
+        User(Symbol),
```
nilehmann: This should be a `FluxDefId` to make sure you can generate a unique identifier in Lean, no?

```suggestion
        User(FluxDefId),
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -149,6 +150,9 @@ pub mod fixpoint {
                 DataSort::Adt(adt_id) => {
                     write!(f, "Adt{}", adt_id.as_u32())
                 }
+                DataSort::User(..) => {
+                    write!(f, "int")
```
nilehmann, line 167: Please add a comment explaining why this is `int`.

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -44,35 +47,153 @@ impl<'genv, 'tcx, 'a> LeanEncoder<'genv, 'tcx, 'a> {
...
         let mut file = fs::File::create(defs_path)?;
-        writeln!(file, "mutual")?;
-        for fun_def in defs {
-            writeln!(file, "{}", lean_format::LeanFunDef(fun_def, self.genv))?;
+        if has_opaques {
+            writeln!(file, "import {}.InferredInstance", pascal_project_name.as_str())?;
+        }
+        if !data_decls.is_empty() {
+            writeln!(file, "-- STRUCT DECLS --")?;
+            writeln!(file, "mutual")?;
```
nilehmann, line 151: This may be fine for now, but defining all types to be mutually recursive (when they are not) will put everything in the same recursor (i.e., inductive principle), which will make it very annoying to work with.

crates/flux-infer/src/lean_format.rs
```diff
@@ -5,17 +5,73 @@ use itertools::Itertools;
...
+        match self.0 {
+            DataSort::User(symbol) => write!(f, "{}", symbol),
+            _ => write!(f, "{}", self.0.display().to_string().replace("$", "_")),
+        }
+    }
+}
+
+impl<'a, 'genv, 'tcx> fmt::Display for LeanDataDecl<'a, 'genv, 'tcx> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.0.ctors.is_empty() {
```
nilehmann: A data type could have no constructors (an empty type), so this is not enough to distinguish between opaque sorts and data types. I think it make sense to have them separate, i.e, create a new `fixpoint::SortDecl`

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1815,6 +1844,16 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         Ok((defs, consts))
     }
 
+    pub fn const_decl_to_fixpoint(
```
```suggestion
    pub fn func_decl_to_fixpoint(
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1815,6 +1844,16 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+        &mut self,
+        def_id: FluxDefId,
+        scx: &mut SortEncodingCtxt,
+    ) -> fixpoint::ConstDecl {
+        let name = *self.const_env.fun_def_map.get(&def_id).unwrap();
```
```suggestion
        let name = self.const_env.fun_def_map.get[&def_id];
```

nilehmann (commented):
This is very exciting @petros-marko! A couple of comments.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -159,6 +160,13 @@ pub mod fixpoint {
                 DataSort::Adt(adt_id) => {
                     write!(f, "Adt{}", adt_id.as_u32())
                 }
+                // There's no way to declare opaque sorts in the fixpoint horn syntax so we encode user
+                // declared opaque sorts, type parameter sorts, and (unormalizable) type alias sorts as
+                // integers. Well-formedness should ensure values of these sorts are used "opaquely",
+                // i.e., the only values of these sorts are variables.
```
nilehmann: This comment is wrong; it should only refer to opaque sorts. Also, sort parameters and (unnormalizable) type alias sorts are still encoded as `int` in `sort_to_fixpoint`, so part of this comment should go there.

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -44,35 +50,193 @@ impl<'genv, 'tcx, 'a> LeanEncoder<'genv, 'tcx, 'a> {
...
+            for data_decl in data_decls {
+                writeln!(file, "{}", lean_format::LeanDataDecl(data_decl, self.genv))?;
+            }
+            writeln!(file, "end")?;
+        }
+        if !structs.is_empty() {
+            writeln!(file, "-- CONSTRUCTORS & Projections --")?;
```
nilehmann: When you define `structure MyStruct`, Lean generates a constructor called `MyStruct.mk`. Can you use that? The same for fields, each field has an accessor called `MyStruct.field`.

nilehmann: You can also change the name of the default constructor like so

```lean
structure MyStruct where
  otherCtorName::
  field1: Int
  field2: Int
```
That will generate a constructor called `MyStruct.otherCtorName`

crates/flux-infer/src/fixpoint_encoding.rs
petros-marko: Oops, you're right; I just took it as it was. I'll fix it before merging.

crates/flux-infer/src/lean_encoding.rs
petros-marko: This was a hack admittedly. Since the fixpoint representations of the constructors and projections have enough info to recover the struct they refer to, I will attempt to fix it in the lean encoding of the *uses* of these things.

PR #1349, opened 2025-10-23
[DO NOT MERGE] Prove very basic things about an int vector

petros-marko:
Start writing a vector implementation that's refined by the vector's elements.
Prove a couple of simple facts about such a vector holding i32s using lean. The reason why the lemmas are specialized to ints is that we aren't allowed to write generic refinement functions that are not opaque.

This feels very clunky. Mainly I'm putting this up for review to get feedback on whether this is what you were picturing for writing these proofs.

I did Nico's suggested trick for defining the lean-land `get` function for lists. It's necessary because even though it's only used in a context where we know that the index is in bounds, the definition doesn't reflect that in any way, so it has to do something when it's out of bounds.


petros-marko:
Yeah, it’s been merged into one of the demo repos.

PR #1351, opened 2025-10-23
Factor out fixpoint BoundVar

cole-k:
I want to be able to substitute over `BoundVar`s in the `fixpoint` encoding (which presently only can be introduced in `fixpoint`'s own output), but this is difficult when the `BoundVar` is a part of the polymorphic `T::Var`. So I've factored it out into its own enumeration in `fixpoint::Expr<T>`.

This also includes a helper function for getting free vars (which are unrelated to `BoundVar`s) from these exprs and a helper function for performing substitution over bound vars. I need both in my branch, although I could remove these functions and only include them in my brach.

I tested with `cargo x run tests/tests/pos/surface/loop01.rs` to make sure that parsing is unchanged.

### DIFF
--- a/lib/liquid-fixpoint/src/parser.rs
+++ b/lib/liquid-fixpoint/src/parser.rs
@@ -498,6 +498,7 @@ where
     fn push_scope(&mut self, names: &[String]) {
         self.scopes.push(names.iter().cloned().collect());
     }
+
     fn pop_scope(&mut self) {
         self.scopes.pop();
     }


lib/liquid-fixpoint/src/parser.rs
```diff
@@ -477,14 +486,30 @@ where
...
             Ok((sorts, expr))
         } else {
             Err(ParseError::err("expected (lambda (params) body)"))
         }
     }
+
+    fn push_scope(&mut self, names: &[String]) {
+        self.scopes.push(names.iter().cloned().collect());
+    }
+    fn pop_scope(&mut self) {
```
nilehmann, line 502:
```suggestion

    fn pop_scope(&mut self) {
```

PR #1354, opened 2025-10-30
Metrics report

nilehmann:


ranjitjhala:
@nilehmann do you want to take a look? I'm working on the playground in the meantime.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -35,11 +35,16 @@ impl<T: Types> Constraint<T> {
 
     /// Returns true if the constraint has at least one concrete RHS ("head") predicates.
     /// If `!c.is_concrete` then `c` is trivially satisfiable and we can avoid calling fixpoint.
-    pub fn is_concrete(&self) -> bool {
+    /// Returns the number of concrete, non-trivial head predicates in the constraint.
+    pub fn concrete_head_count(&self, count: &mut usize) {
```
nilehmann: hmm, it feels wrong to expose `count`. I typically use this pattern

```rust
pub fn concrete_head_count(&self) -> usize {
    fn go(cstr: &Constraint, count: &mut usize) { ... }
    let mut count = 0;
    go(self, &mut count);
    count
}
```

nilehmann (commented):
I can't approve because it's my own PR, but it LGTM

lib/liquid-fixpoint/src/constraint.rs
ranjitjhala: I like that, will do.

PR #1357, opened 2025-11-05
Add V0 of `#[flux::no_panic]`

ninehusky:
cc @nilehmann 

Adds a `no_panic` annotation which yields an error iff:
1. A function annotated with `no_panic` has an explicit `panic` inside.
2. A function annotated with `no_panic` calls a function which is **not** annotated with `no_panic`.

The next steps will involve figuring out which functions from `core` do panic, and manually annotating those functions with `no_panic` if flux machinery prevents them from panicking (like with [unwrap](https://github.com/flux-rs/flux/blob/6b080b32801f923bfb590ac48e729de38f829f21/lib/flux-core/src/option.rs#L21).

### DIFF
--- a/crates/flux-middle/src/queries.rs
+++ b/crates/flux-middle/src/queries.rs
@@ -269,7 +269,7 @@ pub struct Queries<'genv, 'tcx> {
     fn_sig: Cache<DefId, QueryResult<rty::EarlyBinder<rty::PolyFnSig>>>,
     lower_late_bound_vars: Cache<LocalDefId, QueryResult<List<ty::BoundVariableKind>>>,
     sort_decl_param_count: Cache<FluxDefId, usize>,
-    no_panic: Cache<DefId, bool>
+    no_panic: Cache<DefId, bool>,
 }
 
 impl<'genv, 'tcx> Queries<'genv, 'tcx> {
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -890,7 +890,6 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             }
         }
 
-
         let actuals =
             unfold_local_ptrs(infcx, env, fn_sig.skip_binder_ref(), actuals).with_span(span)?;
         let actuals = infer_under_mut_ref_hack(infcx, &actuals, fn_sig.skip_binder_ref());
--- a/crates/flux-metadata/src/lib.rs
+++ b/crates/flux-metadata/src/lib.rs
@@ -176,7 +176,7 @@ pub struct Tables<K: Eq + Hash> {
     func_sort: UnordMap<FluxId<K>, rty::PolyFuncSort>,
     func_span: UnordMap<FluxId<K>, Span>,
     sort_decl_param_count: UnordMap<FluxId<K>, usize>,
-    no_panic: UnordMap<K, bool>
+    no_panic: UnordMap<K, bool>,
 }
 
 impl CStore {


crates/flux-refineck/src/checker.rs
```diff
@@ -789,6 +790,26 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            if let Some(callee_def_id) = callee_def_id {
+                let callee_no_panic = if let Some(local_callee) = callee_def_id.as_local() {
+                    genv.no_panic(local_callee)
+                } else {
+                    false
+                };
+
+                if !callee_no_panic {
+                    // This needs to be a different shape, I imagine.
+                    return Err(CheckerError {
```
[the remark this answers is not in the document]
ninehusky: (this doesn't compile, btw, it's just placeholder text)

crates/flux-refineck/src/checker.rs
```diff
@@ -789,6 +789,18 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
         let genv = self.genv;
         let tcx = genv.tcx();
 
+        let no_panic = genv.no_panic(self.def_id);
+        if no_panic && let Some(callee_def_id) = callee_def_id {
+            let callee_no_panic = callee_def_id
+                .as_local()
+                .is_some_and(|local| genv.no_panic(local));
```
nilehmann: This won't work for functions defined in other crates, i.e., if you import a function `foo` from another crate, `foo`'s def id won't be local. In particular, this won't work if you call a function in `core`.

We need to add a _query_. There isn't much documentation about queries. I suppose the best place to look at is this PR by @petros-marko that adds a new query https://github.com/flux-rs/flux/pull/1317. Also, happy to chat about it.

ninehusky: I think I get the general shape of what adding a query would look like. Making sure we're talking about the same thing: the query would be something of the type `fn is_no_panic(genv, def_id) -> bool`?

crates/flux-middle/src/queries.rs
```diff
@@ -193,6 +193,7 @@ pub struct Providers {
     pub item_bounds:
         fn(GlobalEnv, MaybeExternId) -> QueryResult<rty::EarlyBinder<List<rty::Clause>>>,
     pub sort_decl_param_count: fn(GlobalEnv, FluxId<MaybeExternId>) -> usize,
+    pub no_panic: fn(GlobalEnv, MaybeExternId) -> QueryResult<bool>,
```
nilehmann: Since you are inlining the implementation, you don't need a provider. I'll mark all the parts you can remove.

```suggestion
```

crates/flux-middle/src/queries.rs
```diff
@@ -230,6 +231,7 @@ impl Default for Providers {
             item_bounds: |_, _| empty_query!(),
             constant_info: |_, _| empty_query!(),
             sort_decl_param_count: |_, _| empty_query!(),
+            no_panic: |_, _| empty_query!(),
```
nilehmann:
```suggestion
```

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -72,6 +72,11 @@ pub fn provide(providers: &mut Providers) {
     providers.default_assoc_refinement_body = default_assoc_refinement_body;
     providers.item_bounds = item_bounds;
     providers.sort_decl_param_count = sort_decl_param_count;
+    providers.no_panic = no_panic;
+}
+
+fn no_panic(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<bool> {
+    genv.no_panic(def_id.local_id())
```
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -877,6 +876,18 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+
+        if no_panic.unwrap_or(false)
+            && let Some(callee_def_id) = callee_def_id
+            && genv.def_kind(callee_def_id).is_fn_like()
+        {
+            let callee_no_panic = genv.no_panic(callee_def_id).unwrap_or(false);
+            if "refine" == M::NAME && !callee_no_panic {
+                genv.sess().emit_err(errors::PanicError { span });
+            }
+        }
```
nilehmann: You could guard the entire thing behind `"refine" == M:NAME`. I don't feel strongly about this.

```suggestion
        if M::NAME == "refine" {
            let no_panic = genv.no_panic(self.checker_id.root_id());

            if no_panic.unwrap_or(false)
                && let Some(callee_def_id) = callee_def_id
                && genv.def_kind(callee_def_id).is_fn_like()
            {
                let callee_no_panic = genv.no_panic(callee_def_id).unwrap_or(false);
                if !callee_no_panic {
                    genv.sess().emit_err(errors::PanicError { span });
                }
            }
        }

```

crates/flux-middle/src/queries.rs
```diff
@@ -583,6 +587,24 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
         })
     }
 
+    pub(crate) fn no_panic(&self, genv: GlobalEnv, def_id: DefId) -> QueryResult<bool> {
```
nilehmann: This is infallible, so we don't need to return a result here.

```suggestion
    pub(crate) fn no_panic(&self, genv: GlobalEnv, def_id: DefId) -> bool {
```

crates/flux-middle/src/queries.rs
```diff
@@ -583,6 +587,24 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
+                |def_id| {
+                    match def_id {
+                        MaybeExternId::Local(def_id) => Ok(genv.fhir_attr_map(def_id).no_panic()),
+                        MaybeExternId::Extern(def_id, _) => {
+                            Ok(genv.fhir_attr_map(def_id).no_panic())
+                        }
+                    }
+                },
+                |def_id| genv.cstore().no_panic(def_id),
+                |def_id| genv.cstore().no_panic(def_id).unwrap_or(Ok(false)),
```
nilehmann: The point of the last case is that def_id is _not_ in the crate store, so querying it again will return `None`. We must return a default value here.


```suggestion
                |def_id| false,
```

PR #1358, opened 2025-11-05
Invariants about `char` needed for `core`

ranjitjhala:


crates/flux-middle/src/rty/mod.rs
```diff
@@ -2746,6 +2747,29 @@ fn uint_invariants(uint_ty: UintTy, overflow_mode: OverflowMode) -> &'static [In
...
 }
 
+fn char_invariants() -> &'static [Invariant] {
+    static INVARIANTS: LazyLock<[Invariant; 2]> = LazyLock::new(|| {
+        [
+            Invariant {
+                pred: Binder::bind_with_sort(
+                    Expr::le(
+                        Expr::cast(Sort::Char, Sort::Int, Expr::nu()),
+                        Expr::constant((char::MAX as usize).into()),
```
nilehmann: Can this be `char::MAX as u32`? That feels more correct

nilehmann (changes requested):
@ranjitjhala Char to int cast is defined like so:

```rust
fn char_uint_cast(idx: &Expr, uint_ty: UintTy) -> Ty {
    let idx = Expr::cast(rty::Sort::Char, rty::Sort::Int, idx.clone());
    Ty::indexed(BaseTy::Uint(uint_ty), idx)
}
```

That's wrong if the target uint is `u8` or `u16`. This was wrong before.

crates/flux-middle/src/rty/mod.rs
ranjitjhala: Right, I suppose we don't have an `into` for `u32` but I'll add it.

ranjitjhala:
I was assuming that `rustc` would yell if you try to cast a `char` to a `u8` or `u16`?

ranjitjhala:
Should we only use this indexed version if the "width" is >= 32 bits?

crates/flux-middle/src/rty/mod.rs
ranjitjhala: done.

crates/flux-refineck/src/checker.rs
```diff
@@ -2015,7 +2015,15 @@ fn uint_char_cast(idx: &Expr) -> Ty {
 
 fn char_uint_cast(idx: &Expr, uint_ty: UintTy) -> Ty {
     let idx = Expr::cast(rty::Sort::Char, rty::Sort::Int, idx.clone());
-    Ty::indexed(BaseTy::Uint(uint_ty), idx)
+    if let Some(w) = uint_ty.bit_width()
```
ranjitjhala: @nilehmann this is the updated version, probably want to double check.

tests/tests/pos/surface/char03.rs
```diff
@@ -15,7 +15,7 @@ pub fn cast_to_u8(x: char) -> u8 {
     x
 }
 
-#[spec(fn (x:char) -> char[x])]
+#[spec(fn (x:char{ cast(x) < 256 }) -> char[x])]
```
ranjitjhala, line 18: this test had to be adjusted accordingly, as it was bogus!

crates/flux-refineck/src/checker.rs
```suggestion
    if uint_bith_width(uint_ty) >= 32 {
```

nilehmann (approved):
Can you add a comment to `uint_to_char` saying that's fine because Rust only allows `u8 -> char` casts?

PR #1367, opened 2025-11-12
Implement binary set operators and literals

nilehmann:


ranjitjhala:
Nice! Did you do set literals too? I have a branch with that already...

PR #1370, opened 2025-11-13
Refine what we consider a function for the purpose of metrics

nilehmann:


crates/flux-driver/src/callbacks.rs
```diff
@@ -303,7 +315,10 @@ impl<'genv, 'tcx> CrateChecker<'genv, 'tcx> {
                 // Make sure we run conversion and report errors even if we skip the function
                 force_conv(self.genv, def_id.resolved_id()).emit(&self.genv)?;
                 let Some(local_id) = def_id.as_local() else { return Ok(()) };
-                refineck::check_fn(self.genv, &mut self.cache, local_id)
+                if is_fn_with_body {
+                    refineck::check_fn(self.genv, &mut self.cache, local_id)?;
```
ranjitjhala, line 322: What would we not check here?

nilehmann: This is skipping trait methods without a default body (there's nothing to check). The check was originally inside `check_fn`, I just moved it here.

PR #1372, opened 2025-11-13
Minor fixes needed for demo

petros-marko:
Mark structures with ext, and skip polymorphic functions in lean defs since they get inlined anyway.


crates/flux-infer/src/lean_format.rs
```diff
@@ -70,6 +70,7 @@ impl<'a> fmt::Display for LeanSortVar<'a> {
 impl<'a, 'genv, 'tcx> fmt::Display for LeanDataDecl<'a, 'genv, 'tcx> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         if self.0.ctors.len() == 1 {
+            writeln!(f, "@[ext]")?;
```
nilehmann, line 73: What's `@[ext]`? Just curious 🧐

nilehmann (approved):

crates/flux-infer/src/lean_format.rs
petros-marko: It allows you to prove that two objects of some structure type are equal by proving that their corresponding fields are equal.

PR #1376, opened 2025-11-15
Augment `no_panic` to work over modules

ninehusky:


crates/flux-middle/src/queries.rs
```diff
@@ -590,11 +590,37 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
             def_id.dispatch_query(
                 genv,
                 |def_id| {
-                    let local_id = match def_id {
+                    let mut current_id = match def_id {
```
nilehmann: You can use `def_id.local_id()` instead of matching

crates/flux-middle/src/queries.rs
```diff
@@ -590,11 +590,37 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
+
+                        // Move to the next parent
+                        if let Some(parent) = genv.tcx().opt_local_parent(current_id) {
+                            current_id = parent;
+                        } else {
+                            break; // Reached the top
+                        }
+                    }
+
+                    false
```
nilehmann: Return the value of the flag here, rather than reading it in the checker. This will ensure that a crate checked with the flag on will have the right value.

nilehmann (approved):
@ninehusky LGTM. Is there anything else you'd like to add to this PR, or would you like to merge it?

PR #1377, opened 2025-11-17
Add extern specs for all `core` functions used by ring buffer

ninehusky:
Adds `no_panic` specs for the following functions:
```
core::slice::<impl [T]>::len: 9
core::num::<impl usize>::saturating_sub: 1
core::slice::<impl [T]>::split_at: 1
core::slice::index::<impl core::ops::Index<I> for [T]>::index: 2
core::ops::Fn::call: 1
core::ops::FnMut::call_mut: 1
```

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -19,6 +19,8 @@ jobs:
 
       - name: Install mdBook
         uses: peaceiris/actions-mdbook@v1.2.0
+        with:
+          mdbook-version: "0.4.48"
 
       - run: mdbook build book
 
--- a/book/book.toml
+++ b/book/book.toml
@@ -1,7 +1,6 @@
 [book]
 authors = ["Ranjit Jhala"]
 language = "en"
-multilingual = false
 src = "src"
 title = "The Flux Book"
 
--- a/crates/flux-desugar/src/lib.rs
+++ b/crates/flux-desugar/src/lib.rs
@@ -96,12 +96,13 @@ pub fn desugar<'genv>(
             fhir::Node::ForeignItem(genv.alloc(item))
         }
         rustc_hir::Node::Ctor(rustc_hir::VariantData::Tuple(_, _, _)) => fhir::Node::Ctor,
-        node => {
-            if let Some(ident) = node.ident() {
-                return Err(query_bug!(def_id, "unsupported item {ident:?}"));
-            } else {
-                return Err(query_bug!(def_id, "unsupported item"));
-            }
+
+        _ => {
+            return Err(query_bug!(
+                def_id,
+                "unexpected {} found in desugaring",
+                genv.tcx().def_descr(def_id.to_def_id())
+            ));
         }
     };
     nodes.insert(def_id, node);
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -396,25 +396,6 @@ fn conv_generic_param_kind(kind: &fhir::GenericParamKind) -> rty::GenericParamDe
     }
 }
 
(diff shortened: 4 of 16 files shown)


crates/flux-bin/src/lib.rs
```diff
@@ -29,6 +29,9 @@ pub struct FluxMetadata {
     /// If present, only check files that match any of the glob patterns. Patterns are checked
     /// relative to the location of the manifest file.
     pub include: Option<Vec<String>>,
+    /// If present, every function in the module is implicitly labeled with a `no_panic` by default.
+    /// This means that the only way a function can panic is (1) if it panics, or (2) it calls an external function without this attribute.
```
enjhnsn2: "This means that the only way a function can panic is (1) if it panics..." 
I think you can elide this comment.

tests/tests/neg/surface/no_panic01.rs
```diff
@@ -0,0 +1,19 @@
+pub fn this_might_panic() -> i32 {
+    let v = vec![1, 2, 3];
+    v[0]
+}
+
+#[cfg_attr(flux, flux::no_panic)]
```
enjhnsn2, line 6: why is this not just `#[flux::no_panic]`?

tests/tests/pos/surface/no_panic03.rs
```diff
@@ -0,0 +1,22 @@
+#[cfg_attr(flux, flux::no_panic)]
```
enjhnsn2, line 1: echo above comment about this being `#[flux::no_panic]`

tests/tests/pos/surface/no_panic03.rs
```diff
@@ -0,0 +1,22 @@
...
+mod my_other_module {
+    pub fn this_wont_panic() -> i32 {
+        3
+    }
+}
+
+
+#[cfg_attr(flux, flux::no_panic)]
```
enjhnsn2, line 9: #[flux::no_panic]

enjhnsn2 (changes requested):
Mostly looks good. Specs appear to be sound, but are currently trusted, but it would be good to at least attempt to verify them at some point.

tests/tests/neg/surface/no_panic01.rs
ninehusky: @nilehmann deferring to you for a smarter explanation. From what I remember, this has to do with Rust's mechanism for attaching attributes to modules? There's something about `#[no_panic]` becoming a "macro within a macro" which causes issues when showing up as a module annotation?

nilehmann: Since this is a test, this is actually not needed, because we never run `rustc` directly on tests. When using this for real, you can't write `flux::no_panic` directly because during normal compilation, rustc will complain about `flux::no_panic` not existing.

enjhnsn2: wait, but you can do `flux_rs::no_panic`, correct? Isn't `flux_rs` the standard way to invoke preconditions/postconditions?

nilehmann: Yes and no. For this particular example, `flux_rs::no_panic` would work, but it won't work for _file module declaration_ (the common version of declaring modules), i.e., the following doesn't work.

```
#[flux_rs::no_panic]
mod my_module;
```

crates/flux-bin/src/lib.rs
ninehusky: you got it o7

nilehmann:
> @enjhnsn2 @nilehmann are we good to punt the conversation re: `cfg_attr(flux, no_panic)` to a different PR? if so, this one is ready for review.

That part is already merged so I don't know why we were discussing it at all here 😅

ninehusky:
OK, marking this ready for review then. I think the `gh-pages` step in CI is failing on this branch, but the last PR may have fixed that?

PR #1384, opened 2025-11-20
Propagate sort arguments from function applications to lean

petros-marko:
`rty::ExprKind::App` knows the sort arguments of the function that's being applied, but we used to drop them in the fixpoint encoding. In this pr
- we keep the sort arguments in the fixpoint encoding
- we drop them when producing the string that is actually sent to fixpoint
- we use them in function applications in lean

There are several places in `fixpoint_encoding.rs` where we create a `fixpoint::Expr::App` where the sort arguments are not obviously available, and for now we just give no arguments in those cases. This is no worse than before, but propagating the sort arguments when they are available is enough to fix the type class inference issue I was running into before.

This PR also 
- renames the type arguments of the lean representations of smt arrays to match the scheme of other generated code (this is necessary because applying sort arguments in lean requires knowing their corresponding names)
- changes representation of fixpoint bools to `Prop` in lean. This is basically what we were doing before, but making it explicit is what allows us to declare an opaque `is_sorted` predicate and use it in lean.

### DIFF
--- a/lib/liquid-fixpoint/src/constraint.rs
+++ b/lib/liquid-fixpoint/src/constraint.rs
@@ -212,7 +212,7 @@ pub enum Expr<T: Types> {
     // having bound variables for [`Expr::Exists`]. We reuse these as well
     // for kvar solutions, which is a bit of a hack.
     BoundVar(BoundVar),
-    App(Box<Self>, Vec<Sort<T>>, Vec<Self>),
+    App(Box<Self>, Option<Vec<Sort<T>>>, Vec<Self>),
     Neg(Box<Self>),
     BinaryOp(BinOp, Box<[Self; 2]>),
     IfThenElse(Box<[Self; 3]>),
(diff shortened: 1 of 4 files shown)


nilehmann:
re ci: I've heard that GitHub limits non-authenticated requests. We should add a token to the request. This is not the first time we have had a transient failure like this. I triggered a re-run and now tests are failing for some other reason.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -212,7 +212,7 @@ pub enum Expr<T: Types> {
     // having bound variables for [`Expr::Exists`]. We reuse these as well
     // for kvar solutions, which is a bit of a hack.
     BoundVar(BoundVar),
-    App(Box<Self>, Vec<Self>),
+    App(Box<Self>, Vec<Sort<T>>, Vec<Self>),
```
nilehmann: Can you make this an `Option`? I want to differentiate missing info from an empty list.

```suggestion
    App(Box<Self>, Option<Vec<Sort<T>>>, Vec<Self>),
```

PR #1386, opened 2025-11-21
`invariant!` macro that generates qualifiers + assertions

ranjitjhala:
So we can now write stuff like

```rust
#[spec(fn (n: usize) -> usize[n])]
pub fn test(n: usize) -> usize {
    let mut i = n;
    let mut res = 0;
    while i > 0 {
        invariant!(res:int; res + i == n);
        i -= 1;
        res += 1;
    }
    res
}
```

Things that should be improved:

- Would be nice to default `res`, `i` etc. to `int` without having to spell it out above;
- Unclear how this would work with _other_ "free-looking" symbols e.g. set-operators, as those only get resolved later in the pipeline... 

otoh, those "free names" are not valid rust ...

... but on the _third_ hand it means you can't use this mechanism to get an invariant/qualifier `s1 = s2 | s3` (where `|` is the set-union operation...) or even `n = pow2(i)` where `fib` is a `def` function... hmm.

### DIFF
--- a/crates/flux-syntax/src/parser/mod.rs
+++ b/crates/flux-syntax/src/parser/mod.rs
@@ -23,7 +23,7 @@ use crate::{
         ParamMode, Path, PathSegment, PrimOpProp, Qualifier, QualifierKind, QuantKind, RefineArg,
         RefineParam, RefineParams, Requires, Sort, SortDecl, SortPath, SpecFunc, Spread, StructDef,
         TraitAssocReft, TraitRef, Trusted, Ty, TyAlias, TyKind, UnOp, VariantDef, VariantRet,
-        WhereBoundPredicate, free_vars,
+        WhereBoundPredicate,
     },
     symbols::{kw, sym},
     token::{self, Comma, Delimiter::*, IdentIsRaw, Or, Token, TokenKind},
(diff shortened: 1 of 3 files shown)


ranjitjhala:
@nilehmann I guess I wanted a first basic version; need to think more about how to get this working without the `res: int` but it wasn't terrible imo.

ranjitjhala:
hi @nilehmann pinging here re this PR

nilehmann:
@ranjitjhala I'll take a look at this today.

crates/flux-syntax/src/surface.rs
```diff
@@ -872,3 +880,28 @@ impl<T, P> Punctuated<T, P> {
         v
     }
 }
+
+/// Collects all free variables in an expression.
+/// A free variable is an `ExprKind::Path` with a single identifier segment.
+pub fn free_vars(expr: &Expr) -> FxHashSet<Ident> {
```
nilehmann: Maybe we can make this a method of `Expr`

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -445,19 +448,63 @@ fn parse_reft_func(cx: &mut ParseCtxt) -> ParseResult<SpecFunc> {
...
-    let name = parse_ident(cx)?;
-    let params = parens(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?;
+    let mut name = parse_ident(cx)?;
+    let mut params = parens(cx, Comma, |cx| parse_refine_param(cx, RequireSort::Yes))?;
     let expr = parse_block(cx)?;
     let hi = cx.hi();
-    Ok(Qualifier { global: !local, name, params, expr, span: cx.mk_span(lo, hi) })
+
+    if let QualifierKind::Hint = kind {
+        for ident in free_vars(&expr) {
```
nilehmann: Not that it matters, but we can make this linear by iterating over `params`, removing elements from `free_vars`, and then extending `params` with all remaining elements.

```rust
let mut fvars = free_vars(&expr);
for param in params {
    fvars.remove(param.name);
}
params.extend(fvars.into_iter().map(...))
```

crates/flux-syntax/src/surface.rs
ranjitjhala: good idea, let me do that.

ranjitjhala: done.

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: indeed -- much nicer, thank you!!!

PR #1387, opened 2025-12-02
`#[no_panic]` --> `#[no_panic_if(cond)]`

ninehusky:
Implements conditional panics, which is the starter step for us to target things like `no_panic` on traits/closures/etc.

The rough implementation roadmap in my head:
- [ ] Change the `no_panic` attribute so that it's accompanied with an `Expr` describing some condition. `no_panic` without an expr desugars to `no_panic_if(true)`.
- [ ] In the same place where we check preconditions/refinements, if a function is annotated with `no_panic_if(cond)`, then append that as a precondition. e.g., if I have some function which requires `len > 0`, and a `no_panic` spec saying `idx < len`, then the precondition to prove is `len > 0 && idx < len`.

EDIT: Freezing this for now so that we can work on the more specific solution of marking closures as `no_panic`. We can come back to this later.

### DIFF
--- a/crates/flux-driver/src/collector/mod.rs
+++ b/crates/flux-driver/src/collector/mod.rs
@@ -17,7 +17,6 @@ use flux_middle::Specs;
 use flux_syntax::{
     ParseResult, ParseSess,
     surface::{self, NodeId, Trusted},
-    symbols::sym::{self, no_panic},
 };
 use rustc_ast::{MetaItemInner, MetaItemKind, tokenstream::TokenStream};
 use rustc_data_structures::fx::FxIndexMap;
@@ -28,8 +27,8 @@ use rustc_hir::{
     def::DefKind,
     def_id::{CRATE_DEF_ID, DefId, LocalDefId},
 };
-use rustc_middle::{query::queries::has_alloc_error_handler, ty::TyCtxt};
-use rustc_span::{DUMMY_SP, Ident, Span, Symbol, SyntaxContext};
+use rustc_middle::ty::TyCtxt;
+use rustc_span::{Ident, Span, Symbol, SyntaxContext};
 
 use crate::collector::detached_specs::DetachedSpecsCollector;
 type Result<T = ()> = std::result::Result<T, ErrorGuaranteed>;


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -671,7 +671,13 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
...
 
         let body_id = self.tcx().hir_node_by_def_id(fn_id.local_id()).body_id();
-        let no_panic = self.genv().no_panic(fn_id);
+
+        let no_panic = if let Some(e) = fn_sig.no_panic_if {
+            let expr = self.conv_expr(&mut env, &e)?;
+            expr
+        } else {
+            if self.genv().no_panic(fn_id) { Expr::tt().clone() } else { Expr::ff().clone() }
+        };
```
nilehmann, line 679:
```suggestion
        let no_panic = if let Some(e) = fn_sig.no_panic_if {
            self.conv_expr(&mut env, &e)?
        } else {
            if self.genv().no_panic(fn_id) { Expr::tt() } else { Expr::ff() }
        };
```

crates/flux-syntax/src/symbols.rs
```diff
@@ -25,6 +25,7 @@ symbols! {
         Set,
         int,
         no_panic,
+        no_panic_if,
```
nilehmann: I believe we are not using this symbol

crates/flux-driver/src/collector/mod.rs
```diff
@@ -756,6 +789,23 @@ impl FluxAttrs {
...
+        let span = if has {
+            self.map
+                .get(attr_name!(NoPanic))
+                .and_then(|attrs| attrs.first())
+                .map(|attr| attr.span)
+        } else {
+            None
+        };
+        (has, span)
+    }
```
nilehmann: Maybe we can just return `Option<Span>` with `Some` representing when there's a `no_panic` attr (plus the `Span`).

```suggestion
    fn has_no_panic(&self) -> Option<Span> {
        self.map
            .get(attr_name!(NoPanic))
            .and_then(|attrs| attrs.first())
            .map(|attr| attr.span)
    }
```

PR #1388, opened 2025-12-03
Render kvar-solutions in vscode 

ranjitjhala:
e.g. we now get something like; also split the extension into a various smaller files with claude; (perhaps a bit of "regrouping" is in order...)

Screenshot 2025-12-02 at 4 21 36 PM

### DIFF
--- a/crates/flux-middle/src/rty/mod.rs
+++ b/crates/flux-middle/src/rty/mod.rs
@@ -9,7 +9,7 @@ pub mod canonicalize;
 mod expr;
 pub mod fold;
 pub mod normalize;
-pub mod pretty;
+mod pretty;
 pub mod refining;
 pub mod region_matching;
 pub mod subst;
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -214,19 +214,14 @@ pub struct KvarSolutionTrace {
 impl KvarSolutionTrace {
     pub fn new(cx: &PrettyCx, kvid: rty::KVid, bind_expr: &rty::Binder<rty::Expr>) -> Self {
         let mut args = Vec::new();
-        let body = rty::pretty::nested_with_bound_vars(
-            cx,
-            "",
-            bind_expr.vars(),
-            Some("".to_string()),
-            |prefix| {
+        let body = cx
+            .nested_with_bound_vars("", bind_expr.vars(), Some("".to_string()), |prefix| {
                 for arg in prefix.split(',').map(|s| s.trim().to_string()) {
                     args.push(arg);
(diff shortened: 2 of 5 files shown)


crates/flux-middle/src/rty/mod.rs
```diff
@@ -9,7 +9,7 @@ pub mod canonicalize;
 mod expr;
 pub mod fold;
 pub mod normalize;
-mod pretty;
+pub mod pretty;
```
ranjitjhala: is this ok @nilehmann ? needed to use `nested_with_bound_vars`...

crates/flux-refineck/src/lib.rs
```diff
@@ -142,6 +143,13 @@ pub fn check_fn(
                     FixpointQueryKind::Body,
                 )
                 .emit(&genv)?;
+
+            // DUMP SOLUTION
+            let tcx = genv.tcx();
+            let hir_id = tcx.local_def_id_to_hir_id(def_id);
+            let body_span = tcx.hir_span_with_body(hir_id);
```
nilehmann, line 150: Is the span really relevant here? I haven't checked the typescript, but I presume you are filtering out the solution based on the span of the cursor, so you need the solution to cover the entire function span? Sounds a bit brittle

crates/flux-middle/src/rty/mod.rs
nilehmann: Maybe we can make `nested_with_bound_vars` a method of `PrettyCx`?

crates/flux-refineck/src/lib.rs
ranjitjhala: Yes, that is exactly what I'm doing -- we want the `solution` event to specify which bit (`Span`) of the code it is the solution for, and this is the only way I know. (This is the same code as what is used for the `-Finclude=span:tests/tests/pos/detached/detach00.rs:13:3` btw...

ranjitjhala: The TypeScript uses this `Span` to create a map from `lineNumber -> Solution` to be used to render the environment at that line...

nilehmann: I suppose a more principled way would be to associate with each event not just the span but also the id of the associated item (the function name), then we could attach the solution to the function and not a span. But I guess this is fine...

ranjitjhala: Even so then in the editor (TS) you have to somehow recover which function you're in from the cursor position, which could get hacky...

ranjitjhala: I suppose if we had a _real_ language server then we wouldn't need these shenanigans...

PR #1391, opened 2025-12-03
Change `no_panic` checking mechanism to constraints

ninehusky:
Changes `no_panic` checking so that we emit constraints within `check_call`. A future PR will handle giving `no_panic` functions the relevant associated traits so that we can handle closure calls.

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -950,14 +950,8 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             {
                 at.check_pred(
                     Expr::implies(
-                        match no_panic {
-                            true => Expr::tt(),
-                            false => Expr::ff(),
-                        },
-                        match fn_sig.no_panic() {
-                            true => Expr::tt(),
-                            false => Expr::ff(),
-                        },
+                        if no_panic { Expr::tt() } else { Expr::ff() },
+                        if fn_sig.no_panic() { Expr::tt() } else { Expr::ff() },
                     ),
                     ConstrReason::NoPanic(callee_def_id),
                 );


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -998,7 +998,7 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
 
         let output = self.conv_fn_output(env, &decl.output)?;
 
-        Ok(rty::FnSig::new(safety, abi, requires.into(), inputs.into(), output))
+        Ok(rty::FnSig::new(safety, abi, requires.into(), inputs.into(), output, false))
```
nilehmann: In this one, we should get the attrs for the function and use `true` if it has `no_panic`

crates/flux-middle/src/rty/canonicalize.rs
```diff
@@ -292,6 +292,7 @@ impl PolyFnSig {
                 preds.into(),
                 inputs.into(),
                 fn_sig.output().clone(),
+                false,
```
nilehmann: Here we are transforming an `FnSig` so we should keep the original one, i.e., `self.no_panic`

crates/flux-middle/src/rty/mod.rs
```diff
@@ -2527,7 +2529,7 @@ impl CoroutineObligPredicate {
             Binder::bind_with_vars(FnOutput::new(self.output.clone(), vec![]), List::empty());
 
         PolyFnSig::bind_with_vars(
-            FnSig::new(Safety::Safe, rustc_abi::ExternAbi::RustCall, List::empty(), inputs, output),
+            FnSig::new(Safety::Safe, rustc_abi::ExternAbi::RustCall, List::empty(), inputs, output, false),
```
nilehmann: This is about coroutines (a.k.a, async code), so we can obviate for now and leave it as `false`.

crates/flux-refineck/src/checker.rs
```diff
@@ -466,7 +466,7 @@ fn promoted_fn_sig(ty: &Ty) -> PolyFnSig {
     let inputs = rty::List::empty();
     let output =
         Binder::bind_with_vars(FnOutput::new(ty.clone(), rty::List::empty()), rty::List::empty());
-    let fn_sig = crate::rty::FnSig::new(safety, abi, requires, inputs, output);
+    let fn_sig = crate::rty::FnSig::new(safety, abi, requires, inputs, output, false);
```
nilehmann: Interesting, I suppose promoted constats never panic, so in principle this could always be `true`. We can try that and see whether any test breaks.

crates/flux-middle/src/rty/refining.rs
```diff
@@ -341,7 +341,7 @@ impl Refine for ty::FnSig {
         // single hole for the "requires"; then we "fill" the hole with a KVAR
         // and generate a PolyFnSig with the hoisted variables
         // see `into_bb_env` in `type_env.rs` for an example.
-        Ok(rty::FnSig::new(self.safety, self.abi, List::empty(), inputs, output))
+        Ok(rty::FnSig::new(self.safety, self.abi, List::empty(), inputs, output, false))
```
nilehmann: This one is for creating a default signature for functions without an annotation, so conservatively, we should always pick `false`.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -2771,6 +2774,7 @@ impl EarlyBinder<PolyVariant> {
                     variant.requires.clone(),
                     inputs,
                     output,
+                    false,
```
nilehmann, line 2806: Constructing an enum will never panic, so this can be `true`

crates/flux-middle/src/rty/mod.rs
```diff
@@ -690,7 +690,7 @@ impl FnTraitPredicate {
         let inputs = self.tupled_args.expect_tuple().iter().cloned().collect();
         let ret = self.output.clone().shift_in_escaping(1);
         let output = Binder::bind_with_vars(FnOutput::new(ret, vec![]), List::empty());
-        FnSig::new(Safety::Safe, rustc_abi::ExternAbi::Rust, List::empty(), inputs, output)
+        FnSig::new(Safety::Safe, rustc_abi::ExternAbi::Rust, List::empty(), inputs, output, false)
```
nilehmann: Aha! This is the signature we create when a function takes a closure as argument. For now, we can keep it as `false` to be conservative, but some of the magic will happen here.

crates/flux-middle/src/rty/mod.rs
```diff
@@ -741,6 +741,7 @@ pub fn to_closure_sig(
         fn_sig.requires.clone(), // crate::rty::List::empty(),
         inputs.into(),
         output,
+        false,
```
nilehmann, line 757: This is the other place where closures are relevant. We can keep it as `false` for now to be conservative.

crates/flux-fhir-analysis/src/conv/mod.rs
ninehusky: How would we find this? Would this require adding a flag like this to the `FnDecl`? The only other way to get a `DefId`-like thing from the arguments provided is to do `self.owner().def_id()`, but that doesn't seem right to me.

nilehmann: You can modify the function to take a MaybeExternId. We have one handy when we call it.

ninehusky: Got it. We also call it in `conv_requires`, but I assume we can pass in `None` in that situation.

crates/flux-refineck/src/checker.rs
ninehusky: Nothing fails!

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: You mean in the `BareFn` case in `conv_ty`?

Then I'll change my answer. We can add what the value of `no_panic` should be as an argument to `conv_fn_decl`, pass `false` for the `BareFn` case and read the attrs in `conv_fn_sig`

ninehusky: Done!

crates/flux-refineck/src/checker.rs
```diff
@@ -883,7 +883,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 && let Some(callee_def_id) = callee_def_id
                 && genv.def_kind(callee_def_id).is_fn_like()
             {
-                let callee_no_panic = genv.no_panic(callee_def_id);
+                let callee_no_panic = fn_sig.skip_binder_ref().skip_binder_ref().no_panic()
+                    || genv.no_panic(callee_def_id);
+
```
ninehusky: @nilehmann here's my best guess for how this should get checked. the `no_panic`ness of a function call can hopefully go through the infrastructure that we added in `FnSig`.

crates/flux-refineck/src/checker.rs
```diff
@@ -883,7 +883,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 && let Some(callee_def_id) = callee_def_id
                 && genv.def_kind(callee_def_id).is_fn_like()
             {
-                let callee_no_panic = genv.no_panic(callee_def_id);
+                let callee_no_panic = fn_sig.skip_binder_ref().skip_binder_ref().no_panic()
+                    || genv.no_panic(callee_def_id);
```
nilehmann: Why did you add the `|| genv.no_panic(callee_def_id)`? The idea is to stop relying on `genv.no_panic` so we can eventually remove it. Also, we should read the "no panicness" of the current function from the signature instead of `genv.no_panic`. I think the best way to do that is to store the `FnSig` in `Checker`. If you look a `Checker::new`, you'll notice we pass the signature when we construct it, but we only store the output. I'd store the entire signature instead of just the output.

crates/flux-refineck/src/checker.rs
```diff
@@ -883,7 +883,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 && let Some(callee_def_id) = callee_def_id
                 && genv.def_kind(callee_def_id).is_fn_like()
             {
-                let callee_no_panic = genv.no_panic(callee_def_id);
+                let callee_no_panic = fn_sig.skip_binder_ref().skip_binder_ref().no_panic()
```
nilehmann: To avoid skipping binders, we should move all this logic (everything inside `M::NAME == "refine"`) to after we instantiate the signature. I'd put it after `let mut at = infcx.at(span)`.

crates/flux-refineck/src/checker.rs
ninehusky: > The idea is to stop relying on `genv.no_panic` so we can eventually remove it.
Ah, right. So the idea here is that any time we construct a `FnSig` we add the no-panicness to it. Will hack on this!

crates/flux-refineck/src/checker.rs
```diff
@@ -954,6 +939,47 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             .deeply_normalize(&mut infcx.at(span))
             .with_span(span)?;
 
+        if M::NAME == "refine" {
```
nilehmann, line 944: If you move everything after `let mut at = infcx.at(span);` you can use `at` directly

crates/flux-refineck/src/checker.rs
```diff
@@ -954,6 +939,47 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                                false => Expr::ff(),
+                            },
+                            match sig.no_panic() {
+                                true => Expr::tt(),
+                                false => Expr::ff(),
+                            },
+                        ),
+                        ConstrReason::NoPanic(callee_def_id),
+                    );
+                }
```
nilehmann: You don't have to do anything with predicates yet; we will take care of that when we implement the associated refinement part of this

```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -954,6 +939,47 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                    Expr::implies(
+                        match no_panic {
+                            true => Expr::tt(),
+                            false => Expr::ff(),
+                        },
+                        match fn_sig.no_panic() {
+                            true => Expr::tt(),
+                            false => Expr::ff(),
+                        },
+                    ),
```
nilehmann, line 955: Maybe

```suggestion
                    Expr::implies(
                        if no_panic { Expr::tt() } else { Expr::ff() },
                        if fn_sig.no_panic() { Expr::tt() } else { Expr::ff() },
                    ),
```

tests/tests/pos/surface/no_panic05.rs
```diff
@@ -0,0 +1,24 @@
+#![feature(stmt_expr_attributes)]
```
nilehmann: This test doesn't make too much sense with what's implemented in the PR

PR #1392, opened 2025-12-04
Automatically convert `&mut` to `&strg` if possible

### LINKED ISSUE #671: Unfolding of mutable references
The following code is safe, but flux flags that it may integer underflow:
```Rust
pub(crate) struct DummyStruct {
    offset: usize,
    index: usize,
}

impl DummyStruct {
    fn set_safe_offset(&mut self) {
        let new_offset = if self.offset <= self.index {
            0
        } else {
            self.offset - self.index
        };
        self.offset = new_offset;
    }
}

```
fails with:
```
error[E0999]: arithmetic operation may overflow
   --> src/main.rs:157:13
    |
    |             self.offset - self.index
    |             ^^^^^^^^^^^^^^^^^^^^^^^^

```
This does not happen if the function takes a immutable parameter. The following code passes:
```Rust
impl DummyStruct {
    fn get_safe_offset(&self) -> usize {
        if self.offset <= self.index {
            0
        } else {
            self.offset - self.index
        }
    }
}
```

ranjitjhala:
Fixes #671 

[Zulip](https://flux-rs.zulipchat.com/#narrow/channel/486098-general/topic/fn.20dostuff.28x.3A.20RingBuffer.29.20ensures.20x.3A.20RingBuffer/with/561754889)

But currently chokes on 

1. closures
2. slices(?)
3. the following test `tests/tests/pos/surface/join01.rs` which ALSO fails if we "manually" do the equivalent of auto-strong...

```
// #[flux::sig(fn(&mut i32[0], bool) -> i32{v : v >= 0})]

#[flux::sig(fn(r1: &mut i32[0], bool) -> i32{v : v >= 0} ensures r1: i32[0])]
pub fn foo(r1: &mut i32, b: bool) -> i32 {
    let mut x = 1;
    let r;
    if b {
        r = &mut *r1;
    } else {
        r = &mut x;
    }
    *r
}
```

@nilehmann does this ring any immediate bells?

```
rjhala@pozole ~/r/flux (auto-strong)> cargo x run tests/tests/pos/surface/join01.rs
...
error: internal compiler error: crates/flux-refineck/src/type_env.rs:620:18: unexpected types: `&'?7 mut i32[] | *` - `ptr(mut['?7], r1)`
  --> tests/tests/pos/surface/join01.rs:6:5
   |
 6 | /     if b {
 7 | |         r = &mut *r1;
 8 | |     } else {
 9 | |         r = &mut x;
10 | |     }
   | |_____^
   |
   = note: -Ztrack-diagnostics: created at crates/flux-refineck/src/type_env.rs:620:18


thread 'rustc' (24922391) panicked at crates/flux-refineck/src/type_env.rs:620:18:
Box<dyn Any>
```

### DIFF
--- a/crates/flux-middle/src/rty/mod.rs
+++ b/crates/flux-middle/src/rty/mod.rs
@@ -3135,7 +3135,6 @@ pub fn auto_strong(
     {
         return fn_sig;
     }
-    println!("TRACE: auto_strong (0) {fn_sig:?}");
     let kind = BoundReftKind::Anon;
     let mut vars = fn_sig.vars().to_vec();
     let fn_sig = fn_sig.skip_binder();
@@ -3182,7 +3181,5 @@ pub fn auto_strong(
 
     // 4. Reconstruct fn sig with new inputs and output and vars
     let fn_sig = FnSig { inputs: List::from_vec(strg_inputs), output, ..fn_sig };
-    let fn_sig = Binder::bind_with_vars(fn_sig, vars.into());
-    println!("TRACE: auto_strong (1) {fn_sig:?}");
-    fn_sig
+    Binder::bind_with_vars(fn_sig, vars.into())
 }


crates/flux-infer/src/infer.rs
```diff
@@ -366,15 +366,14 @@ impl<'infcx, 'genv, 'tcx> InferCtxt<'infcx, 'genv, 'tcx> {
...
-    ) -> U
+    ) -> InferResult<U>
     where
         T: TypeFoldable,
     {
         self.ensure_resolved_evars(|infcx| {
             let t = t.replace_bound_refts_with(|sort, mode, _| infcx.fresh_infer_var(sort, mode));
             Ok(f(infcx, t))
         })
-        .unwrap()
```
nilehmann, line 377: I prefer to keep the `unwrap` here or turning this into a `query_bug!` (i.e., `.map_err(|e| query_bug!(...))`. Otherwise, I feel the error message you get is confusing.

crates/flux-refineck/src/type_env.rs
```diff
@@ -404,7 +404,7 @@ impl<'a> TypeEnv<'a> {
         for constraint in ensures {
             match constraint {
                 Ensures::Type(path, ty) => {
-                    let actual_ty = self.get(path);
+                    let actual_ty = self.get(path).unblocked();
```
nilehmann: Please leave a `// HACK` comment here.

crates/flux-rustc-bridge/src/ty/mod.rs
```diff
@@ -933,7 +933,8 @@ impl<'tcx> ToRustc<'tcx> for Ty {
             TyKind::Uint(uint_ty) => rustc_ty::Ty::new_uint(tcx, *uint_ty),
             TyKind::Adt(adt_def, args) => {
                 let adt_def = adt_def.to_rustc(tcx);
-                let args = tcx.mk_args_from_iter(args.iter().map(|arg| arg.to_rustc(tcx)));
+                let args =
+                    tcx.mk_args_from_iter(args.iter().map(|arg: &GenericArg| arg.to_rustc(tcx)));
```
nilehmann:
```suggestion
                    tcx.mk_args_from_iter(args.iter().map(|arg| arg.to_rustc(tcx)));
```

PR #1393, opened 2025-12-05
Instead of running lake with --dir change directories for executing lean

petros-marko:
This resolves the versioning issue that was sometimes causing grind to appear to succeed when editing the proof in VSCode but fail when lean was run by flux to check the proofs.


petros-marko:
> Is this because some cache or lock file wasn't being saved in the right place or something?

Maybe? Or maybe when you create a project with `lake` it defaults to the latest version of Lean and so when running VSCode in the project the language server picks up that version, but when flux invoked `lean` from outside the project it defaulted to some other version? Are these the same thing 🤷🏽?

PR #1394, opened 2025-12-05
Make kvar gen single before encoding constraint for lean

petros-marko:
Maybe this isn't the most elegant way to go about this, but it work. Before creating the fixpoint constraint that gets sent to lean, we can set the KVarEncoding for all KVars to be `Single` and then only one fixpoint kvar will be created for each rty kvar.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -1168,6 +1168,9 @@ impl KVarGen {
         rty::Expr::kvar(kvar)
     }
 
+    // This function is called before we encode a constraint that will be translated to Lean.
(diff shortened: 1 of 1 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1167,6 +1167,12 @@ impl KVarGen {
         let kvar = rty::KVar::new(kvid, flattened_self_args, exprs);
         rty::Expr::kvar(kvar)
     }
+
+    pub(crate) fn make_all_single(&mut self) {
```
nilehmann, line 1174: Please add a comment to this function explaining why we need it and that it's kind of a hack.

PR #1397, opened 2025-12-10
Source level binders

ranjitjhala:
Various changes to render vscode "traces" nicely, using names from program source.

**Return Binder**
Screenshot 2025-12-11 at 8 16 21 AM

**Loops `usize`**
Screenshot 2025-12-11 at 8 16 37 AM

**Loops `Vec<..>`**
Screenshot 2025-12-11 at 8 16 50 AM

**Struct `strg`**
Screenshot 2025-12-11 at 8 17 02 AM

**Struct immutable**
Screenshot 2025-12-11 at 8 17 32 AM

**Tuple**
Screenshot 2025-12-11 at 8 18 15 AM

**Slice**
Screenshot 2025-12-11 at 8 19 11 AM

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -1005,8 +1005,6 @@ impl KVarEncodingCtxt {
     }
 }
 
-/// TODO(source-level-binders): remove local_var_gen,
-/// make reverse_map an `IndexVec`, make `LocalVarInfo` type `Option<rty::Expr>`
 /// Environment used to map from [`rty::Var`] to a [`fixpoint::LocalVar`].
 struct LocalVarEnv {
     local_var_gen: IndexGen<fixpoint::LocalVar>,
--- a/crates/flux-middle/src/rty/expr.rs
+++ b/crates/flux-middle/src/rty/expr.rs
@@ -980,6 +980,7 @@ impl NameProvenance {
         }
     }
 }
+
 impl KVar {
     pub fn new(kvid: KVid, self_args: usize, args: Vec<Expr>) -> Self {
         KVar { kvid, self_args, args: List::from_vec(args) }
--- a/crates/flux-refineck/src/type_env.rs
+++ b/crates/flux-refineck/src/type_env.rs
@@ -945,7 +945,6 @@ impl TypeEnvTrace {
         env: &TypeEnv,
     ) -> Self {
         let mut bindings = vec![];
-        // let cx = PrettyCx::default(genv).hide_regions(true);
         env.bindings
             .iter()
             .filter(|(_, binding)| !binding.ty.is_uninit())


ranjitjhala:
@nilehmann I think you should probably review this before we proceed to modifying the lean output...?

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1433,7 +1441,8 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                 Ok(rty::TyOrCtor::Ctor(rty::Binder::bind_with_sort(ty, sort)))
             }
             fhir::BaseTyKind::Slice(ty) => {
-                let bty = rty::BaseTy::Slice(self.conv_ty(env, ty)?).shift_in_escaping(1);
+                let name = name.map(|sym| Self::suffix_symbol(sym, "elem"));
```
nilehmann, line 1444: Do you have examples of this showing up in the extension? I'd like to see how it looks. I'm a bit wary of overly long names. Maybe we should cut the recursion at some point. Although you may not have deeply nested stuff in practice.

crates/flux-middle/src/rty/binder.rs
```diff
@@ -283,10 +287,79 @@ impl From<Sort> for BoundVariableKind {
...
 }
 
 impl_slice_internable!(BoundVariableKind);
+
+// --------------------------------------------------------------------------------------------------
+
+#[derive(Clone, Debug)]
+pub struct BinderProvenance {
+    /// Whence?
+    pub span: Option<Span>,
```
nilehmann: This span is never used. We could make `BinderProvenance` the enum

```rust
enum BinderProvenance {
    Unknown,
    UnfoldBoundReft(BoundReftKind),
}
```

Also, more than `BinderProvenance`, this is a `NameProvenance` and I'd define it in `refine_tree.rs`.

crates/flux-middle/src/rty/canonicalize.rs
```diff
@@ -66,6 +67,7 @@ pub struct Hoister<D> {
 pub trait HoisterDelegate {
     fn hoist_exists(&mut self, ty_ctor: &TyCtor) -> Ty;
     fn hoist_constr(&mut self, pred: Expr);
+    fn set_name(&mut self, name: Symbol);
```
nilehmann: I'd not expose this as part of `HositerDelegate`, but rather something internal of the delegate. More concretely:

1. Make the `delegate` field in `Hoister` public
2. When you have a concrete delegate, you can set the name directly by accessing the delegate, e.g., in `into_bb_env` instead of
   ```rust
   hoister.hoist_at_name(name, ty)
   ```
   we do
   ```rust
   hoister.delegate.name = Some(name);
   hoister.hoist(ty)
   ```

crates/flux-refineck/src/type_env/place_ty.rs
```diff
@@ -267,15 +267,19 @@ impl PlacesTree {
...
-            binding.ty = f(&binding.ty)?;
+    fn try_fmap_mut<E>(&mut self, mut f: impl FnMut(&Loc, &Ty) -> Result<Ty, E>) -> Result<(), E> {
+        self.map.iter_mut().try_for_each(|(loc, binding)| {
+            binding.ty = f(loc, &binding.ty)?;
             Ok(())
         })
+        // self.map.values_mut().try_for_each(|binding| {
+        //     binding.ty = f(&binding.ty)?;
+        //     Ok(())
+        // })
```
nilehmann:
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -596,7 +609,7 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             &mut infcx,
             CheckerId::DefId(def_id),
             inherited,
-            &body_root.body,
+            &body_root.body, // TODO:source-level-binder
```
nilehmann: ?

crates/flux-infer/src/infer.rs
```diff
@@ -1078,7 +1099,11 @@ impl<'a, E: LocEnv> Sub<'a, E> {
         let vars = a
             .vars()
             .iter()
-            .map(|kind| Expr::fvar(infcx.define_var(kind.expect_sort())))
+            .map(|kind| {
+                let sort = kind.expect_sort();
+                let kind = kind.expect_bound_reft_kind();
```
nilehmann:
```suggestion
                let (sort, _, kind) = kind.expect_refine();
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -132,11 +132,17 @@ impl Cursor<'_> {
...
-        self.ptr = self.ptr.push_node(NodeKind::ForAll(fresh, sort.clone()));
+        self.ptr = self
+            .ptr
+            .push_node(NodeKind::ForAll(fresh, sort.clone(), provenance));
         fresh
     }
+    /// Temporary "backwards compatible" wrapper, should only use `define_var`...
+    pub(crate) fn define_unknown_var(&mut self, sort: &Sort) -> Name {
+        self.define_var(sort, BinderProvenance::unknown())
+    }
```
nilehmann:
```suggestion
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -475,8 +481,8 @@ impl Node {
                 }
                 Some(constr)
             }
-            NodeKind::ForAll(name, sort) => {
-                cx.with_name_map(*name, |cx, fresh| -> QueryResult<_> {
+            NodeKind::ForAll(name, sort, provenance) => {
+                cx.with_name_map(*name, provenance.clone(), |cx, fresh| -> QueryResult<_> {
```
nilehmann: I'd make it `Copy`

```suggestion
                cx.with_name_map(*name, *provenance, |cx, fresh| -> QueryResult<_> {
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -764,20 +770,24 @@ struct RcxBind {
...
-    pub fn new(genv: GlobalEnv, cursor: &Cursor) -> Self {
+    pub fn new<'genv, 'tcx>(
+        genv: GlobalEnv<'genv, 'tcx>,
+        cursor: &Cursor,
+    ) -> (Self, PrettyCx<'genv, 'tcx>) {
         let parents = ParentsIter::new(NodePtr::clone(&cursor.ptr)).collect_vec();
         let mut bindings = vec![];
         let mut exprs = vec![];
-        let cx = &PrettyCx::default(genv).show_kvar_args();
+        let mut cx0 = PrettyCx::default(genv).show_kvar_args();
```
nilehmann: I'd create the context outside and pass `&mut PrettyCx` as an argument to this function.

crates/flux-infer/src/fixpoint_encoding.rs
nilehmann, line 1: Can we exclude the changes to this file from this PR and add them later with the changes required for Lean?

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: Also, we don't need to do this now, but perhaps a cleaner way to implement this is to have a representation of the "list of projections into the type". A bit like [PlaceElem](https://flux-rs.github.io/flux/doc/flux_rustc_bridge/mir/enum.PlaceElem.html) but tailored for this use case. That way, we can be more flexible when generating the name if we have all the context. We could implement the same idea in the hoister.

ranjitjhala: Yes -- adding the `elem` stuff is debatable. I have some screenshots let me post them above.

crates/flux-middle/src/rty/binder.rs
ranjitjhala: Hmm `refine_tree` is in `flux-infer` but I need it in `flux-middle` is that possible?

crates/flux-refineck/src/checker.rs
ranjitjhala: some junk note; cleaned up!

crates/flux-infer/src/infer.rs
ranjitjhala: had to make `expect_refine` public; assume that's ok!

crates/flux-middle/src/rty/binder.rs
nilehmann: Ah, you need it for the pprint stuff. Then it's ok to leave it in this crate. But still, this is not about binders (in the sense of an `rty::Binder`) but about names.

ranjitjhala: I can put in a separate file in this crate?

nilehmann: I suppose we can put it next to the definition of `Name`

ranjitjhala: Ah, of course!

ranjitjhala:
Ok @nilehmann -- I think I applied all your suggestions -- except for the `elem` thing, you can see what you think from the screenshots at the top...

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1005,6 +1005,8 @@ impl KVarEncodingCtxt {
     }
 }
 
+/// TODO(source-level-binders): remove local_var_gen,
+/// make reverse_map an `IndexVec`, make `LocalVarInfo` type `Option<rty::Expr>`
```
```suggestion
```

crates/flux-infer/src/refine_tree.rs
```diff
@@ -764,20 +766,19 @@ struct RcxBind {
 }
 
 impl RefineCtxtTrace {
-    pub fn new(genv: GlobalEnv, cursor: &Cursor) -> Self {
+    pub fn new<'genv, 'tcx>(cx: &mut PrettyCx<'genv, 'tcx>, cursor: &Cursor) -> Self {
```
nilehmann: I believe these lifetimes are unnecessary 

```suggestion
    pub fn new(cx: &mut PrettyCx, cursor: &Cursor) -> Self {
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -942,6 +942,52 @@ newtype_index! {
...
+        NameProvenance::UnfoldBoundReft(kind)
+    }
+
+    pub fn opt_symbol(&self) -> Option<Symbol> {
+        match &self {
+            NameProvenance::UnfoldBoundReft(BoundReftKind::Named(name)) => Some(*name),
+            _ => None,
+        }
+    }
+}
```
nilehmann, line 982:
```suggestion
}

```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -942,6 +942,52 @@ newtype_index! {
...
+
+impl NameProvenance {
+    pub fn unknown() -> Self {
+        NameProvenance::Unknown
+    }
+
+    pub fn bound_reft_kind(kind: BoundReftKind) -> Self {
+        NameProvenance::UnfoldBoundReft(kind)
+    }
```
nilehmann: These methods feel unnecessary now; we can use the enum variants directly

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1728,9 +1783,9 @@ pub(crate) mod pretty {
                 | ExprKind::InternalFunc(..) => debug_nested(cx, &e),
                 ExprKind::KVar(kvar) => {
                     let kv = format!("{:?}", kvar.kvid);
-                    let mut strs = vec![kv];
+                    let mut strs = vec![kv.clone()];
```
nilehmann: I believe this clone is unnecessary

```suggestion
                    let mut strs = vec![kv];
```

crates/flux-refineck/src/type_env.rs
```diff
@@ -930,10 +941,11 @@ impl TypeEnvTrace {
         genv: GlobalEnv,
         local_names: &UnordMap<Local, Symbol>,
         local_decls: &IndexVec<Local, LocalDecl>,
+        cx: PrettyCx,
         env: &TypeEnv,
     ) -> Self {
         let mut bindings = vec![];
-        let cx = PrettyCx::default(genv).hide_regions(true);
+        // let cx = PrettyCx::default(genv).hide_regions(true);
```
nilehmann:
```suggestion
```

crates/flux-rustc-bridge/src/mir.rs
```diff
@@ -346,6 +346,11 @@ impl Place {
         projection.push(PlaceElem::Deref);
         Place { local: self.local, projection }
     }
+
+    // TODO(source-level-binders): use the bits from the `projection` too?
+    pub fn name(&self, local_names: &UnordMap<Local, Symbol>) -> Option<Symbol> {
+        local_names.get(&self.local).copied()
```
nilehmann: hmm, maybe we should only return `Some` if `self.projection.is_empty()`.

crates/flux-infer/src/refine_tree.rs
```diff
@@ -764,20 +766,19 @@ struct RcxBind {
...
         parents.into_iter().rev().for_each(|ptr| {
             let node = ptr.borrow();
             match &node.kind {
-                NodeKind::ForAll(name, sort) => {
-                    let bind = RcxBind {
-                        name: format_cx!(cx, "{:?}", ^name),
-                        sort: format_cx!(cx, "{:?}", sort),
-                    };
+                NodeKind::ForAll(name, sort, provenance) => {
+                    cx.with_name_provenance(*name, provenance.clone());
```
nilehmann:
```suggestion
                    cx.with_name_provenance(*name, *provenance));
```

nilehmann (approved):

crates/flux-rustc-bridge/src/mir.rs
ranjitjhala: That's reasonable!

PR #1400, opened 2025-12-12
Misc updates to Lean integration

ranjitjhala:
1. Allow using the cargo.toml, instead of fiddling with flags...

```
[package.metadata.flux]
enabled = true
cache = true
lean = "check"
```

2. Generate hyperlinks from `#[proven_externally(proof)]` to the actual proof-definition in `.lean`

3. Remove unnecessary `.to_str()` inside the `lean_encoding.rs`

4. Better error message instead of a `panic!` when missing `-Flean=check` flag.

### DIFF
--- a/crates/flux-driver/src/collector/mod.rs
+++ b/crates/flux-driver/src/collector/mod.rs
@@ -549,14 +549,6 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
                     FluxAttrKind::TrustedImpl(b.into())
                 })?
             }
-            //  ("proven_externally", hir::AttrArgs::Delimited(_)) => {
-            //      let span = dargs.tokens.get(0).unwrap().span();
-            //      FluxAttrKind::ProvenExternally(span)
-            // }
-            // ("proven_externally", hir::AttrArgs::Delimited(dargs)) if dargs.tokens.len() == 1 => {
-            //      let span = dargs.tokens.get(0).unwrap().span();
-            //      FluxAttrKind::ProvenExternally(span)
-            // }
             ("proven_externally", _) => {
                 let span = attr_item_inner_span(attr_item);
                 FluxAttrKind::ProvenExternally(span)


crates/flux-driver/src/collector/mod.rs
```diff
@@ -549,7 +550,14 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
                     FluxAttrKind::TrustedImpl(b.into())
                 })?
             }
-            ("proven_externally", hir::AttrArgs::Empty) => FluxAttrKind::ProvenExternally,
+            ("proven_externally", hir::AttrArgs::Delimited(dargs)) if dargs.tokens.len() == 1 => {
+                let span = dargs.tokens.get(0).unwrap().span();
```
ranjitjhala: This seems *way* more complicated than you'd think because all the *other* span values floating around are somehow for the associated function object, not just the token for `proven_externally` or `proof` ... puzzling.

crates/flux-refineck/src/lib.rs
```diff
@@ -135,13 +136,13 @@ pub fn check_fn(
         )
         .map_err(|err| err.emit(genv, def_id))?;
 
-        if genv.proven_externally(def_id) {
-            if flux_config::emit_lean_defs() {
+        if genv.proven_externally(def_id).is_some() {
+            if flux_config::lean().is_emit() {
```
petros-marko, line 139: I think it'd make sense to also check this flag in `fixpoint_encoding.rs` (in `generate_and_check_lean_lemmas`) where we actually do the checking. If `-Flean="emit"` then we can just emit the theorems and not do any checking.

Maybe we can split the generation and checking at this level and avoid checking the flag multiple times if you prefer that.

petros-marko (approved):
Mostly LGTM. Left a small comment.

crates/flux-driver/src/collector/mod.rs
nilehmann: Let me know if you get the correct spans with the changes in https://github.com/flux-rs/flux/pull/1402. You should be able to use `attr_item.span` to get the entire span or call `attr_item_inner_span` to get the span without the `#[` and `]`.

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -16,28 +17,30 @@ use crate::{
...
-impl<'genv, 'tcx, 'a> LeanEncoder<'genv, 'tcx, 'a> {
-    pub fn new(
-        genv: GlobalEnv<'genv, 'tcx>,
-        lean_path: &'a Path,
-        project_name: String,
-        defs_file_name: String,
-    ) -> Self {
+impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
+    pub fn new(genv: GlobalEnv<'genv, 'tcx>, project_name: String, defs_file_name: String) -> Self {
+        let lean_path = genv
```
nilehmann: I believe `cargo-flux` is in a better position to decide where to put lean files, e.g., in a multi-crate workspace, `cargo-flux` could decide to put lean files inside each crate. So, I'd add a flag to Flux to set the path and let `cargo-flux` choose what the path should be.

crates/flux-middle/src/queries.rs
```diff
@@ -79,6 +79,10 @@ pub enum QueryErr {
         trait_id: DefId,
         name: Symbol,
     },
+    /// Need to `-Flean=check` for externally proven items
+    MissingLeanCheck {
+        def_id: DefId,
+    },
```
nilehmann: We don't need this to be a `QueryErr`; it can be a "normal error" that we emit right away.

crates/flux-refineck/src/lib.rs
ranjitjhala: Excellent suggestion! The first part - its why I added the `emit` in the first place...

But I don't follow your meaning re "split ... at *this* level" ?

ranjitjhala: Oh do you mean, just move the check into `generate_and_check_lean_lemmas` (instead of its current home in `check_fn` ?)

petros-marko: Yeah exactly.

crates/flux-infer/src/lean_encoding.rs
ranjitjhala: So to make sure I follow, essentially use the same logic as `log_dir` to implement a `lean_dir` ?

[The extra wrinkle here is that for the hyperlinks to work I need absolute paths but presumably there is some API to get that ...]

nilehmann: Yes, that's what I'm suggesting.

In flux land, I'd accept both relative and absolute paths. If absolute, we don't do anything special. If relative, we consider it relative to the `working_dir`, so we can just append it to `tcx.sess.opts.working_dir`

In `cargo-flux`, we can pass either an absolute or a relative path, depending on which one is easier.

ranjitjhala: Ok I think that's all done -- I *believe* that `path.join(rest)` takes care of the relative/absolute business?

PR #1401, opened 2025-12-12
Fix parsing constants in fixpoint

cole-k:
Caught a bit of a nasty bug when I was parsing a value greater than `i32::MAX` but that still fit into an `f32`.

Not sure why the code was written like this but it should be (more) correct now. Not fully tested other than my code works now.

This still leaves the upper half of `u128` as an edge case, but now the code panics even if it can be parsed as an `f64`. I suppose it would be more robust to parse the leading `-` out, but I'm lazy. I'm willing to change it to be more robust though.

### DIFF
--- a/lib/liquid-fixpoint/src/parser.rs
+++ b/lib/liquid-fixpoint/src/parser.rs
@@ -186,13 +186,7 @@ where
             }
             Sexp::Atom(Atom::Q(s)) => Ok(Expr::Constant(Constant::String(self.parser.string(s)?))),
             Sexp::Atom(Atom::B(b)) => Ok(Expr::Constant(Constant::Boolean(*b))),
-            Sexp::Atom(Atom::I(i)) => {
-                if *i >= 0 {
-                    Ok(Expr::Constant(Constant::Numeral(*i as u128)))
-                } else {
-                    Ok(Expr::Neg(Box::new(Expr::Constant(Constant::Numeral(-i as u128)))))
-                }
-            }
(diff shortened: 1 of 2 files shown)


lib/liquid-fixpoint/src/sexp.rs
```diff
@@ -121,9 +121,9 @@ impl<'a> Parser<'a> {
             "true" => Ok(Sexp::Atom(Atom::B(true))),
             "false" => Ok(Sexp::Atom(Atom::B(false))),
             _ => {
-                if let Ok(i) = s.parse::<i32>() {
+                if let Ok(i) = s.parse::<i128>() {
```
nilehmann: I believe numeric constants cannot be negative. Can you corroborate this @ranjitjhala?

ranjitjhala: That is correct https://github.com/ucsd-progsys/liquid-fixpoint/blob/9e5603ce755fc178066e4482498691376dbde20e/src/Language/Fixpoint/Parse.hs#L762-L766

cole-k: Updated.

lib/liquid-fixpoint/src/parser.rs
```diff
@@ -448,7 +450,6 @@ where
 
     fn parse_func_sort(&self, items: &[Sexp]) -> Result<Sort<T>, ParseError> {
         if let Sexp::Atom(Atom::I(params)) = &items[0]
-            && *params >= 0
```
cole-k, line 451: I assume this also won't be negative?

PR #1403, opened 2025-12-14
Source level binders in Lean VCs

ranjitjhala:
1. Differentiate between `Lean` and `Fixpoint` encoding,
2. Disable `imm` for `Lean` encoding to reduce binders,
3. Use `NameProvenance` when rendering `Lean` VC.

We get VCs like this now.

Screenshot 2025-12-14 at 1 22 31 PM

@petros-marko please take a look to see if I haven't mangled something!

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -25,7 +25,7 @@ use flux_middle::{
     query_bug,
     rty::{
         self, ESpan, EarlyReftParam, GenericArgsExt, InternalFuncKind, Lambda, List,
-        NameProvenance, ProvenanceMap, SpecFuncKind, VariantIdx, fold::TypeFoldable as _,
+        NameProvenance, PrettyMap, PrettyVar, SpecFuncKind, VariantIdx, fold::TypeFoldable as _,
     },
 };
 use itertools::Itertools;
@@ -686,7 +686,7 @@ where
             let kvar_decls = self.kcx.encode_kvars(&self.kvars, &mut self.scx);
             self.ecx.errors.to_result()?;
 
-            let lean_encoder = LeanEncoder::new(self.genv, self.ecx.local_var_env.provenance_map);
+            let lean_encoder = LeanEncoder::new(self.genv, self.ecx.local_var_env.pretty_var_map);
             lean_encoder
                 .encode_constraint(def_id, &kvar_decls, &constraint)
                 .map_err(|_| query_bug!("could not encode constraint"))?;
@@ -735,6 +735,13 @@ where
             self.comments.push(format!("Tag {idx}: {tag:?}"));
             idx
         })
+    }
+
+    pub(crate) fn with_early_param(&mut self, param: &EarlyReftParam) {
+        self.ecx
+            .local_var_env
+            .pretty_var_map
+            .set(PrettyVar::Param(*param), Some(param.name));
     }
 
     pub(crate) fn with_name_map<R>(
@@ -1030,7 +1037,7 @@ struct LocalVarEnv {
     /// kvars (which can be arbitrary expressions) as local variables; thus we
     /// need to keep the output as an [`rty::Expr`] to reflect this.
     reverse_map: UnordMap<fixpoint::LocalVar, rty::Expr>,
-    provenance_map: ProvenanceMap<fixpoint::LocalVar>,
+    pretty_var_map: PrettyMap<fixpoint::LocalVar>,
 }
 
 impl LocalVarEnv {
@@ -1040,7 +1047,7 @@ impl LocalVarEnv {
             fvars: Default::default(),
             layers: Vec::new(),
             reverse_map: Default::default(),
-            provenance_map: ProvenanceMap::new(),
+            pretty_var_map: PrettyMap::new(),
         }
     }
 
@@ -1058,7 +1065,8 @@ impl LocalVarEnv {
         let fresh = self.fresh_name();
         self.fvars.insert(name, fresh);
         self.reverse_map.insert(fresh, rty::Expr::fvar(name));
-        self.provenance_map.set(fresh, provenance);
+        self.pretty_var_map
+            .set(PrettyVar::Local(fresh), provenance.opt_symbol());
         fresh
     }
 
--- a/crates/flux-infer/src/lean_format.rs
+++ b/crates/flux-infer/src/lean_format.rs
@@ -1,8 +1,10 @@
 use core::fmt;
 use std::fmt::Write;
 
-use flux_common::dbg::as_subscript;
-use flux_middle::{global_env::GlobalEnv, rty::ProvenanceMap};
+use flux_middle::{
+    global_env::GlobalEnv,
+    rty::{PrettyMap, PrettyVar},
+};
 use itertools::Itertools;
 use liquid_fixpoint::{FixpointFmt, Identifier, ThyFunc};
 
@@ -13,7 +15,7 @@ use crate::fixpoint_encoding::fixpoint::{
 
 pub struct LeanCtxt<'a, 'genv, 'tcx> {
     pub genv: GlobalEnv<'genv, 'tcx>,
-    pub provenance_map: &'a ProvenanceMap<LocalVar>,
+    pub pretty_var_map: &'a PrettyMap<LocalVar>,
 }
 
 pub struct WithLeanCtxt<'a, 'b, 'genv, 'tcx, T> {
@@ -210,9 +212,6 @@ impl LeanFmt for Var {
                     write!(f, "{path}_{}", def_id.name())
                 }
             }
-            Var::Local(local_var) => {
-                write!(f, "{}", cx.provenance_map.get(*local_var))
-            }
             Var::DataCtor(adt_id, _) | Var::DataProj { adt_id, field: _ } => {
                 write!(
                     f,
@@ -221,9 +220,11 @@ impl LeanFmt for Var {
                     self.display().to_string().replace("$", "_")
                 )
             }
+            Var::Local(local_var) => {
+                write!(f, "{}", cx.pretty_var_map.get(&PrettyVar::Local(*local_var)))
+            }
             Var::Param(param) => {
-                // We start with a `₀` prefix to distinguish from the `LocalVar`
-                write!(f, "{}₀{}", param.name, as_subscript(param.index))
+                write!(f, "{}", cx.pretty_var_map.get(&PrettyVar::Param(*param)))
             }
             _ => {
                 write!(f, "{}", self.display().to_string().replace("$", "_"))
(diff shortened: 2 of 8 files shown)


crates/flux-middle/src/rty/expr.rs
```diff
@@ -942,29 +942,11 @@ newtype_index! {
...
-                }
-            })
-            .collect()
-    }
-}
+// impl Name {
+//     pub fn as_subscript(&self) -> String {
+//         as_subscript(self.as_usize())
+//     }
+// }
```
petros-marko: Nit: we can delete this too, right?

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -131,7 +131,8 @@ pub enum SortCtor<T: Types> {
 #[derive_where(Hash, Clone, Debug)]
 pub enum Pred<T: Types> {
     And(Vec<Self>),
-    KVar(T::KVar, Vec<T::Var>),
+    // KVar(T::KVar, Vec<T::Var>),
+    KVar(T::KVar, Vec<Expr<T>>),
```
petros-marko: Maybe we should have a comment explaining that for the Fixpoint backend the arguments are expected to be variables?

petros-marko (approved):
LGTM. I only left a couple of nits that shouldn't be blocking imo.

Thanks for also doing the simplifying of the constraints too, it's so much more readable!

lib/liquid-fixpoint/src/constraint.rs
ranjitjhala: excellent suggestion, thanks!!

crates/flux-middle/src/rty/expr.rs
ranjitjhala: good catch, thanks!

crates/flux-common/src/dbg.rs
```diff
@@ -40,7 +40,7 @@ impl SpanTrace {
         let file = SpanTrace::span_file(tcx, span);
         SpanTrace { file, start_line, start_col, end_line, end_col }
     }
-    pub fn from_pathbuf(path: &PathBuf, start_line: usize, start_col: usize, len: usize) -> Self {
+    pub fn from_pathbuf(path: &Path, start_line: usize, start_col: usize, len: usize) -> Self {
```
nilehmann: Maybe

```suggestion
    pub fn from_path(path: &Path, start_line: usize, start_col: usize, len: usize) -> Self {
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1239,10 +1258,15 @@ pub struct ExprEncodingCtxt<'genv, 'tcx> {
     /// invariants for an extern spec on an enum.
     def_id: Option<MaybeExternId>,
     infcx: rustc_infer::infer::InferCtxt<'tcx>,
+    backend: Backend,
```
nilehmann, line 1266: Since we have the concept of a backend now, maybe we could remove the `make_all_single` hack and pass the backend to `KVarEncodingCtxt::declare`

lib/liquid-fixpoint/src/format.rs
```diff
@@ -267,12 +267,13 @@ impl<T: Types> fmt::Display for Pred<T> {
...
             }
-            Pred::KVar(kvid, vars) => {
+            Pred::KVar(kvid, args) => {
                 write!(
                     f,
                     "(${} {})",
                     kvid.display(),
-                    vars.iter().map(Identifier::display).format(" ")
+                    args.iter().join(" "),
+                    // args.iter().map(Identifier::display).format(" ")
```
nilehmann:
```suggestion
```

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -131,7 +131,8 @@ pub enum SortCtor<T: Types> {
 #[derive_where(Hash, Clone, Debug)]
 pub enum Pred<T: Types> {
     And(Vec<Self>),
-    KVar(T::KVar, Vec<T::Var>),
+    /// When translating to `Fixpoint`, the arguments should always be `Expr::Var`.
+    KVar(T::KVar, Vec<Expr<T>>),
```
nilehmann, line 135: Ping @cole-k to see if anything breaks for him.

crates/flux-infer/src/lean_format.rs
```diff
@@ -1,59 +1,92 @@
...
+use rustc_data_structures::unord::UnordMap;
 
 use crate::fixpoint_encoding::fixpoint::{
     BinOp, BinRel, ConstDecl, Constant, Constraint, DataDecl, DataField, DataSort, Expr, FunDef,
     KVarDecl, Pred, Sort, SortCtor, SortDecl, Var,
 };
 
+pub struct LeanCtxt<'a, 'genv, 'tcx> {
+    pub genv: GlobalEnv<'genv, 'tcx>,
+    pub provenance_map: &'a UnordMap<crate::fixpoint_encoding::fixpoint::LocalVar, NameProvenance>,
```
nilehmann: Import this

```suggestion
    pub provenance_map: &'a UnordMap<LocalVar, NameProvenance>,
```

crates/flux-infer/src/lean_format.rs
```diff
@@ -163,16 +211,30 @@ impl<'a, 'genv, 'tcx> fmt::Display for LeanVar<'a, 'genv, 'tcx> {
...
                     f,
                     "{}.{}",
                     LeanSortVar(&DataSort::Adt(*adt_id)),
-                    self.0.display().to_string().replace("$", "_")
+                    self.display().to_string().replace("$", "_")
                 )
             }
+            Var::Param(param) => {
+                // We start with a `₀` prefix to distinguish from the `LocalVar`
+                write!(f, "{}₀{}", param.name, as_subscript(param.index))
```
nilehmann: When I saw `n₀₀` in the image, I found it a bit confusing. Some other ideas that I don't love

* `n'₀`: `'` is not valid in identifiers in Rust, so we'll never have conflicts
* `nₚ₀`: `p` for parameter?

But ultimately, I think the better solution is to not use the index in the variable and have a separate counter per name.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1805,12 +1830,16 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         sort: &rty::Sort,
         scx: &mut SortEncodingCtxt,
         bindings: &mut Vec<fixpoint::Bind>,
-    ) -> QueryResult<fixpoint::Var> {
+    ) -> QueryResult<fixpoint::Expr> {
         let farg = self.expr_to_fixpoint(arg, scx)?;
+        // If we're translating to Lean, no need to do any ANF-ing.
+        if matches!(self.backend, Backend::Lean) {
+            return Ok(farg);
+        }
```
nilehmann, line 1843: Maybe we should put this check in the caller? I don't feel strongly about it, though.

crates/flux-common/src/dbg.rs
```diff
@@ -208,3 +208,45 @@ macro_rules! _debug_assert_eq3 {
...
+                '5' => '⁵',
+                '6' => '⁶',
+                '7' => '⁷',
+                '8' => '⁸',
+                '9' => '⁹',
+                _ => c,
+            }
+        })
+        .collect()
+}
```
nilehmann: I believe we are not using this, and sadly, it looks like superscripts are not valid in Lean identifiers :( 

```suggestion
```

crates/flux-infer/src/lean_format.rs
ranjitjhala: Yes, let me work on the separate-counter. I think that is clearest.

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: Was a bit icky to do that because then you have to pass both the `arg` and `farg` into the `imm` to save the backwards map; i.e. too much of the logic bleeds into the caller...

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: Ok, I did that and the tests pass but I'm quite unfamiliar with that code so you might want to double check this!

lib/liquid-fixpoint/src/constraint.rs
cole-k: This is preferred in fact.

crates/flux-infer/src/lean_format.rs
ranjitjhala: Ok I've used the prefix thing, its a bit ugly because you have to mix both the `EarlyReftParam(Param)` and the `LocalVar` into the same API which gets a bit icky...?

ranjitjhala:
@nilehmann lmk if you have any more comments, thanks!

ranjitjhala:
Now looks like this 
Screenshot 2025-12-15 at 2 24 34 PM

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1030,10 +1057,16 @@ impl LocalVarEnv {
...
+    fn insert_fvar_map(
+        &mut self,
+        name: rty::Name,
+        provenance: NameProvenance,
+    ) -> fixpoint::LocalVar {
         let fresh = self.fresh_name();
         self.fvars.insert(name, fresh);
         self.reverse_map.insert(fresh, rty::Expr::fvar(name));
+        self.pretty_var_map
+            .set(PrettyVar::Local(fresh), provenance.opt_symbol());
```
nilehmann, line 1069: hmmm, I believe you are conflating things. One thing is to map a LocalVar to a Symbol, and something different is to generate unique identifiers from repeated symbols. They don't have to be mixed, and the second can happen only during lean encoding.

nilehmann: @ranjitjhala

> The issue is when you have `EarlyReft` named `n` AND you have
other `LocalVar` who may ALSO be named `n`; you want to make
sure you don't end up doing some weird shadowing.

Yes, but you don't care where the `n` came from when generating the unique identifier. 
There are two independent things involved

1. Associate a source name with a variable. For early params, you can just look at the early reft. For local vars, you can look at the `LocalVar -> Symbol` mapping.
2. Disambiguate a name by putting a prefix. At this point, you don't care where the name came from.

nilehmann: oh sorry. I understand now. The issue is with getting what's the identifier you assigned to a var.

ranjitjhala:
Ok am merging this following the discussion here [#lean-ffi > general chat @ 💬](https://flux-rs.zulipchat.com/#narrow/channel/531635-lean-ffi/topic//near/563950619) we could revert and remove the `EarlyReft` into their own "namespace" with a `+` subscript or such if we want to make the code less icky.

PR #1406, opened 2025-12-16
Stop emitting type classes for opaque sorts and functions

petros-marko:
Updated:
- emit one file per opaque sort if it doesn't exist and define using `sorry`
- emit one file per opaque function if it doesn't exist and define using `sorry`
- import all definitions that actually get used into one file for sorts and one file for functions

Much better indeed :)
___________
Instead of the type-class idea, we create (separately for opaque sorts and opaque functions)
- a fully generated file with all the definitions just pointing to a user-edited file
- a user edited file (not overwritten once it exists, but initially generated) where the actual definitions are given by the user

We *can* get rid of the generated file. If we just don't emit it and have theorem/proof files import the user editable file directly, I think we'll start getting errors where lean is failing to infer types correctly. The proper way to get rid of it is to do a very rudimentary analysis (e.g. search it for the name of the definition) on the user-edited file and just put a `def {xyz} := sorry` whenever a new opaque sort or function (xyz in this case) is defined. That'd also get rid of the errors (missing definitions) that make VSCode freeze.

To be honest I'm all for getting rid of the generated file if we do this sort of analysis, but I wanted to give people a chance to voice thoughts, hence this PR.

### DIFF
--- a/crates/flux-infer/src/lean_encoding.rs
+++ b/crates/flux-infer/src/lean_encoding.rs
@@ -13,7 +13,6 @@ use flux_middle::{
     queries::{QueryErr, QueryResult},
     rty::PrettyMap,
 };
-use itertools::Itertools;
 
 use crate::{
     fixpoint_encoding::fixpoint,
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Thanks @petros-marko -- are you doing the separate-files thing too, yes?

petros-marko:
> Thanks @petros-marko -- are you doing the separate-files thing too, yes?

Yes, I won't merge this. I'll just do the separate files.

PR #1407, opened 2025-12-17
Support for recursive function definitions and PLE

ranjitjhala:
So we can now check stuff like the test below. Stepping stone to support for reflecting recursive ADTs...

```rust
defs! {
    #[recursive]
    fn sum_to(n: int) -> int {
        if n <= 0 {
            0
        } else {
            n + sum_to(n - 1)
        }
    }
}

#[flux::opts(ple = "true")]
#[spec(fn () -> usize[sum_to(5)])]
fn test_sum_to() -> usize {
    15
}
```


crates/flux-syntax/src/symbols.rs
```diff
@@ -18,6 +18,7 @@ symbols! {
         Reft: "reft",
         Invariant: "invariant",
         RefinedBy: "refined_by",
+        Recursive: "recursive",
```
nilehmann, line 21: Maybe we should make this a symbol so we can use `recursive` as an identifier.

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -163,6 +172,18 @@ pub struct FunDef<T: Types> {
     pub body: Expr<T>,
     #[derive_where(skip)]
     pub comment: Option<String>,
+    pub recursive: bool,
+}
+
+impl<T: Types> FunDef<T> {
+    pub fn as_constant(&self) -> FunConstant<T> {
```
nilehmann, line 179: see https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

```suggestion
    pub fn to_constant(&self) -> FunConstant<T> {
```

lib/liquid-fixpoint/src/lib.rs
```diff
@@ -155,6 +155,15 @@ pub struct ConstDecl<T: Types> {
...
 }
 
+#[derive_where(Hash)]
+pub struct FunConstant<T: Types> {
+    pub name: T::Var,
+    pub args: Vec<Sort<T>>,
+    pub out: Sort<T>,
+    #[derive_where(skip)]
+    pub comment: Option<String>,
+}
```
nilehmann, line 165: We already have `ConstDecl`, do we need this extra one?

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -52,6 +52,8 @@ enum SyntaxAttr {
     Hide,
     /// a `#[opaque]` attribute
     Opaque,
+    /// a `#[recursive]` attribute
+    Recursive,
```
nilehmann, line 56: The same note about `Hide` applies to `Recursive` here. We could keep them next to each other and update the comment.

crates/flux-middle/src/rty/normalize.rs
```diff
@@ -1,10 +1,13 @@
...
-use rustc_data_structures::{fx::FxIndexSet, unord::UnordMap};
+use rustc_data_structures::{
+    fx::FxIndexSet,
+    graph::{DirectedGraph, Successors, scc::Sccs},
+    unord::UnordMap,
+};
+use rustc_hash::FxHashSet;
 use rustc_hir::def_id::{CrateNum, DefIndex, LOCAL_CRATE};
 use rustc_macros::{TyDecodable, TyEncodable};
-use toposort_scc::IndexGraph;
```
nilehmann, line 7: Is this dependency used somewhere else? Can we remove it?

crates/flux-middle/src/rty/normalize.rs
```diff
@@ -50,6 +53,15 @@ pub struct NormalizeInfo {
     pub rank: usize,
     /// whether or not this function is uninterpreted by default
     pub hide: bool,
+    /// whether or not this function is recursive
+    pub recursive: bool,
```
nilehmann, line 57: It feels a bit redundant to have `inline`, `hide`, and `recursive`. There's probably some generalization one could do, but I can't see what it is that yet.

nilehmann (commented):
* I like the idea of having to mark recursive functions (at least some form of mark), but it's a bit odd that you can have non-recursive functions marked as `#[recursive]`.

* With the new "each function in a different lean file", one could go and adjust the termination proof, right?

* I just tested it, and I can see that fixpoint will happily accept non-terminating functions and loop forever. How should I think about the soundness of that? And LH has a termination checker, right?

PR #1408, opened 2025-12-18
Fix fixpoint-rs to work with exprs as kvar args

nilehmann:


lib/liquid-fixpoint/src/constraint_solving.rs
```diff
@@ -472,7 +468,7 @@ impl<T: Types> Expr<T> {
         }
     }
 
-    pub(crate) fn substitute(&self, v_from: &T::Var, v_to: &T::Var) -> Self {
+    pub(crate) fn substitute(&self, v_from: &T::Var, v_to: &Expr<T>) -> Self {
```
nilehmann, line 471: ping @cole-k as I believe you were using this function.

cole-k: Should be fine since I can just lift from `T::Var` to `Expr<T>`.

PR #1409, opened 2025-12-18
Take kvar solutions from fixpoint and encode them in lean

petros-marko:
What the title says :)

There are a couple of hacks, that can be improved...

### DIFF
--- a/lib/liquid-fixpoint/src/constraint.rs
+++ b/lib/liquid-fixpoint/src/constraint.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashSet, hash::Hash};
 
 use derive_where::derive_where;
-use itertools::Itertools;
 
 use crate::{ThyFunc, Types};
 


crates/flux-middle/src/rty/expr.rs
```diff
@@ -1004,7 +1004,7 @@ impl<V: Eq + Hash + Copy + Into<usize>> PrettyMap<V> {
     pub fn get(&self, key: &PrettyVar<V>) -> String {
         match self.map.get(key) {
             Some(s) => s.clone(),
-            None => format!("a₊{}", key.as_subscript()),
+            None => format!("a'{}", key.as_subscript()),
```
petros-marko, line 1007: `+` not allowed in lean variable names; this was causing trouble in the expressions generated from kvar solutions.

crates/flux-refineck/src/lib.rs
```diff
@@ -163,6 +159,70 @@ pub fn check_fn(
 
             let errors = answer.errors;
             report_fixpoint_errors(genv, def_id, errors)
+        } else {
```
petros-marko: Is there a way to avoid doing this twice? Both `execute_fixpoint_query` and `execute_lean_query` currently consume their arguments, but maybe we can get away with just borrowing?

petros-marko:
> > @cole-k is there a reason that for Expr::Exists you chose to go without storing the bound variable names? I added them in and I don't want to break anything...
> 
> Did I add them? There's no reason why we couldn't have the names, but you might want to switch from using the [`BoundVar`](https://github.com/flux-rs/flux/blob/2ce18aaf692071b03577c2b56c5daa9b01a2fc2c/lib/liquid-fixpoint/src/constraint.rs#L212-L215) to some sort of named representation. Although in `rty` we do the same thing you've done: we store the names at the binder and use a nameless representation throughout the rest of the AST.

Oh, it looks like Ranjit added them. I saw them used in the `decoding` module and I assumed it was you.

crates/flux-refineck/src/lib.rs
nilehmann: @petros-marko, we can make `FixpointCtxt::check` take a `&mut self` instead of taking ownership (that will work without any extra changes) and then call it inside `execute_lean_query` (at least in principle). But maybe we should sync about this. Running `FixpointCtxt::check` adds extra stuff to encode, and I'm not sure how this affects the lean encoding (i.e., do lean encoding on the same `FixpointCtxt` that we run fixpoint)

nilehmann: Or maybe we just need to split `FixpointCtxt::check` into the part that builds the `Task` and the part that executes it. Then, for lean, we can (1) build the task, (2) execute it (without taking ownership), and (3) extract from the task what's needed for the lean encoding.

petros-marko: Ok, I'll figure out how to do it :) Looks OK otherwise? It's definitely kind of hacky.

crates/flux-middle/src/rty/expr.rs
ranjitjhala: Wait, this seems very old -- I removed all this `+` stuff?

petros-marko: Interesting. I just merged main into my branch and I didn't get any conflicts or anything...

crates/flux-refineck/src/lib.rs
nilehmann: You'll also need to do something about the bound vars. Adding the names to the exist is not enough because bound variables are being represented with debruijn indices. So you either change everything to be named or deal with the debruijn to name mapping during lean encoding.

petros-marko: I think it's fine. We only ever introduce `Expr::BoundVars` during parsing. But then we convert the `fixpoint` representation that gets parsed to `rty`, and my code goes back to `fixpoint` from there, and the `BoundVars` get wiped. It's maybe a bit inefficient to do this back-and-forth, but I wanted to take advantage of the kvar-solution-grouping that was implemented over `rty`. Maybe I'm missing something though.

nilehmann: I think is silly to go to rty to then just go back to fixpoint. The kvar solution groupping is not necessary because we are not expanding kvars for lean so its a nop. But even if we were, staying in fixpoint means we don't need to worry about the quirks of going back and forth.

petros-marko: Wait, grouping is not a no-op. We get the solutions by running in fixpoint mode (expanded), but we use the solutions in lean mode, where we don't expand them. Or are you saying that the solution for the first kvar of the range is good enough and we can ignore the solutions from the "auxiliary" ones?

At any rate, if you think going back-and-forth is silly, that's a fair criticism either way 😄

nilehmann: It will be a no-op with the refactoring I proposed above because we will run fixpoint in the same FixpointCtxt that we use to encode into lean.

petros-marko: Can we even do that? I thought liquid-fixpoint needs the expanded kvars. I guess since we only care about the acyclic kvar solutions it's ok if we don't?

Also, is it easier to do a quick zoom call to figure this out? I had a different idea for the refactoring which seems like it'd work and is very easy but doesn't do what you propose here.

nilehmann: Ah, you mean the a-normalized version. Yeah you are right, hmmm

nilehmann: Maybe you didn't mean that, but it will be an issue with the strategy I proposed.

nilehmann: OK, then maybe just implementat whatever you had in mind so we can have something working soon. But as a side note, how far are we from using your rust implementation of the non-cyclic kvars?

petros-marko: Ok, I just pushed it.

I had misunderstood the solving algorithm over the  summer so it doesn't go as far as it could. My guess is that it wouldn't be very hard to write the correct version if we decide it's worth the effort.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1499,12 +1513,15 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
             rty::ExprKind::GlobalFunc(SpecFuncKind::Def(def_id)) => {
                 fixpoint::Expr::Var(self.declare_fun(*def_id))
             }
+            rty::ExprKind::Exists(expr) => {
+                let expr = self.body_to_fixpoint(expr, scx)?;
```
nilehmann, line 1524: I believe my poor little heart can't take this 🫠. `fixpoint::Exists` uses names when going from rty, but debruijn when parsing!

This is getting too much bandwidth. We should probably discuss alternatives over zoom

petros-marko: Admittedly weird. I knew we'd eventually reach your breaking point with regard to hacks! haha. Happy to chat today (now-ish) or leave it for tomorrow (if we're planning to meet as regularly scheduled)

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -628,18 +631,24 @@ where
...
+        let expr = self.ecx.body_to_fixpoint(solution, &mut self.scx)?;
+        Ok((fixpoint::KVid::from_usize(kvid.as_usize()), expr))
+    }
+
+    fn parse_kvar_solutions(&mut self, result: &FixpointResult<TagIdx>) -> QueryResult<Solution> {
+        let mut solutions = vec![];
+        for b in result.solution.iter().chain(&result.non_cuts_solution) {
+            solutions.push(self.convert_kvar_bind(b)?);
+        }
+        Ok(self.kcx.group_kvar_solution(solutions))
```
nilehmann, line 648: Why did you change this function?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -628,18 +631,24 @@ where
...
-            result
-                .solution
-                .iter()
-                .chain(&result.non_cuts_solution)
-                .map(|b| self.convert_kvar_bind(b))
-                .try_collect_vec()?,
-        ))
+    pub(crate) fn kvar_solution_for_lean(
+        &mut self,
+        kvid: &rty::KVid,
```
```suggestion
        kvid: rty::KVid,
```

crates/flux-infer/src/fixpoint_encoding.rs
petros-marko: Borrow checker didn't like the closure after `self` became `&mut`.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -297,118 +293,16 @@ impl<T: Types> Expr<T> {
...
+                for (var, _sort) in binder {
+                    inner.remove(var);
+                }
+                vars.extend(inner);
             }
         };
         vars
     }
-
-    pub fn substitute_bvar(&self, subst_layer: &[Expr<T>], current_level: usize) -> Self {
```
petros-marko, line 308: As far as I could tell, this was only ever used on @cole-k's branch in `hoist_exists`. With the named representation for `Expr::Exists`, `hoist_exists` can be implemented directly, without this helper method, so I thought it was OK to remove. @cole-k lmk if that's not the case.

nilehmann (approved):
@petros-marko I think it's fine to merge this. We can improve upon it later if https://github.com/ucsd-progsys/liquid-fixpoint/pull/818#pullrequestreview-3608044217 ever gets merged (or if we end up implementing ANF-ing in fixpoint).

I just want confirmation from @cole-k that the removal of `Expr::BoundVar` doesn't break his thing.

petros-marko:
> @petros-marko I think it's fine to merge this. We can improve upon it later if [ucsd-progsys/liquid-fixpoint#818 (review)](https://github.com/ucsd-progsys/liquid-fixpoint/pull/818#pullrequestreview-3608044217) ever gets merged (or if we end up implementing ANF-ing in fixpoint).
> 
> I just want confirmation from @cole-k that the removal of `Expr::BoundVar` doesn't break his thing.

I talked with @cole-k yesterday, and it seems like this is fine 👍🏽 by him. I wanted to ask @ranjitjhala if the following changes should be made in this PR, a new PR, or his PR (it's going to be messy in any case):
 - This PR further relies on encoding boolean expressions as `Prop`s (which your PR #1410) undoes. I think the solution is to add another field to `LeanCtxt` that determines what we should translate booleans to.
 - It also just defines the solved KVars in the VC files, and that'd cause name clashes when applied together with #1410, because we'd be importing all the proof files into `Basic.lean`. We should wrap everything in each VC file in a namespace to avoid that. The corresponding proof file can just open the namespace at the beginning.

ranjitjhala:
On further thought -- maybe best to do them LATER, as otherwise there will be MORE stuff to deal with when merging :-)

i.e. lets just merge this as is, and then deal with the other stuff in #1410?

PR #1410, opened 2025-12-19
Localize lean queries

ranjitjhala:
This PR refactors the `lean-encoding.rs` substantially so that we now generate a directory like

```
lean_proofs/
   LeanProofs/
      Flux/
        Struct/
        Fun/
        VC/
      User/
        Struct/
        Fun/
        Proof/
```

Where the 
- Flux has stuff that is regenerated and should not be user edited
- User has stuff that should be user edited 
- `VC`s are in the directory of their name
- `Proof` has the corresponding proofs 
- Each `VC` imports the relevant bits needed *just* for that VC

So crucially, there are no "global" files that import all definitions (which had the drawback where a single `sorry` can cause ALL the `Proof` to fail...)

@petros-marko -- can you push the full proofs for `flux-demo` into some branch? (Or lmk which branch has them, so I can update that repo to use this stuff in a matching PR there? Currently in `main` I don't see the definitions for various things like `svec_svec_slice` and so on...?)

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -377,7 +377,6 @@ impl SortEncodingCtxt {
     }
 
     pub fn declare_adt(&mut self, did: DefId) -> AdtId {
-        // todo!("HEREHEREHEREHERE");
         if let Some(idx) = self.adt_sorts.get_index_of(&did) {
             AdtId::from_usize(idx)
         } else {


petros-marko:
> @petros-marko -- can you push the full proofs for `flux-demo` into some branch? (Or lmk which branch has them, so I can update that repo to use this stuff in a matching PR there? Currently in `main` I don't see the definitions for various things like `svec_svec_slice` and so on...?)

I'll need to fix them first; we've made some changes (the type class stuff, these changes here, etc). If it's ok, I'll do that after I merge #1409 because that'll also break some things.

ranjitjhala:
Btw, for a bit of bikeshedding, I wonder if we should further tweak the directory structure to make it totally clear what is (to be) user-edited and what is flux-generated.

```
lean_proofs/
   LeanProofs/
      user/
         Fun/
         Struct/
         Proof/
      flux/
         Fun/
         Struct/
         VC/
```

petros-marko (commented):
LGTM. This is much nicer, thank you!

I have a quick question: how are struct names generated? We had talked about this being a problem potentially if say constraint A only talks about struct Foo and constraint B talks about Foo and Bar, it might be that when encoding A we call Foo Adt0 but when we encode B we call Bar Adt0, in which case some file might be overwritten. Are struct names generated from the pretty var map now?

petros-marko:
> Yes, merge #1409 first -- I wonder what the best way is to handle THIS PR? Is there some git trick to somehow apply the changes "on top" of whatever you did in 1409?

I think I'll need to make some relatively big changes anyway (because Nico has very reasonable objections 😄) so you should go ahead with this and I'll manage the conflicts; they shouldn't be very bad anyway.

petros-marko:
> Btw, for a bit of bikeshedding, I wonder if we should further tweak the directory structure to make it totally clear what is (to be) user-edited and what is flux-generated.
> 
> ```
> lean_proofs/
>    LeanProofs/
>       user/
>          Fun/
>          Struct/
>          Proof/
>       flux/
>          Fun/
>          Struct/
>          VC/
> ```

I think this is good, but we could have better names for the Struct and Fun directories on the user side to explain that they correspond to the opaque defs in Flux...

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -42,370 +177,390 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
...
-        }
-        writeln!(structs_file, "-- STRUCT DECLS --")?;
-        writeln!(structs_file, "mutual")?;
-        let cx = LeanCtxt { genv: self.genv, pretty_var_map: &self.pretty_var_map };
-        for data_decl in data_decls {
-            writeln!(structs_file, "{}", WithLeanCtxt { item: data_decl, cx: &cx })?;
+        let name = &data_decl.name;
+        let file = &LeanFile::Struct(name.clone());
+        let path = self.path(file);
+        if !path.exists() {
```
nilehmann: I'm not sure if I'm reading this correctly, but if a struct changes but the name remains the same, then this code will leave the file untouched, right? Shouldn't we regenerate it?

ranjitjhala: Good catch! Yes everything that should be under Flux/ should be regenerated (if it doesn't already exist) so maybe we

1. Should use the Flux vs User subdirectories
2. Erase and regenerate Flux/* when rerunning

?

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -1,421 +1,580 @@
...
-        for data_decl in data_decls {
-            writeln!(structs_file, "{}", WithLeanCtxt { item: data_decl, cx: &cx })?;
+        let name = &data_decl.name;
+        let file = &LeanFile::Struct(name.clone());
+        let path = self.path(file);
+        // No need to regenerate if created in this session; but otherwise regenerate as struct may have changed
+        if !was_created_in_session(&path) {
+            let mut file = create_file_with_dirs(path)?;
+            // import prelude
+            writeln!(file, "{}", self.import(&LeanFile::Fluxlib))?;
```
nilehmann: We are not currently using any parallelism, but this will have races in the filesystem if we ever do.

A better approach is to put everything in a temporary directory and then swap it (as atomically as possible) at the end. Then we don't need an in-memory lock because file creation is atomic. Gemini explains it more eloquently https://gemini.google.com/share/9027e898b136

ranjitjhala: What about the User/ files that should be preserved (and not regenerated?) Should these be first "copied into" the staging area?

ranjitjhala: Btw I'm confused (even after reading the Gemini explanation) about one thing: how does this solve the races problem? Won't you just get the races in the temp directory? 

Presumably this is the "atomic check and create". But then why not just do that in the main directory - what does the "staging" buy? Per Gemini it is mostly to avoid OTHER consumers from seeing partial results (but that's not an issue for us?)

nilehmann: Probably better not to touch user files because it could cause conflicts if the user is editing them. That makes it a bit harder to do the "commit" atomically because `Basic.lean` is outside of `Flux`, so we can't get atomicity just with a directory rename, but the atomicity of that step is less important.

nilehmann: > Btw I'm confused (even after reading the Gemini explanation) about one thing: how does this solve the races problem? Won't you just get the races in the temp directory?
> 
> Presumably this is the "atomic check and create". But then why not just do that in the main directory - what does the "staging" buy? Per Gemini it is mostly to avoid OTHER consumers from seeing partial results (but that's not an issue for us?)

In the general case, the staging area helps because you can do the "commit" atomically, but that is not too important for us IMO (it'd be more correct though). In our case, the staging area helps because we don't have to worry about the case where the file is already in the filesystem, but it's the first time you write it in the current Flux invocation.

ranjitjhala: Ok so the summary is

1. Create and Write in STAGING dir
2. SKIP (1) for "User" files if already present in REAL directory 
3. Swizzle STAGING to REAL at the end to "commit"

I suppose we still need global state to 
1. Know what the staging directory is,
2. When to do the swizzle

nilehmann: Ah I didn't consider that the user files have to be created the first time. But yeah, for user files we can create them _in place_ because we don't have to consider the case where we have to regenerate them. Your plan sounds reasonable.

And yes, we need global state for the path of the tempdir, but I was thinking we can add it to the global env. So we create the temp dir when we create the global env and commit it when we finalize it.

ranjitjhala: Uff -- after implementing this whole thing, I realized your 2-step approach is doomed: we generate AND check the lean proofs incrementally (on a per `DefId` basis) so we need ALL the dependencies in the right place at the time when we're checking the proofs, meaning you need to generate the Vc and the other defs IN the final place (so they can import the relevant stuff when the proof is being checked by lean on the fly...)

ranjitjhala: ^ @nilehmann

ranjitjhala: Hmm unless one can pass an external
path to lean for the dependencies ...

nilehmann: Ah thats annoying. But can't we check everything at the end?

I actually never liked that we call lean from the flux binary. IMO the flux binary should only generate the files and calling lean should be handled by the "build system", i.e., by `cargo flux`.

ranjitjhala:
I don't think the chatgpt thing is making a meaningful distinction between "compiler" and "cargo flux". 

I personally make no distinction between the two -- all I care about is there is a single uniform & integrated mode of interaction (e.g. in vscode) where if I go and write a rust function with "external_proof" I get a red squiggle, which when I click and write the proof and come back, goes away (and returns if I do something to "break" the proof).

To achieve the above, you could do the lean-checks in the `flux binary` OR `cargo flux` it doesn't really matter, I think (but the chat-gpt argument basically goes against _both_ afaict.)

However, if we want to do the lean-checks in `cargo flux` there's a lot more machinery that would have to be setup to track which lean-files should be checked, and to do the error reporting etc. and IMO that should not be in this PR which is already quite substantial.

Sadly, I find that the approach of passing in an "external" lean directory doesn't work -- those need to be full fledged lean packages...

So I'd like to just revert this to the original (write-everything-in-place) approach, after merging in the KVAR related edits, and then we can come back to this question later. (Its pretty trivial to add the 2-phase business, the tedious bit seems to be figuring out the checker invocations.)

ranjitjhala:
I suppose in _this_ PR itself, its probably quite easy to 

1. Log a list of lean queries that should be made and the associated error messages;
2. Then add a thing at the end of `check_crate` that fires off the relevant queries and emits the appropriate errors?

WDYT?

nilehmann:
I agree that it will take a lot more setup to call lean from cargo flux, but I feel we should work towards that. Calling lean at the very end from the flux driver it's a step in that direction.

Would it work if we check the entire lake project at the end of check_crate in a single call? Regardless of the distinction between compiler and build system maybe this is something we disagree on in terms of the UX.

nilehmann:
> I suppose in _this_ PR itself, its probably quite easy to
> 
>     1. Log a list of lean queries that should be made and the associated error messages;
> 
>     2. Then add a thing at the end of `check_crate` that fires off the relevant queries and emits the appropriate errors?
> 
> 
> WDYT?

This sounds reasonable. My original concern was that an in-memory lock doesn't protect you from accessing the file system in parallel. This is not an immediate problem for us, but it'd be nice to parallelize flux eventually. The staging area was a suggestion to address that. Adding this extra tracking to run "lean queries" at the end sounds like it will allows to to implement the staging area idea.

petros-marko:
> @petros-marko -- I think we still want the VCs with the `exists k0, ...` as before -- since in general the fixpoint solutions may not work. (I mean, if they did, why bother with lean anyways?) So really we only want the non-cut kvar solutions, but
> we may have to ponder this a bit...

I agree that we want the `exists ...` for the kvars solved with predicate abstraction; I just thought that if fixpoint fails it'd only give the solutions for the acyclic kvars, which are always correct. If that's not the case, I can put in a quick fix tomorrow.

petros-marko:
Ok, then I think the right thing to do is to ignore everything but the acyclic ones for lean's purposes. I'll get that done tomorrow. I can do it on top of this branch to not give you more conflicts.

petros-marko:
> Or do you want to go a step further and ONLY leave the exists for the cyclic vars? (Even so, let’s write out the solution…).

That's that I was thinking, but I don't mind the idea of just putting in comments about cyclic / non-cyclic and keeping the exists around.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -377,6 +377,7 @@ impl SortEncodingCtxt {
     }
 
     pub fn declare_adt(&mut self, did: DefId) -> AdtId {
+        // todo!("HEREHEREHEREHERE");
```
```suggestion
```

PR #1413, opened 2025-12-24
Split kvars into cut/non-cut during formatting for Lean

petros-marko:
Partition kvar solutions into cut and non-cut by checking for the presence of existential quantifiers.

I didn't want to change the parse_kvar_solutions to return them separately to avoid messing up Cole's code, but it's always an option if you think that's a better approach🤷🏽.

This will wait until #1410 in.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -229,11 +229,10 @@ impl KvarSolutionTrace {
 }
 
 impl SolutionTrace {
-    pub fn new(genv: GlobalEnv, cut_solution: &Solution, non_cut_solution: &Solution) -> Self {
+    pub fn new<T>(genv: GlobalEnv, ans: &Answer<T>) -> Self {
         let cx = &PrettyCx::default(genv);
-        let res = cut_solution
-            .iter()
-            .chain(non_cut_solution.iter())
+        let res = ans
+            .solutions()
             .map(|(kvid, bind_expr)| KvarSolutionTrace::new(cx, *kvid, bind_expr))
             .collect();
         SolutionTrace(res)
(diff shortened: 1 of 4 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -377,6 +379,7 @@ impl SortEncodingCtxt {
     }
 
     pub fn declare_adt(&mut self, did: DefId) -> AdtId {
+        // todo!("HEREHEREHEREHERE");
```
```suggestion
```

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -595,10 +594,7 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
         Ok(())
     }
 
-    fn generate_vc_file(
-        &self,
-        kvar_solutions: HashMap<fixpoint::KVid, FixpointSolution>,
-    ) -> Result<(), io::Error> {
+    fn generate_vc_file(&self, kvar_solutions: KVarSolutions) -> Result<(), io::Error> {
```
nilehmann:
```suggestion
    fn generate_vc_file(&self, kvar_solutions: KVarSolutions) -> io::Result<()> {
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -229,10 +229,11 @@ impl KvarSolutionTrace {
 }
 
 impl SolutionTrace {
-    pub fn new(genv: GlobalEnv, solution: &Solution) -> Self {
+    pub fn new(genv: GlobalEnv, cut_solution: &Solution, non_cut_solution: &Solution) -> Self {
```
nilehmann: I'd pass the entire `Answer` here and add a method to it to iterate over all the solutions.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -648,14 +651,28 @@ where
...
+        self.parse_kvar_solutions(&result.solution)
+    }
+
+    fn parse_non_cut_kvar_solutions(
+        &mut self,
+        result: &FixpointResult<TagIdx>,
+    ) -> QueryResult<Solution> {
+        self.parse_kvar_solutions(&result.non_cuts_solution)
+    }
+
```
nilehmann: Having these methods feels a bit unnecessary.

nilehmann (approved):
Some cosmetic suggestions.

PR #1415, opened 2026-01-05
Only copy when source exists

petros-marko:
In `LeanEncoding::finalize` I was getting a crash because the `src` path is not always there. This PR just skips the renaming step if the directory is missing.


nilehmann (approved):
LGTM

PR #1417, opened 2026-01-06
Add syntactic inference for `no-panic`

ninehusky:
We _know_ we can mark a function as `no_panic` automatically if all of its dependencies are `no_panic`. This PR tracks the optimization to lessen the burden of declaring a module `no_panic`. I wonder what a way to evaluate this would be... maybe I can run on Tock with and without this optimization and see how many `extern_specs` I need to add?


crates/flux-refineck/src/checker.rs
```diff
@@ -948,10 +949,17 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 && let Some(callee_def_id) = callee_def_id
                 && genv.def_kind(callee_def_id).is_fn_like()
             {
+                // try to use flux-opt to infer the no_panicness of the callee.
+                let callee_no_panic = fn_sig.no_panic()
```
ninehusky: This maybe could be pretty expensive. `infer_no_panics` doesn't "remember" any information between calls, so it has to construct a call graph every time we run `check_call`. We could try to fold some of this logic into the `GlobalEnv` by making this a new query (or editing the existing `no_panic` query, which we aren't really using anyways).

cc @nilehmann

ninehusky:
@nilehmann wanted to ping this and bring it to your attention! don't worry if you can't get to it till monday.

ninehusky:
@nilehmann I believe this should be ready for your eyes again!

ninehusky:
Superceded by #1587

PR #1418, opened 2026-01-07
Misc support for generators 

### LINKED ISSUE #1411: ICE when using #[tokio::main] with no refinements
Hi Flux team,

I encountered an internal compiler error in Flux while compiling a minimal program that uses `#[tokio::main]`, despite the absence of any refinements.

This is with a standard Cargo setup following the Flux install guide
(`flux-rs` as a dependency and `package.metadata.flux` enabled).

---

## Minimal repro (`src/main.rs`)

```rust
#[tokio::main]
async fn main() {
    println!("Hello, world!");
}
```

---

## Command

```text
cargo flux
```

---

## Error

```text
error: internal compiler error: crates/flux-middle/src/rty/mod.rs:2093:17:
TODO: Generator DefId(0:4 ~ example1[d7a2]::main::{closure#0}) ResumeTy []

thread 'rustc' (72460) panicked at crates/flux-middle/src/rty/mod.rs:2093:17:
Box<dyn Any>
```

---

## Environment

- rustc: 1.94.0-nightly (21ff67df1 2025-12-15)
- OS: macOS (Apple Silicon / arm64)
- tokio: 1.48.0

---

Thanks!

ranjitjhala:
Fixes #1411


crates/flux-infer/src/infer.rs
```diff
@@ -1014,6 +1014,17 @@ impl<'a, E: LocEnv> Sub<'a, E> {
...
             (BaseTy::Never, BaseTy::Never) => Ok(()),
+            (
+                BaseTy::Coroutine(did1, resume_ty_a, tys_a, _),
+                BaseTy::Coroutine(did2, resume_ty_b, tys_b, _),
+            ) if did1 == did2 => {
+                debug_assert_eq!(tys_a.len(), tys_b.len());
+                for (ty_a, ty_b) in iter::zip(tys_a, tys_b) {
+                    self.tys(infcx, ty_a, ty_b)?;
+                }
+                self.tys(infcx, resume_ty_b, resume_ty_a)?; // TODO(RJ) Nico: is this contravariant?
```
ranjitjhala: @nilehmann pls double check this bit!

crates/flux-middle/src/rty/mod.rs
```diff
@@ -2100,8 +2111,11 @@ impl<'tcx> ToRustc<'tcx> for BaseTy {
                 let preds = tcx.mk_poly_existential_predicates(&preds);
                 ty::Ty::new_dynamic(tcx, preds, re.to_rustc(tcx))
             }
-            BaseTy::Coroutine(def_id, resume_ty, upvars) => {
-                bug!("TODO: Generator {def_id:?} {resume_ty:?} {upvars:?}")
+            BaseTy::Coroutine(did, _, _, args) => {
+                ty::Ty::new_coroutine(tcx, *did, args.to_rustc(tcx))
+                // bug!(
```
ranjitjhala: should we nuke this comment? Mostly this PR is adding the "breadcrumbs" to be able to `to_rustc` back the `coroutine` stuff, following what was done for `closure`.

nilehmann: yes, we can remove the comment

crates/flux-infer/src/infer.rs
nilehmann: Should we make this invariant just to be sure? We could revisit when we have a concrete example that fails. However, it may be hard to attribute an error to the variance once we see one.

In general, I wish we could make everything invariant and deal with *trivial cycles* in the constraint. These really ought to be the *same type*

crates/flux-middle/src/rty/fold.rs
```diff
@@ -1207,3 +1208,44 @@ TrivialTypeTraversalImpls! {
...
+                {
+                    ::core::ops::ControlFlow::Continue(())
+                }
+            }
+        )+
+    };
+}
+
+// For things that can be cloned...
+TrivialClonedTypeTraversalImpls! {
```
nilehmann: I believe you added this for `CoroutineObligPredicate`, but `CoroutineObligPredicate` doesn't need to be `TypeVisitable` or `TypeFoldable`, so we could remove those derives and this macro. And for that matter, it also doesn't need to be encodable, so we could define it as

```rust
#[derive(Clone, PartialEq, Eq, Hash, Debug)]
pub struct CoroutineObligPredicate {
    pub def_id: DefId,
    pub resume_ty: Ty,
    pub upvar_tys: List<Ty>,
    pub output: Ty,
    pub args: flux_rustc_bridge::ty::GenericArgs,
}
```

ranjitjhala: Wow, why on earth did I even think it needed those other things? Perhaps because I copied from the ClosureOblig...

PR #1420, opened 2026-01-08
`no_panic` associated refinement on `impl Fn` types

ninehusky:


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1498,6 +1498,10 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
             fhir::Res::Def(DefKind::TyParam, param_id)
             | fhir::Res::SelfTyParam { trait_: param_id } => {
                 let predicates = type_param_predicates(tcx, param_id);
+                if assoc_ident == Ident::from_str("no_panic") {
+                    // haha I assume here is where we'd do something like "register no_panic for a function?"
```
ninehusky: @nilehmann see this! here's what i was thinking. the `flux-desugar` crate also seemed like a likely area for me to start, but i'm not sure if this is where i should proceed.

ninehusky: never mind, found out about builtin and am moving my "best guess" to there.

crates/flux-middle/src/builtin_assoc_refts.rs
```diff
@@ -27,6 +27,21 @@ impl<'tcx> GlobalEnv<'_, 'tcx> {
 
                 let mut map = UnordMap::default();
 
+                // if it's a function, automatically include a no panic assoc reft if
+                // the `requires` for that function includes it.
```
ninehusky: the "if" is actually outdated here. i _think_ it's fine for every `Fn` to have a `no_panic` associated refinement?

crates/flux-middle/src/builtin_assoc_refts.rs
```diff
@@ -92,6 +116,9 @@ impl<'tcx> GlobalEnv<'_, 'tcx> {
                 .bytes();
             let body = rty::Expr::constant(rty::Constant::from(size));
             rty::Lambda::bind_with_vars(body, List::empty(), rty::Sort::Int)
+        } else if alias_reft.assoc_id.name() == Symbol::intern("no_panic") {
+            let body = rty::Expr::tt(); // TODO: fix!
```
ninehusky: This is maybe me not knowing how the associated refinement will actually come into play. I haven't yet actually handed the error handling to the refinement checker (i.e., i don't somehow mark functions that panic as making `f::no_panic()` return false).

crates/flux-refineck/src/checker.rs
```diff
@@ -939,25 +939,79 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             .deeply_normalize(&mut infcx.at(span))
             .with_span(span)?;
 
-        let mut at = infcx.at(span);
-
         if M::NAME == "refine" {
+            if let Some(callee_def_id) = callee_def_id
+                && genv.def_kind(callee_def_id).is_fn_like()
+            {
+                // does the callee require no panic?
```
ninehusky: Here I'm just looking up essentially: does the function `require no_panic` anywhere in the signature? It's kind of suspicious looking. I feel like I'm missing something.

tests/tests/neg/surface/no_panic05.rs
```diff
@@ -0,0 +1,20 @@
...
+    2
+}
+
+fn foo() {
+    // This is fine because `blah` never gets called.
+    // let blah = |x| panic!("blah");
+
+    // the line below: `foo` is in trouble because we've promised that `foo` can't panic.
+    // because of this promise, the argument to `bar` must be a function that cannot panic.
+    // there may be some sort of analysis that can prove this, but we'll need to merge #1387 to get that infrastructure.
```
ninehusky, line 11: the analysis i'm referring to here is some subset of (1) the syntactic no_panic checker and (2) `no_panic_if(cond)` automatically inferring that `might_panic` panics when `false`.

ninehusky:
@nilehmann here's my first stab at this. i have a deep feeling something is wrong (elaborated a bit on confusing bits in the comments). let's chat about this tomorrow! (today technically)

crates/flux-rustc-bridge/src/lowering.rs
```diff
@@ -824,7 +824,10 @@ impl<'tcx> Lower<'tcx> for rustc_ty::Ty<'tcx> {
             }
             rustc_ty::Closure(did, args) => {
                 let args = args.lower(tcx)?;
-                Ok(Ty::mk_closure(*did, args))
+                // NOTE (Andrew): Even if the closure inherits `no_panic` from its parent,
+                // we can't get that information from this alone. unless...
```
ninehusky: @nilehmann we're hitting a case right now where this is causing `pos/surface/no_panic05.rs` to fail. did you have a sense for how to deal with this? i'm not too looped into what this crate does.

crates/flux-refineck/src/checker.rs
```diff
@@ -939,25 +939,80 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                let parent = tcx.parent(callee_def_id);
+                if tcx.is_lang_item(parent, LangItem::Fn) {
+                    let assoc_refts = genv.assoc_refinements_of(parent).with_span(span)?;
+                    if assoc_refts.find(Symbol::intern("no_panic")).is_some() {
+                        callee_no_panic = true;
+                    }
+                }
+
+                infcx.at(span).check_pred(
+                    Expr::implies(no_panic, if callee_no_panic { Expr::tt() } else { Expr::ff() }),
```
nilehmann: The idea here is to use the two `no_panic` expressions (caller and callee) as is in the implication.

crates/flux-middle/src/queries.rs
```diff
@@ -865,11 +865,31 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
+
+                    let mut fn_sig = genv
                         .lower_fn_sig(def_id)?
                         .skip_binder()
                         .refine(&Refiner::default_for_item(genv, def_id)?)?
                         .hoist_input_binders();
+
+                    if is_fn_call {
+                        let mut inner_sig = fn_sig.clone().skip_binder();
+                        inner_sig.no_panic = Expr::tt();
```
nilehmann: This has to be `Self::no_panic()` instead of `true`.

crates/flux-refineck/src/checker.rs
```diff
@@ -939,25 +939,80 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+                //             );
+                //         }
+                //     }
+                // }
+            }
+
             let no_panic = self.fn_sig.no_panic();
 
-            if no_panic
+            if no_panic == Expr::tt()
```
ninehusky: todo: this is syntactically checking if the expr is true, which is wrong. let's just emit the implication no matter what of `self.fn_sig.no_panic()` implies `self.no_panic` or whatever the variables are (see nico comment below)

crates/flux-refineck/src/checker.rs
ninehusky: done

crates/flux-middle/src/builtin_assoc_refts.rs
```diff
@@ -27,6 +27,21 @@ impl<'tcx> GlobalEnv<'_, 'tcx> {
...
                 let mut map = UnordMap::default();
 
+                // if it's a function, automatically include a no panic assoc reft if
+                // the `requires` for that function includes it.
+                if let Some(fn_def_id) = tcx.lang_items().fn_trait() {
+                    map.insert(
+                        fn_def_id,
+                        AssocRefinements {
+                            items: List::from_arr([AssocReft::new(
+                                FluxDefId::new(fn_def_id, Symbol::intern("no_panic")),
```
nilehmann, line 37: You can add `no_panic` as a predefined symbol in `flux_syntax::symbols`

crates/flux-middle/src/builtin_assoc_refts.rs
```diff
@@ -92,6 +116,20 @@ impl<'tcx> GlobalEnv<'_, 'tcx> {
                 .bytes();
             let body = rty::Expr::constant(rty::Constant::from(size));
             rty::Lambda::bind_with_vars(body, List::empty(), rty::Sort::Int)
+        } else if alias_reft.assoc_id.name() == Symbol::intern("no_panic") {
```
nilehmann: We also need to check that `alias_reft.assoc_id.parent()` is `LangItem::Fn`

crates/flux-middle/src/builtin_assoc_refts.rs
```diff
@@ -27,6 +27,21 @@ impl<'tcx> GlobalEnv<'_, 'tcx> {
 
                 let mut map = UnordMap::default();
 
+                // if it's a function, automatically include a no panic assoc reft if
+                // the `requires` for that function includes it.
+                if let Some(fn_def_id) = tcx.lang_items().fn_trait() {
```
nilehmann: I think we should add the associated refinement to `FnOnce` such that both `Fn` and `FnMut` have access to it.

crates/flux-refineck/src/checker.rs
```diff
@@ -939,25 +939,30 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
-                        if fn_sig.no_panic() { Expr::tt() } else { Expr::ff() },
-                    ),
+                let mut callee_no_panic = fn_sig.no_panic();
+                let parent = tcx.parent(callee_def_id);
+                if tcx.is_lang_item(parent, LangItem::Fn) {
+                    let assoc_refts = genv.assoc_refinements_of(parent).with_span(span)?;
+                    if assoc_refts.find(Symbol::intern("no_panic")).is_some() {
+                        callee_no_panic = Expr::tt();
+                    }
+                }
```
nilehmann: This should not be necessary
```suggestion
```

crates/flux-middle/src/queries.rs
```diff
@@ -865,11 +865,32 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
-                    let fn_sig = genv
+                    let tcx = genv.tcx();
+                    // is the function Fn::call?
+                    let is_fn_call = tcx.opt_associated_item(def_id).is_some_and(|ai| {
+                        let is_callable_trait =
+                            ai.trait_container(tcx).is_some_and(|trait_def_id| {
+                                tcx.is_lang_item(trait_def_id, rustc_hir::LangItem::Fn)
+                            });
+
+                        let is_call_method = ai.name() == Symbol::intern("call");
```
nilehmann: Call is a predefined symbol. You can access it as `flux_syntax::symbols::sym::call`. By convention, we do `use flux_syntax::symbols::sym` and then use the symbol as `sym::call`.

crates/flux-middle/src/queries.rs
```diff
@@ -865,11 +865,32 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
...
+                    let mut fn_sig = genv
                         .lower_fn_sig(def_id)?
                         .skip_binder()
                         .refine(&Refiner::default_for_item(genv, def_id)?)?
                         .hoist_input_binders();
+
+                    if is_fn_call {
+                        let mut inner_sig = fn_sig.clone().skip_binder();
+                        inner_sig.no_panic =
+                            if self.no_panic(genv, def_id) { Expr::tt() } else { Expr::ff() };
```
nilehmann: As I mentioned before, the expression here should be `Self::no_panic()`, i.e., you have to construct the expression using `Expr::alias(...)`.

crates/flux-middle/src/builtin_assoc_refts.rs
ninehusky: I did this for `FnOnce`!

crates/flux-middle/src/builtin_assoc_refts.rs
ninehusky: I haven't pushed this yet -- I think what you're referring to is adding `no_panic` as a reserved symbol the same way that e.g. `requires` is.

Right now, adding this change makes it so that when we have something like `#[flux_rs::sig(.... requires F::no_panic())]`, it complains when parsing the `ident` for `no_panic` because that's a reserved keyword.

Do you think we should make that change here? Maybe a different PR because the syntax for how to specify no panic will still undergo some iteration?

nilehmann: I mean adding it as a symbol, not a keyword, i.e., to this list

https://github.com/flux-rs/flux/blob/ef6eaa1525da52a6182df458c80b25203ee08a51/crates/flux-syntax/src/symbols.rs#L23-L28

PR #1422, opened 2026-01-08
Add lean-check to CI

ranjitjhala:


.github/workflows/ci.yml
```diff
@@ -76,6 +76,46 @@ jobs:
           cd tock
           cargo flux -p cortexm -p rv32i
 
+  lean-demo:
+    runs-on: ubuntu-latest
```
nilehmann: Should we run this on Mac and Windows as well? Maybe at least Mac, since we know it works there already.

.github/workflows/ci.yml
```diff
@@ -76,6 +76,46 @@ jobs:
...
+          echo ~/.cargo/bin >> $GITHUB_PATH
+
+      - name: Clone flux-to-lean-demo
+        run: |
+          git clone https://github.com/flux-rs/flux-to-lean-demo
+
+      - name: Check
+        run: |
+          cd flux-to-lean-demo
+          ./runlean.sh
```
nilehmann: We could run the first `cargo flux` in a separate step to check that it doesn't produce any errors, but we would need to somehow ignore the missing proofs errors.

```bash
cargo flux
cp -r LeanUserProofs/* lean_proofs/LeanProofs/User/
cd lean_proofs
lake build
cd ..
cargo flux
```

.github/workflows/ci.yml
ranjitjhala: good idea, done!

PR #1425, opened 2026-01-09
Add spec for convert/TryInto

ranjitjhala:
I'm not quite sold on this spec, but here it is.

### DIFF
--- a/tests/tests/pos/vec/vec00.rs
+++ b/tests/tests/pos/vec/vec00.rs
@@ -55,6 +55,3 @@ pub fn pop2<T>(vec: &mut Vec<T>) -> Option<(T, T)> {
     let v2 = vec.pop().unwrap();
     Some((v1, v2))
 }
-
-#[spec(fn (x: &mut Vec<i32>[@n]) ensures x: Vec<i32>[#m])]
-fn silly(x: &mut Vec<i32>) {}


lib/flux-core/src/convert/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+trait TryInto<T>: Sized {
+    /// The type returned in the event of a conversion error.
+    // #[stable(feature = "try_from", since = "1.34.0")]
+    // type Error;
+
+    /// Performs the conversion.
+    // #[stable(feature = "try_from", since = "1.34.0")]
+    #[spec(fn(Self[@s]) -> Result<T[Self::into_val(s)], Self::Error>)]
+    fn try_into(self) -> Result<T, Self::Error>;
+}
```
nilehmann, line 8: This is a recurrent issue, but this specification prevents you from using any existing implementation that hasn't been specified yet. And as always, we have two options

1. Use a predicate instead of a value
    ```rust
    #[assoc(fn into_val(x: Self, into: T) -> bool { true })]
    trait TryInto<T>: Sized {
        #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v) }], Self::Error>)]
    }
    ```
    note that this doesn't prevent you from having the *singleton* spec in the impl
    ```rust
    #[assoc(fn into_val(x: MyNumber, y: int) -> bool { x.val == y })]
    impl TryInto<i32> for MyNumber {
        type Error = ();

        #[spec(fn(self: MyNumber[@s]) -> Result<i32[s.val], Self::Error>)]
        fn try_into(self) -> Result<i32, Self::Error> {
            Ok(self.0)
        }
    }
    ```
2. Use an uninterpreted function
    ```rust
    defs! { fn into_val<T>() -> T }
    #[assoc(fn into_val(x: Self) -> T { into_val() } )]
    trait TryInto<T>: Sized {
        #[spec(fn(Self[@s]) -> Result<T[Self::into_val(s)], Self::Error>)]
    }
    ```


1 is more obviously correct for me and more flexible; you may genuinely not have enough expressiveness to express the equality. 2 is maybe nicer to use. One may also consider turning pattern 2 into a feature.

ranjitjhala: IIRC (2) is a non-starter because you cannot prove that the UIF spec actually holds for any predefined type in any existing implementation so you have to basically `trust` it? (Or I suppose it would work as long as you never had to analyze the _code_ of any existing implementation...? I seem to recall lots of stuff breaking when we tried it for some other trait.)

I _thought_ (1) was a non-starter for this particular use-case because I needed to know "the value" but I was wrong, you can just shuffle the specs around to use constraints instead of exact indices... (i.e. instead of `ty[into_val(x)]` you write `ty{v:into_val(x, v)}`...) so have made that change!

ranjitjhala:
merging this lmk if you want to revert!

PR #1426, opened 2026-01-09
Close non-cut kvar solutions & small fixes

petros-marko:
Before sending kvar solutions to lean, we close over their free variables. Only do this for non-cut solutions since cut kvar solutions should already be closed(?).

Also, inherit a map from global variables to FluxDefIds from the `ExprEncodingContext` to be able to map global variables to their proper names.

~This PR does *not* yet address the problem of missing type parameters discussed [here](https://flux-rs.zulipchat.com/#narrow/channel/531635-lean-ffi/topic/Kvar.20Solutions.20from.20Fixpoint/near/567239318).~

It also turns any free-floating `Sort::Var`-s into `Sort::Int` and explicitly tells lean the sorts when encoding function applications in order to force it to infer the right types.

This all feels pretty hacky if I'm being honest.

Fixes the way we encode struct constructors in Lean, because it was broken.

### DIFF
--- a/crates/flux-infer/src/lean_encoding.rs
+++ b/crates/flux-infer/src/lean_encoding.rs
@@ -90,6 +90,7 @@ fn run_lean(genv: GlobalEnv, def_id: DefId) -> io::Result<()> {
     let proof_path = LeanFile::Proof(def_id).path(genv);
     let out = Command::new("lake")
         .arg("--quiet")
+        .arg("--log-level=error")
         .arg("lean")
         .arg(proof_path)
         .stdout(Stdio::piped())


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1295,11 +1308,45 @@ pub struct ExprEncodingCtxt<'genv, 'tcx> {
...
+        let closed_cut_solutions = cut_solutions
+            .into_iter()
+            .map(|(k, v)| (k, (vec![], v)))
+            .collect();
+
+        let mut closed_non_cut_solutions: HashMap<fixpoint::KVid, ClosedSolution> = HashMap::new();
+        for (kvid, solution) in non_cut_solutions {
+            let bound_vars: HashSet<&_> = solution.0.iter().map(|(var, _)| var).collect();
+            let vars = solution.1.free_vars();
+            let free_vars = vars.iter().filter(|var| {
```
nilehmann, line 1335: Iterating over the `HashSet` makes it non-deterministic. The same for the HashMaps for solutions (which were already wrong)

nilehmann (changes requested):

crates/flux-infer/src/fixpoint_encoding.rs
petros-marko: I've made changes to fix this (coming soon), but why does it matter here? The order in which we get the free varsity (or the kvar solutions for that matter) doesn't affect anything, so why does it need to be deterministic?

nilehmann: Doesn't the iteration order affect the order of the vector which affect how the parameters are generated in the lean file?

petros-marko: I think we only do the iteration nondeterministically once, afterwards we collect them in a vector and work from there, so the order should be fixed.

nilehmann: But the order of the vector could potentially be different in every run, so the generated lean file is different every time.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -702,23 +712,30 @@ where
...
         let sort_deps =
             SortDeps { opaque_sorts, data_decls: task.data_decls, adt_map: self.scx.adt_sorts };
+        let mut fun_decl_map = HashMap::new();
+        for (def_id, var) in self.ecx.const_env.fun_decl_map {
+            let fixpoint::Var::Global(idx, _) = var else {
+                bug!("non global var encountered for function")
+            };
+            fun_decl_map.insert(idx.index(), def_id);
+        }
+        let fun_deps = FunDeps { define_funs: task.define_funs, fun_decl_map };
```
nilehmann: I feel it's cleaner to recover this when we parse the solution instead of *patching* it during lean encoding.

lib/liquid-fixpoint/src/parser.rs
```diff
@@ -306,13 +307,29 @@ where
...
-                    .to_vec()
-                    .iter()
-                    .map(|sexp| self.parse_expr_possibly_nested(sexp))
-                    .try_collect()?;
-                Ok(Expr::App(Box::new(exp1), None, args))
+                let Sexp::List(inner) = &items[0] else {
+                    Err(ParseError::err("Expected list for inner func"))?
+                };
+                match &inner[1] {
+                    Sexp::Atom(Atom::S(s)) if s == "apply" => {
```
nilehmann, line 314: Is this required together with the int hack? I somehow thought that making everything int solved the inference issue. Or are there two unrelated issues?

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -257,7 +316,7 @@ impl BoundVar {
 pub enum Expr<T: Types> {
     Constant(Constant<T>),
     Var(T::Var),
-    App(Box<Self>, Option<Vec<Sort<T>>>, Vec<Self>),
+    App(Box<Self>, Option<Vec<Sort<T>>>, Vec<Self>, Option<Sort<T>>),
```
nilehmann, line 319: Without fully understanding how the elaborated solutions work, I feel that this is somewhat redundant, and the list of instantiated sorts should be enough. Are elaborated sort instantiations not a thing in fixpoint?

crates/flux-infer/src/fixpoint_encoding.rs
petros-marko: Sure, that can work.

lib/liquid-fixpoint/src/parser.rs
petros-marko: They're not unrelated, but I think it's required. Basically, enabling the flag in fixpoint that dumps the solutions that include sorts also makes applications of monomorphised functions explicit like this (I think), so we need to parse it.

lib/liquid-fixpoint/src/constraint.rs
petros-marko: When you say "list of instantiated sorts" you mean the type parameters of the function being applied right? 

That's what I was hoping we could get, but I couldn't figure it out. The solutions dumped by fixpoint don't tell us (as far as I can tell) what the type arguments should be.  

When applying a generic function (e.g. `T -> Vec<T> -> Int`) the sort we get back is with T already applied (e.g. `Vec<Int> -> Int`), so I don't think we can easily tell when we should have a type parameter and when we should not (unless we want to run some sort of unification since we do have the actual signatures of the functions).

crates/flux-infer/src/lean_format.rs
```diff
@@ -32,9 +33,24 @@ pub struct LeanCtxt<'a, 'genv, 'tcx> {
...
+impl<'a, 'genv, 'tcx> LeanCtxt<'a, 'genv, 'tcx> {
+    pub(crate) fn with_bool_mode(&self, bool_mode: BoolMode) -> Self {
+        LeanCtxt {
+            genv: self.genv,
+            pretty_var_map: self.pretty_var_map,
+            adt_map: self.adt_map,
+            fun_decl_map: self.fun_decl_map,
+            kvar_solutions: self.kvar_solutions,
+            bool_mode,
+        }
```
nilehmann: Does the update syntax work for this?
```suggestion
        LeanCtxt { bool_mode, ..*self }
```

PR #1429, opened 2026-01-16
Modify `cargo x test` to emit lean

ranjitjhala:
This PR lets you write

1. `cargo x test --emit-lean` to crank out the lean_proofs for each `pos` test in its own little sub-dir hanging off `tests/lean_bench/`

2. `python3 tools/run_lean.py tests/lean_bench/` to then run `lake build` in each sub-directory under `lean_bench`

(If nothing else, the above exercise shows various things that need to be added to the lean-translation...)


tests/tests/compiletest.rs
```diff
@@ -36,16 +39,23 @@ impl Args {
...
+                "--emit-lean" => {
+                    emit_lean = true;
+                }
                 _ => {}
             }
         }
+        // Also check for --emit-lean as a standalone arg (not followed by a value)
+        if env::args().any(|a| a == "--emit-lean") {
+            emit_lean = true;
+        }
```
nilehmann: We could require this to be passed as `--emit-lean true` to keep the parser simpler. But that also requires parsing the value, so maybe it's not simpler.

tests/tests/compiletest.rs
```diff
@@ -76,15 +86,109 @@ fn test_runner(_: &[&()]) {
 
     let path: PathBuf = ["tests", "pos"].iter().collect();
     if path.exists() {
-        config.mode = Mode::Ui;
-        config.src_base = path;
-        compiletest_rs::run_tests(&config);
+        if args.emit_lean {
```
nilehmann: Nit: i'd structure the code like

```rust
if args.emit_lean {
    run_tests_with_lean_emit(config, &path);
} else {
    // the old code
}
```

Then you can move the `Config`

tests/tests/compiletest.rs
```diff
@@ -76,15 +86,109 @@ fn test_runner(_: &[&()]) {
...
+        if let Some(stem) = rel_path.file_stem() {
+            lean_dir.push(stem);
+        }
+
+        // Create the directory if it doesn't exist
+        if let Err(e) = fs::create_dir_all(&lean_dir) {
+            eprintln!("Failed to create directory {:?}: {}", lean_dir, e);
+            failures.push(test_path.clone());
+            continue;
+        }
```
nilehmann: Shouldn't `flux` create the directory as needed?

tests/tests/compiletest.rs
```diff
@@ -76,15 +86,109 @@ fn test_runner(_: &[&()]) {
...
+    use std::panic;
+
+    let lean_bench_dir = PathBuf::from("lean_bench");
+    let mut failures: Vec<PathBuf> = Vec::new();
+    let mut total_tests = 0;
+
+    for entry in WalkDir::new(pos_path)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| e.path().extension().is_some_and(|ext| ext == "rs"))
```
nilehmann: I don't like this too much. It's a bit redundant to walk the directory _and_ use compiletest. Since we are already walking the dir, I'd call the `flux` binary directly instead of going through `compiletest`. I'd also not put this as part of the test binary and instead implement the logic directly in the `xtask` crate. But I don't know how much we want to evolve this *feature*, so maybe it's ok as is.

nilehmann (approved):
I don't like it, but there's also no much harm in it, so 🤷🏽‍♂️

ranjitjhala:
Another option is to NOT mess around with `compiletest.rs` and instead just do everything externally in a script -- is that preferable?

nilehmann:
> Another option is to NOT mess around with `compiletest.rs` and instead just do everything externally in a script -- is that preferable?

I like having scripts in Rust, I think that's easier to change and maintain. `xtask` is basically a script written in Rust. So I sort meant an external script but written in Rust inside xtask.

Issue #1430, opened 2026-01-16
Lean VC Gen


ranjitjhala:
Track the various glitches in the current lean-vc generation

- [ ] lean name clash
- [ ] constants
- [ ] struct constructor identifiers

Based on analysis of `lean_bench_log.txt` (437 tests, 295 passed, 142 failed).

## 1. Name Conflicts with Lean Builtins (~14 occurrences)

Generated function/type names that clash with Lean's standard library:

- `Or`, `Max`, `Min`, `Id`, `Add`, `Sub`, `Div`, `Empty`, `Append`, `Exists`, `Zero`

Examples:

```
error: LeanProofs/Flux/VC/Or.lean:4:4: `Or` has already been declared
error: LeanProofs/Flux/VC/Max.lean:4:4: `Max` has already been declared
error: LeanProofs/Flux/VC/Id.lean:4:4: `Id` has already been declared
error: LeanProofs/Flux/VC/Add.lean:4:4: `Add` has already been declared
error: LeanProofs/Flux/VC/Sub.lean:4:4: `Sub` has already been declared
error: LeanProofs/Flux/VC/Div.lean:4:4: `Div` has already been declared
error: LeanProofs/Flux/VC/Empty.lean:4:4: `Empty` has already been declared
error: LeanProofs/Flux/VC/Append.lean:4:4: `Append` has already been declared
error: LeanProofs/Flux/VC/Exists.lean:4:4: `Exists` has already been declared
error: LeanProofs/Flux/VC/Zero.lean:4:4: `Zero` has already been declared
error: LeanProofs/Flux/VC/Min.lean:4:4: `Min` has already been declared
```

**Affected tests:**

- abstract_refinements/test00 (`Max`)
- abstract_refinements/test03 (`Or`)
- detached/detach00 (`Id`)
- detached/detach_enum00 (`Zero`)
- detached/detach_macro (`Id`)
- enums/list00 (`Empty`, `Append`)
- surface/arg_syntax (`Exists`)
- surface/check_overflow05 (`Add`)
- surface/closure12 (`Add`)
- surface/constr00 (`Sub`)
- surface/explicit_variant_val (`Max`)
- surface/fndef01 (`Id`)
- surface/fndef02 (`Div`)
- surface/if-then-else (`Min`)

## 2. Unknown Identifier `c0`/`c1` (~12+ occurrences)

Constants aren't being properly declared or imported:

```
[...]

PR #1436, opened 2026-01-16
Fix mangled module names

ranjitjhala:
Fixes a bunch of the "mangled module name" errors in https://github.com/flux-rs/flux/issues/1430


ranjitjhala:
@nilehmann am merging this, but lmk if you hate the `regex` dependency and we'll revert!

PR #1440, opened 2026-01-19
Execute all ghost statements in exit basic blocks

nilehmann:


ranjitjhala:
bit surprising we hadn't hit this before tbh!

PR #1441, opened 2026-01-20
Add Definitions for Interpreted Constants

petros-marko:
For #1432.

Handle constants as follows:
- interpreted rust constants (e.g. INT_MAX) -> add quantifier to the constraint
- ~generic inequality relations -> add them to the prelude~
- ~everything else -> emit the same as an opaque function would~

### DIFF
--- a/crates/flux-infer/src/lean_encoding.rs
+++ b/crates/flux-infer/src/lean_encoding.rs
@@ -615,10 +615,6 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
                 file,
                 "def SmtMap_select {{ t0 t1 : Type }} [Inhabited t0] [BEq t0] [Inhabited t1] (m : SmtMap t0 t1) (k : t0) := m k"
             )?;
-            writeln!(file, "def gt {{ t0 : Type }} [LT t0] (l r : t0) : Prop := l > r")?;
-            writeln!(file, "def ge {{ t0 : Type }} [LE t0] (l r : t0) : Prop := l ≥ r")?;
-            writeln!(file, "def lt {{ t0 : Type }} [LT t0] (l r : t0) : Prop := l < r")?;
-            writeln!(file, "def le {{ t0 : Type }} [LE t0] (l r : t0) : Prop := l ≤ r")?;
         }
         Ok(())
     }


petros-marko:
> Fixes the build for all tests mentioned on the issue, but I'm not sure it's the right handling in every case. For example `issue-1025.rs` which uses trait methods it seems to me that the correct handling would be to actually emit an opaque function (e.g. using `axiom` instead of def). There are cases though (e.g. `primops00.rs`) where I think this is the right approach.
> 
> I just wanted to get this out there to get some feedback on whether this makes sense for how to handle the different cases.

```rust
const RUST_CNST : i64 = 10;

defs! {
    fn t() -> int {
        RUST_CNST
    }
}

#[sig(fn() -> i64[t()])]
fn test00() -> i64 {
    10
}
```

with ^ this example, we can make the generated function `t` dependent on the constant too, which means my approach of just adding a quantifier to the VC doesn't work. For interpreted constants I guess the right answer is to just emit them together with a definition then?

petros-marko:
> For example `issue-1025.rs` which uses trait methods it seems to me that the correct handling would be to actually emit an opaque function (e.g. using `axiom` instead of def)

This test is kind of contrived, because we define the two refts on the trait, but then provide no implementation and the test doesn't even use them (it just makes a random assertion); that is to say, adding some entirely opaque functions to correspond to these constants as I suggested above is definitely not correct. Maybe not the highest priority, but we should come up with a way to handle traits in this encoding. Typeclasses perform a similar function, but I don't know if it's an exact fit.

petros-marko:
> We should get the RUST_CNST example working - that's a likely common case. What does the fixpoint-hs encoding look like there?

```
(constant c1 int)  ;; rust const: RUST_CNST
(define_fun f$t$0 () int (c1))  ;; flux def: FluxId { parent: DefId(0:0 ~ primop_prop_let[17cd]), name: "t" }

(constraint
 (forall ((_$ int) ((= c1 10)))
  (tag ((= 10 (f$t$0 ))) "0")))
```

I think defining the constant the same way we define functions is the way to go.

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -518,6 +619,10 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
                 file,
                 "def SmtMap_select {{ t0 t1 : Type }} [Inhabited t0] [BEq t0] [Inhabited t1] (m : SmtMap t0 t1) (k : t0) := m k"
             )?;
+            writeln!(file, "def gt {{ t0 : Type }} [LT t0] (l r : t0) : Prop := l > r")?;
+            writeln!(file, "def ge {{ t0 : Type }} [LE t0] (l r : t0) : Prop := l ≥ r")?;
+            writeln!(file, "def lt {{ t0 : Type }} [LT t0] (l r : t0) : Prop := l < r")?;
+            writeln!(file, "def le {{ t0 : Type }} [LE t0] (l r : t0) : Prop := l ≤ r")?;
```
nilehmann: Does defining these functions make a difference in any of the tests? 

When there's a clear order for a sort, we emit a binary operator. These uifs are meant for cases where there's no natural order, so I'd expect there won't be a `[LT t0]` instance in Lean either. If emitting these definitions is making a difference, we may need to adjust the encoding to emit the interpreted operator such that it works for fixpoint as well.

crates/flux-infer/src/lean_format.rs
```diff
@@ -32,6 +32,7 @@ pub struct LeanCtxt<'a, 'genv, 'tcx> {
     pub genv: GlobalEnv<'genv, 'tcx>,
     pub pretty_var_map: &'a PrettyMap<LocalVar>,
     pub adt_map: &'a FxIndexSet<DefId>,
+    pub pretty_const_map: &'a FxIndexMap<GlobalVar, DefId>,
```
nilehmann: Can we add the `DefId` to the `Var::Global` as we do for functions instead?

crates/flux-infer/src/lean_encoding.rs
petros-marko: Doesn't make a difference as far as I can tell. It was just left over from when I tried to bring a broader solution. I'll remove them for now.

nilehmann (approved):

PR #1446, opened 2026-01-22
Add braces to nontrivial var-less canonical types (#1444)

### LINKED ISSUE #1444: Bug in implementation of `Pretty` for `CanonicalTy`
I'm getting the following when printing a type

```
[i32 | *; 0]
```

The type inside the array is missing the curly braces. The problem is in this implementation

https://github.com/flux-rs/flux/blob/main/crates/flux-middle/src/rty/canonicalize.rs#L428-L471

If `constr.pred().is_trivially_true()` is false and `vars.is_empty()` is true, then it'll print the constraint type without curly braces.

cole-k:
Fixes #1444

### DIFF
--- a/crates/flux-middle/src/rty/canonicalize.rs
+++ b/crates/flux-middle/src/rty/canonicalize.rs
@@ -443,7 +443,7 @@ mod pretty {
                             if constr.pred().is_trivially_true() {
                                 w!(cx, f_body, "{:?}", &constr.ty)
                             } else {
-                                w!(cx, f_body, "{:?} | {:?}", &constr.ty, &constr.pred)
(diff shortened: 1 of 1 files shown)


crates/flux-middle/src/rty/canonicalize.rs
```diff
@@ -443,7 +443,7 @@ mod pretty {
                             if constr.pred().is_trivially_true() {
                                 w!(cx, f_body, "{:?}", &constr.ty)
                             } else {
-                                w!(cx, f_body, "{:?} | {:?}", &constr.ty, &constr.pred)
+                                w!(cx, f_body, "{{{:?} | {:?}}}", &constr.ty, &constr.pred)
```
nilehmann, line 446: Won't this give you double braces in case `!vars.is_empty()`?

PR #1462, opened 2026-02-04
Basic support for checking raw pointer read/write 

ranjitjhala:
1. Track an `int` index with the pointer, to track how much is "valid" to write,
2. Check at deref that the associated index is (strictly) positive,
3. Defer tracking provenance etc. to later, perhaps once we have more experience with what is needed.

e.g. the following now works, with an error reported without the preconditions.

```rust
#[flux::opts(check_raw_pointer = "checked")]
#[flux::spec(fn (ptr: *mut{v: v > 0} T, value: T))]
fn write<T>(ptr: *mut T, value: T) {
    unsafe {
        *ptr = value;
    }
}

#[flux::opts(check_raw_pointer = "checked")]
#[flux::spec(fn (ptr: *mut{v: v > 0} T, value: T))]
fn write<T>(ptr: *mut T, value: T) {
    unsafe {
        *ptr = value;
    }
}
```

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -1401,9 +1401,6 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 Ok(ty)
             }
             Rvalue::RawPtr(kind, place) => {
-                // TODO:RAW-PTR
-                env.lookup_place(&mut infcx.at(stmt_span), place)
-                    .with_span(stmt_span)?;
                 // ignore any refinements on the type stored at place
                 let ty = &env.lookup_rust_ty(genv, place).with_span(stmt_span)?;
                 let ty = self.refine_default(ty).with_span(stmt_span)?;
--- a/crates/flux-refineck/src/primops.rs
+++ b/crates/flux-refineck/src/primops.rs
@@ -49,11 +49,6 @@ pub(crate) struct Pre {
     pub pred: Expr,
 }
 
-fn is_cmp(op: &mir::BinOp) -> bool {
-    use mir::BinOp::*;
-    matches!(op, Eq | Ne | Lt | Le | Gt | Ge)
-}
-
 pub(crate) fn match_bin_op(
     op: mir::BinOp,
     bty1: &BaseTy,
@@ -68,12 +63,7 @@ pub(crate) fn match_bin_op(
         OverflowMode::None => &OVERFLOW_NONE_BIN_OPS,
         OverflowMode::StrictUnder => &OVERFLOW_STRICT_UNDER_BIN_OPS,
     };
-    // TODO(HACK): for tests/tests/pos/surface/binop.rs:133:5
-    if (matches!(bty1, BaseTy::RawPtr(..)) || matches!(bty2, BaseTy::RawPtr(..))) && is_cmp(&op) {
-        MatchedRule { precondition: None, output_type: rty::Ty::bool() }
-    } else {
-        table.match_inputs(&op, [(bty1.clone(), idx1.clone()), (bty2.clone(), idx2.clone())])
-    }
+    table.match_inputs(&op, [(bty1.clone(), idx1.clone()), (bty2.clone(), idx2.clone())])
 }
 
 pub(crate) fn match_un_op(
@@ -460,26 +450,29 @@ fn mk_bit_xor_rules() -> RuleMatcher<2> {
     }
 }
 
+// the (a: T, b: S) case ensures the last case is "unreachable", hence the "allow"
+#[allow(unreachable_code)]
 /// `a == b`
 fn mk_eq_rules() -> RuleMatcher<2> {
     primop_rules! {
         fn(a: T, b: T) -> bool[E::eq(a, b)]
         if T.is_integral() || T.is_bool() || T.is_char() || T.is_str()
-
-        fn(a: T, b: T) -> bool
+        fn(a: T, b: S) -> bool
     }
 }
 
+#[allow(unreachable_code)]
 /// `a != b`
 fn mk_ne_rules() -> RuleMatcher<2> {
     primop_rules! {
         fn(a: T, b: T) -> bool[E::ne(a, b)]
         if T.is_integral() || T.is_bool()
 
-        fn(a: T, b: T) -> bool
+        fn(a: T, b: S) -> bool
     }
 }
 
+#[allow(unreachable_code)]
 /// `a <= b`
 fn mk_le_rules() -> RuleMatcher<2> {
     primop_rules! {
@@ -488,10 +481,11 @@ fn mk_le_rules() -> RuleMatcher<2> {
 
         fn(a: bool, b: bool) -> bool[E::implies(a, b)]
 
-        fn(a: T, b: T) -> bool
+        fn(a: T, b: S) -> bool
     }
 }
 
+#[allow(unreachable_code)]
 /// `a >= b`
 fn mk_ge_rules() -> RuleMatcher<2> {
     primop_rules! {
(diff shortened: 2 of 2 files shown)


ranjitjhala:
@nilehmann ping -- I'd like to merge this one in to do more stuff on core.

crates/flux-desugar/src/resolver.rs
```diff
@@ -157,6 +157,8 @@ impl<'genv, 'tcx> CrateResolver<'genv, 'tcx> {
         );
         self.func_decls
             .insert(Symbol::intern("cast"), fhir::SpecFuncKind::Cast);
+        self.func_decls
+            .insert(Symbol::intern("ptr_size"), fhir::SpecFuncKind::PtrSize);
```
nilehmann: @copilot add `ptr_size` to the list of symbols (not keywords) in `crates/flux-syntax/src/symbols.rs` and use the predefined symbol here as `sym::ptr_size` instead of interning.

crates/flux-refineck/src/checker.rs
```diff
@@ -1742,6 +1747,16 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
 
+    fn check_lookup_place(
+        &mut self,
+        infcx: &mut InferCtxt<'_, 'genv, 'tcx>,
+        env: &mut TypeEnv,
+        span: Span,
+        place: &Place,
+    ) -> InferResult<Ty> {
+        env.lookup_place(&mut infcx.at(span), place)
+    }
```
nilehmann: @copilot, this function is not necessary anymore. Inline it and then remove it.

crates/flux-infer/src/infer.rs
```diff
@@ -79,6 +79,7 @@ pub enum ConstrReason {
     Goto(BasicBlock),
     Overflow,
     Underflow,
+    RawDeref,
```
nilehmann: @copilot, remove this variant and update the code accordingly.

crates/flux-refineck/src/lib.rs
```diff
@@ -326,6 +327,13 @@ mod errors {
         pub span: Span,
     }
 
+    #[derive(Diagnostic)]
+    #[diag(refineck_raw_deref_error, code = E0999)]
+    pub struct RawDerefError {
+        #[primary_span]
+        pub span: Span,
+    }
```
nilehmann: @copilot, removing `ConstrReason::RawDeref` makes this struct unnecesary. Please remove it.

crates/flux-refineck/locales/en-US.ftl
```diff
@@ -42,6 +42,9 @@ refineck_overflow_error =
 refineck_underflow_error =
     arithmetic operation may underflow
 
+refineck_raw_deref_error =
+    raw pointer dereference may be unsafe
+
```
nilehmann: @copilot remove this message as is not longer needed.

crates/flux-infer/src/infer.rs
```diff
@@ -933,6 +933,16 @@ impl<'a, E: LocEnv> Sub<'a, E> {
...
             }
             (BaseTy::Slice(ty_a), BaseTy::Slice(ty_b)) => self.tys(infcx, ty_a, ty_b),
+
+            (BaseTy::RawPtr(ty_a, mut_a), BaseTy::RawPtr(ty_b, mut_b)) => {
+                debug_assert_eq!(mut_a, mut_b);
+                self.tys(infcx, ty_a, ty_b)?;
+                if matches!(mut_a, Mutability::Mut) {
+                    self.tys(infcx, ty_b, ty_a)?;
+                }
+                Ok(())
```
nilehmann, line 943: This makes me nervous. Did you add it because `as_mut_ptr` returns whatever refined type was in the input, and you wanted to keep using it the pointer at that type?

crates/flux-middle/src/rty/expr.rs
```diff
@@ -756,6 +756,8 @@ pub enum SpecFuncKind {
     Thy(liquid_fixpoint::ThyFunc),
     /// User-defined function. This can be either a function with a body or a UIF.
     Def(FluxDefId),
+    /// Built-in UIF for pointer size: `RawPtr -> Int`
+    PtrSize,
```
nilehmann: Why is this not an `InternalFunc`?

crates/flux-refineck/src/primops.rs
```diff
@@ -63,7 +68,12 @@ pub(crate) fn match_bin_op(
         OverflowMode::None => &OVERFLOW_NONE_BIN_OPS,
         OverflowMode::StrictUnder => &OVERFLOW_STRICT_UNDER_BIN_OPS,
     };
-    table.match_inputs(&op, [(bty1.clone(), idx1.clone()), (bty2.clone(), idx2.clone())])
+    // TODO(HACK): for tests/tests/pos/surface/binop.rs:133:5
+    if (matches!(bty1, BaseTy::RawPtr(..)) || matches!(bty2, BaseTy::RawPtr(..))) && is_cmp(&op) {
+        MatchedRule { precondition: None, output_type: rty::Ty::bool() }
+    } else {
+        table.match_inputs(&op, [(bty1.clone(), idx1.clone()), (bty2.clone(), idx2.clone())])
+    }
```
nilehmann: You can make the fallback rule for each primop more general like so:

```
        fn(a: T, b: S) -> bool
```

crates/flux-refineck/src/checker.rs
```diff
@@ -857,7 +857,7 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
         let obligations = infcx
             .at(span)
             .ensure_resolved_evars(|infcx| {
-                let ret_place_ty = env.lookup_place(infcx, Place::RETURN)?;
+                let ret_place_ty = env.lookup_place(&mut infcx.at(span), Place::RETURN)?;
```
```suggestion
                let ret_place_ty = env.lookup_place(infcx, Place::RETURN)?;
```

crates/flux-refineck/src/checker.rs
```diff
@@ -1400,6 +1401,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 Ok(ty)
             }
             Rvalue::RawPtr(kind, place) => {
+                // TODO:RAW-PTR
+                env.lookup_place(&mut infcx.at(stmt_span), place)
+                    .with_span(stmt_span)?;
```
```suggestion
```

crates/flux-middle/src/rty/expr.rs
ranjitjhala: indeed, because I think I'd forgotten those exist! fixed below.

crates/flux-infer/src/infer.rs
ranjitjhala: Sort of -- I was just trying to mirror the story for `&mut T` here? 

The option is to _remove_ subtyping between raw-ptr, but that seemed to be worse, no?

ranjitjhala: Or is your point that we IGNORE the refinements for stuff read through raw-ptr so it doesn't matter e.g. 
https://github.com/flux-rs/flux/blob/9c27760a7bbbda9c95bceb21f831334bee11408e/crates/flux-refineck/src/checker.rs#L1403-L1404

crates/flux-refineck/src/primops.rs
ranjitjhala: Ok did that but had to add some "allow(unreachable)" stuff.

crates/flux-refineck/src/checker.rs
ranjitjhala: done

ranjitjhala:
@nilehmann I did the changes, except the ptr-subtyping, see discussion above.

crates/flux-infer/src/infer.rs
nilehmann: I don't know what is my point haha, but we dont have a good story for this. I mean once you go to a raw pointer all bets are off, so the pointee type is not very meaningful. Unless we restrict the use of raw pointers somehow

ranjitjhala: I don't have a strong feeling either way, the main bit in this PR is just tracking the index on the `ptr` not so much the data inside, so I can just remove all subtyping on the payload. OTOH it seems no worse than the situation right now (i.e. no "more unsound") so we could keep it in too. Whichever is quickest to merge this PR imo!

nilehmann: Yeah, you are right. let's merge as is.

PR #1464, opened 2026-02-06
inline box_args()

ranjitjhala:


crates/flux-middle/src/rty/canonicalize.rs
```diff
@@ -204,12 +204,16 @@ impl<D: HoisterDelegate> TypeFolder for Hoister<D> {
...
+                // HACK(RJ): inlining box_args() to work around panic in `core::alloc`
+                if let [GenericArg::Ty(boxed), GenericArg::Ty(alloc)] = &args[..] {
+                    let args = List::from_arr([
+                        GenericArg::Ty(boxed.fold_with(self)),
+                        GenericArg::Ty(alloc.clone()),
+                    ]);
+                    BaseTy::Adt(adt_def.clone(), args)
+                } else {
+                    BaseTy::Adt(adt_def.clone(), args.clone())
+                }
```
petros-marko: This (as far as I understand) looks like what @cole-k  had suggested. When I talked with @nilehmann about that, he said this wasn't the proper fix; am I missing some difference that means this is actually fine or are we just saying we're OK with this as a temporary hack?

ranjitjhala: I don't know what the correct fix is -- will have to understand @nilehmann 's comment on zulip closely for that perhaps -- but its literally inlining the old function and not crashing so it seems like a strict improvement on what is currently there?

nilehmann: This is just kicking the can down the road. There are other places where we expect boxes to have `GenericArg::Ty` so it'll eventually fail somewhere else. So the issue is having a `GenericArg::Base` in the first place. This is ok as a temporary patch though

ranjitjhala:
> @ranjitjhala can you check if the alternative implementation in #1483 works?

Ok @nilehmann so to check this I should try to 

1. *locally* merge your PR #1483 into my `box-args-hack` branch yes? OR
2. Just run your PR #1483 standalone?

ranjitjhala:
Yes, that PR seems to be working! (but lets just keep this around in case I've mucked something up and am using the wrong branch or something silly...?)

ranjitjhala:
Can also close, yes @nilehmann ?

nilehmann:
Yes, superseded by #1483

PR #1481, opened 2026-02-11
Static Specs

ranjitjhala:
Adds basic support for `static_spec`. 

The *unsound* part is that they are not (yet) actually *checked* against the definition of the static.

Lets us check stuff like

```rust
#[flux::spec([u32{v:v < 100}; 3])]
static HOG: [u32; 3] = [67, 67, 67];

#[flux::sig(fn () -> u32{v:v < 300})]
pub fn test_hog() -> u32 {
    HOG[0] + HOG[1] + HOG[2]
}
``` 

~~will return to checking the def in due course.~~

### DIFF
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -454,7 +454,11 @@ fn generics_of(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::Gener
                 .ok_or_else(|| query_bug!(def_id.local_id(), "no generics for {def_id:?}"))?;
             conv::conv_generics(genv, generics, def_id, is_trait)
         }
-        DefKind::OpaqueTy | DefKind::Closure | DefKind::TraitAlias | DefKind::Ctor(..) | DefKind::Static { .. } => {
+        DefKind::OpaqueTy
+        | DefKind::Closure
+        | DefKind::TraitAlias
+        | DefKind::Ctor(..)
+        | DefKind::Static { .. } => {
             let rustc_generics = genv.lower_generics_of(def_id);
             refining::refine_generics(genv, def_id.resolved_id(), &rustc_generics)
         }
--- a/crates/flux-middle/src/metrics.rs
+++ b/crates/flux-middle/src/metrics.rs
@@ -97,7 +97,7 @@ pub enum TimingKind {
     /// Total time taken to run the complete Flux analysis on the crate
     Total,
     /// Time taken to check the body of a function
-    CheckFn(LocalDefId),
+    CheckBody(LocalDefId),
     /// Time taken to run a single fixpoint query
     FixpointQuery(DefId, FixpointQueryKind),
 }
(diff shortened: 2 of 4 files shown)


crates/flux-desugar/src/desugar/lift.rs
```diff
@@ -543,12 +544,20 @@ impl<'genv> RustItemCtxt<'_, 'genv, '_> {
...
     }
 
+    pub(crate) fn lift_foreign_static(&mut self) -> fhir::Item<'genv> {
+        fhir::Item {
+            owner_id: self.owner,
+            generics: self.lift_generics_inner(hir::Generics::empty()),
+            kind: fhir::ItemKind::Static(None),
+        }
+    }
+
```
nilehmann: This function is not being used

crates/flux-driver/src/collector/mod.rs
```diff
@@ -527,6 +549,9 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
             ("constant", hir::AttrArgs::Delimited(dargs)) => {
                 self.parse(dargs, ParseSess::parse_constant_info, FluxAttrKind::Constant)?
             }
+            ("static_spec", hir::AttrArgs::Delimited(dargs)) => {
```
nilehmann: Why not `spec`? We could peek the parser based on the `DefKind`

tests/tests/neg/detached/detach_static00.rs
```diff
@@ -0,0 +1,29 @@
+mod bob {
+
+    static BOB: usize = 42;
```
nilehmann: Is this intentionally 42?

crates/flux-desugar/src/lib.rs
```diff
@@ -91,9 +91,18 @@ pub fn desugar<'genv>(
...
+                    let item =
+                        cx.with_rust_item_ctxt(owner_id, None, |cx| Ok(cx.lift_foreign_static()))?;
+                    fhir::Node::Item(genv.alloc(item))
+                }
+                _ => {
+                    let item = cx
+                        .with_rust_item_ctxt(owner_id, None, |cx| cx.lift_foreign_item(*foreign))?;
+                    fhir::Node::ForeignItem(genv.alloc(item))
+                }
+            }
```
nilehmann: Move this match inside `lift_foreign_item`

crates/flux-driver/src/collector/mod.rs
```diff
@@ -156,6 +156,9 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
 
                 self.collect_constant(owner_id, attrs)?;
             }
+            ItemKind::Static(..) => {
```
nilehmann: Should we report an error for mutable statics with specs?

crates/flux-refineck/src/checker.rs
```diff
@@ -1844,7 +1844,7 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             let param_env = tcx.param_env(self.checker_id.root_id());
             let typing_env = infcx.region_infcx.typing_env(param_env);
             if let Ok(val) = tcx.const_eval_resolve(typing_env, uneval, constant.span) {
-                return Ok(self.check_const_val(val, ty));
+                return Ok(self.check_const_val(val, ty)?);
```
```suggestion
                return self.check_const_val(val, ty);
```

PR #1488, opened 2026-02-13
Do not mark all non-queried items as ignored

petros-marko:
Instead look for unsupported errors, and make those errors reported at the use-site.

This fixes a bunch of use-of-ignored-item errors that cropped up in VecDeque after removing the encode_and_save_metadata early exit hack.

### DIFF
--- a/crates/flux-middle/locales/en-US.ftl
+++ b/crates/flux-middle/locales/en-US.ftl
@@ -17,7 +17,7 @@ middle_query_ignored_item =
     use of ignored item
 
 middle_query_not_included_item =
-    use of item that was not included when checking external crate
+    use of item defined in external crate, but was not included when checking that crate
 
 # Query Errors reported at use site
 
@@ -30,7 +30,7 @@ middle_query_ignored_at =
     .label = help: try ignoring or trusting this code
 
 middle_query_not_included_at =
-    use of {$kind} `{$name}` that was not included
+    use of {$kind} `{$name}` that was not included when checking external crate
     .label = help: try including the file or module where `{$name}` is defined
     .help = when checking external crate, include the file or module where the excluded item is defined
 


crates/flux-middle/src/global_env.rs
```diff
@@ -79,11 +79,14 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
         key: K,
         query: impl FnOnce(Self, K) -> QueryResult<R>,
     ) -> QueryResult<R> {
-        if !self.inner.queries.queried(key.def_id()) {
-            return Err(QueryErr::Ignored { def_id: key.def_id() });
+        match query(self, key) {
```
nilehmann: This is not going to change the behavior. At this point, the error has already been emitted. If you run `cargo flux -p core` with this PR, you still get a bunch `use of unsupported ...`

petros-marko: I did run it; I'm not getting any errors.

nilehmann: Running this PR on verify-rust-std/sync-2025-11-25

image

petros-marko: Ah, that one was already marked as `ignore` on my branch.

Then I'm not sure where we should intercept the error to avoid having this happen.

nilehmann: If you pass `-Ztrack-diagnostics`, then errors show where they were emitted (e.g., `RUSTFLAGS=-Ztrack-diagnostics cargo flux -p core`). Most of the errors we get in core are emitted in `crates/flux-fhir-analysis/src/wf/mod.rs:101:14`

ranjitjhala:
final version 

Screenshot 2026-02-19 at 9 15 57 AM

PR #1497, opened 2026-02-18
Escaping var hack

ranjitjhala:
Isolates the stuff needed to work around the "escaping-vars" errors in `core` (which cause a crash inside rustc) from #1464


nilehmann:
> @nilehmann ping -- are you working on this?

No, but I don't like the hack because it potentially leaves stuff unnormalized. I believe we can more fundamentally fix the issue by _erasing_ more lifetimes.

PR #1503, opened 2026-02-20
Lean cache

ranjitjhala:
It is rather slow to run lean on each query; so cache.

0. Add a `lean_status` to the `FixpointResult` (should probably rename to `VcQueryResult` 
1. A cache entry has `lean_status = Valid` iff its hash matches the current task AND a                                                                                                              prior run successfully ran check_proof for it
2. If the task hash changed, `run_task_with_cache` will overwrite the entry with default 
`lean_status = Invalid`, so stale Valid is never visible.
3. So in `check_crate` we can safely skip check_proof by checking with a key-lookup ONLY (we don't even _have_ the task-hash there...) as `execute_lean_query` already enforces hash correctness in 2.

### DIFF
--- a/crates/flux-infer/src/lean_encoding.rs
+++ b/crates/flux-infer/src/lean_encoding.rs
@@ -349,7 +349,6 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
         self.generate_lib_if_absent()?;
         self.generate_vc_file()?;
         self.generate_proof_if_absent()?;
-        // self.record_proof()?;
         Ok(())
     }
 


crates/flux-infer/src/lean_encoding.rs
```diff
@@ -349,7 +349,7 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
         self.generate_lib_if_absent()?;
         self.generate_vc_file()?;
         self.generate_proof_if_absent()?;
-        self.record_proof()?;
+        // self.record_proof()?;
```
petros-marko:
```suggestion
```

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -765,3 +751,46 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
         Ok(())
     }
 }
+
+fn hyperlink_proof(genv: GlobalEnv, def_id: MaybeExternId) {
+    let vc_name = vc_name(genv, def_id.resolved_id());
+    let proof_name = format!("{vc_name}_proof");
+    let path = LeanFile::Proof(def_id.resolved_id()).path(genv);
+    if let Some(span) = genv.proven_externally(def_id.local_id()) {
+        let dst_span = SpanTrace::from_path(&path, 3, 5, proof_name.len());
```
petros-marko, line 759: What are the 3 and 5 magic numbers here?

ranjitjhala: Probably line/column offsets within the file?

PR #1504, opened 2026-02-20
Add flag for `dump_constraint` to FluxMetadata

ranjitjhala:
bit tedious to have to faff around with the flags otherwise.


nilehmann:
I don't think this flag should be exposed in `Cargo.toml` because it's not about the crate but about a single invocation of `cargo flux`, e.g., you shouldn't commit with the flag enabled. The proper thing to do here is to have something similar to [`config.toml`](https://doc.rust-lang.org/cargo/reference/config.html) with a different set of flags.

In the meantime, you can `export FLUXFLAGS=-Fdump-constraint` if you don't want to type that every time.

ranjitjhala:
Ha, yes, that makes sense (on both counts!)

PR #1509, opened 2026-02-24
Fix support for poly-opaque-sorts

ranjitjhala:


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -154,7 +160,8 @@ pub mod fixpoint {
     pub enum DataSort {
         Tuple(usize),
         Adt(AdtId),
-        User(FluxDefId),
+        // User(OpaqueId, FluxDefId),
```
petros-marko: Nit: the comment can go?

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -402,8 +410,14 @@ impl SortEncodingCtxt {
         self.tuples.insert(arity);
     }
 
-    pub fn declare_opaque_sort(&mut self, def_id: FluxDefId) {
-        self.user_sorts.insert(def_id);
+    pub fn declare_opaque_sort(&mut self, def_id: FluxDefId) -> OpaqueId {
+        if let Some(idx) = self.opaque_sorts.get_index_of(&def_id) {
+            OpaqueId::from_usize(idx)
+        } else {
+            let opaque_id = OpaqueId::from_usize(self.adt_sorts.len());
```
petros-marko: This should be `self.opaque_sorts.len()` not `self.adt_sorts.len()` right?

crates/flux-infer/src/lean_encoding.rs
```diff
@@ -284,14 +284,20 @@ impl<'genv, 'tcx> LeanEncoder<'genv, 'tcx> {
...
         }
     }
 
     fn datasort_name(&self, sort: &fixpoint::DataSort) -> String {
-        let name = format!("{}", WithLeanCtxt { item: sort, cx: &self.lean_cx() });
-        snake_case_to_pascal_case(&name)
+        if let fixpoint::DataSort::User(opaque_id) = sort {
+            let (flux_id, _) = &self.sort_deps.opaque_sorts[opaque_id.as_usize()];
+            return snake_case_to_pascal_case(flux_id.name().as_str());
+        } else {
```
petros-marko: Why do we need a special case for opaque sorts here? The other case calls format on `sort` again, which should have a case for opaque sorts already.

petros-marko (approved):
I still think we could get away without introducing `OpaqueId` as a new thing; `FluxDefId` should be OK since these are global defs. If you and/or @nilehmann think this is the better way to do it, though, the changes to make this work LGTM.

crates/flux-infer/src/lean_encoding.rs
ranjitjhala: hmm good catch let me check! .. I was getting some infinite loop but maybe that's now broken by the other thing..

ranjitjhala: good catch! fixed.

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: yup! fixed.

petros-marko:
> Hmm I think you may be right; the catch is when you want to slurp the result from fixpoint back into flux-land (to recover the `FluxDefId` it is convenient to work with a plain `OpaqueId` ... ?

Yeah, fair enough. We could do something hacky like encode the whole def path in the name we send to fixpoint, but that's not a nicer solution than the OpaqueId.

crates/flux-infer/src/lean_format.rs
```diff
@@ -105,7 +107,10 @@ impl LeanFmt for DataField {
 impl LeanFmt for DataSort {
     fn lean_fmt(&self, f: &mut fmt::Formatter, cx: &LeanCtxt) -> std::fmt::Result {
         match self {
-            DataSort::User(def_id) => write!(f, "{}", def_id.name()),
+            DataSort::User(opaque_id) => {
+                let (def_id, _) = cx.opaque_adt_map.get(opaque_id.as_usize()).unwrap();
+                write!(f, "{}", snake_case_to_pascal_case(def_id.name().as_str()))
```
nilehmann, line 112: Not an immediate problem, but this will cause name clashes once we support the following

```rust
mod a {
    defs! {
        opaque sort MySort;
    }
}
mod b {
    defs! {
        opaque sort MySort;
    }
}
```
We can internally represent this as two different sorts (that's the idea behind `FluxDefId`), but the name resolution part is not implemented. I would like to fix that at some point so we can have proper namespacing for `defs!`.

ranjitjhala:
Hmm I thought that the Name was ultimately going through that pascalcase
business and so getting fully qualified? Let me check …

- Ranjit.

crates/flux-infer/src/lean_format.rs
nilehmann: > Hmm I thought that the Name was ultimately going through that pascalcase
business and so getting fully qualified? Let me check …

But you are only passing the `name` part of the `FluxDefId` and not the `parent`, which is what disambiguates between two sorts with the same name but defined in different places.

It's currently impossible to have two `FluxDefId` with the same `name` and different `parent` because of the resolution part, but it'd be nice to fix that.

ranjitjhala: Got it! Agree 100%

PR #1526, opened 2026-03-02
Merge `no_panic_if(true)`, `no_panic` behavior

### LINKED ISSUE #1521: Treat `no_panic` as `no_panic_if(true)`
Morally, a `#[flux_rs::no_panic]` should be equivalent to a `no_panic_if(true)`. In our (my 😛) implementation, however, we're still treating `no_panic` (bool) and `no_panic_if` (`Expr`) as two separate things, which makes it hard to keep track of when to convert one to the other.

For example, when a closure inherits the `no_panic`ness from its parent, it _only_ looks at the first parent who's marked with a `no_panic`; if a parent function is annotated `no_panic_if(phi)`, the closure's `no_panic` spec will be `false`.

ninehusky:
Closes #1521.

### DIFF
--- a/crates/flux-middle/src/global_env.rs
+++ b/crates/flux-middle/src/global_env.rs
@@ -27,7 +27,7 @@ use crate::{
     queries::{DispatchKey, Providers, Queries, QueryErr, QueryResult},
     query_bug,
     rty::{
-        self, QualifierKind,
+        self, Expr, QualifierKind,
         refining::{Refine as _, Refiner},
     },
 };
@@ -459,6 +459,12 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
     /// Whether the function is marked with `#[flux::no_panic]`
     pub fn no_panic(self, def_id: impl IntoQueryParam<DefId>) -> bool {
         self.inner.queries.no_panic(self, def_id.into_query_param())
+    }
+
+    pub fn no_panic_if(self, def_id: impl IntoQueryParam<DefId>) -> Expr {
+        self.inner
+            .queries
+            .no_panic_if(self, def_id.into_query_param())
     }
 
     pub fn is_box(&self, res: fhir::Res) -> bool {
--- a/crates/flux-middle/src/global_env.rs
+++ b/crates/flux-middle/src/global_env.rs
@@ -27,7 +27,7 @@ use crate::{
     queries::{DispatchKey, Providers, Queries, QueryErr, QueryResult},
     query_bug,
     rty::{
-        self, Expr, QualifierKind,
+        self, QualifierKind,
         refining::{Refine as _, Refiner},
     },
 };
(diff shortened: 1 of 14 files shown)


ninehusky:
This is a pretty messy PR so far, but there are a lot of changes I had to make.

Right now, I'm in the middle of reorganizing the plumbing of `genv::no_panic` so that when it's doing the parent-walk, it returns when it finds _either_ a parent item who's annotated with `flux::no_panic`, or an `FnSig`, in which case we can just return the `fn_sig.no_panic()`. While doing this generalization of `no_panic` to return an `Expr` instead of a boolean, I ran into a query cycle where `conv_fn_sig` is calling `genv::no_panic`, and vice versa. 

So I think I'm violating some scheduling invariant. Maybe we need to move `no_panic_if` so that it's an attribute again, like with the regular `no_panic`? I don't remember all the details why the two annotations are handled so differently.

ninehusky:
Okay, I think getting a basic version that doesn't die is alright. Some todos to continue with:
- [ ] support multi-closure inheritance (see the "grandchild" test of the last commit)
- [ ] add the condition Nico told me about the two binders of the `FnSig`
- [ ] write test to see if more complex `no_panic_if` expressions blow the system up, e.g., `no_panic_if(T::foo_no_panic)`
- [ ] do a merge with the syntactic-inference PR as a stress test. ideally the `alarm.rs` closure false positive should shut up.

crates/flux-middle/src/global_env.rs
```diff
@@ -439,6 +439,23 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
             .variants_of(self, def_id.into_query_param())
     }
 
+    pub fn parent_of_closure(self, def_id: DefId) -> DefId {
```
nilehmann, line 442: I believe you can reuse `tycx.typeck_root_def_id`

crates/flux-refineck/src/checker.rs
```diff
@@ -1335,7 +1336,26 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+            bug!("missing parent fn_sig for {closure_id:?}");
+        };
+        assert!(
+            !parent_sig
+                .skip_binder_ref()
+                .skip_binder_ref()
+                .no_panic()
+                .has_escaping_bvars()
+        );
+        let no_panic = parent_sig.skip_binder().skip_binder().no_panic();
```
nilehmann, line 1357: `check_closure_body` is only being called from `check_rvalue_closure`, where you compute the same thing again. Take `no_panic` as an argument here so you only compute it once.

crates/flux-refineck/src/checker.rs
```diff
@@ -1363,7 +1383,27 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
         // (4) Return the closure type
-        let no_panic = self.genv.no_panic(*did);
+
+        let tcx = self.genv.tcx();
+        let my_id = self.checker_id.root_id().to_def_id();
+        let parent_id = if matches!(tcx.def_kind(my_id), DefKind::Closure) {
+            self.genv.parent_of_closure(my_id)
+        } else {
+            my_id
+        };
```
nilehmann, line 1393: Were you getting errors without this check?

nilehmann (changes requested):

PR #1527, opened 2026-03-02
Translate Flux Bool to Prop in Lean and open Classical everywhere

petros-marko:
- remove BoolMode from lean formatter; default to translating all occurrences of bool to Prop
- mark definitions with `noncomputable` to stop lean from complaining if they indeed happen to be non-computable
- open Classical in all lean files

~TODO~
- ~test this out a little~
- ~I'm slightly concerned that the translation might break down when bools appear in non-positive positions, but I don't know of a useful reason we'd do that~

Tested by emitting lean files in `flux-demo` (for sorting related PCs). It seems to work fine. The second point I'm still not sure about, but I think LEM would allow us to do what we need.

### DIFF
--- a/crates/flux-infer/src/lean_format.rs
+++ b/crates/flux-infer/src/lean_format.rs
@@ -622,7 +622,7 @@ impl LeanFmt for KVarDecl {
 impl LeanFmt for (&KVid, &ClosedSolution) {
     fn lean_fmt(&self, f: &mut fmt::Formatter, cx: &LeanCtxt) -> fmt::Result {
         let (kvid, (implicit, (explicit, inner))) = self;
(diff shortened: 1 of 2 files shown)


crates/flux-infer/src/lean_format.rs
```diff
@@ -559,7 +526,7 @@ impl LeanFmt for Expr {
 impl LeanFmt for FunDef {
     fn lean_fmt(&self, f: &mut fmt::Formatter, cx: &LeanCtxt) -> fmt::Result {
         let FunDef { name, sort, comment: _, body } = self;
-        write!(f, "def ")?;
+        write!(f, "noncomputable def ")?;
```
nilehmann, line 529: Is this necessary as soon as you open Classical, or do you need it for some specific functions?

PR #1530, opened 2026-03-16
Fix filling in elided lifetime holes

### LINKED ISSUE #1529: ICE in `struct_compat.rs` due to elided lifetimes
We ICE under cases like this, which mirror what I saw while trying to verify `Iter::no_panic` in vtock:

```
struct MyIter<'a, T>(&'a T);

#[flux_rs::assoc(fn next_no_panic() -> bool)]
trait MyIterator {
    #[flux_rs::sig(fn(&mut Self) -> bool)]
    #[flux_rs::no_panic_if(Self::next_no_panic())]
    fn next(&mut self) -> bool;
}

#[flux_rs::assoc(fn next_no_panic() -> bool { true })]
impl<'a, T> MyIterator for MyIter<'a, T> {
    #[flux_rs::sig(fn(&mut Self) -> bool)]
    #[flux_rs::no_panic_if(Self::next_no_panic())]
    fn next(&mut self) -> bool {
        true
    }
}

// Because Flux doesn't support lifetimes inside expressions, I'm running into an ICE
// when I try to e.g. write `no_panic_if` specs for `Iter<'a, u8>`.

// Here's Claude's hypothesis:
// The elided lifetime on `MyIter<u8>` is the trigger: Flux creates ReVar(?0) for `'a`
// then panics in replace_holes when it can't find it in holes.regions.
#[flux_rs::sig(fn(_) -> bool)]
#[flux_rs::no_panic_if(<MyIter<u8> as MyIterator>::next_no_panic())]
pub fn foo(_x: u8) -> bool {
    true
}

```

From reading what Claude said, I think the solution here is to go into `struct_compat` where Flux fills in lifetimes and tell it to not fill in region holes for lifetimes?
```
   // struct_compat.rs:149
   fn fold_region(&mut self, r: &rty::Region) -> rty::Region {
       if let rty::Region::ReVar(vid) = r {
           self.regions
               .get(vid)
               .copied()
               .unwrap_or_else(|| bug!("unfilled region hole {vid:?}")) // <-- panics here!
       } else {
           *r
       }
   }
```

ninehusky:
Closes #1529

### DIFF
--- a/crates/flux-fhir-analysis/src/conv/struct_compat.rs
+++ b/crates/flux-fhir-analysis/src/conv/struct_compat.rs
@@ -148,10 +148,7 @@ impl TypeFolder for &Holes {
 
     fn fold_region(&mut self, r: &rty::Region) -> rty::Region {
         if let rty::Region::ReVar(vid) = r {
-            self.regions
-                .get(vid)
-                .copied()
-                .unwrap_or_else(|| bug!("unfilled region hole {vid:?}"))
+            self.regions.get(vid).copied().unwrap_or(rty::ReErased)
         } else {
             *r
         }
--- a/tests/tests/pos/surface/no_panic07.rs
+++ b/tests/tests/pos/surface/no_panic07.rs
@@ -1,0 +1,29 @@
+struct MyIter<'a, T>(&'a T);
+
+#[flux_rs::assoc(fn next_no_panic() -> bool)]
+trait MyIterator {
+    #[flux_rs::sig(fn(&mut Self) -> bool)]
+    #[flux_rs::no_panic_if(Self::next_no_panic())]
+    fn next(&mut self) -> bool;
+}
+
+#[flux_rs::assoc(fn next_no_panic() -> bool { true })]
+impl<'a, T> MyIterator for MyIter<'a, T> {
+    #[flux_rs::sig(fn(&mut Self) -> bool)]
+    #[flux_rs::no_panic_if(Self::next_no_panic())]
+    fn next(&mut self) -> bool {
+        true
+    }
+}
+
+// Because Flux doesn't support lifetimes inside expressions, I'm running into an ICE
+// when I try to e.g. write `no_panic_if` specs for `Iter<'a, u8>`.
+
+// Here's Claude's hypothesis:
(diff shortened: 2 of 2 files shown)


crates/flux-fhir-analysis/src/conv/struct_compat.rs
```diff
@@ -148,10 +148,7 @@ impl TypeFolder for &Holes {
 
     fn fold_region(&mut self, r: &rty::Region) -> rty::Region {
         if let rty::Region::ReVar(vid) = r {
-            self.regions
-                .get(vid)
-                .copied()
-                .unwrap_or_else(|| bug!("unfilled region hole {vid:?}"))
+            self.regions.get(vid).copied().unwrap_or(rty::ReErased)
```
nilehmann, line 151: I dont like filling arbitrary holes with erased lifetimes. This can mask other bugs where we do need to fill the lifetime. I think we should instead generate erased lifetimes from the get go when inside associated refinements.

Did claude generate the fix?

tests/tests/pos/surface/no_panic07.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Because Flux doesn't support lifetimes inside expressions, I'm running into an ICE
+// when I try to e.g. write `no_panic_if` specs for `Iter<'a, u8>`.
+
+// Here's Claude's hypothesis:
+// The elided lifetime on `MyIter<u8>` is the trigger: Flux creates ReVar(?0) for `'a`
+// then panics in replace_holes when it can't find it in holes.regions.
+#[flux_rs::sig(fn(_) -> bool)]
+#[flux_rs::no_panic_if(<MyIter<u8> as MyIterator>::next_no_panic())]
```
nilehmann, line 26: Note that this associated refinement is monomorphic and you can replace it by its definition, i.e., by true

nilehmann (changes requested):

crates/flux-fhir-analysis/src/conv/struct_compat.rs
ninehusky: > Did claude generate the fix?

yes 😆

> I think we should instead generate erased lifetimes from the get go when inside associated refinements

I'll dig into this a bit

PR #1531, opened 2026-03-18
Add `Span` for better ICE reports

ranjitjhala:
It's super tedious to have to schlep through stack traces when we have random ICE's inside the projection code; for some reason the `tracked_span_bug` was not working, so I've added `Span` everywhere.


nilehmann (commented):
Hmm im not a fan of threading spans. Doesn't it work to add a track_span call around the places that are failing?

nilehmann:
You need to have this function in the stack for tracked_span_bug to work

 https://github.com/flux-rs/flux/blob/d6cec33f756f4e508533169baead05f25b14ccb5/crates/flux-common/src/bug.rs#L16

PR #1533, opened 2026-03-19
Visit `no_panic_if` while walking `fn_sig`s

### LINKED ISSUE #1532: ICE in `conv/mod.rs` when referencing refinements in `no_panic_if` `Expr`s
Consider:

```
#[flux_rs::no_panic_if(n != 0)]
#[flux_rs::sig(fn (x: i32[@n]) -> i32[9])]
fn foo(x: i32) -> i32 {
    if x == 0 {
        panic!("AHH!");
    }
    9
}
```

When running this, we get:

```

error: internal compiler error: crates/flux-fhir-analysis/src/conv/mod.rs:224:32: node without elaborated sort for `FhirId { owner: Rust(Local(DefId(0:3 ~ no_panic07[818a]::foo))), local_id: 10 }`


thread 'rustc' (121173246) panicked at crates/flux-fhir-analysis/src/conv/mod.rs:224:32:
Box<dyn Any>

```

You can kind of get around the problem temporarily by moving the `n != 0` to a `requires` clause, but the point of the `no_panic_if` annotations are that you can express special preconditions that are optional if you don't care about panicking.

ninehusky:
Closes #1532.

The change to `wf/mod.rs`, to my understanding, helps map the nodes in the `no_panic_if` expression to their sorts. 

And the change to `fhir/visit.rs` also seemed relevant to add. I couldn't pinpoint exactly why, so I had Claude make `neg/surface/no_panic08.rs` which demonstrates an error that we only emit if we visit the `no_panic_if` expression inside `fhir/visit.rs`.

### DIFF
--- a/tests/tests/neg/surface/no_panic08.rs
+++ b/tests/tests/neg/surface/no_panic08.rs
@@ -1,0 +1,5 @@
+#[flux::sig(fn[hrn p: int -> bool](x: i32[@n]) -> bool)]
+#[flux_rs::no_panic_if(true || p(n))] //~ ERROR illegal use of refinement parameter
+fn foo(x: i32) -> bool {
+    true
+}
--- a/tests/tests/pos/surface/no_panic08.rs
+++ b/tests/tests/pos/surface/no_panic08.rs
@@ -12,6 +12,6 @@ struct Foo;
 
 impl Drop for Foo {
     #[flux_rs::sig(fn(&mut Self[@s]))]
-    #[flux_rs::no_panic_if(s != State::Bad)] // bogus field access
+    #[flux_rs::no_panic_if(s != State::Bad)] // `s` should be a `State`, but my changes seem to have made it a `Foo`...
     fn drop(&mut self) {}
 }
--- a/tests/tests/pos/surface/no_panic08.rs
+++ b/tests/tests/pos/surface/no_panic08.rs
@@ -1,0 +1,17 @@
+#[flux_rs::refined_by(n: int)]
+pub enum State {
+    #[flux_rs::variant(State[0])]
+    Good,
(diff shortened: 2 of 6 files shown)


tests/tests/neg/surface/no_panic08.rs
```diff
@@ -0,0 +1,5 @@
+#[flux::sig(fn[hrn p: int -> bool](x: i32[@n]) -> bool)]
+#[flux_rs::no_panic_if(true || p(n))] //~ ERROR illegal use of refinement parameter
```
nilehmann, line 2: Tests not related to refinement type checking should go in `tests/tests/neg/error_messages` in the corresponding directory for the *phase*, e.g., this should go in `wf`

tests/tests/pos/surface/no_panic08.rs
```diff
@@ -0,0 +1,17 @@
...
+    Bad,
+}
+
+#[flux_rs::refined_by(state: State)]
+#[flux_rs::opaque]
+struct Foo;
+
+impl Drop for Foo {
+    #[flux_rs::sig(fn(&mut Self[@s]))]
+    #[flux_rs::no_panic_if(s != State::Bad)] // `s` should be a `State`, but my changes seem to have made it a `Foo`...
```
nilehmann, line 15: This isn't related to your changes; it's about how we do sort checking. The sort of a variable is always the struct it refines, but when the `refined_by` has a single field, we sometimes *coerce* it to the sort of the field, i.e., we inject a projection. For this example, this would mean transforming the expression into `s.state != State::Bad`. The issue here is that we are not doing the coercion for binary operators. This could be fixed, but it's unrelated to the changes in the PR

PR #1537, opened 2026-03-31
Improve flux-core specs for saturating_add/saturating_sub

enjhnsn2:
This PR extends the `flux-core` specs for saturating addition/subtraction.
In particular, it:
1. Adds a `saturating_add` spec for `usize`
2. Extends the `saturating_sub` spec for `usize`
3. Extends these specs to cover `u32`
4. Adds positive and negative tests for all of the above


lib/flux-core/src/num/mod.rs
```diff
@@ -2,6 +2,26 @@ use flux_attrs::*;
 
 #[extern_spec(core::num)]
 impl usize {
+    /// Core impl: https://github.com/rust-lang/rust/blob/0e95a0f4c677002a5d4ac5bc59d97885e6f51f71/library/core/src/num/uint_macros.rs#L2384-L2400
     #[no_panic]
-    fn saturating_sub(self, other: usize) -> usize;
+    #[sig(fn(num: usize, rhs: usize) -> usize[if num < rhs { 0 } else { num - rhs }])]
```
nilehmann: @copilot change the annotations in this file to use spec instead of sig

nilehmann: @copilot remove the fixpoint binary

PR #1539, opened 2026-04-01
Add `no_panic` specs for `Option<T>`

ninehusky:
Adds `no_panic` specs for `Option::map, `Option::map_or`, and `Option::inspect`.

On my [fork](https://github.com/ninehusky/flux/tree/ninehusky-with-full-stack-trace) of Flux the two tests I've included pass, but they're failing right now. I need to make more progress on vtock verification for tonight, but I'll dig into why the tests currently fail later. (If I had to guess, we need to merge one or more of #1521, #1526)?


nilehmann:
I'm closing this because of inactivity. Most of the specs here were also added in other PR.

PR #1540, opened 2026-04-01
Add saturating_add/saturating_sub specs for u64

sweiglbosker:
- Adds `saturating_add` and `saturating_sub` specs for `u64` based on the existing specs for `usize` and `i32`
- Adds positive and negative tests for these specs


nilehmann (approved):
Thanks @sweiglbosker, this looks great!

PR #1544, opened 2026-04-04
Improve flux-core num specs

enjhnsn2:
This PR:
1. implements `uint_spec!` and `int_spec` macros to ergonomically implement specs for numeric methods.
2. Extend existing numeric spec to all types using these macros.
3. Add `checked_add`, `checked_sub`, `count_ones`, `count_zeroes`, and `abs` methods.
4. Tests for all of the above.

Notes:
1. To capture overflowing behavior in these functions (in particular, `abs`), I did not include the overflowing condition in the precondition, but instead the postcondition captures the overflowing behavior (how it works in release mode). Let me know if there is a better way to handle this.
2. These specs were in part written by Claude as an ongoing effort to play with Claude + Flux. That being said, I manually audited the specs and (heavily) edited them, so if they're wrong, you can blame me.

### DIFF
--- a/lib/flux-core/src/num/mod.rs
+++ b/lib/flux-core/src/num/mod.rs
@@ -39,12 +39,12 @@ macro_rules! int_spec {
 
             /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L496-L499
             #[no_panic]
-            #[spec(fn(num: $T, rhs: $T) -> Option<$T[num + rhs]>[num + rhs <= $T::MAX])]
+            #[spec(fn(num: $T, rhs: $T) -> Option<$T[num + rhs]>[num + rhs >= $T::MIN && num + rhs <= $T::MAX])]
             fn checked_add(self, rhs: $T) -> Option<$T>;
 
             /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L646-L649
             #[no_panic]
-            #[spec(fn(num: $T, rhs: $T) -> Option<$T[num - rhs]>[num - rhs >= $T::MIN])]
+            #[spec(fn(num: $T, rhs: $T) -> Option<$T[num - rhs]>[num - rhs >= $T::MIN && num - rhs <= $T::MAX])]
             fn checked_sub(self, rhs: $T) -> Option<$T>;
 
             /// # Incomplete: result is bounded but not precisely determined by the input value.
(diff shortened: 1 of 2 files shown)


lib/flux-core/src/num/mod.rs
```diff
@@ -1,40 +1,98 @@
...
+            #[no_panic]
+            #[spec(fn(num: $T, rhs: $T) -> $T[if num + rhs < $T::MIN { $T::MIN } else if num + rhs > $T::MAX { $T::MAX } else { num + rhs }])]
+            fn saturating_add(self, rhs: $T) -> $T;
+
+            /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L1864-L1866
+            #[no_panic]
+            #[spec(fn(num: $T, rhs: $T) -> $T[if num - rhs < $T::MIN { $T::MIN } else if num - rhs > $T::MAX { $T::MAX } else { num - rhs }])]
+            fn saturating_sub(self, rhs: $T) -> $T;
+
+            /// Panics if `self == T::MIN` (overflow when negating in debug mode). If in release mode, returns `T::MIN` on overflow.
```
nilehmann, line 35: Interesting behavior. It'd be interesting to honor the `#[rustc_inherit_overflow_checks]` annotation here. But also, when running flux we always pass `-Check-overflow=off`, so I'm not sure what the correct `no_panic` specification would be for this. But probably out of scope for this PR.

lib/flux-core/src/num/mod.rs
```diff
@@ -1,40 +1,98 @@
...
-    #[spec(fn(num: usize, rhs: usize) -> usize[if num + rhs > usize::MAX { usize::MAX } else { num + rhs }])]
-    fn saturating_add(self, rhs: usize) -> usize;
-}
+macro_rules! int_spec {
+    ($T:ident) => {
+        #[extern_spec(core::num)]
+        impl $T {
+            /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L1818-L1820
+            #[no_panic]
+            #[spec(fn(num: $T, rhs: $T) -> $T[if num + rhs < $T::MIN { $T::MIN } else if num + rhs > $T::MAX { $T::MAX } else { num + rhs }])]
```
nilehmann: maybe add a `clamp` `flux::defs!`?

lib/flux-core/src/num/mod.rs
```diff
@@ -1,40 +1,98 @@
...
-    fn saturating_add(self, rhs: u32) -> u32;
+            /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L646-L649
+            #[no_panic]
+            #[spec(fn(num: $T, rhs: $T) -> Option<$T[num - rhs]>[num - rhs >= $T::MIN])]
+            fn checked_sub(self, rhs: $T) -> Option<$T>;
+
+            /// # Incomplete: result is bounded but not precisely determined by the input value.
+            /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/num/int_macros.rs#L82
+            #[no_panic]
+            #[spec(fn(num: $T) -> u32{v: v <= $T::BITS})]
```
nilehmann, line 53: pretty cool that `$T::BITS` works!

PR #1545, opened 2026-04-06
Add flux-core spec for Result

enjhnsn2:
`Result::expect`, `unwrap`, `unwrap_err`, and `expect_err` all carry preconditions (`[true]`/`[false]` index) that statically guarantee the correct variant is present, making panic impossible. Without `#[no_panic]`, Flux couldn't leverage this and callers would accumulate unnecessary panic obligations.

- **`expect` / `unwrap`** — require `Result<T, E>[true]`; `Ok` is proven, panic unreachable
- **`unwrap_err` / `expect_err`** — require `Result<T, E>[false]`; `Err` is proven, panic unreachable

Consistent with the existing `Option::unwrap` pattern in `option.rs`.

### DIFF
--- a/lib/flux-core/src/result.rs
+++ b/lib/flux-core/src/result.rs
@@ -52,24 +52,28 @@ impl<T, E> Result<T, E> {
     fn map_err<F, O: FnOnce(E) -> F>(self, op: O) -> Result<T, F>;
 
     /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/result.rs#L1113
+    #[no_panic]
     #[spec(fn(Result<T, E>[true], &str) -> T)]
     fn expect(self, msg: &str) -> T
     where
         E: core::fmt::Debug;
 
     /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/result.rs#L1161
+    #[no_panic]
     #[spec(fn(Result<T, E>[true]) -> T)]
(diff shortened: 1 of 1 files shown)


enjhnsn2:
@copilot for the functions where the precondition implies panic-safety, can you mark them as no_panic?

lib/flux-core/src/result.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/result.rs#L1113
+    #[no_panic]
+    #[spec(fn(Result<T, E>[true], &str) -> T)]
+    fn expect(self, msg: &str) -> T
+    where
+        E: core::fmt::Debug;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/result.rs#L1161
+    #[no_panic]
+    #[spec(fn(Result<T, E>[true]) -> T)]
```
nilehmann, line 63: I'm worried this (and the other unwrapers) is going to be too annoying. But we can relax it later if necessary (i.e., put the `true` in the `no_panic_if`)

enjhnsn2: > I'm worried this (and the other unwrapers) is going to be too annoying. But we can relax it later if necessary (i.e., put the `true` in the `no_panic_if`)

Ya, I just followed the existing semantics of `Option::unwrap`. I agree that we should lift these to `no_panic_if` at some point. Andrew also said that using `no_panic_if` with indices is a little finicky right now, so idk

PR #1549, opened 2026-04-07
Fix code in equality chapter

sweiglbosker:
- enum was declared as `Roles` but code uses `Role`
- `Role::User` is used in some parts of the code even though such a variant does not exist. (I imagine it was renamed to `Role::Member` after adding the `User` struct)
- add `is_{admin,guest}` methods for `User`. The fact that these did not exist caused an error that prevented earlier verifier failures. I also added specs to these to make the later assertions fail due to lack of a `new` spec, which is the behavior described in the chapter
- Similiar fixes for the `Role` `PartialEq` impl. The broken neq comparison is still there, I just fixed it up enough so it doesn't stop the rest of the code blocks from checking due to undeclared symbols.
- Also added `configure` and `delete` methods for `User`. The fact that they are missing prevents other code blocks from being checked and I don't think they were omitted on purpose.
- Bonus fix in ownership chapter: test was checking the wrong function

### DIFF
--- a/book/src/tutorial/02-ownership.md
+++ b/book/src/tutorial/02-ownership.md
@@ -303,7 +303,7 @@ hence, that upon return `x: i32[11]`.
 ```rust, editable
 fn test_incr() {
   let mut z = 10;
-  incr_inv(&mut z);
+  incr(&mut z);
   assert(z == 11);
 }
 ```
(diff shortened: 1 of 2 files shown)


nilehmann (approved):
Awesome, thanks!

PR #1552, opened 2026-04-08
Improve slice spec: add specs for `last`, `get`, `split_first/last`, `split_at_checked`, and mut counterparts

enjhnsn2:
Extends `flux-core`'s slice extern specs with refinement types for several previously unspecced methods, and adds `#[no_panic]` to existing specs that were missing it.

## New specs

- `last`, `last_mut` — `Option<&T>[n != 0]`, symmetric with `first`
- `first_mut` — mutable counterpart of `first`
- `get`, `get_mut` — `Option<...>[<I as SliceIndex<[T]>>::in_bounds(idx, n)]`, reusing the existing `in_bounds` predicate from the `Index` impl
- `split_first`, `split_last` — `Option<(&T, &[T][n - 1])>[n != 0]`, tail/init length tracked precisely
- `split_at_mut` — mutable counterpart of `split_at`, same precondition `mid <= n`
- `split_at_checked`, `split_at_mut_checked` — `Option<(&[T][mid], &[T][n - mid])>[mid <= n]`

## `#[no_panic]` retrofits

Added `#[no_panic]` to pre-existing specs that were missing it: `is_empty`, `first`, `iter`, `as_ptr`, `as_mut_ptr` — all unconditionally panic-free.

## Other

- Migrated `#[sig(...)]` → `#[spec(...)]` on all pre-existing specs
- Moved inline `cfg(flux_sysroot_test)` tests from `mod.rs` into the dedicated `tests/` directory

```rust
// get: provably Some when in bounds
fn get_in_bounds(xs: &[i32], i: usize) {
    if i < xs.len() {
        assert(xs.get(i).is_some()); // verified ✓
    }
}

// split_first: tail length tracked precisely
fn split_first_tail_len(xs: &[i32]) {
    let n = xs.len();
    if let Some((_, tail)) = xs.split_first() {
        assert(tail.len() == n - 1); // verified ✓
    }
}

// split_at_checked: lengths flow through the Option
fn split_at_checked_lengths(xs: &[i32], mid: usize) {
    let n = xs.len();
    if let Some((left, right)) = xs.split_at_checked(mid) {
        assert(left.len() == mid);      // verified ✓
        assert(right.len() == n - mid); // verified ✓
    }
}
```

### DIFF
--- a/lib/flux-core/src/slice/mod.rs
+++ b/lib/flux-core/src/slice/mod.rs
@@ -6,72 +6,51 @@ use flux_attrs::*;
 #[extern_spec(core::slice)]
 impl<T> [T] {
     #[no_panic]
-    #[sig(fn(&Self[@n]) -> usize[n])]
+    #[spec(fn(&Self[@n]) -> usize[n])]
     fn len(&self) -> usize;
 
-    #[sig(fn(&Self[@n]) -> bool[n == 0])]
+    #[spec(fn(&Self[@n]) -> bool[n == 0])]
     fn is_empty(&self) -> bool;
 
-    #[sig(fn(&Self[@n]) -> Option<&T>[n != 0])]
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/slice/mod.rs#L154
+    #[spec(fn(&Self[@n]) -> Option<&T>[n != 0])]
     fn first(&self) -> Option<&T>;
 
-    #[sig(fn(&Self[@n]) -> Iter<T>[0, n])]
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/slice/mod.rs#L280
+    #[no_panic]
+    #[spec(fn(&Self[@n]) -> Option<&T>[n != 0])]
+    fn last(&self) -> Option<&T>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/slice/mod.rs#L570
+    #[no_panic]
+    #[spec(fn(&Self[@n], I[@idx]) -> Option<&I::Output>[<I as SliceIndex<[T]>>::in_bounds(idx, n)])]
+    fn get<I: SliceIndex<[T]>>(&self, index: I) -> Option<&I::Output>;
+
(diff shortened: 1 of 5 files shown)


enjhnsn2:
@copilot This PR was written by a junior developer with little experience in Flux or formal verification. Can you check it very thoroughly (especially for unsoundness)?

enjhnsn2:
@copilot The junior developer retrofitted no_panic to some old specs. Can you check them?

lib/flux-core/src/slice/mod.rs
```diff
@@ -6,72 +6,81 @@ use flux_attrs::*;
...
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/slice/mod.rs#L570
+    #[no_panic]
+    #[spec(fn(&Self[@n], I[@idx]) -> Option<&I::Output>[<I as SliceIndex<[T]>>::in_bounds(idx, n)])]
+    fn get<I: SliceIndex<[T]>>(&self, index: I) -> Option<&I::Output>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/slice/mod.rs#L596
+    #[no_panic]
+    #[spec(fn(&mut Self[@n], I[@idx]) -> Option<&mut I::Output>[<I as SliceIndex<[T]>>::in_bounds(idx, n)])]
+    fn get_mut<I: SliceIndex<[T]>>(&mut self, index: I) -> Option<&mut I::Output>;
```
nilehmann, line 44: These don't panic if the corresponding method in the `SliceIndex` implementation for `I` doesn't panic. I believe none of the current implementations panics. Downstream users cannot add implementations because of the orphan rule, but technically, the standard library could add a new implementation that panics, making this signature wrong.

I'm ok with playing a bit fast and lose, but maybe there's a way we can make the assumption more explicit?

enjhnsn2: > These don't panic if the corresponding method in the `SliceIndex` implementation for `I` doesn't panic. I believe none of the current implementations panics. Downstream users cannot add implementations because of the orphan rule, but technically, the standard library could add a new implementation that panics, making this signature wrong.
> 
> I'm ok with playing a bit fast and lose, but maybe there's a way we can make the assumption more explicit?

I'll see if there is a more obviously sound way to do this and get back to you

enjhnsn2: @nilehmann The best I could do was indeed check that SliceIndex is a sealed trait and doesn't panic, and document that fact in the comments.

nilehmann (approved):

PR #1553, opened 2026-04-09
Improve option spec

enjhnsn2:
Added some boring specs for common `Option<T>` functions.
Mostly supercedes #1539 but I couldn't get `Option::inspect` to work. 
It seems to fail with: 
```
fails with "cannot determine corresponding unrefined predicate" for FnOnce(&T) + [const] Destruct bounds
```

Interestingly, some functions with `[const] Destruct` work like `map`, but the implicit `()` return type in `inspect` seems to cause issues.


enjhnsn2:
@copilot an undergrad wrote this code, can you thoroughly inspect it? In particular, ensure that the specs written are sound and clear as possible.

enjhnsn2:
Whoops, I branched off `improve_slice_spec` instead of `main`. Let me fix that in a new PR.

PR #1554, opened 2026-04-09
Improve option spec 

enjhnsn2:
Added some boring specs for common `Option<T>` functions.
Mostly supercedes #1539 but I couldn't get `Option::inspect` to work. 
It seems to fail with: 
```
fails with "cannot determine corresponding unrefined predicate" for FnOnce(&T) + [const] Destruct bounds
```

Interestingly, some functions with `[const] Destruct` work like `map`, but the implicit `()` return type in `inspect` seems to cause issues.

### DIFF
--- a/lib/flux-core/src/option.rs
+++ b/lib/flux-core/src/option.rs
@@ -58,6 +58,33 @@ impl<T> Option<T> {
     #[spec(fn(Option<T>[@a], Option<T>[@b]) -> Option<T>[a || b])]
     fn or(self, optb: Option<T>) -> Option<T>;
 
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/option.rs#L1309
+    #[no_panic]
+    #[spec(fn(Option<T>[@b], E) -> Result<T, E>[b])]
+    fn ok_or<E>(self, err: E) -> Result<T, E>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/option.rs#L1334
+    #[flux_rs::no_panic_if(F::no_panic())]
+    #[spec(fn(Option<T>[@b], F) -> Result<T, E>[b] where F: FnOnce() -> E)]
+    fn ok_or_else<E, F: FnOnce() -> E>(self, err: F) -> Result<T, E>;
+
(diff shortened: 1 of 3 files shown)


lib/flux-core/src/option.rs
```diff
@@ -32,4 +47,51 @@ impl<T> Option<T> {
...
+    fn ok_or<E>(self, err: E) -> Result<T, E>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/option.rs#L1334
+    #[flux_rs::no_panic_if(F::no_panic())]
+    #[spec(fn(Option<T>[@b], F) -> Result<T, E>[b] where F: FnOnce() -> E)]
+    fn ok_or_else<E, F: FnOnce() -> E>(self, err: F) -> Result<T, E>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/c6a955468b025dbe3d1de3e8f3e30496d1fb7f40/library/core/src/option.rs#L1504
+    #[flux_rs::no_panic_if(F::no_panic())]
+    #[spec(fn(Option<T>[@b], F) -> Option<U>{s: s => b} where F: FnOnce(T) -> Option<U>)]
```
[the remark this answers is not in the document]
enjhnsn2: not true you goofy bot. In flux `T{v: pred}` binds v to T's index, not to its value.

nilehmann, line 73: @ranjitjhala @petros-marko here's a more mundane function where we could say something stronger if we could refer to the output of the closure.

nilehmann (approved):
LGTM

Unsure what's going on with `inspect`. But const trait bounds (in their current form) have only recently been added to nightly, and I have only added enough support to make our test suite pass.

lib/flux-core/src/option.rs
nilehmann: @enjhnsn2, curious about how you arrived at this specification. Did claude suggest it as is? Thinking about it as written breaks my head, but the contrapositive is fine.

enjhnsn2: This is a claude suggestion, although I did manually check this one since it looks wierd. The "real" spec would be something like     
```rust
#[spec(fn(Option<T>[@b], F) -> Option<U>{s: b && F::output } where F: FnOnce(T) -> Option<U>)]
```
where `F::output` is the index contained withing the closure's return type.
The spec as-is just implies that if the output is true, then the input must have been true.

Issue #1546, opened 2026-04-07
Flux-core support for pointers and allocation primitives


enjhnsn2:
I have ambitions of verifying [Bumpalo](https://github.com/fitzgen/bumpalo), a memory allocator used by Firefox, Cranelift, and wasm-bindgen. I have done some quick prototyping and it seems pretty doable, but it requires much more robust support of the core crates ptr/nonnull and alloc. These specs should also be very useful for other projects (e.g., the flan work).

The list of extern specs looks something like:

Ptr<T>
- [ ] *const T::sub / *mut T::sub                                                                                     
- [ ] *const T::offset / *mut T::offset                                                                               
 - [ ] *const T::byte_add / *mut T::byte_add                                                                           
 - [ ] *const T::byte_sub / *mut T::byte_sub                                                                           
 - [ ] ptr::copy                                                                                                       
 - [ ] ptr::copy_nonoverlapping
 - [ ] *const T::offset_from / *mut T::offset_from                                                                     
 - [ ] ptr::write_bytes                                                                                                
 - [ ] *const T::byte_offset / *mut T::byte_offset
 - [ ] *const T::is_null / *mut T::is_null                                                                             
 - [ ] ptr::swap / ptr::swap_nonoverlapping
 - [ ] *mut T::offset_from                                                                                             
                  
NonNull<T>                                                                                                        
  - [ ] new                                                                                                             
  - [ ] new_unchecked                                                                                                   
  - [ ] as_ptr                                                                                                          
  - [ ] add / sub                                                                                                       
  - [ ] offset        
  - [ ] byte_add / byte_sub / byte_offset
  - [ ] read / write                     
  - [ ] offset_from                                                                                                     
  - [ ] slice_from_raw_parts
  - [ ] len                                                                                                             
                  
  Allocator / Layout
  - [x] Layout::new                                                                                                     
  - [x] Layout::from_size_align
  - [x] Layout::size                                                                                                    
  - [x] Layout::align 
  - [x] Layout::array
  - [ ] Allocator::allocate
  - [ ] Allocator::deallocate
  - [ ] Allocator::grow                                                                                                 
  - [ ] Allocator::shrink

Notes: 
- This will require choosing indices for each of these types (or in the case of Ptr<T>, better indices).
- Ptr and NonNull should be almost completely redundant.

ranjitjhala:
Exciting stuff lmk if you want to chat tomorrow or we’d

- Ranjit.

PR #1555, opened 2026-04-10
Add `align_of` associated refinement

sweiglbosker:
This is plumbing to tackle some of #1546 (the layout stuff, mostly)

- Added `align_of` builtin associated refinement
- Added `core::mem::align_of` spec using said refinement
- Move the `size_of` tests that were testing the flux core specs to an `extern_specs` test


nilehmann (approved):
Cool

PR #1556, opened 2026-04-10
Remove extra semicolons

sweiglbosker:
Some bad programmer added broken code in  https://github.com/flux-rs/flux/pull/1549...


nilehmann (approved):
Some bad reviewer didn't notice XD

PR #1559, opened 2026-04-12
Add flux core specs for `alloc::Layout`

sweiglbosker:
Adds some initial specs for alloc::Layout:
- `Layout::new`
- `Layout::from_size_align`, `Layout::from_size_align_unchecked`
- `Layout::size`, `Layout::align`
- `Layout::array`

Related: https://github.com/flux-rs/flux/issues/1546

### DIFF
--- a/lib/flux-core/src/alloc/layout.rs
+++ b/lib/flux-core/src/alloc/layout.rs
@@ -15,7 +15,7 @@
 
     // See: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L559-L589
     fn is_valid_array_layout(n: int, size: int, align: int) -> bool {
-        size == 0 || is_valid_layout(size * n, align)
+        size == 0 || size * n + align - 1 <= isize::MAX
     }
 )]
 
--- a/lib/flux-core/src/alloc/layout.rs
+++ b/lib/flux-core/src/alloc/layout.rs
@@ -1,0 +1,53 @@
+#![flux::defs(
+    fn pow2(x: int) -> bool { pow2bv(bv_int_to_bv64(x)) }
+    fn pow2bv(x: bitvec<64>) -> bool {
+        bv_and(x, bv_sub(x, bv_int_to_bv64(1))) == bv_int_to_bv64(0)
+    }
+    fn is_valid_layout(s: int, a: int) -> bool {
+        (s + a - 1 <= isize::MAX) && a > 0 && pow2(a)
+    }
+)]
+
+use flux_attrs::*;
+
+#[extern_spec(core::alloc)]
+#[opaque]
+#[refined_by(size: int, align: int)]
+#[invariant(align > 0 && pow2(align))]
+#[invariant(size + align - 1 <= isize::MAX)]
+struct Layout;
+
+#[extern_spec(core::alloc)]
+impl Layout {
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L206-L208
+    #[no_panic]
+    #[spec(fn() -> Self[T::size_of(), T::align_of()])]
+    const fn new<T>() -> Self;
+
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L59-L66
+    #[no_panic]
+    #[spec(fn(s: usize, a: usize) -> Result<Self[s,a], _>[is_valid_layout(s, a)])]
+    const fn from_size_align(size: usize, align: usize) -> Result<Layout, LayoutError>;
+
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L132-L144
+    #[no_panic]
+    #[spec(fn(s: usize, a: usize) -> Self[s, a] requires is_valid_layout(s, a))]
+    const unsafe fn from_size_align_unchecked(size: usize, align: usize) -> Self;
+
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L175-L177
+    #[no_panic]
+    #[spec(fn(&Self[@s, @a]) -> usize[s])]
+    const fn size(&self) -> usize;
+
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L187-L189
+    #[no_panic]
+    #[spec(fn(&Self[@s, @a]) -> usize[a])]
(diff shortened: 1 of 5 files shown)


lib/flux-core/src/alloc/layout.rs
```diff
@@ -0,0 +1,53 @@
+#![flux::defs(
+    fn pow2(x: int) -> bool { pow2bv(bv_int_to_bv64(x)) }
+    fn pow2bv(x: bitvec<64>) -> bool {
+        bv_and(x, bv_sub(x, bv_int_to_bv64(1))) == bv_int_to_bv64(0)
+    }
+    fn is_valid_layout(s: int, a: int) -> bool {
+        (s + a - 1 <= isize::MAX) && a > 0 && pow2(a)
+    }
```
nilehmann, line 14: Please comment on what this function is doing, in particular the `s + a - 1 <= iszie::MAX` part. It wasn't obvious to me that it was doing the right thing. Maybe you can point to this comment [max_size_for_alignment](https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L80-L93).

nilehmann (commented):
I'm concerned the bitvectors may blow up, but it may be okay. @enjhnsn2 thoughts

tests/tests/neg/extern_specs/flux_core_alloc00.rs
```diff
@@ -0,0 +1,68 @@
+extern crate flux_core;
```
nilehmann, line 1: I've been feeling a bit uncomfortable with tests on extern specs because they are as good as our understanding of what they are supposed to return. But what if we run them? Runtime refinement checking is an open problem in general, but we could make it work for these assert-driven tests.

`compiletest_rs` supports running the generated binaries. We would need to adjust the flux binary to emit binaries and figure out what to do with neg tests

@enjhnsn2 thoughts?

enjhnsn2: Yes, currently the tests for extern tests only check whether the spec itself is written coherently (e.g., if you can actually use it to prove something meaningful) and does not check whether the spec reflects reality. We've talked often about converting refinements to tests, but maybe its time to actually do it :)

enjhnsn2: I think this would require 3 steps:
1. Convert static `assert` to dynamic `assert!`
2. Supply a generator for inputs to functions based on the precondition
3. Convert post-conditions to asserts.

We had a janky version of this in WaVe, maybe we could port it over?

lib/flux-core/src/alloc/layout.rs
```diff
@@ -0,0 +1,53 @@
...
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L187-L189
+    #[no_panic]
+    #[spec(fn(&Self[@s, @a]) -> usize[a])]
+    const fn align(&self) -> usize;
+
+    // Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/layout.rs#L559-L589
+    #[no_panic]
+    #[spec(fn(n: usize) -> Result<Self[n * T::size_of(), T::align_of()],_>[
+        T::size_of() == 0 || n * T::size_of() + T::align_of() - 1 <= isize::MAX
+    ])]
```
enjhnsn2, line 62: This predicate can be wrapped up into a `flux_def!`.

enjhnsn2:
> I'm concerned the bitvectors may blow up, but it may be okay. @enjhnsn2 thoughts

These specs are the cleanest "complete" specs that you can implement.
One alternative option (that is ugly but may be more performant) is to implement the alignment `pow2` constraints as a finite enumeration over integers, e.g. `align ==1 || align==2 || align==4...` . With this, I think we can completely avoid int->bv and bv->int conversions.

I'm also not too concerned about constraint blow-up in this module `Layout` in particular. These specs will only be used if you're trying to verify a memory allocator and are unlikely to contaminate other code bases.

sweiglbosker:
Are we also worried about how clunky this is? The scenario in my head is
that someone wants to create a layout aligned to pagesize, so
we mark `getpagesize()`  trusted and refine the return type, but this would
require duplicating all the defs here in that module, which seems a bit annoying.

tests/tests/neg/extern_specs/flux_core_alloc00.rs
nilehmann: Yes, that sounds about what I had in mind. Steps 1 and 2 should be straightforward, as that's just testing infrastructure, no messing with Flux internals. 

Btw, I think we can get pretty far with just 1 and 2 by having explicit asserts. Property-based specification testing, you may say

nilehmann:
The finite enumeration came up in my discussion with Copilot. That could be our escape hatch in case bit vectors blow up. But we won't know which specification will be useful in practice until we try to prove something. And I agree that people are not randomly using `Layout` so this is only going to come up when you are doing layouty things.

I'm ok with the specs as is. We can revisit later once we have a specific verification task.

tests/tests/neg/extern_specs/flux_core_alloc00.rs
enjhnsn2: @nilehmann Ranjit redirected a strong UCSD masters student to me that could perhaps work on this? I was going to have him verify a data structures library, but if this doesn't require too much digging in Flux, perhaps this could be a good option?

nilehmann: Yeah, I think this could be a cool project for them!

enjhnsn2:
> Are we also worried about how clunky this is? The scenario in my head is that someone wants to create a layout aligned to pagesize, so we mark `getpagesize()` trusted and refine the return type, but this would require duplicating all the defs here in that module, which seems a bit annoying.

The concern is that the conversions between bitvector and int could make the SMT solver hang. 
In general, there are certain operations (bitvector-int conversions, multiplication, division) that are hard for SMT solvers to handle, so if possible, we avoid them.

That being said, I think the spec is fine as implemented in this particular case.

sweiglbosker:
> The concern is that the conversions between bitvector and int could make the SMT solver hang.
> In general, there are certain operations (bitvector-int conversions, multiplication, division) that are hard for SMT solvers to handle, so if possible, we avoid them.
> 
> That being said, I think the spec is fine as implemented in this particular case.

Yeah, what I mention there is an unrelated gripe; I was wondering if there is way to do
the bitvector stuff more concisely or share the defs across translation units. (Or if that is even a valid concern)

lib/flux-core/src/alloc/layout.rs
nilehmann: @sweiglbosker, don't know if it's worth the trouble, but the refactor is doing redundant work because `T::align_of` is always going to be valid.

sweiglbosker: Good point!

sweiglbosker:
Also https://github.com/flux-rs/flux/pull/1562 should probably be landed before this so I can clean up the tests here

enjhnsn2 (approved):

nilehmann (approved):

PR #1561, opened 2026-04-13
Support Fn/FnMut/FnOnce bounds with full signatures

petros-marko:
This PR adds support for function trait bounds with full function signatures (e.g. `where F: Fn(x: i32[@n]) -> i32[n+1] requires ... ensures ...`). This enables the following test to pass:

```
#[flux_rs::refined_by(i: int)]
struct Wrapper {
    #[flux_rs::field(i32[i])]
    inner: i32,
}

impl Wrapper {
    #[flux_rs::spec(fn[hrn upd: (int, int) -> bool](s: &mut Self[@slf], f: F)
        ensures s : Self{v : upd(v, slf)}
        where F: Fn(x: &mut i32[@old]) ensures x : i32{v : upd(v, old)}
    )]
    fn with_mut<F>(&mut self, f: F)
    where
        F: Fn(&mut i32),
    {
        f(&mut self.inner);
    }
}

#[flux_rs::spec(fn() -> i32[100])]
fn test() -> i32 {
    let mut w = Wrapper { inner: 99 };
    w.with_mut(|x| *x += 1);
    w.inner
}
```

It adds a new field to `surface::TraitRef` to represent an Fn trait bound (if available) and carries it through desugaring and conversion. This part I understand. The changes in `checker.rs` and `infer.rs`, I'm much more suspicious of, and I have a much looser grasp on. I'll continue to look at them, but I thought I should put this out too.

### DIFF
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -872,6 +872,17 @@ impl<'a, E: LocEnv> Sub<'a, E> {
                 )?;
                 Ok(())
             }
+            (
+                TyKind::StrgRef(_, path, ty),
+                TyKind::Indexed(BaseTy::Ref(_, bound, Mutability::Mut), idx),
+            ) => {
+                // Allow a strong reference to be used where an `&mut` is expected by
+                // unfolding the strong ref and checking the unfolded inner type.
+                self.idxs_eq(infcx, &Expr::unit(), idx);
+                self.env.unfold_strg_ref(infcx, path, ty)?;
+                let ty = self.env.get(path);
+                self.tys(infcx, &ty, bound)
+            }
 
             (TyKind::Indexed(bty_a, idx_a), TyKind::Indexed(bty_b, idx_b)) => {
                 self.btys(infcx, bty_a, bty_b)?;
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -404,19 +404,6 @@ impl<'infcx, 'genv, 'tcx> InferCtxt<'infcx, 'genv, 'tcx> {
         }
         let evars = &mut self.inner.borrow_mut().evars;
         if let ExprKind::Var(Var::EVar(evid)) = b.kind()
-            && let EVarState::Unsolved(marker) = evars.get(*evid)
-        {
-            let _ = marker.has_free_vars(a);
-            evars.solve(*evid, a.clone());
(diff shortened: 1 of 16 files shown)


crates/flux-infer/src/infer.rs
```diff
@@ -404,8 +404,7 @@ impl<'infcx, 'genv, 'tcx> InferCtxt<'infcx, 'genv, 'tcx> {
         }
         let evars = &mut self.inner.borrow_mut().evars;
         if let ExprKind::Var(Var::EVar(evid)) = b.kind()
-            && let EVarState::Unsolved(marker) = evars.get(*evid)
-            && !marker.has_free_vars(a)
+            && let EVarState::Unsolved(_) = evars.get(*evid)
```
petros-marko, line 407: This must be bad, right?

nilehmann:
To begin with, I don't like going back to having a separate `FnPredicate`. But setting that aside, it didn't take me too much time to find examples that break the implementation

Changing the refinement to `Fn: Fn(x: &mut u32[@old])` (`u32` instead of `i32`) gives an ICE. Changing the bound in the refinement to be `FnMut` gives a refinement type error. With the current implementation, both cases are rejected as bad signatures. The error is not very good, but at least I understand why we get the specific error we get.

Removing the `!marker.has_free_vars(a)` is also wrong. The issue is that the extra unfold that was added generates an evar in the wrong context. The error was adding that extra case.

I'm not sure what to do about this. I don't want to spend time reading slop.

petros-marko:
I think not reading slop is the right thing to do. I definitely thought there were things wrong here, so I appreciate you pinpointing it quickly, I think it'd have taken me much longer to do.

I guess the real question is: what's the right way to go about adding support for fuller closure signatures, and how far is it from this?

nilehmann:
@petros-marko I think what I would like to understand is if we can make sense of `ensures` for closures in a world where `F: Fn(A) -> B` is represented as two predicates `F: Fn<A>` and `<F as Fn<A>>::Output = B` (the way Rust represents it). We have, in the past, represented fn trait bounds as a separate predicate in the list of clauses, but at some point, I refactored it to delay the construction of the `FnPredicate` to the point where we check it in the hope that we would eventually be able to remove the special casing we have for fn trait bounds. But maybe having a separate clause is the right thing to do. This also relates to the change we wanted to make to have refined tuples.

We should have a design meeting.

PR #1562, opened 2026-04-14
Allow lossless casts from signed to unsigned

### LINKED ISSUE #1558: Be more precise about signed/unsigned casts of integer constants
There is room to be less conservative when reasoning about casts between signed and unsigned integer constants.

```rust
// i.e, this will currently fail
assert(1u8 as i8 == 1)
```

I bring this up after encountering code that does stuff like `isize::MAX as usize`, which seems to be a common pattern.

sweiglbosker:
A cast from a signed integer to an unsigned integer preserves idx in two cases:
- The value is known to be non-negative and the destination is at least as wide as the source
- The value is known to be non-negative and known to fit in the destination

Therefore we can use a "lazy" type of the form: `u<n>{v: 0 <= idx && idx <= u<n>::MAX => v == idx}`

The performance of this probably needs investigation, but I think int -> uint casts are rare enough that there shouldn't be a huge regression.

Fixes: #1558

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -1631,7 +1631,9 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                     (Uint!(uint_ty, idx), RustTy::Int(int_ty)) => {
                         uint_int_cast(idx, *uint_ty, *int_ty)
                     }
-                    (Int!(_, _), RustTy::Uint(uint_ty)) => Ty::uint(*uint_ty),
+                    (Int!(int_ty, idx), RustTy::Uint(uint_ty)) => {
+                        int_uint_cast(idx, *int_ty, *uint_ty)
+                    }
                     (TyKind::Discr(adt_def, _), RustTy::Int(int_ty)) => {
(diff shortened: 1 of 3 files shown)


enjhnsn2:
> A cast from a signed integer to an unsigned integer preserves idx in two cases:
> 
>     * The value is known to be non-negative and the destination is at least as wide as the source
> 
>     * The value is known to be non-negative and known to fit in the destination
> 
> 
> Therefore we can use a "lazy" type of the form: `u<n>{v: 0 <= idx && idx <= u<n>::MAX => v == idx}`
> 
> The performance of this probably needs investigation, but I think int -> uint casts are rare enough that there shouldn't be a huge regression.
> 
> Fixes: #1558

I don't think performance should be a concern here, as this can (at max) add 1 linear numeric constraint when we cast. And we should already be doing this anyways :)

enjhnsn2 (approved):

nilehmann:
> The performance of this probably needs investigation, 

We have one real benchmark (vtock) and CI time didn't go up for it, so I'd say we are good :)

> but I think int -> uint casts are rare enough that there shouldn't be a huge regression.

That's because they don't have flux to prove their int to uint conversions :)

PR #1565, opened 2026-04-19
RawPtr rework

sweiglbosker:
Refine pointers with base, addr, size

Examples:

```rust
  #[spec(fn(*const[@p] i32))]                                                                                                                                                 
  fn a(p: *const i32) {}                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                     
  #[spec(fn(*const[@base, @addr, @size] i32))]                                                                                                                                
  fn b(p: *const i32) {}           
                                                                                                                                                                                   
  #[flux::sig(fn({*const[@base, @addr, @size] i32 | size > 0}))]                                                                                                                   
  fn c(p: *const i32) {}            
                                                                                                                                                                                                                                                                                                                  
  #[flux::sig(fn(*const{p: p.size > 0} i32))]    
  fn d(p: *const i32) {}                                                                                                                                                 
```

Related: #1546


nilehmann (commented):
The implementation looks good to me, but I don't know whether `ptr_size` and `ptr_offset` are the best way to represent pointers. Why not `ptr_addr`, `ptr_start`, and `ptr_size`?. Or even `ptr_end`, I believe @ranjitjhala added `ptr_size` without thinking too much about it, so we don't have to necessarily keep it. 

I don't have a strong opinion, so I'll let @enjhnsn2 chime in

sweiglbosker:
> @nilehmann commented on this pull request.                                                                                                                                         
>                                                                                                                                                                                    
> The implementation looks good to me, but I don't know whether `ptr_size` and `ptr_offset` are the best way to represent pointers. Why not `ptr_addr`, `ptr_start`, and `ptr_size`?.
>  Or even `ptr_end`, I believe @ranjitjhala added `ptr_size` without thinking too much about it, so we don't have to necessarily keep it.                                           
>                                                                                                                                                                                    
> I don't have a strong opinion, so I'll let @enjhnsn2 chime in 

Yeah, I haven't really settled on the exact representation yet, and opened this mostly to get feedback on implementation; I will probably make changes as I implement more.

If anyone has feedback though, here are the initial design notes I have jotted down:
- ptr_addr: Int
- ptr_offset: Int
- ptr_extent: Int (size of the allocation)
- ptr_len: Int
- ptr_align: Int (needed for nonoverlapping precondition. Even if this is redundant, I feel this is better than expressing in terms of address or start and offset because it is usually statically known)

I think my naive philosophy behind leaning towards the offset / extent / len choice is that it seems like otherwise you would end up with a bunch of expressions like `end - start == x` instead of nice independent ints.

I still include ptr_addr to reason about non_null (originally I had ptr_start but this wasnt used for anything directly). I guess there could be an is_null uif but I want to provide the ability to reason about the address arithmetically even if I avoid it as much as possible for core specs.

Having this much provenance info might be a bit annoying and lead to large specs though...

nilehmann:
Regarding the implementation. If we are going to add more UIFs, I'd try to reduce the duplication. I think we could have an inner enum for the *pointer builtins*

enjhnsn2:
Hmm, so I think it would be best if our indices mapped to the values used to reason about provenance in the Rust docs as closely as possible. 
In this vein, I think we need to track 3 indices: `base`, `ptr`, and `size`.
I believe we do not need to explicitly track size of the pointer itself and its alignment, that should all be statically determined through the type system, and we can write generic specs using `T::size_of()` and `T::align_of()`.

I also think we should probably not use UIFs for these at all.
In my last meeting with Stefan we agreed that using UIFs might be a good idea because they are more flexible, but the more I think about it, I think the 3 provenance indices are the "right" indices.
Our one concern with the 3 provenance indices is that all specs with pointers now need to track these 3 indices, but I think thats just kind of unavoidable (from an ergonomics perspective, flux is also good at ignoring indices when you don't care about them, e.g., you can replace and index with an underscore, like a return argument that you are ignoring).

Hopefully this pivot shouldn't be _too_ bad, and only requires changing the syntax of the specs and removing the UIF boilerplate (sorry stefan!)


Links:
Ptr safety rules: https://doc.rust-lang.org/std/ptr/index.html
Provenance definition: https://doc.rust-lang.org/std/ptr/index.html#provenance
Allocation definition: https://doc.rust-lang.org/std/ptr/index.html#allocation

nilehmann:
Something that might not be clear is that the UIF approach was already implemented and @sweiglbosker simply added a new one. In the current implementation, UIFs on pointers are hardcoded inside Flux.

Fundamentally, there's not much difference between UIFs and multiple indices (formally, a struct with multiple fields). They are equally expressible, although in practice, specs would look different. At this point, I'm more worried about having something to experiment with, so I'd stay with UIFs because it's the path of least resistance (already implemented). I don't think it'd be too hard to transition away from UIF's latter (assuming the IA will translate the specs into the new syntax XD).

enjhnsn2:
> Something that might not be clear is that the UIF approach was already implemented and @sweiglbosker simply added a new one. In the current implementation, UIFs on pointers are hardcoded inside Flux.
> 
> Fundamentally, there's not much difference between UIFs and multiple indices (formally, a struct with multiple fields). They are equally expressible, although in practice, specs would look different. At this point, I'm more worried about having something to experiment with, so I'd stay with UIFs because it's the path of least resistance (already implemented). I don't think it'd be too hard to transition away from UIF's latter (assuming the IA will translate the specs into the new syntax XD).

How much work do you think it would take to add built-in indices?
I think if its <= 2x as much work, we should still use the indices, as 1) the ergonomics will be much better (uniformity with the other specs will make life much easier) and 2) the 3 provenance indices are essentially official "rust semantics" (in the sense that they are the 3 values you need to prove Rust safety rules).

nilehmann:
Adjusting the current implementation to have the three UIFs you mention (`ptr_base`, `ptr_addr`, and `ptr_size`), it's like under 30 minutes work. Re-implementing everything to use an index with three fields, it's maybe like a couple of days work (I don't think it'll be too terrible).


---

As a side note, I've been thinking for a while about having a Lean-like syntax where `e1.foo(e2, ...)` gets translated internally to `foo(e1, e2, ...)`, which would make functions (not just UIFs) a lot nicer in general.

enjhnsn2:
> Adjusting the current implementation to have the three UIFs you mention (`ptr_base`, `ptr_addr`, and `ptr_size`), it's like under 30 minutes work. Re-implementing everything to use an index with three fields, it's maybe like a couple of days work (I don't think it'll be too terrible).
> 
> As a side note, I've been thinking for a while about having a Lean-like syntax where `e1.foo(e2, ...)` gets translated internally to `foo(e1, e2, ...)`, which would make functions (not just UIFs) a lot nicer in general.

Alright, I suppose UIFS are fine for now, but we should move towards using base, addr, and size as the 3 UIFS.

And yes, that lean syntax would be immeasurably better that the UIF/flux_def syntax we have now

sweiglbosker:
> Hopefully this pivot shouldn't be too bad, and only requires changing the syntax of the specs and removing the UIF boilerplate (sorry stefan!)

No worries haha, this is exactly why I opened this tiny pr.

> I believe we do not need to explicitly track size of the pointer itself and its alignment, that should all be statically determined through the type system, and we can write generic specs using T::size_of() and T::align_of()

I think you are wrong about the statically known part (pointers don't have to be correctly aligned), but right that it doesn't need its own field and we don't need to track it directly. (As in we can use the `addr % T::align_of() == 0` when we require a pointer with valid alignment). Alternatively we can just ignore alignment altogether for now (which we already do for read/write, which are ub for unaligned pointers). 

> In this vein, I think we need to track 3 indices: base, ptr, and size.

Yeah, I like that a lot more than my idea :)

nilehmann:
My impression is that alignment bugs are common and they happen when you cast a pointer to a type with stricter alignment (they could also happen with `repr(packed)`, but my bet is that those are less common). But I agree with Evan in that once the pointer is aligned it remains aligned.

nilehmann:
There may be some trade off with lazy vs strict alignment checking, e.g., if you check alignment when casting from `*const u8` to `*const u32` you may forget about alignment from there on

nilehmann:
btw, this reminds me that Zig tracks alignment in the type system. When you have `*T`, that's guaranteed to have the natural alignment of `T`. If you want a different alignment, you can specify it explicitly, e.g., `*align(1) u32`. Two pointers with the same pointee and different alignment are different. By default, you can go from a pointer with larger alignment to one with smaller, and they have mechanisms to go the other way, either adding runtime checks or causing UB when wrong.

enjhnsn2:
I think for now we can avoid explicitly tracking alignment and just use `%` when necessary. 
Because modulus is such an expensive operation, it might be worth tracking alignment in the future with an explicit argument. I think there are some clever ways to "cache the misalignment" to avoid doing lots of `%` operations, but we should only bother with that if alignment checks become a burden,

sweiglbosker:
> flux is also good at ignoring indices when you don't care about them, e.g., you can replace and index with an underscore, like a return argument that you are ignoring).

Doesn't this only work with generics?

Anyway, I messed around with the "ADTish" idea I mentioned in the original pr description. Is this close to what you had in mind?
Basically, indexing a type with multiple indices already desugars to `Indexed(BaseTy, Record([field1, field2, ...])` so I just reuse that. I also added hardcoded field projections, so stuff like `p.base` `p.addr` `p.size` works. It lowers to Tuple(3) for the solver, which is the same as the other aggregate types. RawPtr is still a primitive sort so the single index syntax also works.

The only weirdness to this approach is that the RawPtr record doesn't have a defid, so we need to handle that and lower to a `Expr::tuple` instead of `Expr::ctor_struct`.

Example:

```rs                                                                                                   
  #[spec(fn(*const[@p] i32))]                                                                                                                                                 
  fn a(p: *const i32) {}                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                     
  #[spec(fn(*const[@base, @addr, @size] i32))]                                                                                                                                
  fn b(p: *const i32) {}           
                                                                                                                                                                                   
  #[flux::sig(fn({*const[@base, @addr, @size] i32 | size > 0}))]                                                                                                                   
  fn c(p: *const i32) {}            
                                                                                                                                                                                                                                                                                                                  
  #[flux::sig(fn(*const{p: p.size > 0} i32))]    
  fn d(p: *const i32) {}                                                                                                                                                 
```

enjhnsn2:
> > flux is also good at ignoring indices when you don't care about them, e.g., you can replace and index with an underscore, like a return argument that you are ignoring).
> 
> Doesn't this only work with generics?
> 
> Anyway, I messed around with the "ADTish" idea I mentioned in the original pr description. Is this close to what you had in mind? Basically, indexing a type with multiple indices already desugars to `Indexed(BaseTy, Record([field1, field2, ...])` so I just reuse that. I also added hardcoded field projections, so stuff like `p.base` `p.addr` `p.size` works. It lowers to Tuple(3) for the solver, which is the same as the other aggregate types. RawPtr is still a primitive sort so the single index syntax also works.
> 
> The only weirdness to this approach is that the RawPtr record doesn't have a defid, so we need to handle that and lower to a `Expr::tuple` instead of `Expr::ctor_struct`.
> 
> Example:
> 
> ```rust
>   #[spec(fn(*const[@p] i32))]                                                                                                                                                 
>   fn a(p: *const i32) {}                                                                                                                                          
>                                                                                                                                                                                                                                                                                                                                      
>   #[spec(fn(*const[@base, @addr, @size] i32))]                                                                                                                                
>   fn b(p: *const i32) {}           
>                                                                                                                                                                                    
>   #[flux::sig(fn({*const[@base, @addr, @size] i32 | size > 0}))]                                                                                                                   
>   fn c(p: *const i32) {}            
>                                                                                                                                                                                                                                                                                                                   
>   #[flux::sig(fn(*const{p: p.size > 0} i32))]    
>   fn d(p: *const i32) {}                                                                                                                                                 
> ```

That implementation looks good. The 4 examples you provide cover all the bases for how we might want to refer to ptr's indices. W.r.t. to the struct without a DefId, I'll let Nico say whether that is good or bad, as the implications of that choice go a little beyond my knowledge of Flux internals

nilehmann:
The implementation points in the right direction, but there are some issues with it. I don't have time now to give a detailed review, but the short version is that _pointer structs_ should be represented in `rty` instead of using `Expr::tuple`. Ideally, I'd like to generalize `SortCtor::User` to do this, but I think it's fine to have pointers hardcoded.

lib/flux-core/src/ptr/mod.rs
```diff
@@ -2,17 +2,28 @@
...
 #[extern_spec(core::ptr)]
-#[spec(fn (src: *const{p: ptr_size(p) > 0} T) ->  T)]
+#[spec(fn (src: *const{p: p.addr >= p.base && p.size > 0} T) -> T)]
 unsafe fn read<T>(src: *const T) -> T;
 
 #[cfg(flux)]
 #[extern_spec(core::ptr)]
-#[spec(fn (dst: *mut{p: ptr_size(p) > 0} T, src: T))]
+#[spec(fn (dst: *mut{p: p.addr >= p.base && p.size > 0} T, src: T))]
 unsafe fn write<T>(dst: *mut T, src: T);
```
enjhnsn2, line 29: I think these specs need to be stronger. The correct spec, I believe, is `size_of(T) == 0 || (p.addr >= 0 && p.size >= size_of(T) && addr % align_of(T) == 0)` based on the pointer validity rules herehttps://doc.rust-lang.org/std/ptr/index.html.

 This property should for sure be pulled out to a `flux_def`

enjhnsn2: Although I suppose we can leave this to a future PR, since the specs were already weak, and it might be nice to just get this PR merged. I'll leave it to Nico.

enjhnsn2: I think I see the problem: The current spec is measuring size in elements of `T`, whereas we need to measure it in bytes, so that we might handle some of the pointer calculations that can potentially misalign pointers.

lib/flux-core/src/slice/mod.rs
```diff
@@ -101,10 +101,10 @@ impl<T> [T] {
...
 
     #[no_panic]
-    #[spec(fn(&Self[@n]) -> *const{p: ptr_size(p) == n} T)]
+    #[spec(fn(&Self[@n]) -> {base:int, addr:int. *const[base, addr, n] T | addr >= base})]
     fn as_ptr(&self) -> *const T;
 
     #[no_panic]
-    #[spec(fn(&mut Self[@n]) -> *mut{p: ptr_size(p) == n} T)]
+    #[spec(fn(&mut Self[@n]) -> {base:int, addr:int. *mut[base, addr, n] T | addr >= base})]
     fn as_mut_ptr(&mut self) -> *mut T;
```
enjhnsn2, line 109: A more precise spec could say `{base:int, addr:int. *mut[base, addr, n] T | base >= 0 && base % align_of(T) == 0 && base == addr && size == n * size_of(T)})`.

I don't love the multiplication in that spec, but we can optimize this later if it becomes a problem.

enjhnsn2 (commented):
I'll wait for Nico to pass judgement on the internal implementation, but I took a look at the flux_core specs on existing functions and they could use a little work. I'm happy to punt on the flux-core specs until after this is merged though, as it isn't really any more unsound than it was previously.

lib/flux-core/src/ptr/mod.rs
sweiglbosker: I have a branch stacked on this pr for core specs and that is pretty much what I went with. Should I include those changes here too? They don’t touch any of the same files

enjhnsn2: I think just leave it until we merge this.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -706,25 +710,36 @@ pub enum Ctor {
...
     fn is_enum(&self) -> bool {
         matches!(self, Self::Enum(..))
     }
+
+    pub fn opt_def_id(&self) -> Option<DefId> {
+        match self {
+            Self::Struct(def_id) | Self::Enum(def_id, _) => Some(*def_id),
+            Self::RawPtr => None,
+        }
+    }
```
nilehmann, line 729: The `def_id` and `variant_idx` are always accessed together (expect for one case). We could have a single function that returns an `Option<(DefId, VariantIdx)>` to make it clear it doesn't make sense to access the variant if there's no `DefId`.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -1485,6 +1535,19 @@ pub(crate) mod pretty {
                     w!(cx, f, "({:?} is {:?}::{:?})", idx, def_id, ^variant_idx)
                 }
                 ExprKind::Ctor(ctor, flds) => {
+                    if matches!(ctor, Ctor::RawPtr) {
+                        let fields = iter::zip(
+                            [
+                                Symbol::intern("base"),
+                                Symbol::intern("addr"),
+                                Symbol::intern("size"),
```
nilehmann: We could add these to the table in `flux-syntax/src/symbols.rs` (I believe `size` is already available through the re-export form `rustc`)

nilehmann (changes requested):
Hi, sorry for the delay!

Changes look good, but there are too many places where we are hardcoding (1) a mapping from index to RawPtrField and (2) the sort of each field. We should have a single source of truth for both. One idea could be to have them as methods in `RawPtrField`. We could have, for example, a method that iterates over all fields in the right order, and a method that returns the sort for that field (always int for now).

The other thing is that it may be cleaner to represent `FieldProj::RawPtr` with a `u32` as we do for the other cases instead of `RawPtrField`, but I'm not fully convinced about this. Perhaps what we need is not a `RawPtrField` but something akin to `AdtSortDef` but for `RawPtr`. This could be a collection of free functions

sweiglbosker:
Sorry for the delay on my end as well! I think I've cleaned up the things mentioned, let me know if there any other changes I should make.

Also, the lean-demo CI is failing because it seems to be using a version of flux core that depends on the old ptr UIFs. Is there something I should do about that?

ranjitjhala:
@nilehmann how do you feel about disabling the lean CI job till the dust settles a bit eg on this and the @petros-marko 's thing? (Though it's odd the demo should fail unless it's that silliness re order of kvar parameters...)

sweiglbosker:
> @nilehmann how do you feel about disabling the lean CI job till the dust settles a bit eg on this and the @petros-marko 's thing? (Though it's odd the demo should fail unless it's that silliness re order of kvar parameters...)

Based on the error message, I think the failure is probably coming from CI using flux from this branch with an older version of flux core that still relies on the old pointer UIFs. I'm not super familiar with what is going on with the flux+lean stuff but maybe it can be adjusted to pull flux and flux-core from the same revision?

nilehmann (approved):
Ok, lgtm. I'm going to merge it

I'm not sure what's going on with the lean demo. But if it keeps failing in `main`, we can turn it off until the dust settles, as Ranjit suggested.

PR #1573, opened 2026-04-22
Flux-core specs for fallible numeric conversion (`TryFrom`)

enjhnsn2:
This PR greatly expands Flux-core's support for fallible numeric conversions (i.e., `TryFrom`). In particlar, it adds:
1. A general spec for `TryFrom`, which more or less mirrors `TryInto`.
2. Macros to generate `TryFrom` specs for concrete numeric types. These macros mirror those found in core.
3. Tests for the above.

In theory, it would also be nice to implement `TryInto`, however `TryInto` is usually implemented via a blanket spec on `TryFrom`, and I have not had luck getting that working.


lib/flux-core/src/convert/mod.rs
```diff
@@ -6,3 +8,11 @@ trait TryInto<T>: Sized {
     #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>)]
     fn try_into(self) -> Result<T, Self::Error>;
 }
+
+/// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L687
+#[extern_spec(core::convert)]
+#[assoc(fn from_val(x: T, into: Self) -> bool { true })]
+trait TryFrom<T>: Sized {
+    #[spec(fn(T[@s]) -> Result<Self{v: <Self as TryFrom<T>>::from_val(s, v)}, Self::Error>)]
```
nilehmann: Does this work?

```suggestion
    #[spec(fn(T[@s]) -> Result<Self{v: Self::from_val(s, v)}, Self::Error>)]
```

lib/flux-core/src/convert/num.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use core::num::TryFromIntError;
+
+use flux_attrs::*;
+
+// no possible bounds violation — always Ok
+macro_rules! try_from_unbounded {
+    ($Src:ident => $($Dst:ident),+) => {$(
+        #[extern_spec(core::convert)]
+        impl TryFrom<$Src> for $Dst {
```
nilehmann, line 12: We must specify the `from_val` here, otherwise it defaults to `true`.

```suggestion
        #[extern_spec(core::convert)]
        #[assoc(fn from_val(x: T, into: Self) -> bool { x == into })]
        impl TryFrom<$Src> for $Dst {
```

lib/flux-core/src/convert/num.rs
```diff
@@ -0,0 +1,175 @@
...
+            fn try_from(value: $Src) -> Result<$Dst, TryFromIntError>;
+        }
+    )+}
+}
+
+// only negative bound — signed source, unsigned dest of same or wider width
+macro_rules! try_from_lower_bounded {
+    ($Src:ident => $($Dst:ident),+) => {$(
+        #[extern_spec(core::convert)]
+        impl TryFrom<$Src> for $Dst {
```
nilehmann, line 26: Same here

```suggestion
        #[assoc(fn from_val(x: T, into: Self) -> bool { x == into })]   
        impl TryFrom<$Src> for $Dst {
```

lib/flux-core/src/convert/num.rs
```diff
@@ -0,0 +1,175 @@
...
+            fn try_from(value: $Src) -> Result<$Dst, TryFromIntError>;
+        }
+    )+}
+}
+
+// only upper bound — unsigned source
+macro_rules! try_from_upper_bounded {
+    ($Src:ident => $($Dst:ident),+) => {$(
+        #[extern_spec(core::convert)]
+        impl TryFrom<$Src> for $Dst {
```
nilehmann, line 40:
```suggestion
        #[assoc(fn from_val(x: T, into: Self) -> bool { x == into })]  
        impl TryFrom<$Src> for $Dst {
```

lib/flux-core/src/convert/num.rs
```diff
@@ -0,0 +1,175 @@
...
+            fn try_from(value: $Src) -> Result<$Dst, TryFromIntError>;
+        }
+    )+}
+}
+
+// both bounds — signed source, narrower dest
+macro_rules! try_from_both_bounded {
+    ($Src:ident => $($Dst:ident),+) => {$(
+        #[extern_spec(core::convert)]
+        impl TryFrom<$Src> for $Dst {
```
nilehmann, line 54:
```suggestion
        #[assoc(fn from_val(x: T, into: Self) -> bool { x == into })] 
        impl TryFrom<$Src> for $Dst {
```

nilehmann (commented):
I haven't gone through all the macro invocations, but I assume Copilot is better than me at catching typos?

lib/flux-core/src/convert/num.rs
nilehmann: btw, not in this case, but one has to be careful about this because not specifying the associated refinements could make the annotations unsound. We should be able to catch when the unsoundness occurs, but we currently don't do any checks.

nilehmann (approved):

PR #1574, opened 2026-04-23
Refinement suggestions

cole-k:
# Basic usage
Compile with `--wick` for `xtask` (`--features=wick` otherwise). Running flux will now produce suggestions in addition to the error message.

There is now a `no_suggestions` attribute, flux.toml option, and flag to disable suggestions (granularly with the attribute, global for all else). This does nothing if not compiled with the `wick` feature.

Flux is more unstable with this feature. I've noticed crashes resulting from the WKVars auto-generated in types.

However, modulo errors caused by the removal of the 1-tuple optimization, Flux without the feature is the same as before.

# Errors
* RE: vtock, the crash is caused by the 1-tuple optimization being removed. I tested in a branch with the 1-tuple optimization added back in and it gets past the crash. It doesn't finish running though (even on main) on my machine, so I can't say for sure it's OK

# Commits
I've painstakingly separated out my changes into separate commits, the heaviest of which are the ones which add wkvars and then implement solving for them.

I recommend reviewing *everything else* first, because it's even possible for us to merge those changes into flux in separate commits if so desired.

# TODOs
There are a few TODOs in the PR. Not all of them need to be addressed and I'm pretty sure none are critical for flux to work with or without suggestions enabled.

I do need to remove (commented-out) print statements and do some more code tidying. But I'd like this PR to get review on the important parts rather than continuing to wait for me to tidy it up.

# Next steps
Aside from the TODOs, which represent smaller amounts of work, here are a few things I think should be addressed.
* Polymorphism
  - e.g. `Range` `..` doesn't work (cause a crash) because we don't/can't generate polymorphic sorts when querying Z3.
* Z3 dependency
  - In an ideal world we shouldn't need to compile against Z3. Some options include switching to calling out to Z3 like what fixpoint does or rolling our own QE.
* UIFs in QE
  - We have to fall back to a naive form of suggestion when QE fails due to UIFs. We can explore Ackermannization as well as other what other improvements would be necessary for QE to handle other theories/things Z3 doesn't like.
* Only prune invariants
  - I'm pretty sure we can fix some failures (e.g. those discussed in [#reconciliation > Special case solving for functions that return bool?](https://flux-rs.zulipchat.com/#narrow/channel/526882-reconciliation/topic/Special.20case.20solving.20for.20functions.20that.20return.20bool.3F/with/564841045)[#reconciliation > Special case solving for functions that return bool?](https://flux-rs.zulipchat.com/#narrow/channel/526882-reconciliation/topic/Special.20case.20solving.20for.20functions.20that.20return.20bool.3F/with/564841045)) by only pruning invariants, i.e. keeping all other assumptions
* Translate theory improvements
  - In the paper we made a lot of (good) simplifications to the theory. Now that the code is more manageable, I think it's worth trying to clean up what it's doing using these simplifications.

### DIFF
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -615,7 +615,9 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
             #[cfg(feature = "wick")]
             let mut fn_sig = fn_sig.hoist_input_binders();
             #[cfg(feature = "wick")]
-            if !genv.no_suggestions(def_id.local_id()) {
+            if !genv.no_suggestions(def_id.local_id())
+                && !matches!(fhir_node, fhir::Node::TraitItem(..) | fhir::Node::ForeignItem(..) | fhir::Node::ImplItem(..))
+            {
                 let id = match def_id {
                     MaybeExternId::Extern(_local_id, def_id) => def_id,
                     MaybeExternId::Local(local_id) => local_id.into(),
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -611,16 +611,15 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
             let fn_sig = AfterSortck::new(genv, &wfckresults)
                 .into_conv_ctxt()
                 .conv_fn_sig(def_id, fhir_fn_sig)?;
-            if fn_sig.vars().is_empty() {
-                fn_sig = fn_sig.hoist_input_binders();
+            let fn_sig = struct_compat::fn_sig(genv, fhir_fn_sig.decl, &fn_sig, def_id)?;
+            let mut fn_sig = fn_sig.hoist_input_binders();
+            if !genv.no_suggestions(def_id.local_id()) {
+                let id = match def_id {
+                    MaybeExternId::Extern(_local_id, def_id) => def_id,
+                    MaybeExternId::Local(local_id) => local_id.into(),
+                };
+                fn_sig = fn_sig.add_weak_kvars(genv, id)?;
             }
-            // println!("adding wkvars for {:?}", def_id);
-            fn_sig = fn_sig.hoist_input_binders();
-            let id = match def_id {
-                MaybeExternId::Extern(_local_id, def_id) => def_id,
-                MaybeExternId::Local(local_id) => local_id.into(),
-            };
-            fn_sig = fn_sig.add_weak_kvars(genv, id)?;
 
             if config::dump_rty() {
                 let generics = genv.generics_of(def_id)?;
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -611,8 +611,16 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
             let fn_sig = AfterSortck::new(genv, &wfckresults)
                 .into_conv_ctxt()
                 .conv_fn_sig(def_id, fhir_fn_sig)?;
-            let fn_sig = struct_compat::fn_sig(genv, fhir_fn_sig.decl, &fn_sig, def_id)?;
-            let fn_sig = fn_sig.hoist_input_binders();
+            if fn_sig.vars().is_empty() {
+                fn_sig = fn_sig.hoist_input_binders();
+            }
+            // println!("adding wkvars for {:?}", def_id);
+            fn_sig = fn_sig.hoist_input_binders();
+            let id = match def_id {
+                MaybeExternId::Extern(_local_id, def_id) => def_id,
+                MaybeExternId::Local(local_id) => local_id.into(),
+            };
+            fn_sig = fn_sig.add_weak_kvars(genv, id)?;
 
             if config::dump_rty() {
                 let generics = genv.generics_of(def_id)?;
--- a/crates/flux-fhir-analysis/src/lib.rs
+++ b/crates/flux-fhir-analysis/src/lib.rs
@@ -612,7 +612,9 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
                 .into_conv_ctxt()
                 .conv_fn_sig(def_id, fhir_fn_sig)?;
             let fn_sig = struct_compat::fn_sig(genv, fhir_fn_sig.decl, &fn_sig, def_id)?;
+            #[cfg(feature = "wick")]
             let mut fn_sig = fn_sig.hoist_input_binders();
+            #[cfg(feature = "wick")]
             if !genv.no_suggestions(def_id.local_id()) {
                 let id = match def_id {
                     MaybeExternId::Extern(_local_id, def_id) => def_id,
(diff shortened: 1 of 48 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -2353,7 +2353,10 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
         self_args: usize,
         scx: &mut SortEncodingCtxt,
     ) -> Option<fixpoint::Var> {
-        if !wkvid.parent_fn.is_local() {
```
cole-k: @nilehmann is what I wrote equivalent?

We don't have to filter `Extern` wkvars, but my thinking is that we should keep suggestions local to the current crate. Of course in the future we might want to allow users to tweak this.

cole-k:
@nilehmann The PR as a whole is ready for review. I want to know if you have any insight on the vtock error.

This only happens when we remove the 1-tuple optimization.

```
error: internal compiler error: crates/flux-infer/src/fixpoint_encoding.rs:770:43: fixpoint crash: CrashInfo([Array [Array [Array [Number(-1), Null], String("crash: SMTLIB2 respSat = Error \"line 166 column 1506: unknown constant fld3$36$8 (Int) declared: (declare-fun fld3$36$8 (Adt3) Adt5) \"")]], String("HasCallStack backtrace:\n  collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception\n  toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:89:42 in ghc-internal:GHC.Internal.Exception\n  throw, called at src/Language/Fixpoint/Types/Errors.hs:151:7 in liquid-fixpoint-0.9.6.3.5-inplace:Language.Fixpoint.Types.Errors\n  die, called at src/Language/Fixpoint/Smt/Interface.hs:496:19 in liquid-fixpoint-0.9.6.3.5-inplace:Language.Fixpoint.Smt.Interface\n  respSat, called at src/Language/Fixpoint/Smt/Interface.hs:465:17 in liquid-fixpoint-0.9.6.3.5-inplace:Language.Fixpoint.Smt.Interface\n  smtCheckUnsat, called at src/Language/Fixpoint/Solver/Monad.hs:194:16 in liquid-fixpoint-0.9.6.3.5-inplace:Language.Fixpoint.Solver.Monad\n\n:1:1-1:1: Error\n  crash: SMTLIB2 respSat = Error \"line 166 column 1506: unknown constant fld3$36$8 (Int) declared: (declare-fun fld3$36$8 (Adt3) Adt5) \"")])
    --> arch/cortex-m/src/mpu.rs:1169:5
     |
1169 |     fn configure_mpu(&self, config: &RArray<CortexMRegion>, id: usize, is_dirty: bool) {
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

SMT2 without 1-tuple opt
[configure-mpu-bad.smt2.txt](https://github.com/user-attachments/files/27045873/configure-mpu-bad.smt2.txt)

SMT2 with 1-tuple opt
[configure-mpu-good.smt2.txt](https://github.com/user-attachments/files/27045874/configure-mpu-good.smt2.txt)

The error happens in the SMT query that fixpoint makes to Z3 and I honestly have no clue what it means since we're defining the function. In my tests, you don't need a constant declared before you define it as a function, unless I'm mistaken?

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -612,7 +612,18 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
...
             let fn_sig = struct_compat::fn_sig(genv, fhir_fn_sig.decl, &fn_sig, def_id)?;
-            let fn_sig = fn_sig.hoist_input_binders();
+            #[cfg(feature = "wick")]
+            let mut fn_sig = fn_sig.hoist_input_binders();
+            #[cfg(feature = "wick")]
+            if !genv.no_suggestions(def_id.local_id())
+                && !matches!(fhir_node, fhir::Node::TraitItem(..) | fhir::Node::ForeignItem(..) | fhir::Node::ImplItem(..))
+            {
+                let id = match def_id {
+                    MaybeExternId::Extern(_local_id, def_id) => def_id,
```
cole-k, line 622: Use resolved_id

crates/flux-infer/src/fixpoint_encoding/decoding.rs
```diff
@@ -187,7 +162,17 @@ where
...
+                        // HACK: this is a free var generated from an existential
+                        // we got from fixpoint. We just will add enough to it so it
+                        // is likely to not conflict with exisitng free vars, but in an
+                        // ideal world we would properly add it to the fvars map and
+                        // make a fresh name for it.
+                        //
+                        // But since we just need to be consistent about the value we
+                        // give these vars, this should suffice for now --- they won't
+                        // actually appear in any solutions because we anti-unify.
+                        Ok(rty::Expr::fvar(rty::Name::from_u32(fname.as_u32() + 100_000)))
```
cole-k, line 174: Check that we still need this

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -912,11 +973,14 @@ where
     where
         Tag: std::fmt::Debug,
     {
-        *self.tags_inv.entry(tag).or_insert_with(|| {
-            let idx = self.tags.push(tag);
-            self.comments.push(format!("Tag {idx}: {tag:?}"));
-            idx
-        })
+        // Once we added refinement suggestions, we now need to have unique tags for each
```
cole-k, line 976: See above comment.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -687,16 +730,41 @@ where
...
+                        let tag = self.tags[tag_idx];
+                        #[cfg(not(feature = "wick"))]
+                        let possible_solutions = Default::default();
+                        #[cfg(feature = "wick")]
+                        let possible_solutions = if let Some(suggestion_ctx) = &suggestion_ctx {
+                            find_possible_solutions(self, tag_idx, &suggestion_ctx)
+                        } else {
+                            Default::default()
+                        };
+                        FixpointCheckError::new(tag, possible_solutions)
```
cole-k, line 766: Ideally: dedup the errors by tag but somehow have different suggestions per error.

QUESTION: when do we have multiple heads per the same tag?

The issue that may arise is if we fold all suggestions for fixpoint tags that correspond to the same rty tag, then we're pretending like all of these suggestions are for the SAME error but they are not.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -2248,6 +2339,57 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+    /// Currently returns an Option in case weak kvars generated automatically
+    /// don't track the sorts of their arguments, but there is no reason why
+    /// they shouldn't --- we could probably unwrap the Option.
+    fn define_const_for_wkvar(
+        &mut self,
+        wkvid: &rty::WKVid,
+        self_args: usize,
+        scx: &mut SortEncodingCtxt,
+    ) -> Option<fixpoint::Var> {
+        if !matches!(
```
cole-k, line 2355: This should not appear here but appear in wherever we generate the wkvars

cole-k (commented):
We want three commits in the PR:

* Stuff that is benign
* Stuff behind a feature flag
* Stuff that is neither behind nor behind a feature flag

PR #1575, opened 2026-04-23
Blanket `TryInto` spec

enjhnsn2:
This PR adds a `TryInto` spec for everything that implements `TryFrom`, mirroring `TryInto`'s blanket impl in core. Specifically this PR:
1. adds a `succeeds` assoc to `TryFrom` and `TryInto`  that describes the discriminant on the `Result` type returned from `TryFrom` and `TryInto`
2. Added a blanket spec for `TryInto` that inherits from `TryFrom`, like in core.
3. added tests for the above

### DIFF
--- a/lib/flux-core/src/convert/mod.rs
+++ b/lib/flux-core/src/convert/mod.rs
@@ -4,27 +4,27 @@ use flux_attrs::*;
 
 /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L612
 #[extern_spec(core::convert)]
-#[assoc(fn succeeds(x: Self) -> bool { true })]
-#[assoc(fn into_val(x: Self, into: T) -> bool { true })]
+#[assoc(fn succeeds(s: Self, out: Result) -> bool { true })]
+#[assoc(fn into_val(s: Self, into: T) -> bool { true })]
 trait TryInto<T>: Sized {
-    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>[Self::succeeds(s)])]
+    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>{out: Self::succeeds(s, out)})]
     fn try_into(self) -> Result<T, Self::Error>;
 }
 
 /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L687
 #[extern_spec(core::convert)]
-#[assoc(fn succeeds(x: T) -> bool { true })]
-#[assoc(fn from_val(x: T, into: Self) -> bool { true })]
+#[assoc(fn succeeds(s: T, out: Result) -> bool { true })]
+#[assoc(fn from_val(s: T, into: Self) -> bool { true })]
 trait TryFrom<T>: Sized {
-    #[spec(fn(T[@x]) -> Result<Self{v: Self::from_val(x, v)}, Self::Error>[Self::succeeds(x)])]
+    #[spec(fn(T[@s]) -> Result<Self{v: Self::from_val(s, v)}, Self::Error>{out: Self::succeeds(s, out)})]
     fn try_from(value: T) -> Result<Self, Self::Error>;
 }
 
 /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L740
 #[extern_spec(core::convert)]
-#[assoc(fn succeeds(x: T) -> bool { <U as TryFrom<T>>::succeeds(x) })]
-#[assoc(fn into_val(x: T, into: U) -> bool { <U as TryFrom<T>>::from_val(x, into) })]
+#[assoc(fn succeeds(s: T, out: Result) -> bool { <U as TryFrom<T>>::succeeds(s, out) })]
+#[assoc(fn into_val(s: T, into: U) -> bool { <U as TryFrom<T>>::from_val(s, into) })]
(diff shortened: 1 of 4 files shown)


lib/flux-core/src/convert/mod.rs
```diff
@@ -2,17 +2,29 @@ mod num;
 
 use flux_attrs::*;
 
+/// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L612
 #[extern_spec(core::convert)]
+#[assoc(fn succeeds(x: Self) -> bool { true })]
 #[assoc(fn into_val(x: Self, into: T) -> bool { true })]
 trait TryInto<T>: Sized {
-    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>)]
+    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>[<Self as TryInto<T>>::succeeds(s)])]
```
nilehmann:
```suggestion
    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>[Self::succeeds(s)])]
```

lib/flux-core/src/convert/mod.rs
```diff
@@ -2,17 +2,29 @@ mod num;
...
     fn try_into(self) -> Result<T, Self::Error>;
 }
 
 /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/mod.rs#L687
 #[extern_spec(core::convert)]
+#[assoc(fn succeeds(x: T) -> bool { true })]
 #[assoc(fn from_val(x: T, into: Self) -> bool { true })]
 trait TryFrom<T>: Sized {
-    #[spec(fn(T[@s]) -> Result<Self{v: Self::from_val(s, v)}, Self::Error>)]
+    #[spec(fn(T[@x]) -> Result<Self{v: Self::from_val(x, v)}, Self::Error>[<Self as TryFrom<T>>::succeeds(x)])]
```
nilehmann:
```suggestion
    #[spec(fn(T[@x]) -> Result<Self{v: Self::from_val(x, v)}, Self::Error>[Self::succeeds(x)])]
```

nilehmann (approved):
Looks great!

nilehmann (changes requested):
Actually, I spoke too soon. This makes all existing implementations without extra refinements promise that they always return `Ok`. We should make `succeed` a predicate on the output rather than specifying its exact value, like so

```rust
#[assoc(fn succeeds(x: Self, out: Result) -> bool { true })] 
#[assoc(fn into_val(x: Self, into: T) -> bool { true })]
trait TryInto<T>: Sized {
    #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>{out: Self::succeeds(s, out)})]
}

```

and then

```rust
macro_rules! try_from_lower_bounded {
    ($Src:ident => $($Dst:ident),+) => {$(
        #[extern_spec(core::convert)]
        #[assoc(fn succeeds(x: int, out: Result) -> bool { out.is_ok == x >= 0 })]
        #[assoc(fn from_val(x: int, into: int) -> bool { x == into })]
        impl TryFrom<$Src> for $Dst {
            #[no_panic]
            /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/num.rs#L271
            #[spec(fn($Src[@x]) -> Result<$Dst[x], _>[x >= 0])]
            fn try_from(value: $Src) -> Result<$Dst, TryFromIntError>;
        }
    )+}
}
```

enjhnsn2:
> Actually, I spoke too soon. This makes all existing implementations without extra refinements promise that they always return `Ok`. We should make `succeed` a predicate on the output rather than specifying its exact value, like so
> 
> ```rust
> #[assoc(fn succeeds(x: Self, out: Result) -> bool { true })] 
> #[assoc(fn into_val(x: Self, into: T) -> bool { true })]
> trait TryInto<T>: Sized {
>     #[spec(fn(Self[@s]) -> Result<T{v: Self::into_val(s, v)}, Self::Error>{out: Self::succeeds(s, out)})]
>     fn try_into(self) -> Result<T, Self::Error>;
> }
> ```
> 
> and then
> 
> ```rust
> macro_rules! try_from_lower_bounded {
>     ($Src:ident => $($Dst:ident),+) => {$(
>         #[extern_spec(core::convert)]
>         #[assoc(fn succeeds(x: int, out: Result) -> bool { out.is_ok == x >= 0 })]
>         #[assoc(fn from_val(x: int, into: int) -> bool { x == into })]
>         impl TryFrom<$Src> for $Dst {
>             #[no_panic]
>             /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/convert/num.rs#L271
>             #[spec(fn($Src[@x]) -> Result<$Dst[x], _>[x >= 0])]
>             fn try_from(value: $Src) -> Result<$Dst, TryFromIntError>;
>         }
>     )+}
> }
> ```

ah, you're right. I think you can still avoid the existential if you define the default `succeeds` as a UIF rather than as `true`. 
I even thought of this before-hand, but forgot to check that the Claude output did this smh... The danger of getting too comfortable with LLMs

nilehmann:
We've gone with the uif route before but found some issues when actually trying to use it. I can't remember what is was though 🤔 

@ranjitjhala, do you recall?

ranjitjhala:
Basically you can never ever *prove* the UIF … rendering it useless…?

- Ranjit.

enjhnsn2:
> Basically you can never ever *prove* the UIF … rendering it useless…? - Ranjit.

Well, that's what we'd be going for: the equivalent of `{v: true}` where we can say nothing about that case.

nilehmann (approved):

PR #1576, opened 2026-04-24
Tweak running instructions

ranjitjhala:
To put the `cargo flux` -- the more common use-case -- first.

(Also to trigger the CI on the f2l demo)


ranjitjhala:
subsumed by #1582

PR #1577, opened 2026-04-24
Switch from using hashmap to indexmap

petros-marko:
Had to change a dependency, which I think is why I didn't do it when I first implemented.


ranjitjhala:
awesome! I'll merge this if/when the non-lean CI tasks pass (the lean one will break because...)

nilehmann (approved):

ranjitjhala:
lol looks like the lean thing works now because the params are again in the right order haha!

PR #1578, opened 2026-04-26
Add explicit check for proof matching theorem type

petros-marko:
Instead of just type checking the proof files, add a (generated) file for each theorem-proof pair to `#check` that the proof is indeed a proof of the theorem statement.


nilehmann (approved):
LGTM I don't know if @ranjitjhala wants to take a look.

ranjitjhala:
LGTM too - I thought we'd just add the check to Basic.lean but looks like there is a separate file per check? (Probably better)

PR #1581, opened 2026-05-01
Bitvector related fixes for flux tests with lean-fixpoint

petros-marko:
Convert prelude functions to `abbrev`-s for, correctly translate the zero/sign extensions, fix bad formatting coming from bit vector size being treated as a type argument.


crates/flux-infer/src/lean_format.rs
```diff
@@ -420,6 +420,9 @@ impl LeanFmt for Expr {
                 function.as_ref().lean_fmt(f, cx)?;
                 if let Some(sort_args) = sort_args {
                     for (i, s_arg) in sort_args.iter().enumerate() {
+                        if matches!(s_arg, Sort::BvSize(..)) {
+                            continue;
+                        }
```
nilehmann, line 427: I don't understand what exactly is going on here. Can you leave a comment?

nilehmann: I guess what I don't understand is why we are skipping `Sort::BvSize` here but also adding the `Sort::BvSize(size) => write!(f, "{size}")` case

petros-marko: I added the bv-size case for completeness, although with this filtering it should never happen. I guess it's better to leave the panic there so that we get an error there as opposed to somewhere harder to understand.

nilehmann (approved):

PR #1589, opened 2026-05-09
Include namespace in UnresolvedPath error

### LINKED ISSUE #877: Improve error message when type cannot be resolved
Consider the following

```rust
#[flux::opaque]
#[flux::refined_by(b : bitvec <8>)]
pub struct BV8(u8);

impl BitAnd for BV8 {
    type Output = BV8;
    
    #[flux::trusted]
    #[flux::sig(fn(BV8 [@b1], BV8 [@b2]) -> bv_and(b1, b2))]
    fn bitand(self, rhs: BV8) -> BV8 { BV8(self.0 & rhs.0) }
}
```

It currently fails with 

```
error[E0999]: cannot resolve `bv_and`
  --> attic/playground.rs:11:45
   |
11 |     #[flux::sig(fn(BV8 [@b1], BV8 [@b2]) -> bv_and(b1, b2))]
   |                                             ^^^^^^^^^^^^^^
   |
   = help: flux can only resolve a path if it is present in the definition being refined
```

`bv_and` is parsed as a type, but the error doesn't mention type anywhere. The `help: ` is also not true since we implemented the global resolver so it should be removed. 

We should give an error similar to what rustc gives

```
error[E0999]: cannot resolve type `bv_and` in this scope
  --> attic/playground.rs:11:45
   |
11 |     #[flux::sig(fn(BV8 [@b1], BV8 [@b2]) -> bv_and(b1, b2))]
   |                                             ^^^^^^^^^^^^^^
   |                                             not found in this scope
```

kimjune01:
Shows "cannot resolve type `X` in this scope" instead of just "cannot resolve `X`". Closes #877.

### DIFF
--- a/crates/flux-desugar/src/resolver.rs
+++ b/crates/flux-desugar/src/resolver.rs
@@ -879,16 +879,10 @@ mod errors {
 
     impl UnresolvedPath {
         pub fn new(path: &surface::Path, ns: rustc_hir::def::Namespace) -> Self {
-            use rustc_hir::def::Namespace;
-            let ns = match ns {
-                Namespace::TypeNS => "type",
-                Namespace::ValueNS => "value",
-                Namespace::MacroNS => "macro",
-            };
             Self {
                 span: path.span,
                 path: path.segments.iter().map(|segment| segment.ident).join("::"),
-                ns,
+                ns: ns.descr(),
             }
         }
     }
--- a/crates/flux-desugar/src/resolver.rs
+++ b/crates/flux-desugar/src/resolver.rs
@@ -732,7 +732,7 @@ impl<'a, 'genv, 'tcx> ItemResolver<'a, 'genv, 'tcx> {
                 .path_res_map
                 .insert(path.node_id, partial_res);
         } else {
-            self.errors.emit(errors::UnresolvedPath::new(path));
+            self.errors.emit(errors::UnresolvedPath::new(path, ns));
         }
     }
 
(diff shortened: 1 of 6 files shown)


nilehmann:
Thanks @kimjune01

Looks like CI errors are not related to your changes but are due to a change in liquid fixpoint.

@ranjitjhala do you know what's going on? 

```
error: internal compiler error: crates/flux-infer/src/fixpoint_encoding.rs:892:39: failed to run fixpoint: fixpoint: Uncaught exception ghc-internal:GHC.Internal.IO.Exception.IOException:
                                
                                user error (unrecognized option `--sortedsolution'
                                
                                Use --help for usage information.)
                                
                                HasCallStack backtrace:
                                  ioError, called at src/Language/Fixpoint/Types/Config.hs:466:27 in liquid-fixpoint-0.9.6.3.6-inplace:Language.Fixpoint.Types.Config
```

crates/flux-desugar/src/resolver.rs
```diff
@@ -871,15 +871,24 @@ mod errors {
...
 
     impl UnresolvedPath {
-        pub fn new(path: &surface::Path) -> Self {
+        pub fn new(path: &surface::Path, ns: rustc_hir::def::Namespace) -> Self {
+            use rustc_hir::def::Namespace;
+            let ns = match ns {
+                Namespace::TypeNS => "type",
+                Namespace::ValueNS => "value",
+                Namespace::MacroNS => "macro",
+            };
```
nilehmann: There's a method that does exactly this

https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/def/enum.Namespace.html#method.descr

nilehmann (changes requested):
Ok, I fixed CI in https://github.com/flux-rs/flux/pull/1590

PR looks good, just one nit.

crates/flux-desugar/src/resolver.rs
kimjune01: Good call — replaced the manual match with `ns.descr()`.

nilehmann (approved):
Thanks, @kimjune01 looks good!

Issue #858, opened 2024-10-22
Do not pretty print fixpoint constraints by default


nilehmann:
We currently pretty print fixpoint horn constraints in the `Display` implementation, i.e., we add newlines and indentation to make the expression easier to read. This is wasteful if we are not looking at the constraint and it has caused problems in the past (e.g., https://github.com/flux-rs/flux/issues/857#issuecomment-2430070077). I still think is valuable to pretty print them in cases where we do look at them. We should disable pretty printing by default and implement a flag to enable it on demand. 

The `PadAdapter` is probably not the most efficient implementation. We could use normal functions instead of implementing `Display` and use the `write!` macro to avoid wrapping the formatter to get the padding behavior.

nilehmann:
`PadAdapter` was removed in https://github.com/flux-rs/flux/commit/c5a037f61d520e8b6149db4f5aaee9b64ee1fabf

pqcfox:
Interested in helping out with the remainder of this issue (flag for pretty-printing fixpoint horn constraints)!

Will give this a shot either this weekend or next :)

PR #1592, opened 2026-05-11
Use compact formatting for fixpoint binary communication

kimjune01:
Addresses #858.

Adds a `pretty` flag to `ConstraintFormatter` so constraints can be formatted compactly (spaces instead of newlines/indentation) when writing to the fixpoint binary via stdin. The `Display` impl remains unchanged for debug output.

- `fmt_constraint_compact` skips newlines and indentation
- `Task::fmt_compact()` uses compact constraint formatting with explicit `writeln!` separators between top-level S-expressions
- `Task::run()` now calls `fmt_compact()` instead of `Display`

### DIFF
--- a/lib/liquid-fixpoint/src/lib.rs
+++ b/lib/liquid-fixpoint/src/lib.rs
@@ -322,34 +322,8 @@ impl<T: Types> Task<T> {
         hasher.finish()
     }
 
-    fn fmt_compact(&self) -> String {
-        struct CompactTask<'a, T: Types>(&'a Task<T>);
-
-        impl<T: Types> fmt::Display for CompactTask<'_, T> {
-            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                if self.0.scrape_quals {
-                    writeln!(f, "(fixpoint \"--scrape=both\")")?;
-                }
-                for data_decl in &self.0.data_decls {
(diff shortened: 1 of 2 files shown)


lib/liquid-fixpoint/src/lib.rs
```diff
@@ -322,6 +322,36 @@ impl<T: Types> Task<T> {
...
+                }
+                for kvar in &self.0.kvars {
+                    writeln!(f, "{kvar}")?;
+                }
+                format::fmt_constraint_compact(&self.0.constraint, f)
+            }
+        }
+
+        CompactTask(self).to_string()
+    }
```
nilehmann: This unnecessarily allocates a `String` for the entire constraint instead of formatting it directly into the process's standard input. It is also duplicating the code to format the task. Can we implement it without the duplication and the extra allocation?

kimjune01:
Refactored to eliminate both issues. Moved the formatting logic into a shared `fmt_task(task, formatter, pretty)` function in format.rs. The CompactTask wrapper now just calls `fmt_task(self.0, f, false)` instead of duplicating all the loops, and we write directly to stdin using `write!(w, "{}", format::CompactTask(self))` which formats into the writer's buffer without an intermediate String allocation.

nilehmann (approved):
Thanks @kimjune01

PR #1593, opened 2026-05-11
Add extern spec for `SliceIndex` to `flux-core`

### LINKED ISSUE #1588: Track length of `SliceIndex` output
The following [example](https://flux.goto.ucsd.edu/?code=TcvtCoIwFMbx77uKJ4RQWOLUIJS6kcMYvcwYzQXNURHdeyPSPB_P__ckC9pbe72nwQWvT5lkLKHOhoe6-abx5px2Ls2w2sFUJYlCRtE59Bc16H_AiyGe1QP6p_LWHHUDCpsWa4ktSHCUHBVHzeOnnbAPhy-OZjkOqczzOpoJuVhHSMVs3c-D-AVRsDf7AA==) in the playground fails to check:

```rs
#![allow(unused)]

#[flux_rs::sig(fn() -> i32[10])]
fn mk_ten() -> i32 {
    let my_slice: [u8; 5] = [1, 2, 3, 4, 5];
    let subslice = &my_slice[2..4];

    let n = subslice[0]; // assertion might fail: possible out-of-bounds access
    let m = subslice[1]; // assertion might fail: possible out-of-bounds access
    10
}
```

My intuition is that `subslice` has type `&[u8]`, but its length isn't connected back to the `2..4` that produced it, so the indexing explodes.

I think what's missing is an extern spec on the [`SliceIndex::index` trait method](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html#tymethod.index) saying that its output is valid. Perhaps something in the spirit of `valid_item` in the [iterator extern spec](https://github.com/flux-rs/flux/blob/0dcdeaa512e7661b7079cdfe3b8885c4d6696f67/lib/flux-core/src/iter/traits/iterator.rs#L11).

```rs
// this is not 100% right
#[extern_spec(core::slice)]
#[assoc(fn output_pred(idx: Self, slice: &T, out: Self::Output) -> bool)]
trait SliceIndex<T> {
    #[sig(fn(self, slice: &T) -> &Self::Output{o: <Self as SliceIndex<T>>::output_pred(self, slice, o)})]
    fn index(self, slice: &T) -> &Self::Output;
}

#[assoc(fn output_pred(idx: Range<usize>, slice: [T], out: [T]) -> bool {
    out.len() == idx.end - idx.start // this keeps track of the output?
})]
impl<T> SliceIndex<[T]> for Range<usize> {
    #[sig(fn(self, slice: &T) -> &Self::Output{o: <Self as SliceIndex<T>>::output_pred(self, slice, o)})]
    fn index(self, slice: &T) -> &Self::Output;
}
```

ninehusky:
Closes #1588


ninehusky:
Oops, realized this is working from _my_ main, which is different from the actual main. Will reopen.

PR #1595, opened 2026-05-12
Validate self type in extern specs for trait impls

### LINKED ISSUE #833: Check that the self type in an extern spec for a trait impl matches the external definition.
We currently accept nonsensical extern specs on trait implementations like this one

```rust
use flux_rs::*;

#[extern_spec(std::ops)]
impl<A: Step> Iterator for Range<usize> {}
```
Given the way we expand the macro, `A` won't be reported as non-constrained, and `Range<usize>` passes further validations because it is less generic than the extern implementation so it still implements `Iterator`. We should check that the self-type of extern specs for trait implementations matches the external definition.

kimjune01:
## Summary
- Add `check_extern_impl_self_ty` to validate that the self type in an extern spec matches the external trait impl
- Emit a clear diagnostic (`driver_mismatched_impl_self_ty`) when they diverge, instead of silently accepting a wrong spec
- Fixes #833

## Test plan
- Added negative test `mismatched_impl_self_ty.rs` that verifies the error is emitted for `impl Iterator for Range<usize>` (wrong self type)
- Existing test suite passes

### DIFF
--- a/crates/flux-driver/locales/en-US.ftl
+++ b/crates/flux-driver/locales/en-US.ftl
@@ -89,3 +89,9 @@ driver_mismatched_generics =
     .label = generic parameters don't match the external {$def_descr}
     .extern_def_label = external {$def_descr} found here
     .note = extern specs must exactly match the external definition, including the list of generic parameters and their names
+
+driver_mismatched_impl_self_ty =
+    invalid extern spec for trait impl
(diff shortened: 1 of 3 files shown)


nilehmann (approved):
Thanks @kimjune01

PR #1596, opened 2026-05-12
Emit Primop constants as axioms

petros-marko:
Allows tests/pos/surface/primop_prop_let.rs and tests/pos/surface/primops00.rs to pass.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -840,7 +840,7 @@ where
                 interpreted.push((decl, gvar_eq_map.remove(&gvar).unwrap()));
             }
         }
-        let const_deps = ConstDeps { interpreted };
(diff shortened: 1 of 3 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -852,7 +852,17 @@ where
...
+            .ecx
+            .const_env
+            .const_map
+            .iter()
+            .filter_map(|(key, decl)| {
+                if let ConstKey::PrimOp(op) = key { Some((decl.clone(), op.clone())) } else { None }
+            })
+            .collect();
+        let (mut const_deps, constraint) = self.compute_const_deps(task.constants, task.constraint);
+        const_deps.opaque = primop_consts;
```
nilehmann: Why not compute this inside `compute_const_deps`?

nilehmann (approved):

PR #1601, opened 2026-05-14
add minimal spec for maybeuninit

enjhnsn2:
This PR adds a simple spec for `MaybeUninit<T>` that refines `MaybeUninit` by a bool specifying whether this value has been initialized. The `uninit` constructor sets it to false, `write` sets it to true, and `assume_init` (essentially it's "unwrap" function) requires the index to be true. Simple stuff.


ranjitjhala (approved):
LGTM

ranjitjhala:
The `lean-demo` seems borked but that seems totally unrelated to this PR so am merging.

PR #1605, opened 2026-05-17
un/curry parsing of primop-uif-applications

ranjitjhala:
This fixes-ish a bug in the code to convert fixpoint::Expr -> rty::Expr; the UIF for `>>` and such like bin-op expects two arguments, but the fixpoint-uif encoding is curried and so looks like `((c0 arg1) arg2)`. I'm not entirely sure why *this* needs special casing: why isn't it like _other_ UIFs? 
- Is fixpoint not "uncurrying" the solutions? (seems unlikely, what about other UIFs, which are presumably treated identically by fixpoint?)
- Is this primop-uif encoded in some special "curried" manner (unlike other UIFs which are encoded fully applied?) 

At any rate, this patch deals with the current crash... @cole-k do you have any theories? or @petros-marko ?

### DIFF
--- a/tests/tests/pos/surface/primops00.rs
+++ b/tests/tests/pos/surface/primops00.rs
@@ -72,3 +72,10 @@ pub fn test6(c: char) {
     let c = c | 1;
     flux_rs::assert(c <= 0x10FFFF); // (1)
 }
+
+#[spec(fn (b:bool) -> usize{v: v <= 7})]
+pub fn test7(b: bool) -> usize {
+    let n: usize = 10;
+    let m = if b { n & 7 } else { 6 };
+    m
+}
--- a/crates/flux-infer/src/fixpoint_encoding/decoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding/decoding.rs
@@ -287,16 +287,10 @@ where
                                 // NOTE: Only a few of these are meaningfully needed,
                                 // e.g. ConstKey::Alias because the rty Expr has its
                                 // args as a part of it.
-                                ConstKey::PrimOp(bin_op) => {
-                                    if fargs.len() != 2 {
-                                        Err(FixpointParseError::PrimOpArityMismatch(fargs.len()))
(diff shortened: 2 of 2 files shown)


tests/tests/pos/surface/primops00.rs
```diff
@@ -72,3 +72,10 @@ pub fn test6(c: char) {
     let c = c | 1;
     flux_rs::assert(c <= 0x10FFFF); // (1)
 }
+
+#[spec(fn (b:bool) -> usize{v: v <= 7})]
+pub fn test7(b: bool) -> usize {
+    let n: usize = 10;
+    let m = if b { n & 7 } else { 6 };
```
ranjitjhala, line 79: This branch introduces a KVar which makes this test crash if you run with `-Fdump-checker-trace=info`

cole-k:
I'm pretty sick so getting out of bed to quickly answer because I think I know the answer to this one. I definitely saw curried outputs from fixpoint and I wrote this code to get around it https://github.com/flux-rs/flux/blob/ck-refinement-suggestions/lib/liquid-fixpoint/src/constraint.rs#L851

I want to say it happened probably at the same time we started getting all of the casts appearing in its outputs but my memory could be wrong.

May be slow to answer follow-up questions, but hope this helps...

Edit: this is where I added the uncurry https://github.com/flux-rs/flux/blob/ck-refinement-suggestions/lib/liquid-fixpoint/src/constraint.rs#L851

ranjitjhala:
@cole-k is your code in `main` or no? (Or will it soon be in `main` and hence will subsume this PR?)

cole-k:
> @cole-k is your code in `main` or no? (Or will it soon be in `main` and hence will subsume this PR?)

No and it won't soon be in `main` unless we're fine with just shoving a bunch of code in under a compilation flag + feature flag to turn off suggestions.

ranjitjhala:
@cole-k merging this per discussion this morning, thanks!!!

PR #1610, opened 2026-05-19
No panic inference

nilehmann:


nilehmann:
@ninehusky, have you had a chance to try this out? If it's working reasonably well in the stuff you are working on, I'd like to merge it.

enjhnsn2:
Performance seems to have regressed and this version of no panic inference cannot handle generic `core` functions. For example, I get a ton of `MightPanic(NotInCallGraph)` errors: 
```
error[E0999]: call to core[4d41]::result::{impl#28}::from_residual may panic: MightPanic(NotInCallGraph)
   --> src/bst.rs:421:28
    |
421 |             free_list[i] = Self::read_arenaid(&b[i * 8..(i + 1) * 8])?;
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0999]: call to core[4d41]::result::{impl#27}::branch may panic: MightPanic(NotInCallGraph)
   --> src/bst.rs:214:9
    |
214 |         Self::validate_key(key)?;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^
```


These panics were handled by my (ugly) implementation

nilehmann:
Your ugly implementation was unsound.

nilehmann:
> Performance seems to have regressed

@enjhnsn2 by performance, do you mean how long it takes to run? Or do you mean this analysis is less precise?

In any case, the old implementation was unsound and a mess. I'm much more confident that this implementation is correct. We can iterate on the precision if people give specific examples.

enjhnsn2:
I meant that it was less precise. I'll try to isolate a few examples that the previous impl could infer that this one cannot.

ninehusky:
Let me look into it now -- will update with numbers.

ninehusky:
OK, I reran both your/my implementation of no-panic inference on an unannotated `capsules/core`.

The breakdown is that my implementation flagged 154 sites, and yours flagged 170 (diff of +16).

A quick summary:
-  148 are flagged in both implementations
- 6 are only flagged in mine
- 22 are only flagged in yours

Only errored in mine (i.e., places that your implementation resolved) (6):
 - `grant::iter`
  - 3x of`.into()`
  - `processbuffer::get`
  - `cmp::min`

Only errored in yours (22):
  - 18× of `slice::len`
  - 2× take_cell::put
  - 1× take_cell::is_none
  - 1× Iterator::take

TL;DR: If the thing blocking merging is "does this work reasonably well", then I'm OK to merge, but let me know if you want to dig into any of these further. In particular, I think that failed inference of `take_cell:*` is a bit concerning because those should all resolve to concrete nodes in the callgraph (they're inside tock).

nilehmann:
@ninehusky do the functions where a call to `slice::len` is flagged as panic also have other calls that do genuily (potentially) panic?

nilehmann:
Also, are the calls to `slice::len` flagged as panic monomorphic?

ninehusky:
> do the functions where a call to slice::len is flagged as panic also have other calls that do genuily (potentially) panic?

Yes. Every function that surfaces a call to `slice::len` as `MightPanic` also has a different `MightPanic` error for a different function. (There is one special case where a `len` error is emitted from a macro).

> are the calls to slice::len flagged as panic monomorphic?

I don't think so? Here is an error that surfaces from calling `len` on a `&[u8]`:

```
error[E0999]: call to core[9ef1]::slice::{impl#0}::len may panic: MightPanic(NotInCallGraph)
   --> capsules/core/src/stream.rs:261:11
    |
261 |     buf[..bs.len()].copy_from_slice(bs);
    |           ^^^^^^^^
```

nilehmann:
> I don't think so? Here is an error that surfaces from calling len on a &[u8]:

That means it's monomorphic: you are calling `len` on a slice `&[T]` where `T = u8`.


This is consistent with what I expected. The issue is that to support monomorphized calls like Evan wants, we need to change the way we represent the panic spec map to index by `Instance` instead of `DefId` and then somehow resolve to the same instance in the `Checker`.

enjhnsn2:
If it helps, I believe instance objects have the defid embedded in them, so it should be easy to map them back to defids if necessary. I was going to fetch some more concrete examples of monomorphization false negatives, but this `len` one is good.

nilehmann:
As discussed in the meeting, I'm going to merge this. We can implement support for monomorphization properly later.

PR #1594, opened 2026-05-12
Add `SliceIndex` associated refinement tracking length

ninehusky:
Closes #1593.

Also adds extern specs for:
- `SliceIndex<[T]> for ops::Range<usize>`
- `SliceIndex<[T]> for ops::RangeTo<usize>`
- `SliceIndex<[T]> for ops::RangeFrom<usize>`

PR #1617, opened 2026-05-22
Slice index specs for Arrays

ranjitjhala:
Extends #1594 to support arrays `[T; N]`. Note that arrays themselves are indexed by unit, but we have access to the `N` and can use the associated refinement for plain slices to specify the contract for arrays...

### DIFF
--- a/lib/flux-core/src/array/mod.rs
+++ b/lib/flux-core/src/array/mod.rs
@@ -8,22 +8,16 @@ impl<T, I, const N: usize> Index<I> for [T; N]
 where
     [T]: Index<I>,
 {
+    #![assoc(
+        fn in_bounds(len: (), idx: I) -> bool {
+            <[T] as Index<I>>::in_bounds(N, idx)
+        }
+
+        fn output_pred(len: (), idx: I, out: <[T] as Index<I>>::Output) -> bool {
+            <[T] as Index<I>>::output_pred(N, idx, out)
+        }
+    )]
+
     #[sig(fn(&Self, {I[@idx] | <[T] as Index<I>>::in_bounds(N, idx)}) -> &<[T; N] as Index<I>>::Output{out: <[T] as Index<I>>::output_pred(N, idx, out)})]
     fn index(&self, index: I) -> &<[T; N] as Index<I>>::Output;
-}
-
-#[cfg(flux_sysroot_test)]
-mod tests {
-    #![allow(dead_code)]
-
-    use flux_attrs::*;
-
-    fn test00(xs: &[i32; 100]) {
-        let _y = &xs[0..10];
-    }
-
-    #[should_fail]
-    fn test01(xs: &[i32; 10]) {
-        let _y = &xs[0..20];
-    }
 }
--- a/lib/flux-core/src/array/mod.rs
+++ b/lib/flux-core/src/array/mod.rs
@@ -1,5 +1,4 @@
 #[cfg(flux)]
-use core::ops;
 use core::ops::Index;
 
 use flux_attrs::*;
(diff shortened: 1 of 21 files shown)


lib/flux-core/src/slice/index.rs
```diff
@@ -3,14 +3,22 @@ use core::ops;
...
     #![assoc(
         fn in_bounds(len: int, idx: I) -> bool {
             <I as SliceIndex<[T]>>::in_bounds(idx, len)
         }
     )]
-
+    #![assoc(
+        fn output_pred(len: int, idx: I, out: <I as SliceIndex<[T]>>::Output) -> bool {
+            <I as SliceIndex<[T]>>::output_pred(idx, len, out)
+        }
```
nilehmann, line 18: The `#![assoc(` attribute accepts a list of functions you can merge them into the same one

tests/tests/pos/surface/slice_len_hack.rs
```diff
@@ -1,5 +1,6 @@
 use flux_rs::{assert, attrs::*};
 extern crate flux_core;
+use std::ops::Index;
```
nilehmann: This looks like a hallucination

lib/flux-core/src/array/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    fn test00(xs: &[i32; 100]) {
+        let _y = &xs[0..10];
+    }
+
+    #[should_fail]
+    fn test01(xs: &[i32; 10]) {
+        let _y = &xs[0..20];
+    }
+}
```
nilehmann: Please remove or move to `tests/tests/[pos,neg]/extern_specs`. I'm removing support for lib tests https://github.com/flux-rs/flux/pull/1618

lib/flux-core/src/array/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use flux_attrs::*;
+
+#[extern_spec(core::array)]
+impl<T, I, const N: usize> Index<I> for [T; N]
+where
+    [T]: Index<I>,
+{
+    #[sig(fn(&Self, {I[@idx] | <[T] as Index<I>>::in_bounds(N, idx)}) -> &<[T; N] as Index<I>>::Output{out: <[T] as Index<I>>::output_pred(N, idx, out)})]
+    fn index(&self, index: I) -> &<[T; N] as Index<I>>::Output;
```
nilehmann: This is unsound if you don't define `in_bounds`

nilehmann (changes requested):

lib/flux-core/src/array/mod.rs
ranjitjhala: But it is defined: it's using the in_bounds from slice?

nilehmann: No, if you don't define an associated refinement in an impl, you get the default in the trait (true in this case). The problem is that we don't check the impl is a subtype of the trait for extern specs (after substituting the definitions of assocs in the trait)

ranjitjhala: Hmm interesting .. let us hope the generic N is in scope where the associated refinement is defined...

tests/tests/pos/surface/slice_len_hack.rs
ranjitjhala: haha more like detritus from some debugging efforts...

lib/flux-core/src/array/mod.rs
ranjitjhala: Ok I pushed a thing, take a look. Bit tedious that we have to _again_ spell out the spec for `fn index` even after defining the assoc-refts?

nilehmann (approved):

lib/flux-core/src/array/mod.rs
nilehmann: Yes, agree it's tedious. The more serious problem is that we are not checking subtyping, though.

ranjitjhala: Is there an issue?

nilehmann: I don't think so

PR #1622, opened 2026-05-27
Specs for to_vec and vec::clone

ranjitjhala:


ranjitjhala:
@nilehmann this is a quick one any objections?

nilehmann (approved):
LGTM, I'm puzzled by the ci error, though. Formatting looks good to me

ranjitjhala:
@nilehmann I added specs for Deref + DerefMut for vec->slice; can you take a look please?

lib/flux-alloc/src/vec/mod.rs
```diff
@@ -59,3 +65,17 @@ impl<'a, T, A: Allocator> IntoIterator for &'a Vec<T, A> {
...
+#[extern_spec]
+impl<T, A: Allocator> Deref for Vec<T, A> {
+    #[spec(fn(&Vec<T, A>[@n]) -> &[T][n])]
+    fn deref(&self) -> &[T];
+}
+
+#[extern_spec]
+impl<T, A: Allocator> DerefMut for Vec<T, A> {
+    #[spec(fn(&mut Vec<T, A>[@n]) -> &mut [T][n])]
+    fn deref_mut(&mut self) -> &mut [T];
```
nilehmann, line 80: This signature is not compatible with the default signature in `DerefMut`, i.e., `fn(&mut Vec[@n]) -> &mut [T][n]` is not a subtype of `fn(&mut Vec) -> &mut [T]`

ranjitjhala: This is true for *every* situation where we put refinements on the `impl` method but leave the `trait` method unrefined (i.e. default), no?

nilehmann: It's not true for traits that take a shared reference. `fn(&T[@n])` is a subtype of `fn(&T)`

ranjitjhala: Is the point that we should not cheat, but instead, introduce an assoc `derefed` etc?

```
    #[spec(fn(&mut Self[@me]) -> &mut Out{v: derefed(me, v)})]
    fn deref_mut(&mut self) -> &mut Out;
```

nilehmann: Yes, that'd be ideal

ranjitjhala: Two things:

1. It looks like we already had a similar spec for `Rc` ...
2. It seems there is no way to write the sort for slices? As in literally how to write

```rust
#[extern_spec]
#[assoc(fn as_deref(me: Self, target: <Self as Deref>::Target) -> bool { target == me })]
impl<T, A: Allocator> Deref for Vec<T, A> {
    #[spec(fn(&Vec<T, A>[@me]) -> &Self::Target{target: <Self as Deref>::as_deref(me, target)})]
    fn deref(&self) -> &[T];
}
```


The `<Self as Deref>::Target` or `Self::Target` or `[T]` all seem to fail in various ways?

ranjitjhala: Do we even have a way to represent `[T]` as a `rty::Sort`?

ranjitjhala: ah, maybe i just use `int` duh.

ranjitjhala: hmm, somehow this is causing an ICE...

```rust
#[extern_spec(core::ops)]
#[assoc(fn as_deref(v: Self, target: Self::Target) -> bool { true })]
trait Deref {
    // #[sig(fn(self: &Self[@v]) -> &Self::Target{target: <Self as Deref>::as_deref(v, target)})]
    // #[sig(fn(self: &Self[@v]) -> &Self::Target)]
    fn deref(&self) -> &Self::Target;
}
```

nilehmann: This is working for me in the playground. Does it not work in main?

```rust
#![feature(allocator_api)]

use flux_rs::attrs::*;

use std::alloc::Allocator;

#[extern_spec]
#[refined_by(len: int)]
struct Vec<T, A: Allocator>;

#[extern_spec(core::ops)]
#[assoc(fn as_deref(v: Self, target: Self::Target) -> bool { true })]
trait Deref {
    #[sig(fn(self: &Self[@v]) -> &Self::Target{target: Self::as_deref(v, target)})]
    fn deref(&self) -> &Self::Target;
}

#[extern_spec(std::vec)]
#[assoc(fn as_deref(v: Self, target: int) -> bool { v.len == target })]
impl<T, A: Allocator> std::ops::Deref for Vec<T, A> {

    #[sig(fn(self: &Self[@v]) -> &[T][v])]
    fn deref(&self) -> &[T];
}
```

https://flux.goto.ucsd.edu/?code=jZFha4MwEIa_-ytuCKKj2w-4rrLC_sHKvohImp5FyIwkp3RI__suunZdy6D5Yi7n-1zeN_FDUZPi3lGqjLFasXWV6pqsjKLeE9SmP1TOIyrm8Hlczueed3IWFIjrk1B6cUEHJtdWviNdSumoblraVduv1FCL0LQsbM-u1wwfpF82C1gjnBn5NSTV1hGi7XwWeMp7q9O6BeWrHQk9HRDeydQLYOX2xHOFuJmqDJ5y2FprYASZSXAUCjvVMLwFNYwRyIoL3-yFmvqghSQgitehnOTJJXD8O-X3Fqf5WZgQmHLHuZUE6C1pGR2vrU6pDqTvdRrCvDA4PEvEsFr99CevzWdnbjKen08iRZxTqK07P0YumdwRSrEpC9n_71V-EIvf

ranjitjhala: CC claims

```

⏺ Now I see the problem clearly. Look at the deref function body:

  fn deref(__self: &Rc<T, A>) -> &T {
      <Rc<T, A> as Deref>::deref::<>(__self)
  }

  This call <Rc<T, A> as Deref>::deref() forces rustc to resolve <Rc<T,A> as Deref>::Target for the return type. With full compilation on (-Ffull-compilation=on), the trait solver
  actually evaluates this. The __FluxExternTraitDeref: Deref<> trait from flux-core introduces a second path for resolving Deref, and the solver constructs an AliasTy using the impl DefId
   of the synthetic trait rather than the proper associated type DefId — hence "unexpected DefKind in AliasTy: Impl { of_trait: true }".

  In short: the Deref trait extern spec in flux-core/src/ops/deref.rs creates a synthetic trait __FluxExternTraitDeref: Deref<> that confuses rustc's new trait solver when
  flux-alloc/rc.rs tries to call <Rc<T,A> as Deref>::deref() during full compilation. The Deref trait has an associated type Target, and the solver somehow resolves it through the
  synthetic trait's impl rather than the real one, producing a malformed AliasTy.

  This is likely a bug in how the extern_spec macro generates code for traits with associated types — or it may be an interaction with the new rustc trait solver on nightly-2025-11-25.
  The Iterator trait extern spec doesn't trigger this because no flux-alloc code calls <T as Iterator>::next() directly in a synthesized function body with full compilation.
```

ranjitjhala: Very mysterious. Your suggestion works fine on `main` BUT somehow not on this branch. Anyways, let me revert this branch to the stuff without `deref` and merge that, and then I can try again in case there was something silly I was doing.

ranjitjhala:
"deprecated" in favor of #1623

PR #1624, opened 2026-05-27
Add spec for `deref` and `deref_mut`

ranjitjhala:
@nilehmann see if this is what you had in mind; still puzzled by why the earlier version was crashing...


lib/flux-core/src/ops/deref.rs
```diff
@@ -0,0 +1,14 @@
...
+#[extern_spec(core::ops)]
+#[assoc(fn as_deref(v: Self, target: Self::Target) -> bool { true })]
+trait Deref {
+    #[sig(fn(self: &Self[@v]) -> &Self::Target{target: Self::as_deref(v, target)})]
+    fn deref(&self) -> &Self::Target;
+}
+
+#[extern_spec(core::ops)]
+trait DerefMut: Deref {
+    #[sig(fn(self: &mut Self[@v]) -> &mut Self::Target{target: Self::as_deref(v, target)})]
```
ranjitjhala, line 12: @nilehmann is it ok to use `Self::as_deref` here? It somehow resolves to the `Deref::as_deref`?

ranjitjhala:
ping @nilehmann

lib/flux-core/src/ops/deref.rs
nilehmann: Yes, that's even the preferred way imo

nilehmann (approved):

lib/flux-core/src/ops/deref.rs
ranjitjhala: How does it "know" to resolve as `Deref::as_deref` instead of failing being unable to find `DerefMut::as_deref` in this case?

PR #1630, opened 2026-05-29
Add a `log_verbose!` macro

ranjitjhala:
You can run with

```
FLUXFLAGS="-Fverbose" cargo flux
```

and it will print out the messages that were marked in the source with

```
dbg::log_verbose!("FLUX checking: {def_id:?} {span:?}");
```

(This lets one know where flux is stuck or ICEs ...)


ranjitjhala:
ping @nilehmann ^^^

nilehmann (approved):

PR #1631, opened 2026-05-29
Add attribute to assume a generic in a function is parametric

nilehmann:
@ranjitjhala you'll like this one


ranjitjhala:
hallelujah!!!

PR #1632, opened 2026-06-02
Use unord containers when trivially possible

nilehmann:


ranjitjhala:
@nilehmann I should wait for this PR or should I use FxIndexMap for the PlacesTree thing?

PR #1637, opened 2026-06-03
Several needed fixes

petros-marko:
- ~correctly name sortedsolution flag for fixpoint~
- parse conditionals in the k-solutions returned from fixpoint
- hide type variables (and let lean fill them in) when they'd cause trouble in the definition of VCs
- correct the check flux makes to determine a successful lean run (stdout is OK as long as it doesn't include a "sorry" use)

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -1481,6 +1481,7 @@ impl KVarSolutions {
                     )
             });
             let free_var_subs = free_vars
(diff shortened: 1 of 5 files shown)


ranjitjhala:
Thanks! Didn't @nilehmann fix the sorted solution thing? I hope this doesn't revert?

nilehmann:
@petros-marko I think you have to update fixpoint locally. The correct flag name is with a dash

nilehmann (approved):

PR #1638, opened 2026-06-04
Produce a better error message when parsing metadata fails

shumbo:
Flux uses rustc's opaque encoder/decoder to write/read saved metadata (`.fluxmeta`) alongside the `.rmeta` files in `target`. The decoder assumes well-formed input, so when the on-disk layout no longer matches the current types (e.g. after changing a data structure like `rty::ExprKind`), decoding panics and aborts with an unhelpful message:

```
thread 'rustc' panicked at /rust/deps/hashbrown-0.16.1/src/raw/mod.rs:37:40:
Hash table capacity overflow
```

This PR catches the decode failure (via `catch_unwind`, temporarily silencing the panic hook so there's no ICE dump) and instead emits a regular Flux diagnostic. The header/`METADATA_VERSION` mismatch path, which previously raised `bug!`, now reports the same error:

```
error[E0999]: failed to decode flux metadata file `<path>.fluxmeta`; this is likely because it was produced by an incompatible version of flux. Run `cargo clean` to remove stale metadata and try again
```

Because `catch_unwind` can't distinguish stale metadata from a genuine decoder bug, the message is deliberately hedged ("likely…"). `cargo clean` is the right first move either way.

### DIFF
--- a/crates/flux-metadata/locales/en-US.ftl
+++ b/crates/flux-metadata/locales/en-US.ftl
@@ -1,3 +1,6 @@
 metadata_decode_file_error = "error when decoding flux metadata file {$path}: {$err}"
 
+metadata_incompatible_metadata =
+    failed to decode flux metadata file `{$path}`; this is likely because it was produced by an incompatible version of flux. Run `cargo clean` to remove stale metadata and try again
+
 metadata_duplicate_spec = "duplicate spec for {$def_name}"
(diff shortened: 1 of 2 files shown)


shumbo:
Also flagging that this doesn't have a test right now, as I couldn't find a reasonable place to add a test. Since the change is small I'm not sure if we need one but I can add a regression test if needed.

nilehmann (approved):
Awesome looks great!

Issue #1635, opened 2026-06-02
Can't prove simple Real property


enjhnsn2:
I have been unable to prove `x*x == x*2` using Reals in Flux, although I can prove `(x+y)+z == x+y+z` and `x + y - y == x`.
```rust
#[flux_rs::refined_by(val: real)]
#[flux_rs::opaque]
pub struct Num(f32);

#[flux_rs::trusted]
impl Num {
    #[flux_rs::sig(fn(x: &Num[@xx], y: &Num[@yy]) -> Num[xx + yy])]
    pub fn add(&self, other: &Num) -> Num {
        Num(self.0 + other.0)
    }

    #[flux_rs::sig(fn(x: &Num[@xx], y: &Num[@yy]) -> Num[xx - yy])]
    pub fn subtract(&self, other: &Num) -> Num {
        Num(self.0 - other.0)
    }

    #[flux_rs::sig(fn(x: &Num[@xx], y: &Num[@yy]) -> Num[xx * yy])]
    pub fn mult(&self, other: &Num) -> Num {
        Num(self.0 * other.0)
    }
}

// works
#[flux_rs::sig(fn(x: &Num[@xx], y: &Num[@yy], z: &Num[@zz]) -> Num[xx + yy + zz])]
pub fn add3(x: &Num, y: &Num, z: &Num) -> Num {
    x.add(&y).add(&z)
}

// error: refinement type error, postcondition can't be proved
// Issue 2: can't prove x + x == 2 * x?
#[flux_rs::sig(fn(x: &Num[@xx]) -> Num[2 * xx])]
pub fn add_mult_test(x: &Num) -> Num {
    x.add(x)
}

// works
#[flux_rs::sig(fn(x: &Num[@xx], y: &Num[@yy]) -> Num[xx])]
pub fn add_sub_test(x: &Num, y: &Num) -> Num {
    x.add(y).subtract(y)
}
```

The failing function generates `reals_test.add_mult_test.smt2`:
```
(constraint
 (forall ((reftgen$xx$0 real) (true))
  (tag ((= (+ reftgen$xx$0 reftgen$xx$0) (* 2 reftgen$xx$0))) "0")))
```
which looks correct.

I'm not quite sure where to start in fixing this, could you perhaps point me in the right direction?

ranjitjhala:
If you (i.e. flux) emitted the constraint

```
(constraint
 (forall ((x real) (true))
  (tag ((= (+ x x) (* 2.0 x))) "0")))
[...]

PR #1639, opened 2026-06-04
Literals of Real sort

### LINKED ISSUE #1634: Can't instantiate non-integer values of sort `Real`
I'm beginning to experiment with verifying numerical code with flux (still just an idea), but the interface for reasoning about Reals/floats is still missing some basic building blocks. In particular, it is not possible to declare a real from a decimal:
```rust
#[flux_rs::refined_by(val: real)]
#[flux_rs::opaque]
pub struct Num(f32);

#[flux_rs::trusted]
impl Num {
    // works fine
    #[flux_rs::sig(fn() -> Num[2])]
    pub fn two() -> Self {
        Num(2.0)
    }

    // Error: unexpected literal
    #[flux_rs::sig(fn() -> Num[3.5])]
    pub fn three_point_five() -> Self {
        Num(3.5)
    }
}
``` 

As I see it, there are 2 ways to handle this:
1. Change Flux's parser such that decimal values can be reals.
2. Add the theory function `fp.to_real` and refine floats by proper IEEE754 floats, as supported by most SMT solvers (more work).

If the answer is (1), maybe this would be a good task for the flux hackathon.

enjhnsn2:
This PR adds support for `real` literals of the form `[3.5]`. In doing so, I also disambiguated some int parsing and generally made it harder to accidentally cause implicit conversions in the SMT solver.
In particular, this PR:
1. Implements Real Literals. The literals are parsed as floats, but are never actually converted to floats (they are kept as strings)
2. Prevents ints from being used as indices for reals `[3]` and only allows decimal indicies `[3.5]`. Previously int literals had a kind of either `Int` or `Real`, which was confusing. 
3. Added a check to prevent float suffixes on real literals `3.5f` in order to prevent confusion (floats are not supported).
4. Added some tests

This should close #1634 and #1635

### DIFF
--- a/crates/flux-fhir-analysis/src/conv/mod.rs
+++ b/crates/flux-fhir-analysis/src/conv/mod.rs
@@ -2077,6 +2077,10 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                     } else {
                         Err(self.emit(errors::InvalidBitVectorConstant::new(span, sort)))?
                     }
+                } else if sort == rty::Sort::Real {
+                    // Sort inference allows Int literals to unify with Real, but we require
+                    // explicit float syntax to avoid silently producing a mistyped constant.
+                    Err(self.emit(errors::IntLiteralInRealContext::new(span, n)))?
                 } else {
                     Ok(rty::Constant::from(n))
                 }
@@ -3161,6 +3165,21 @@ mod errors {
     pub(super) struct GenericsOnForeignTy {
         #[primary_span]
         pub span: Span,
+    }
+
+    #[derive(Diagnostic)]
+    #[diag(fhir_analysis_int_literal_in_real_context, code = E0999)]
+    pub struct IntLiteralInRealContext {
+        #[primary_span]
+        #[label]
+        span: Span,
+        n: u128,
+    }
+
+    impl IntLiteralInRealContext {
+        pub(crate) fn new(span: Span, n: u128) -> Self {
+            Self { span, n }
+        }
     }
 
     #[derive(Diagnostic)]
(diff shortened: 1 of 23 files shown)


crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -2067,26 +2067,25 @@ fn prim_ty_to_bty(prim_ty: rustc_hir::PrimTy) -> rty::BaseTy {
...
+                        && (n == 0 || n.ilog2() < size)
+                    {
+                        Ok(rty::Constant::BitVec(n, size))
+                    } else {
+                        Err(self.emit(errors::InvalidBitVectorConstant::new(span, sort)))?
                     }
+                } else if sort == rty::Sort::Real {
+                    // Sort inference allows Int literals to unify with Real, but we require
+                    // explicit float syntax to avoid silently producing a mistyped constant.
+                    Err(self.emit(errors::IntLiteralInRealContext::new(span, n)))?
```
nilehmann, line 2083: I don't have a strong opinion about this, but do we want to explicitly forbid writing reals without a decimal point?, i.e., if we have `2 + 3.0`, we don't want to infer that the `2` is being used as a real?

We could support that by creating a `Symbol::intern(&format!("{n}.0"))` here, but I can see an argument to force you to write it as `2.0`.

enjhnsn2: I considered that but ended up making the decision to be only the decimal representation. I don't think there is a super strong reason to support integer-reals and it makes the parsing much more confusing. The bigger issue is that Z3 automatically casts between integers and reals, so we want to avoid mixing those whenever possible lol.

ranjitjhala:
I agree with Evan!


- Ranjit.

nilehmann (approved):

crates/flux-fhir-analysis/src/conv/mod.rs
nilehmann: Out of curiosity, do you mean that in z3 is not that integer literals are polymorphic, but rather there's an implicit cast to real (i.e., it mixes the theory of reals and integers)?

enjhnsn2: Ranjit and I digged into this, and we found `Real + 2` and `Real + 2.0` evaluated to different things (one was sat, one was unsat). Ranjit asked Nicolai about it, but I don't know if he's heard anything back yet.

PR #1642, opened 2026-06-05
flux-config flags now appear in main cli help dialog but setting them does nothing

Shoooooon:
Run `cargo flux -h` or `cargo-flux flux -h` to see new output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,6 +492,7 @@ dependencies = [
 name = "flux-config"
 version = "0.1.0"
 dependencies = [
+ "clap",
  "globset",
  "serde",
  "toml",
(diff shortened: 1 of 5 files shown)


cole-k:
This is awesome, nice job! But the commit message looks off?

Shoooooon:
> This is awesome, nice job! But the commit message looks off?

I had to change it. Clap always triggers the parsers of non-optional fields, so `panicking_parser` can only be used for optional fields. Other fields can be set, but setting them has no effect.

nilehmann:
Thanks @Shoooooon!

nilehmann (approved):

PR #1646, opened 2026-06-09
 Skip dummy basic blocks to avoid spurious KVar joins

ranjitjhala:
(This is a redone version of #1644)

## Why?

Rustc often inserts intermediate basic blocks that serve only as merge points with no meaningful statements. When Flux treats these as join points, it emits unnecessary KVars that lose precision. By detecting these dummy blocks and skipping past them during checking, we preserve the type information flowing through the real control-flow edges. For example, consider the test `dummy_join00.rs`, which is

```rust
#[flux::spec(fn (x: i32) -> i32{v: 1 <= v && v <= 4})]
pub fn test(x: i32) -> i32 {
    if x == 0 {
        1
    } else if x == 1 {
        2
    } else if x == 2 {
        3
    } else if x == 3 {
        4
    } else {
        test(x - 1)
    }
}
```

In `main` we get the constraint (cargo x run tests/tests/pos/surface/dummy_join00.rs -- -Fdump-constraint) If you look at `dummy_join00.test.simp.fluxc` on `main`

```
∀ .
  ∀ a0: int.
    (a0 ≠ 0) =>
      (a0 ≠ 1) =>
        (a0 ≠ 2) =>
          (a0 ≠ 3) =>
            ∀ a1: int.
              (1 ≤ a1 ∧ a1 ≤ 4) => $k0(a1)[a0] ~ Goto(bb10) at 9:12: 13:6
          ¬(a0 ≠ 3) => $k0(4)[a0] ~ Goto(bb10) at 10:9: 10:10
          ∀ a1: int.
            $k0(a1)[a0] => $k1(a1)[a0] ~ Goto(bb11) at 7:12: 13:6
        ¬(a0 ≠ 2) => $k1(3)[a0] ~ Goto(bb11) at 8:9: 8:10
        ∀ a1: int.
          $k1(a1)[a0] => $k2(a1)[a0] ~ Goto(bb12) at 5:12: 13:6
      ¬(a0 ≠ 1) => $k2(2)[a0] ~ Goto(bb12) at 6:9: 6:10
      ∀ a1: int.
        $k2(a1)[a0] => (1 ≤ a1 ∧ a1 ≤ 4) ~ Ret at 3:5: 13:6
```

vs with this optimization, we get the constraint

```
true
```

Similarly, when you _do_ need an actual join, e.g. 

```rust
pub fn test2(x: i32) -> i32 {
    let res = if x == 0 {
        1
    } else if x == 1 {
        2
    } else if x == 2 {
        3
    } else {
        4
    };
    res + 1
}
```

where we can't just directly check the output type in each branch, you get a more compact constraint with a single kvar

```
∀ .  
  ∀ a0: int.  
    (a0 ≠ 0) =>  
      (a0 ≠ 1) =>  
        (a0 ≠ 2) => $k0(4)[a0] ~ Goto(bb9) at 25:9: 25:10
        ¬(a0 ≠ 2) => $k0(3)[a0] ~ Goto(bb9) at 23:9: 23:10
      ¬(a0 ≠ 1) => $k0(2)[a0] ~ Goto(bb9) at 21:9: 21:10
    ¬(a0 ≠ 0) => $k0(1)[a0] ~ Goto(bb9) at 19:9: 19:10
    ∀ a1: int.  
      $k0(a1)[a0] => (2 ≤ a1 + 1 ∧ a1 + 1 ≤ 5) ~ Ret at 27:5: 27:12
```

## What

  - Identify "dummy" basic blocks in the MIR — blocks with multiple incoming edges, only nop statements, and a single Goto terminator — and precompute their "real" (transitively reachable non-dummy) successor
  - In check_goto and fold_unfold's goto, skip dummy blocks by resolving directly to the real successor before the is_join_point check, avoiding spurious KVar emission at these artificial join points
  - Make BasicBlockData::terminator private and access it through Body::terminator_for to enforce consistent usage
  - Use BasicBlockData::new constructor in lowering instead of struct literal

## Key changes

  - crates/flux-rustc-bridge/src/mir.rs: Added DummyBlockTarget enum, mk_dummy_basic_blocks / get_transitive_dummy_target for precomputation, Body::resolve_dummy for lookup, and Body::terminator_for accessor
  - crates/flux-refineck/src/checker.rs: check_goto skips dummy blocks before the is_join_point check; check_basic_block and is_exit_block use terminator_for
  - crates/flux-refineck/src/ghost_statements/fold_unfold.rs: Same skip-dummy-block logic in goto; basic_block uses terminator_for
  - crates/flux-rustc-bridge/src/lowering.rs: Uses BasicBlockData::new constructor


ranjitjhala:
@nilehmann can you take a very close look at this?

crates/flux-refineck/src/ghost_statements/fold_unfold.rs
```diff
@@ -449,7 +449,9 @@ impl<M: Mode> FoldUnfoldAnalysis<'_, '_, '_, M> {
     }
 
     fn goto(&mut self, target: BasicBlock, env: Env) -> QueryResult {
-        if self.body.is_join_point(target) {
+        if let Some(real_target) = self.body.resolve_dummy(target) {
```
ranjitjhala: This is the KEY change suggested by CC; the `terminator_for` is not really necessary. 

This is the only thing that seems to work without breaking various CFG invariants, causing infinite loops etc.

nilehmann: I don't feel comfortable with the change. 

The `points_to` analysis also depends on the shape of the CFG. That analysis runs on the original rustc's mir, not the lowered one (to reuse rustc's dataflow analysis framework). We insert ghost statements based on that analysis, so there may be a mismatch.

Why did we drop the original idea of skipping dummy basic blocks in `checker.rs`?

ranjitjhala:
Isn’t that what this is doing? (Not following what you mean by “original”
here)


- Ranjit.

crates/flux-refineck/src/ghost_statements/fold_unfold.rs
ranjitjhala: Do you mean the business where "terminator_for" is hacked to skip ahead?

1. Doesn't any skipping done in checker.rs have the same problem that you might miss fold/unfold ghosts inserted by some earlier analysis?

2. Can this problem be addressed simply by checking that there are no ghosts in a "dummy block"?

nilehmann: Yes to 2. we could go back to the original idea where we determine whether to skip a basic block in checker.rs, but we consider a basic block dummy if it doesn't have ghost statements.

ranjitjhala: I am confused. 

`terminator_for` (which is supposed to determine skipping?) is used/required in two places: 

1. `checker.rs` 
2. `fold_unfold.rs` inside the `fn basic_block`

Should we

(a) NOT use it inside (2) -- if so, doesn't `kind` have to be public?
(b) USE it inside (2) in which case why do we care about the ghost stuff (because that would also have been skipped)?

Is the answer (a) because there are some EXTRA ghost things added OUTSIDE `fold_unfold.rs` (using the plain rust CFG?)

ranjitjhala: Also: if we are to use the Ghost stuff to determine the "dummy-ness" it seems `GhostStatements` should be visible inside `flux-rustc-bridge` since that's where the dummy-ness is?

ranjitjhala: Put another way, it seems the ENTIRE skipping computation should be in `checker.rs` and NOT in `mir.rs`?

nilehmann: There are stuff added to ghost in points_to.rs which uses the rustc_middle::mir. 

To compute dummyness considering ghost we need to move it to `flux-refineck`, i.e., to checker.rs

ranjitjhala:
@nilehmann I have updated the PR to be as minimal as I think: in `checker.rs` inside `fn check_goto` we check if the target is a "dummy-basic-block" (no op-codes, or ghost statements), and if so, we recurse onto its successor. No changes *anywhere* else. 

Its a bit tedious that `is_dummy_basic_block` is split across the `checker.rs` (the bits requiring ghost) and `mir.rs` (the bits about CFG), but it seems simpler that refactoring everything -- though I wonder if the CFG bit can be done entirely inside `checker.rs`?

--- 

## Why the other `terminator_for` method seems doomed

The other `terminator_for` method seems doomed -- I kept hitting all sorts of strange crashes, most recently in `bsearch.rs`.  CC after looking at the CFG shown below explains that we'll get crashes because blocks like `bb16` can "jump" all the way to `bb5` if we use the `terminator_for` to do the "skipping" (instead of inside `check_goto`) which caused a crash in `fn move_to` because I guess we somehow skipped passed a join point?

[bsearch_cfg.pdf](https://github.com/user-attachments/files/28779587/bsearch_cfg.pdf)


```
Let me trace through the crash. Looking at the MIR and the TRACE output:
  - bb18 (preds: bb16, bb17) → nops + goto bb4 → dummy, real successor = bb4
  - bb20 (preds: bb9, bb13) → nops + goto bb21 → dummy, real successor = bb21
  - bb21 (preds: bb2, bb20) → nops + goto bb22 → dummy, real successor = bb22

  What terminator_for does:

  For bb9 (goto → bb20): bb20 is dummy → recurses to terminator_for(bb21) → bb21 is also dummy → recurses to terminator_for(bb22) → bb22's terminator is return. So check_basic_block(bb9) gets
  a return terminator.

  For bb16 (goto → bb18): bb18 is dummy → recurses to terminator_for(bb4) → bb4's terminator is FalseUnwind { real: bb5 }. So check_basic_block(bb16) gets bb4's FalseUnwind terminator, with
  successor bb5.

  The marker crash:

  When check_basic_block(bb16) runs with bb4's FalseUnwind terminator, it does:
  self.markers[bb16] = Some(infcx.marker());
  self.check_successors(infcx, env, bb16, span, successors) // successors = [bb5]

  This processes bb5 and the entire loop body as if they were successors of bb16. Eventually the loop reaches bb16/bb17 again via bb18 (which gets skipped by terminator_for), and these blocks
  again get bb4's terminator with successor bb5. The inference context's marker tree gets deeply nested from repeated re-entry into the loop body. When the checker later tries to revisit bb4
  as a join point and calls infcx.move_to(&marker), the marker from a previous iteration no longer exists in the cursor tree — hence the "impossible case reached" panic at infer.rs:495.

  Root cause: Same fundamental problem as the read_loop.rs stack overflow. terminator_for returns a distant block's terminator, creating phantom edges (bb16 → bb5, bb9 → return) that don't
  exist in the real CFG. This breaks both cycle detection (causing stack overflows in simple cases) and marker/inference-context bookkeeping (causing this crash in more complex cases).
```

ranjitjhala:
@nilehmann have updated to move everything into `checker` and using the `is_join` instead of exposing the predecessor stuff.

nilehmann (approved):

PR #1647, opened 2026-06-09
Lean directory hassles

ranjitjhala:
Check for `lakefile.toml` instead of just the lean-path, as otherwise we sometimes SKIP properly initializing the lean directory, due to that whole renaming stuff... Per CC

Look at generate_lake_project_if_not_present (line 521-535):

```rust
  fn generate_lake_project_if_not_present(&self) -> io::Result<()> {
      let path = project_path(self.genv, FileKind::User);
      if !path.exists() {
          Command::new("lake").arg("new").arg(project())...
      }
      Ok(())
  }
 ```

`project_path(genv, FileKind::User)` returns `final_project_path(genv)` which is lean_parent_dir/lean_proofs — just the directory path.

But `finalize` (line 160-164) is called at the end of compilation:

```rust
  fn finalize(genv: GlobalEnv) -> io::Result<()> {
      let path = project_path(self.genv, FileKind::User);
      if !path.exists() {
          Command::new("lake").arg("new").arg(project())...
      }
      Ok(())
  }
```

project_path(genv, FileKind::User) returns final_project_path(genv) which is lean_parent_dir/lean_proofs — just the directory path.

`rename_dir_contents` calls `fs::create_dir_all(dst)` (line 106), which creates the `lean_proofs` directory.

- So on the first run, `lake new` runs and everything works. But `finalize` also creates the `lean_proofs/` directory via `create_dir_all`. 

- On a subsequent run (or if the directory was left behind from a failed run), `path.exists()` is true because the directory exists, so lake new is skipped — even if the `lakefile.toml` and other lake-generated files are missing (e.g., if someone deleted them, or if only the
  directory structure was created by `rename_dir_contents` without lake ever running).


ranjitjhala:
@petros-marko can you double check this?

petros-marko (approved):
LGTM. Makes sense.

PR #1648, opened 2026-06-09
Add re-run hints to error messages

shumbo:
When verification fails, Flux now attaches a `note:` to the item's first error with a copy-pastable command to re-run the check on just that item.

```
 --> src/lib.rs:7:5
  |
7 |     0
  |     ^ a postcondition cannot be proved
  |
 --> src/lib.rs:5:23
  |
5 | #[spec(fn() -> i32{v: v > 0})]
  |                       ^^^^^
  = note: to rerun: `FLUXFLAGS='-Finclude=def:foo' cargo flux check -p mycrate`
```

When run via the standalone `flux` binary on a single file, the note instead reads:

```
  = note: to rerun pass `'-Finclude=def:foo'`
```

Some notes:

- The hint targets the item with a `def:<path>` include pattern, which is stable across line/column edits. It falls back to a whole-file `glob:<file>` pattern when the def path is empty (or, under `cargo flux`, contains spaces, e.g., trait-impl methods like `<S as T>::f`, since `FLUXFLAGS` is split on spaces).
  - `FLUXFLAGS` can recognize quotes but I didn't do so to minimize the PR. Re-running the file isn't too bad IMO.
- Messages have two forms depending on how Flux was invoked:
  - standalone `flux`: `to rerun pass '-Finclude=def:foo'`
  - `cargo flux` (`FLUX_CARGO=1`): `to rerun: FLUXFLAGS='-Finclude=def:foo' cargo flux check -p <pkg>`
- Enabled by default. Disable with `-Frerun-hint=false`.
- In the implementation, `FluxSession::with_next_err_notes` stages notes for the duration of a closure and `emit_err` attaches them (once) to the first error it emits. I considered threading the note through the `emit_err` call sites and rejected it as too invasive — `next_err_notes` is the staging mechanism I landed on. Open to better ideas.

### DIFF
--- a/crates/flux-driver/src/callbacks.rs
+++ b/crates/flux-driver/src/callbacks.rs
@@ -229,9 +229,17 @@ impl<'genv, 'tcx> CrateChecker<'genv, 'tcx> {
     }
 
     fn check_def_catching_bugs(&mut self, def_id: LocalDefId) -> Result<(), ErrorGuaranteed> {
-        let mut this = std::panic::AssertUnwindSafe(self);
-        let msg = format!("def_id: {:?}, span: {:?}", def_id, this.genv.tcx().def_span(def_id));
-        flux_common::bug::catch_bugs(&msg, move || this.check_def(def_id))?
+        let genv = self.genv;
+        // Stage the re-run hint for the scope of the check so it gets attached as a note to the
+        // first error this item emits.
+        let note = config::rerun_hint()
+            .then(|| rerun_hint_note(genv, def_id))
+            .flatten();
+        genv.sess().with_next_err_notes(note, || {
+            let mut this = std::panic::AssertUnwindSafe(self);
+            let msg = format!("def_id: {:?}, span: {:?}", def_id, this.genv.tcx().def_span(def_id));
+            flux_common::bug::catch_bugs(&msg, move || this.check_def(def_id))
+        })?
     }
 
     fn check_def(&mut self, def_id: LocalDefId) -> Result<(), ErrorGuaranteed> {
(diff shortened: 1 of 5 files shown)


nilehmann:
This is really cool, but Ranjit found the other day that `-Finclude` has a problem. Because `include` has `or` semantics, whenever you have an `include` list in your Cargo.toml, extra `-Finclude` will be conjoined, so it won't have the effect of focusing on a particular item.

Ranjit is trying to fix this as part of https://github.com/flux-rs/flux/pull/1650. We just had a meeting and decided on an approach. Maybe we should wait for that so we can reuse the same logic here.

shumbo:
> This is really cool, but Ranjit found the other day that `-Finclude` has a problem. Because `include` has `or` semantics, whenever you have an `include` list in your Cargo.toml, extra `-Finclude` will be conjoined, so it won't have the effect of focusing on a particular item.
> 
> Ranjit is trying to fix this as part of #1650. We just had a meeting and decided on an approach. Maybe we should wait for that so we can reuse the same logic here.

Sounds good! I will keep an eye out for the PR and rebase when it's done.

crates/flux-driver/src/callbacks.rs
```diff
@@ -229,9 +229,17 @@ impl<'genv, 'tcx> CrateChecker<'genv, 'tcx> {
...
+        // Stage the re-run hint for the scope of the check so it gets attached as a note to the
+        // first error this item emits.
+        let note = config::rerun_hint()
+            .then(|| rerun_hint_note(genv, def_id))
+            .flatten();
+        genv.sess().with_next_err_notes(note, || {
+            let mut this = std::panic::AssertUnwindSafe(self);
+            let msg = format!("def_id: {:?}, span: {:?}", def_id, this.genv.tcx().def_span(def_id));
+            flux_common::bug::catch_bugs(&msg, move || this.check_def(def_id))
+        })?
```
nilehmann, line 242: There's no guarantee that the next error we emit will be about this `def_id`. We need to add the note at the point where we generate the error, somewhere here https://github.com/flux-rs/flux/blob/main/crates/flux-refineck/src/lib.rs#L210

ranjitjhala:
Gents I *desperately* want this now (because it is a _super_ handy and robust way to tell CC what to rerun after tweaking stuff.) what needs to be done @nilehmann ?

ranjitjhala:
Basically we need to update this code to use the `--only-check=def:<PATH>` I think?

nilehmann:
@ranjitjhala https://github.com/flux-rs/flux/pull/1648#discussion_r3413820475 also needs to be fixed.

ranjitjhala:
hmm looks like your link "somewhere here" points outside any functions; can you tell me what you were referring to?

nilehmann:
Ah, I was referring to report_errors which is now a couples of lines below. We need to attach the note at the point we generate error to make sure is the right error.

ranjitjhala:
Got it - I assume we have to pass the def id etc to that method, and not
use some state shenanigans?

- Ranjit.

ranjitjhala:
See #1698 (apologies for the git messup @shumbo !)

PR #1650, opened 2026-06-10
Priority Includes Flag

ranjitjhala:
This adds to the `cargo flux` binary, a flag `--only-check <INCLUDEPATTERN>` flag, which is **overrides** all other include patterns from the `cargo.toml` etc. when checking "local" crates, not external ones (e.g. `flux-core` etc.) 

This allows a user or the `vscode` extension, to *override* any include patterns that may be present in the `cargo.toml` by running

```
cargo flux --only-check <PATTERN>
```


ranjitjhala:
@nilehmann (and @shumbo) here's the PR with the `--only-check` flag. 

I think for #1648 you need only modify the message to be, e.g. 

```
  = note: to rerun: `cargo flux check -p mycrate --only-check def:foo`
```

crates/flux-bin/src/bin/cargo-flux.rs
```diff
@@ -130,6 +143,7 @@ rustflags = [{:?}]
                         .into_flags(&metadata.target_directory, manifest_dir_relative_to_workspace)
                         .iter()
                         .chain(flux_flags.iter().flatten())
+                        .chain(flux_local_flags)
```
nilehmann: I had in mind something different. 

Instead of adding the `only` prefix, we could pass `--only-check` to the `into_flags` method and avoid appending extra includes there if the flag was specified.

ranjitjhala: Makes sense, and a lot cleaner. Let me update!

ranjitjhala: Ok, I updated the code so that we filter out all other `-Finclude=` and just push the only-check pattern for local crates.

ranjitjhala: ping @nilehmann

nilehmann (approved):

PR #1652, opened 2026-06-10
Only check lean proofs for "included" DefIds

ranjitjhala:
ping @nilehmann  and @petros-marko -- any objections?


ranjitjhala:
ugh why are 7 files changed. sigh.

ranjitjhala:
sorry see #1653

PR #1653, opened 2026-06-10
Only check lean proofs for "included" DefIds

ranjitjhala:
Any objections? @petros-marko


petros-marko (approved):
Makes sense, but did this come up at all anywhere?

ranjitjhala:
Yes when you're working on one (broken) proof you don't want to be slowed down by other slow to check proofs...

PR #1657, opened 2026-06-12
Store monomorphized panic spec

nilehmann:
* Make a query for the call graph
* Add NodeKey to distinguish between source items and monomorphized items
* Key PanicSpec by NodeKey
* Recover resolved PanicSpec for callee based on the call graph in checker.rs

### DIFF
--- a/crates/flux-bin/src/lib.rs
+++ b/crates/flux-bin/src/lib.rs
@@ -86,7 +86,7 @@ impl FluxMetadata {
             flags.push(format!("-Fallow-uninterpreted-cast={v}"));
         }
         if let Some(only_check) = only_check {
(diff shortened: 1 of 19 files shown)


nilehmann:
@ninehusky, can you run the same experiment here https://github.com/flux-rs/flux/pull/1610#issuecomment-4596758347 and see how the numbers are with this PR?

enjhnsn2:
@ninehusky Can you take a look at this?
I am currently blocked on this issue (although its not urgent).

nilehmann:
I'm going to merge this because I think it's strictly better. 

We can still iterate on it, so it'd be good if you can run the experiment when you have time @ninehusky

PR #1658, opened 2026-06-12
Higher-order (spec) functions

ranjitjhala:
Allow writing stuff like

```rust
#![flux::defs(

    fn lt(x: int, y: int) -> bool {
        x < y
    }

    fn cmp0(x:int, cmp: fn(int, int) -> bool) -> bool {
        cmp(0, x)
    }

    fn pos(n: int) -> bool {
        cmp0(n, lt)
    }

)]

#[flux::spec(fn() -> i32{v:pos(v)})]
fn test0() -> i32 {
    29
}
```


ranjitjhala:
@nilehmann any thoughts on this?

ranjitjhala:
I don't know if we have a super compelling use case yet, but I think it will be a useful building block.

nilehmann:
I'm not sure about the implications of this. I suppose it will allow writing specs that will make fixpoint-hs crash, but I suppose we can restrict it once we see such errors.

In general, I think we can allow arbitrarily complex expressions and catch unsupported things when we generate the constraint before sending to fixpoint-hs, but it's hard to give good error messages at that point...

PR #1597, opened 2026-05-14
Stop emitting kvar solutions from fixpoint to use in lean

petros-marko:
This will break flux-to-lean demo for now, should be able to merge after we switch those proofs to lean fixpoint tactics.

PR #1659, opened 2026-06-13
Unbounded quantifiers

ranjitjhala:
(Only for external proofs.)

Need a matching PR to test actual lean backend...


ranjitjhala:
Added a test 

https://github.com/flux-rs/flux-to-lean-demo/blob/main/src/ringbuffer.rs#L6-L8

ranjitjhala:
@nilehmann one thing I am unsure about is what to do for the `forall` in fixpoint?

1. Let it through and hope for the best?
2. Complain with some error (maybe but this is problematic while we still rely on fixpoint solutions for the kvars... -- maybe we can throw an error AFTER https://github.com/flux-rs/flux/pull/1597 is merged )
??

crates/flux-middle/src/queries.rs
```diff
@@ -1221,6 +1220,12 @@ impl<'a> Diagnostic<'a> for QueryErrAt {
                         diag.note(err.descr);
                         diag
                     }
+                    // QueryErr::UnboundedQuantifier { span } => {
+                    //     let mut diag = dcx
+                    //         .struct_span_err(cx_span, fluent::middle_query_unsupported_quantifier);
+                    //     diag.note(err.descr);
+                    //     diag
+                    // }
```
nilehmann:
```suggestion
```

crates/flux-middle/src/fhir/visit.rs
```diff
@@ -608,7 +608,7 @@ pub fn walk_expr<'v, V: Visitor<'v>>(vis: &mut V, expr: &Expr<'v>) {
                 vis.visit_expr(&s.expr);
             }
         }
-        ExprKind::BoundedQuant(_, param, _, expr) => {
+        ExprKind::Quant(_, param, _, expr) => {
```
nilehmann: This should visit the sort in the domain

crates/flux-middle/src/rty/fold.rs
```diff
@@ -1057,7 +1057,7 @@ impl TypeSuperVisitable for Expr {
                 args.visit_with(visitor)
             }
             ExprKind::Abs(body) => body.visit_with(visitor),
-            ExprKind::BoundedQuant(_, _, body) => body.visit_with(visitor),
+            ExprKind::Quant(_, _, body) => body.visit_with(visitor),
```
nilehmann: This should visit the sort in the domain

crates/flux-middle/src/rty/fold.rs
```diff
@@ -1120,8 +1120,8 @@ impl TypeSuperFoldable for Expr {
             ExprKind::Hole(kind) => Expr::hole(kind.try_fold_with(folder)?),
             ExprKind::KVar(kvar) => Expr::kvar(kvar.try_fold_with(folder)?),
             ExprKind::Abs(lam) => Expr::abs(lam.try_fold_with(folder)?),
-            ExprKind::BoundedQuant(kind, rng, body) => {
-                Expr::bounded_quant(*kind, *rng, body.try_fold_with(folder)?)
+            ExprKind::Quant(kind, rng, body) => {
```
nilehmann: This should fold the sort in the domain.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -760,6 +760,12 @@ pub enum SpecFuncKind {
     Def(FluxDefId),
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, Debug, TyDecodable)]
```
nilehmann:
```suggestion
#[derive(Clone, PartialEq, Eq, Hash, TyEncodable, Debug, TyDecodable, TypeVisitable, TypeFoldable)]
```

nilehmann (commented):
> @nilehmann one thing I am unsure about is what to do for the forall in fixpoint?
> 
> 1. Let it through and hope for the best?
> 2. Complain with some error (maybe but this is problematic while we still rely on fixpoint solutions for the kvars... -- maybe we can throw an error AFTER https://github.com/flux-rs/flux/pull/1597 is merged )

I don't have much hope for 1., so we must merge #1597 because we use the same constraint for fixpoint-hs and lean.

Also related to this. We already have `ExprKind::ForAll` in `rty`. You can produce one of those by writing a quantified `requires` like so

```rust
#[flux::sig(
    fn(x: i32)
    requires forall y. y >= 0 => y > x
)]
fn requires_negative(x: i32) {
    assert(x + 1 == 1 + x); // make sure there's something to check to avoid optimizing the entire constraint away
}

fn test2() {
    requires_negative(-1);
}
```

When we see such a forall in a head, we *unfold it* and when we see it in an assumption, we drop it. This is done in `head_to_fixpoint` and `assumption_to_fixpoint`. The surface syntax for `requires` ensures that foralls only appear in those two places.

We should probably unify the behaviors. I find this interesting because we can have a function with *bounded refinements* where the definition is proved in Lean, but callers can still use fixpoint.

So we could have something like

1. In lean mode, let anything go through
2. In fixpoint-hs mode, try to see if we can unfold the forall and still generate a horn constraint. Otherwise, report an error (or decay to an uninterpreted blob, or yolo fixpoint-hs mode, ...)

ranjitjhala:
Ok, sounds good. To clarify: you're suggesting 

1. Revert so the "bounded" stuff is separate from the Forall/Exists
2. Reuse the (already present) Forall/Exists for the "unbounded" case
3. Generalize the behavior (where non-head forall is rejected) to ALSO allow it wherever it appears, for the lean backend?

nilehmann:
> Ok, sounds good. To clarify: you're suggesting
> 
>   1. Revert so the "bounded" stuff is separate from the Forall/Exists
>   2. Reuse the (already present) Forall/Exists for the "unbounded" case
>   3. Generalize the behavior (where non-head forall is rejected) to ALSO allow it wherever it appears, for the lean backend?


I don't have strong opinions on how to represent it. We could also merge the existing `ExprKind::ForAll` into `ExprKind::Quant`. The important bit is step 3.

ranjitjhala:
Wait, can you remind me _why_ we have the existing `forall` but special cased to appear only inside `requires`?

ranjitjhala:
(By which I mean, why did the _parser_ special case it, not the assume_to_fixpoint / head_to_fixpoint)

nilehmann:
We wanted bounded refinements at some point, and we realized that in many cases where they are useful, the callee is trusted anyway (there's a database query or whatever behind), so I added syntax that guarantees callers always generate horn constraints.

We never did anything useful with it, though.

ranjitjhala:
@nilehmann -- ok I removed the separate `Exists` and `ForAll` and put them all under the `Quant` with a `QuantKind::Exists` and `QuantKind::Forall` (hopefully preserving the old behavior.)

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1785,6 +1777,25 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
                 };
                 self.expr_to_fixpoint(&expr, scx)?
             }
+            rty::ExprKind::Quant(kind, rty::QuantDom::Unbounded, body)
+                if matches!(kind, flux_middle::fhir::QuantKind::Exists)
```
nilehmann: Why are you allowing exists to go through here?

ranjitjhala: because, for some reason, it was going through earlier?

ranjitjhala: https://github.com/flux-rs/flux/blob/2b948c6e5e786b68fcdf315128fdb278f6a43e81/crates/flux-infer/src/fixpoint_encoding.rs#L1765-L1768

nilehmann: Do we have a test that fails without it? This may be related to @cole-k stuff?

nilehmann: We definitely won't need this for lean after merging https://github.com/flux-rs/flux/pull/1597, because there's no way we can get exists without parsing stuff from fixpoint.

We may still need it for wick, though.

ranjitjhala: Yes, I also thought it may be related to some of the wick stuff.

There are no tests that fail without it.

nilehmann: anywho, I think we should merge #1597 ...

cole-k: I don't remember writing any of this, but it's possible I did if I'm in the git blame. It shouldn't be necessary for wick since we add exists AFTER doing the fixpoint encoding.

I have a branch that does fusion in flux-rs which WOULD require adding exists DURING fixpoint encoding, but it's not merged...

nilehmann: But when you get a solution from fixpoint-hs (with exists) and parse it back to an `rty::Expr`, do we ever call `expr_to_fixpoint` again on that solution (in the process of doing quantifier elimination, maybe)?

nilehmann: Or in wick we never go back to `rty::Expr`?

cole-k: We shouldn't. We should stick to fixpoint for everything we do and I'm pretty sure I split up the fixpoint solution parsing to give back the fixpoint solution in addition to the rty one.

Once we parse to `rty::Expr` we shouldn't return. But that's just based off the top of my head.

ranjitjhala: Sounds like we should nuke it, and can trivially restore later if needed?

ranjitjhala:
ok, I removed that spl case; lmk if there's anything else tbd.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -813,9 +821,9 @@ pub enum ExprKind {
     /// and the places where we generate inference variable for them (where we do need to worry
     /// about the scope).
     Hole(HoleKind),
-    ForAll(Binder<Expr>),
+    // CUT ForAll(Binder<Expr>),
```
nilehmann:
```suggestion
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -813,9 +821,9 @@ pub enum ExprKind {
...
     /// Only for non-cuts solutions from fixpoint
-    Exists(Binder<Expr>),
+    // CUT Exists(Binder<Expr>),
```
nilehmann:
```suggestion
```

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1785,6 +1777,24 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
...
+                };
+                fixpoint::Expr::Quantifier(kind, expr.0, Box::new(expr.1))
+            }
+            rty::ExprKind::Hole(..)
+            | rty::ExprKind::KVar(_)
+            | rty::ExprKind::Local(_)
+            | rty::ExprKind::PathProj(..)
+            | rty::ExprKind::Quant(..)
+            | rty::ExprKind::InternalFunc(_) => {
+                span_bug!(self.def_span(), "unexpected expr: `{expr:?}`")
```
nilehmann, line 1797: Now that we have merged the other PR, we can report a proper error here for `Backend::Fixpoint` when we reach `rty::ExprKind::Quant`

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -1053,7 +1055,9 @@ where
             rty::ExprKind::KVar(kvar) => {
                 preds.push(self.kvar_to_fixpoint(kvar, bindings)?);
             }
-            rty::ExprKind::ForAll(_) => {
+            rty::ExprKind::Quant(QuantKind::Forall, QuantDom::Unbounded, _)
+                if matches!(self.ecx.backend, Backend::Fixpoint) =>
+            {
```
nilehmann: And I'd remove this case and let it fall through, so we report an error inside `expr_to_fixpoint`. We have some tests that depend on this, but I think the behavior of ignoring is just too weird, so I'd mark the functions that rely on this behavior as `#[trusted]`

crates/flux-infer/src/fixpoint_encoding.rs
ranjitjhala: by "proper error" do you mean something other than `span_bug!` ?

crates/flux-syntax/src/parser/mod.rs
```diff
@@ -1545,17 +1547,21 @@ fn parse_bounded_quantifier(cx: &mut ParseCtxt) -> ParseResult<Expr> {
         return Err(lookahead.into_error());
     };
     let param = parse_refine_param(cx, RequireSort::Maybe)?;
```
nilehmann, line 1549: I just realized that parsing a `RequireSort::Maybe` param here is a bit weird because you can write something like `forall i:bool in 0..10 { i > 0 }`. Also, inference will fail for `forall i in 0..10 { 0 >= 0 }`

crates/flux-infer/src/fixpoint_encoding.rs
nilehmann: Yes. It's hard to give a good error at this level, but it's better than an ICE

crates/flux-syntax/src/parser/mod.rs
nilehmann: Actually, bounded quantifiers are wrong because we can write `forall b:bool in 0..10 { b }` which has sort `bool`, but will be expanded to `0 ∧ 1 ∧ ... ∧ 10`

ranjitjhala:
I think the fact x is to just prohibit sort annotations— I can’t think of
any situation where inference fails but we want to accept with an
annotation…?


- Ranjit.

nilehmann:
Yes forbid sort annotations is fine,  but we still need to force the param to be int for bounded quantifiers otherwise you can write

```
forall i in 0..10 { i == true }
```

and it will expand to something ill sorted

crates/flux-syntax/src/parser/mod.rs
ranjitjhala: Ugh. Yes, this is wrong with or without sort annotations; we ought to check that in the `Bounded` case the binder is indeed an int...

nilehmann: haha, vtock uses the sort annotation

ranjitjhala: wow, and here I wanted to infer them because I thought it _does not_ use the sort annotations, sigh.

ranjitjhala:
Have restored support for `for i : int in 10..20 {...}` + various tests to check you don't write the WRONG sort.

crates/flux-fhir-analysis/src/wf/sortck.rs
```diff
@@ -323,8 +323,18 @@ impl<'genv, 'tcx> InferCtxt<'genv, 'tcx> {
...
                 self.check_expr(body, &rty::Sort::Bool)?;
+                if let fhir::QuantDom::Bounded { .. } = dom {
+                    let sort = self.param_sort(param.id);
+                    let sort = self.resolve_vars_if_possible(&sort);
+                    if !matches!(sort, rty::Sort::Int) {
+                        return Err(
+                            self.emit_err(errors::IllSortedQuantifier::new(param.span, sort))
+                        );
+                    }
+                }
```
nilehmann: I believe the following will be a bit more permissive because it will allow the parameter to be implicitly coerced inside the body (it will also accept the case where the parameter is not used)

```suggestion
                let sort = self.param_sort(param.id);
                if self.try_equate(sort, &rty::Sort::Int).is_none() {
                    return Err(
                            self.emit_err(errors::IllSortedQuantifier::new(param.span, sort))
                    );
                }
                self.check_expr(body, &rty::Sort::Bool)?;
```

nilehmann (approved):

crates/flux-fhir-analysis/src/wf/sortck.rs
ranjitjhala: hmm.. but I had an explicit test which checks for whether the parameter is used inside (`ill_sorted_forall02.rs`)?

ranjitjhala: wait, how will the above work for `forall i { i != 10 || i != 20}` ? We have to first check the `body` to figure out  that `i` is an `int` , don't we?

ranjitjhala: oh you _want_ those to be allowed... ok!

nilehmann: Ah, sorry, I meant to guard that by `if let fhir::QuantDom::Bounded { .. } = dom`

nilehmann:
```rust
if let fhir::QuantDom::Bounded { .. } = dom {
  let sort = self.param_sort(param.id);
  if self.try_equate(sort, &rty::Sort::Int).is_none() {
      return Err(
              self.emit_err(errors::IllSortedQuantifier::new(param.span, sort))
      );
  }
}
self.check_expr(body, &rty::Sort::Bool)?;
```

ranjitjhala: yup -- I added that back

ranjitjhala: hmm except I hoisted the 

```
 let sort = self.param_sort(param.id);
```

which is hopefully ok?

PR #1660, opened 2026-06-13
Refinement suggestions (wick)

cole-k:
# Basics

This PR introduces suggestions for Flux, realized as follows

```
  --> benchmarks/pldi23/src/kmeans.rs:6:26
   |
 6 |         let di = x[i] - y[i];
   |                          ^^^ a precondition cannot be proved
   |
  --> benchmarks/pldi23/src/vec.rs:26:57
   |
26 | #[flux::assoc(fn in_bounds(idx: int, len: int) -> bool {idx < len} )]
   |                                                         ^^^^^^^^^
help: try adding the refinement
   |
 2 + #[flux_rs::sig(fn(&Vec<f32>[@x], &Vec<f32>[@y]) -> f32 requires y.len >= x.len)]
   |
```

The mechanism for achieving this involves the use of Z3 bindings (previously used only for `--rust-fixpoint`) and is gated behind a compilation flag `--wick`. Compile using `cargo xtask install --wick`

You can also automatically apply the suggestions provided by running `cargo flux --fix`. I made changes to the pretty printer to ensure that the signatures are valid, but I have not tested this thoroughly, so even though the diags are marked `MachineApplicable` there may be bugs in rendering.

If you use Flux as your LSP, you also will get these suggestions as editor hints.

# PR structure

I have loosely organized this PR into separate commits.

The first is regarding a `no_suggestions` flag and attribute which turns off the suggestion mechanism. 60a59e10fa7a8a0a610fedfb8da8a9dd8fddfb50

The second is all of the changes which should be gated behind the `wick` compilation flag. 2d934a739cfcd5bc7a1f26eab14baed44d5cc72d

The third is the introduction of a new internal `Expr` type. f35e06a2ddbd2d92e26883b638ebcf07eb38a609

The fourth is other changes not gated behind a feature flag, mostly pertaining to pretty printing. fdd9eb0b726bae53e477ff8d26e452d258895234

Everything else is somewhat minor and includes small tweaks and formatting.


crates/flux-bin/src/cargo_flux_opts.rs
```diff
@@ -45,7 +45,21 @@ impl CargoFluxCommand {
...
             CargoFluxCommand::Check(check_opts) => {
-                cmd.arg("check");
+                if check_opts.fix {
+                    // `cargo fix` applies span_suggestion replacements written by flux-driver.
+                    //
+                    // Mechanism: cargo fix sets RUSTC_WORKSPACE_WRAPPER to a self-re-exec proxy
+                    // that runs flux-driver (our RUSTC) with --error-format=json, parses the
+                    // resulting diagnostic JSON for suggested_replacement spans, and applies them
+                    // via rustfix. This is the same pipeline Clippy uses, just with RUSTC= instead
+                    // of RUSTC_WORKSPACE_WRAPPER= on our side; the two env vars are independent.
```
cole-k, line 55: This was written by an LLM. I think it's reasonably correct.

crates/flux-infer/src/fixpoint_encoding/decoding.rs
```diff
@@ -184,7 +187,17 @@ where
...
-                        Err(FixpointParseError::NoLocalVar(*fname))
+                        // HACK: this is a free var generated from an existential
+                        // we got from fixpoint. We just will add enough to it so it
+                        // is likely to not conflict with exisitng free vars, but in an
+                        // ideal world we would properly add it to the fvars map and
+                        // make a fresh name for it.
+                        //
+                        // But since we just need to be consistent about the value we
+                        // give these vars, this should suffice for now --- they won't
+                        // actually appear in any solutions because we anti-unify.
```
cole-k: Still untested whether this is needed...

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -205,14 +222,45 @@ impl Pretty for FnSig {
...
+            join!(", ", self.inputs.iter().map(|input| input.shallow_canonicalize()))
         )?;
-        if !self.requires.is_empty() {
-            w!(cx, f, " requires {:?}", join!(" ∧ ", &self.requires))?;
-        }
-        Ok(())
+        w!(cx, f, " -> ")?;
+        // Format requires *before* entering the output binder context: requires predicates
+        // reference FnArgs variables (from the outer PolyFnSig binder) and pushing the FnRet
+        // layer would shift their debruijn indices.
```
cole-k, line 231: Machine-generated comment. I think it's nonsense? But don't want to remove it if it actually makes sense.

The only reason to format the requires first is because it's in a different context (no output binders introduced).

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -205,14 +222,45 @@ impl Pretty for FnSig {
...
+            .iter()
+            .filter(|r| !r.is_trivially_true())
+            .collect_vec();
+        let requires_str = if !filtered_requires.is_empty() {
+            format_cx!(cx, " requires {:?}", join!(" && ", &filtered_requires))
+        } else {
+            String::new()
+        };
+        // Format the return type and ensures within the output binder context so that bound
+        // variables in the output position get the correct `#` decorator.
```
cole-k, line 243: Machine-generated, but correct.

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -514,10 +593,41 @@ impl Pretty for BaseTy {
...
+                        if !has_default {
+                            break;
+                        }
+                        let earlier_args =
+                            tcx.mk_args_from_iter(args[..arg_idx].iter().map(|a| a.to_rustc(tcx)));
+                        let default_ty = tcx.type_of(param.def_id).instantiate(tcx, earlier_args);
+                        if arg.to_rustc(tcx) != rustc_middle::ty::GenericArg::from(default_ty) {
+                            break;
+                        }
+                        args.pop();
```
cole-k, line 628: The computer managed to hide default args in the output using this. I have no clue how brittle it is, but it at least works for `Vec`?

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -57,19 +58,23 @@ impl<T: Pretty> std::fmt::Debug for Binder<T> {
     }
 }
 
-fn format_fn_root_binder<T: Pretty + TypeVisitable>(
+/// Like [`format_fn_root_binder`] but takes a closure to produce the body string, allowing the
+/// caller to format multiple pieces (e.g. return type, requires, ensures) together under the same
+/// binder context so that bound variables get the correct `#` decorator.
```
cole-k, line 63: The computer wrote this and much of the other pretty printing changes.

There are a lot of changes to pretty printing I had it do, and I did not carefully review them. I'm pretty sure the only place we are using the pretty printing is in suggestions right now, but if you want me to review it more carefully before we merge then I will. I could see this code becoming more important later, and it would be better to document any hacks it made or fix them.

crates/flux-middle/src/big_int.rs
```diff
@@ -69,6 +69,71 @@ impl BigInt {
     pub fn uint_max(bit_width: u32) -> BigInt {
         (u128::MAX >> (128 - bit_width)).into()
     }
+
+    pub fn neg(&self) -> Self {
```
cole-k, line 73: The computer added these for constant folding + simplification in `rty::Expr`. Used for pretty printing.

Seems reasonably legit.

crates/flux-middle/src/rty/expr.rs
```diff
@@ -2060,3 +2362,298 @@ pub(crate) mod pretty {
         }
     }
 }
+
+impl Expr {
+    /// Applies transformations to simplify and humanize canonical Z3 outputs.
+    pub fn prettify(&self) -> Self {
```
cole-k, line 2324: This is all computer-written and very hacky. Can note it in the code OR try and clean it up. I suspect there are things duplicated in what we already have in simplification code...

Presently only used for pretty printing suggestions.

crates/flux-middle/src/global_env.rs
```diff
@@ -44,13 +45,24 @@ pub struct GlobalEnv<'genv, 'tcx> {
     inner: &'genv GlobalEnvInner<'genv, 'tcx>,
 }
 
+pub struct WeakKvarInfo {
+    /// Solutions (if provided as ground truth annotations). Each soution is
```
cole-k, line 50: NOTE: There is no support for ground truth annotations at present. Should I clarify that in a comment?

crates/flux-refineck/src/lib.rs
```diff
@@ -251,6 +319,66 @@ fn report_expected_neg(genv: GlobalEnv, def_id: LocalDefId) -> Result<(), ErrorG
     }))
 }
 
+fn add_fn_fix_diagnostic<'a>(
```
cole-k, line 334: This has been passed through the computer several times and frankly I can't say I understand all of the span hacking it does. But it seems from my testing to work relatively fine.

cole-k (commented):
Some general notes

crates/flux-bin/src/cargo_flux_opts.rs
```diff
@@ -119,22 +133,42 @@ pub struct CompileOpts {
     ///   <pattern>               — bare string treated as a glob
     #[arg(long, value_name = "PATTERN")]
     pub only_check: Option<String>,
+
+    /// Automatically apply Flux's suggested annotations to the source code
+    /// using 'cargo fix'.
+    ///
+    /// Implies --broken-code.
+    #[arg(long)]
+    pub fix: bool,
```
nilehmann, line 142: This should be its own subcommand (i.e., `cargo flux fix`), not a flag on `cargo flux check`

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -612,7 +613,23 @@ fn fn_sig(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::EarlyBinde
...
+                    fhir::Node::TraitItem(..)
+                        | fhir::Node::ForeignItem(..)
+                        | fhir::Node::ImplItem(..)
+                )
+            {
+                let id = match def_id {
+                    MaybeExternId::Extern(_local_id, def_id) => def_id,
+                    MaybeExternId::Local(local_id) => local_id.into(),
+                };
+                fn_sig = fn_sig.add_weak_kvars(genv, id)?;
```
```suggestion
                fn_sig = fn_sig.add_weak_kvars(genv, def_id.resolved_id())?;
```

crates/flux-infer/src/fixpoint_encoding/decoding.rs
```diff
@@ -184,7 +187,17 @@ where
...
+                        Ok(rty::Expr::fvar(rty::Name::from_u32(fname.as_u32() + 100_000)))
```
nilehmann: Are you sure this is needed? 

During parsing, we generate fresh local variables for existentianls and then during `fixpoint_to_expr` we push binders into `local_var_env`.

This is something Petros implemented, and you might not have noticed it.

crates/flux-infer/src/suggestions.rs
```diff
@@ -0,0 +1,220 @@
+use std::collections::HashSet;
```
nilehmann, line 1: I'm not going to look at this file

crates/flux-infer/src/wkvars.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// In theory this could be a Vec<rty::Expr>, but the instantiator
+    /// is configured right now to only return a single solution.
+    args_to_param: &'a HashMap<rty::Expr, rty::Expr>,
+    /// Set of self args
+    self_args: &'a HashSet<rty::Expr>,
+    /// Were any of the self args used in the expr?
+    any_self_args_used: bool,
+    /// In theory, this could (and probably should) map to multiple
+    /// solutions, i.e. a Vec<rty::Expr>.
+    memo: &'a mut HashMap<rty::Expr, rty::Expr>,
```
nilehmann: I believe this can be made unrodered containers

```suggestion
    args_to_param: &'a UnordMap<rty::Expr, rty::Expr>,
    /// Set of self args
    self_args: &'a UnordSet<rty::Expr>,
    /// Were any of the self args used in the expr?
    any_self_args_used: bool,
    /// In theory, this could (and probably should) map to multiple
    /// solutions, i.e. a Vec<rty::Expr>.
    memo: &'a mut UnordMap<rty::Expr, rty::Expr>,
```

crates/flux-middle/src/rty/expr.rs
```diff
@@ -963,6 +1146,87 @@ pub struct KVar {
...
+
+impl TypeFoldable for WKVar {
+    fn try_fold_with<F: FallibleTypeFolder>(&self, folder: &mut F) -> Result<Self, F::Error> {
+        Ok(WKVar {
+            wkvid: self.wkvid.try_fold_with(folder)?,
+            self_args: self.self_args,
+            args: self.args.try_fold_with(folder)?,
+        })
+    }
+}
```
nilehmann: Derive `TypeVisitable` and `TypeFoldable` instead of implementing it by hand.

crates/flux-middle/src/big_int.rs
nilehmann, line 73: Maybe we should start using proper bigints...

crates/flux-refineck/src/lib.rs
```diff
@@ -197,48 +199,114 @@ pub fn check_fn(
...
             | ConstrReason::Subtype(SubtypeReason::Input)
             | ConstrReason::Subtype(SubtypeReason::Requires)
-            | ConstrReason::Predicate => call_error(genv, span, err.dst_span),
-            ConstrReason::Assign => genv.sess().emit_err(errors::AssignError { span }),
+            | ConstrReason::Predicate => call_error(genv, span, tag.dst_span),
+            ConstrReason::Assign => {
+                genv.sess()
+                    .dcx()
+                    .handle()
+                    .create_err(errors::AssignError { span })
```
nilehmann, line 249: This is going to conflict with some open prs.

lib/flux-alloc/flux.toml
```diff
@@ -0,0 +1 @@
+no_suggestions = true
```
nilehmann: Don't put stuff in `flux.toml`. If it's a crate-level configuration, it should go in `Cargo.toml` or as a crate attribute.

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -71,6 +72,251 @@ impl<T: Types> Constraint<T> {
...
+    ///    outputs and letting consumers choose which wkvars to solve in each
+    ///    constr.
+    ///
+    /// ```
+    /// (p || q) => c
+    /// ==
+    /// (p => c) && (q => c)
+    /// ```
+    pub fn wkvars_and_constrs(&self) -> Vec<WKVarAndConstrs<T>>
+where {
```
nilehmann:
```suggestion
    pub fn wkvars_and_constrs(&self) -> Vec<WKVarAndConstrs<T>> {
```

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -460,12 +760,287 @@ impl<T: Types> Expr<T> {
...
+            Expr::Exists(_sorts, expr) => {
+                !expr.wkvars_in_conj().is_empty() || expr.has_wkvar_reachable_by_split()
+            }
+            Expr::And(exprs) => exprs.iter().any(|expr| expr.has_wkvar_reachable_by_split()),
+            _ => false,
+        }
+    }
+
+    pub fn hoist_exists(&self) -> (Vec<VarSorts<T>>, Expr<T>)
+where {
```
nilehmann: What's up with these empty `where` clauses?

lib/liquid-fixpoint/src/sexp.rs
```diff
@@ -141,7 +141,7 @@ impl<'a> Parser<'a> {
             "true" => Ok(Sexp::Atom(Atom::B(true))),
             "false" => Ok(Sexp::Atom(Atom::B(false))),
             _ => {
-                if let Ok(i) = s.parse::<u128>() {
+                if let Ok(i) = s.parse::<i128>() {
```
nilehmann, line 144: Does fixpoint generate negative literals? That looks like a bug in fixpoint.

crates/flux-bin/src/cargo_flux_opts.rs
nilehmann: Saying the two env vars are independent is a bit confusing. My understanding is that `cargo` will pass to the binary specified in `RUSTC_WORKSPACE_WRAPPER`, the value of `RUSTC` if set, no?

crates/flux-bin/src/cargo_flux_opts.rs
cole-k: Clippy uses `--fix`, what's the rationale for having a separate subcommand?

lib/liquid-fixpoint/src/sexp.rs
cole-k: Why wouldn't fixpoint be able to generate negative literals?

crates/flux-bin/src/cargo_flux_opts.rs
nilehmann: I suppose it works, but this is adding the option to `CompileOpts` which is also used  for `cargo flux build` so it'll show up in `cargo flux build --help`. It is also forwarding `FixOpts` to all commands, so if you run `cargo flux --allow-dirty` it will crash when executing `cargo`

lib/liquid-fixpoint/src/sexp.rs
nilehmann: because that's bad smtlib syntax. If you want `-3` you write `(- 3)`

cole-k: Fixpoint does. I did not write this repro.

```haskell
{-# LANGUAGE OverloadedStrings #-}

-- Demonstrates the path from unary negation to an atomic negative integer in
-- the JSON representation used for --json solutions.
module Main where

import qualified Data.Aeson as A
import qualified Data.ByteString.Lazy.Char8 as BL
import qualified Data.HashMap.Strict as HM
import Language.Fixpoint.Solver.Simplify (applyConstantFolding)
import Language.Fixpoint.Types

-- This is the AST produced for unary negation of the positive literal 3.
unary :: Expr
unary = ENeg (ECon (I 3))

-- This is the same folding operation used for ENeg by the solver.
folded :: Expr
folded = case unary of
  ENeg e -> applyConstantFolding Minus (ECon (I 0)) e
  e      -> e

-- Result's JSON instance serializes solution bodies through ScopedExpr and
-- toHornSMT, just as the --json code path does.
resultWithNegative :: Result (Integer, ())
resultWithNegative =
  safe
    { resSolution = HM.singleton (intKvar 0) folded
    }

main :: IO ()
main = do
  print unary
  print folded
  BL.putStrLn (A.encode resultWithNegative)
```

save as `repro-negative-json.hs` and run `stack runghc repro-negative-json.hs` in liquid-fixpoint's root.

Output

```
ENeg (ECon (I 3))
ECon (I (-3))
{"nonCutsSolution":[],"solution":[{"kvar":"k_##0","val":"(lambda () -3)"}],"status":{"contents":{"numBrkt":0,"numChck":0,"numCstr":0,"numIter":0,"numVald":0},"tag":"Safe"}}
```

cole-k: To my eye this is evidence enough but it is not an end-to-end proof… Forgive me if I have not verified it in enough detail.

crates/flux-infer/src/fixpoint_encoding/decoding.rs
cole-k: You're right. I'll leave a note.

cole-k:
@nilehmann can we meet today or tomorrow and finally bring this to a close?

crates/flux-bin/src/cargo_flux_opts.rs
```diff
@@ -372,3 +406,29 @@ impl ManifestOptions {
...
     }
 }
+
+/// Options forwarded verbatim to `cargo fix` when `--fix` is passed.
+/// These are ignored when running `cargo check`.
+#[derive(Default, Debug, clap::Args)]
+#[command(about = None, long_about = None, next_help_heading = "Fix Options")]
+pub struct FixOpts {
+    /// Don't require a clean git working directory when running `--fix`.
+    #[arg(long, requires = "fix")]
```
cole-k, line 416: Try to make sure we die if this is called with anything but `cargo flux` OR `cargo flux check`.

crates/flux-refineck/src/lib.rs
```diff
@@ -291,6 +331,69 @@ fn report_expected_neg(genv: GlobalEnv, def_id: LocalDefId) -> Result<(), ErrorG
...
+        .span_to_snippet(fn_first_line)
+        .unwrap_or_else(|_| panic!("No snippet for span {:?}", fn_first_line));
+    let prefix_spaces = &fn_first_line_snippet[..fn_first_line_snippet
+        .find(|c: char| !c.is_whitespace())
+        .unwrap_or(fn_first_line_snippet.len())];
+    let subst_solutions = &wkvar_subst.subst_instantiations[&wkvid];
+    assert!(subst_solutions.len() == 1);
+
+    // Check if there's an existing spec attribute that needs to be replaced
+    if let Some(old_spec_span) = genv.spec_attr_span(wkvid.parent_fn) {
```
cole-k, line 365: Try getting span from genv.queries.fn_sig(wkvid.parent_fn) and see if this still works.

If we can do this, then rip out the spec_attr_span.

crates/flux-infer/src/fixpoint_encoding/decoding.rs
```diff
@@ -184,7 +187,16 @@ where
...
-                        Err(FixpointParseError::NoLocalVar(*fname))
+                        // This is a free var generated from an existential we
+                        // got from fixpoint. It should be unique (previously we
+                        // had a terrible hack to add 100_000 to ensure it was
+                        // --- this should be no longer necessary; however if
+                        // there are issues with name clashes because the
+                        // uniqueness property is violated, we should probably
+                        // just add make the local var an enum and make this a
+                        // separate enumeration).
+                        Ok(rty::Expr::fvar(rty::Name::from_u32(fname.as_u32())))
```
cole-k, line 198: Why do we get local vars that reach this path?

crates/flux-refineck/src/lib.rs
```diff
@@ -196,89 +201,124 @@ pub fn check_fn(
...
     } else {
         None
     };
-
+    let mut solutions_by_tag: FxHashMap<Tag, (TagIdx, PossibleSolutions)> = FxHashMap::default();
+    for error in errors {
+        if let Some((_, val)) = solutions_by_tag.get_mut(&error.tag) {
+            val.extend(error.possible_solutions);
+        } else {
+            solutions_by_tag.insert(error.tag, (error.tag_idx, error.possible_solutions));
```
cole-k, line 234: Add a vec of tag_idx as well.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -984,11 +1058,23 @@ where
...
+            // of the form `b`. By removing this optimization, we ensure
+            // that we offer suggestions of the form `a => b`, which is
+            // often more desirable.
+            //
+            // rty::ExprKind::BinaryOp(rty::BinOp::Imp, e1, e2) => {
+            //     let (bindings, assumption) =
+            //         self.assumption_to_fixpoint(e1)?;
+            //     let cstr = self.head_to_fixpoint(e2, mk_tag)?;
+            //     Ok(fixpoint::Constraint::foralls(bindings, mk_implies(assumption, cstr)))
+            // }
```
cole-k, line 1077: Remove this and explain what the comment is.

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -2560,3 +2725,54 @@ impl FromSexp<FixpointTypes> for SexpParseCtxt<'_> {
         Err(ParseError::err(format!("Unknown sort: {name}")))
     }
 }
+
+fn parse_wkvars(expr: &mut fixpoint::Expr) {
```
cole-k, line 2729: call it fix_parsing_hack or something

crates/flux-middle/src/rty/expr.rs
```diff
@@ -653,6 +657,184 @@ impl Expr {
...
     }
+
+    /// Replaces [`BoundReftKind::Named(..)`] in [`Var::Bound(..)`] with
+    /// [`BoundReftKind::Anon`]. This is to ensure that expr equality works
+    /// properly --- the names are just annotations that do not change the
+    /// expressions themselves.
+    ///
+    /// If you think you need this function, you may be looking for
+    /// [`Expr::erase_metadata()`].
+    pub fn erase_bound_reft_kind(&self) -> Expr {
```
cole-k, line 668: call anonymize_bound_reft

crates/flux-middle/src/rty/expr.rs
```diff
@@ -2034,3 +2318,298 @@ pub(crate) mod pretty {
         }
     }
 }
+
+impl Expr {
```
cole-k, line 2322: Move to a different file.

cole-k: Along with the other reduction/metadata erasing code above.

crates/flux-middle/src/rty/pretty.rs
```diff
@@ -205,14 +222,45 @@ impl Pretty for FnSig {
...
-            &self.output
+            "fn({:?})",
+            join!(", ", self.inputs.iter().map(|input| input.shallow_canonicalize()))
         )?;
-        if !self.requires.is_empty() {
-            w!(cx, f, " requires {:?}", join!(" ∧ ", &self.requires))?;
-        }
-        Ok(())
+        w!(cx, f, " -> ")?;
+        // Format requires *before* entering the output binder context: requires predicates
```
cole-k, line 229: Kill

crates/flux-middle/src/rty/refining.rs
```diff
@@ -497,3 +503,304 @@ pub fn refine_bound_variables(vars: &[ty::BoundVariableKind]) -> List<rty::Bound
         })
         .collect()
 }
+
+impl rty::PolyFnSig {
```
cole-k, line 507: put in wkvars module in rty.

crates/flux-middle/src/lib.rs
```diff
@@ -392,6 +392,8 @@ pub struct Specs {
     trait_items: UnordMap<OwnerId, surface::TraitItemFn>,
     impl_items: UnordMap<OwnerId, surface::ImplItemFn>,
     pub flux_items_by_parent: FxIndexMap<OwnerId, Vec<surface::FluxItem>>,
+    /// Maps function DefIds to their #[sig(...)] attribute spans (if they have one)
+    spec_attr_spans: UnordMap<DefId, Span>,
```
cole-k, line 396: Try removing this and using fn_sig span.

crates/flux-middle/src/queries.rs
```diff
@@ -1013,6 +1012,24 @@ impl<'genv, 'tcx> Queries<'genv, 'tcx> {
                         });
                     }
 
+                    // We only will add weak kvars if
```
cole-k, line 1015: Remove this? (hooray)

(make sure we don't regress)

lib/liquid-fixpoint/src/constraint.rs
```diff
@@ -52,6 +53,248 @@ impl<T: Types> Constraint<T> {
...
+            }
+        }
+    }
+}
+
+pub type WKVarAndConstrs<T> = (WKVar<T>, FlatConstraint<T>, Vec<FlatConstraint<T>>);
+pub type VarSorts<T> = (<T as Types>::Var, Sort<T>);
+
+#[derive_where(Clone, Debug)]
+pub struct FlatConstraint<T: Types> {
```
cole-k, line 107: // TODO: merge with HornClause

cole-k (commented):
Nico

nilehmann (approved):

crates/flux-bin/src/cargo_flux_opts.rs
cole-k: Also remove the long comment (probably wrong) about RUSTC_WRAPPER etc.

PR #1663, opened 2026-06-16
Core Ptr Specs - part 1

enjhnsn2:
This will be the first of a set of 3-4 PRs merging core library ptr specs.
In particular, this PR adds:
1. `add`, `sub`, `byte_add`, `byte_sub`, `read`, `read_unaligned`, `write`, `write_unaligned`, `Slice::as_ptr`, and `Slice::as_mut_ptr`.
2. flux_defs like `dereferencable`, `valid`, `aligned` and `valid_zst` to help us write clean specs for pointer functions in the future.
3. Tests for the above specs

### DIFF
--- a/lib/flux-core/src/ptr/mod.rs
+++ b/lib/flux-core/src/ptr/mod.rs
@@ -102,7 +102,6 @@ unsafe fn write<T>(dst: *mut T, src: T);
 #[cfg(flux)]
 #[extern_spec(core::ptr)]
 // - `dst` must be valid for writes or `T` must be a ZST.
-// - `dst` must be properly aligned. Use `write_unaligned` if this is not the case.
 // See: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/ptr/mod.rs#L1843-L1846
 #[spec(fn (dst: *mut[@p] T, src: T) requires valid(p, T::size_of()))]
 unsafe fn write_unaligned<T>(dst: *mut T, src: T);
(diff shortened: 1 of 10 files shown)


lib/flux-core/src/ptr/mod.rs
```diff
@@ -6,8 +44,21 @@ macro_rules! ptr_specs {
     ($mutable:tt) => {
         #[extern_spec(core::ptr)]
         impl<T> *$mutable T {
-            #[spec(fn (me: *$mutable[@base, @addr, @size] T, count: usize) -> *$mutable[base, addr + count, size - count] T)]
+            #[spec(fn (me: *$mutable[@p] T, count: usize)
+                -> *$mutable[p.base, p.addr + count * T::size_of(), p.size - count * T::size_of()] T)]
```
nilehmann: I believe this spec doesn't guarantee the following requirement

> If the computed offset is non-zero, then self must be [derived from](https://doc.rust-lang.org/std/ptr/index.html#provenance) a pointer to some [allocation](https://doc.rust-lang.org/std/ptr/index.html#allocation), and the entire memory range between self and the result must be in bounds of that allocation. In particular, this range must not “wrap around” the edge of the address space.

i.e., `p.addr + count * T::size_of()` must be less than or equal to `p.base + p.size`.

enjhnsn2: I believe `p.size` is the "remaining size" in the buffer. so our spec is `requires count * T::size_of() <= p.size`, but ya, it was missing that,

nilehmann (approved):
The suggestion about `write_unaligned` looks correct.

nilehmann (changes requested):
There's some discussion about this in https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/raw.20pointers.20from.20references/with/604906919

We need to decide whether we are sticking with bytes or not

nilehmann (approved):
We decided https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/raw.20pointers.20from.20references/near/604915954

PR #1672, opened 2026-06-18
Ref-to-Ptr Conversion

ranjitjhala:
Given a `Ty` and a `RawPtrKind`, creates a raw pointer to `Ty` with "size = 1"
see test `fn ref_to_ptr_read` in `crates/flux/tests/tests/with_deps/pos/extern_specs/flux_core_ptr01.rs`. Is this legit?


ranjitjhala:
@nilehmann and @enjhnsn2 -- is this PR "stale" (and should be nuked) because we want the size to be in bytes and not `1`?

nilehmann:
Yes, we want size to be in bytes now, and I think we want to go through the `Size::size_of` generic refinement so we dont have to recompute sizes

ranjitjhala:
Hmm so do we want to change the ONE to be something with size_of instead (in this very PR)

sweiglbosker:
Yeah, i think what you want is create a pointer that is sized to `T::size_of()` and aligned to `T::align_of()` (if we still plan to track alignment)

ranjitjhala:
@copilot can you modify this PR so that instead of using the size `Expr::constant(Constant::ONE)` we use the `T::size_of()` (and also the `T::align_of()`) where `T` is the `Ty` ?

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+    let size = Expr::field_proj(nu, rty::FieldProj::RawPtr { field: rty::RawPtrField::Size });
+    // base == addr: the pointer is at the start of its allocation (no offset)
+    // addr != 0:    Rust references are never null, so the derived pointer is non-null
+    // size == T::size_of(): the allocation holds exactly one element of type T (in bytes)
+    // addr % T::align_of() == 0: the address is properly aligned for type T
+    let pred = Expr::and_from_iter([
+        Expr::eq(base, addr.clone()),
+        Expr::ne(addr.clone(), Expr::zero()),
+        Expr::eq(size, size_of_expr),
+        Expr::eq(Expr::binary_op(BinOp::Mod(Sort::Int), addr, align_of_expr), Expr::zero()),
```
ranjitjhala, line 2135: I'm a bit worried this `mod` is exposed? Is this what happens in the other specs? Or can we hide it behind a `final` predicate at the trait level @sweiglbosker ?

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
     }
 }
 
+/// Given a `Ty` and a `RawPtrKind`, creates a raw pointer to `Ty` with `size = T::size_of()`
```
ranjitjhala: @sweiglbosker and/or @nilehmann did you want to eyeball this to see if its what we want?

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+fn raw_ptr_with_size(genv: GlobalEnv, kind: &RawPtrKind, ty: Ty) -> Result<Ty> {
+    let sized_id = genv.tcx().require_lang_item(LangItem::Sized, DUMMY_SP);
+    let pointee_bty = ty
+        .as_bty_skipping_existentials()
+        .unwrap_or_else(|| tracked_span_bug!("expected indexed type in ref-to-raw-ptr cast"))
+        .clone();
+    let ctor = pointee_bty.to_subset_ty_ctor();
+    let args = rty::List::from_arr([GenericArg::Base(ctor)]);
+    let size_of_expr = Expr::alias(
+        AliasReft { assoc_id: FluxDefId::new(sized_id, sym::size_of), args: args.clone() },
```
nilehmann: I don't like that we are calling `FluxDefId::new` here (that's the reason CI is failing). We should put this in `builtin_assoc_refts.rs` so we only have one source of truth for these ids. 

I'm thinking we should have something similar to `LangItem` and have a list of all the built-in assoc refts and then have something analogous to `require_lang_item` that returns the `FluxDefId`.

crates/flux-refineck/src/checker.rs
nilehmann: Hmm, good point/idea. 

The annoying bit is that we would still need to hard-code it in `builtin_assoc_refts.rs`.

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+    );
+    let align_of_expr = Expr::alias(
+        AliasReft { assoc_id: FluxDefId::new(sized_id, sym::align_of), args },
+        rty::List::empty(),
+    );
+
+    let bty = BaseTy::RawPtr(ty, kind.to_mutbl_lossy());
+    let nu = Expr::nu();
+    let base =
+        Expr::field_proj(nu.clone(), rty::FieldProj::RawPtr { field: rty::RawPtrField::Base });
```
```suggestion
    let base =
        Expr::field_proj(&nu, rty::FieldProj::RawPtr { field: rty::RawPtrField::Base });
```

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+    let addr =
+        Expr::field_proj(nu.clone(), rty::FieldProj::RawPtr { field: rty::RawPtrField::Addr });
```
```suggestion
    let addr = Expr::field_proj(&nu, rty::FieldProj::RawPtr { field: rty::RawPtrField::Addr });
```

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2099,48 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
     }
 }
 
+/// Given a `Ty` and a `RawPtrKind`, creates a raw pointer to `Ty` with `size = T::size_of()`
+/// and `addr % T::align_of() == 0`.
+/// see test `fn ref_to_ptr_read` in `crates/flux/tests/tests/with_deps/pos/extern_specs/flux_core_ptr01.rs`
+fn raw_ptr_with_size(genv: GlobalEnv, kind: &RawPtrKind, ty: Ty) -> Result<Ty> {
+    let sized_id = genv.tcx().require_lang_item(LangItem::Sized, DUMMY_SP);
+    let pointee_bty = ty
+        .as_bty_skipping_existentials()
```
nilehmann: This is wrong it'll potentially crash with stuff like

```rust
fn foo(r: &i32{v: v > 0}) -> *const i32 {
    r as *const i32
}
```

crates/flux-refineck/src/checker.rs
ranjitjhala: @copilot CHANGE this so that inside of using `FluxDefId::new` you instead:

1. ADD to the use the API provided in @crates/flux-middle/src/builtin_assoc_refts.rs a way to LOOKUP the `AliasReft` given a `DefId` and `Symbol`
2. USE this updated API to get the `AliasReft` (by querying with (sized_id, sym::size_of)`

crates/flux-refineck/src/checker.rs
ranjitjhala: @nilehmann  -- AFAICT this function is ONLY used privately, after all refinements have been erased -- see lines 1514-1517 above --  so the above crash shouldn't happen? [ That said, maybe its better to do the erasing inside `raw_ptr_with_size` itself?

ranjitjhala: looks like the erasing was factored out to avoid adding a method to `Checker` ...

nilehmann: Oh sorry, I wrote this message but never submitted it

Then we should do:
```rust
let ctor = self
  .default_refiner
  .refine_ty_or_base(&constant.ty)?
  .expect_base();
```
to get a `SubetTyCtor` directly instead of going through a `Ty` and then skipping the binder

crates/flux-refineck/src/checker.rs
sweiglbosker: I think the resulting spec looks good. Maybe mention that `addr != 0` and `base = addr` in the doc comment?

crates/flux-refineck/src/checker.rs
sweiglbosker: Haven't checked but I do think the way the other specs handle alignment isn't great. The way I see this going is:
1. We come up with a smarter way to handle pointer alignment in general (so we can hide the mod from the solver unless absolutely necessary)
2. We wire up this conversion to use whatever that looks like

crates/flux-refineck/src/checker.rs
ranjitjhala: ping @nilehmann ^^

ranjitjhala: ok I edited this with what I think you meant, can you take a look before I merge?

crates/flux-refineck/src/checker.rs
```diff
@@ -1511,9 +1513,13 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
             Rvalue::RawPtr(kind, place) => {
                 // ignore any refinements on the type stored at place
                 let ty = &env.lookup_rust_ty(genv, place).with_span(stmt_span)?;
-                let ty = self.refine_default(ty).with_span(stmt_span)?;
-                let ty = BaseTy::RawPtr(ty, kind.to_mutbl_lossy()).to_ty();
-                Ok(ty)
+                // let ty = self.refine_default(ty).with_span(stmt_span)?;
```
ranjitjhala:
```suggestion
```

tests/tests/with_deps/pos/extern_specs/flux_core_ptr01.rs
```diff
@@ -32,6 +32,8 @@ pub fn test_add_mut_ix(ptr: *mut i32) {
     }
 }
 
+
```
ranjitjhala, line 35:
```suggestion
```

crates/flux-refineck/src/checker.rs
```diff
@@ -2097,6 +2102,44 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
     }
 }
 
+/// Given a `Ty` and a `RawPtrKind`, creates a raw pointer to `Ty` with
+/// base == addr: the pointer is at the start of its allocation (no offset)
+/// addr != 0:    Rust references are never null, so the derived pointer is non-null
+/// size == T::size_of(): the allocation holds exactly one element of type T (in bytes)
+/// addr % T::align_of() == 0: the address is properly aligned for type T
+/// see test `fn ref_to_ptr_read` in `crates/flux/tests/tests/with_deps/pos/extern_specs/flux_core_ptr01.rs`
```
nilehmann: I hate AI comments. A better comment would be:

```
&mut T => *mut{p: p.size == T::size_of() && p.base == p.addr && p.addr % T::align_of() == 0 } T
```

nilehmann (approved):

crates/flux-refineck/src/checker.rs
```diff
@@ -2103,11 +2103,10 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
 }
 
-/// Given a `Ty` and a `RawPtrKind`, creates a raw pointer to `Ty` with
-/// base == addr: the pointer is at the start of its allocation (no offset)
-/// addr != 0:    Rust references are never null, so the derived pointer is non-null
-/// size == T::size_of(): the allocation holds exactly one element of type T (in bytes)
-/// addr % T::align_of() == 0: the address is properly aligned for type T
+/// Converts a reference into a raw-ptr, tracking size etc.
+///
+///     &mut T => *mut{p: p.size == T::size_of() && p.base == p.addr && p.addr % T::align_of() == 0 } T
```
sweiglbosker, line 2108: p.addr != 0?

PR #1673, opened 2026-06-18
Use correct `DefId` (sized_id) when building BUILTIN assoc-refts

ranjitjhala:
This fixes a bug where `cargo flux` finishes fine but `cargo flux --only-check <PATH>` was ICE-ing because of missing stuff e.g.

```
error: internal compiler error: crates/flux-middle/src/rty/mod.rs:813:32: caller should guarantee existence of associated refinement
```

The bug was because FLUX could not find the assoc refinement `size_of`.

Per CC

```
 The `def_id` here is the argument to `builtin_assoc_refts(self, def_id: DefId)` — it's whatever `DefId` was passed to look up. But this code is inside a
  `OnceLock::get_or_init` closure, so it only runs once — on the very first call. 
  The `def_id` captured in the closure is the `def_id` from that first call,
  which might not be `sized_id` at all.

  Compare with line 36 for `FnOnce`, which correctly uses `fn_once_id`:
  `FluxDefId::new(fn_once_id, sym::no_panic),  // correct: uses fn_once_id`

  And line 94 in builtin_assoc_reft_sort, which also correctly uses `sized_id`:
  `FluxDefId::new(sized_id, sym::size_of),  // correct: uses sized_id`

  The bug: Lines 50 and 54 use def_id (the function parameter, captured by the OnceLock closure) instead of sized_id. The first time builtin_assoc_refts is
  called, def_id gets baked into the FluxDefId for size_of and align_of. If that first call happens to pass sized_id, it works. If a different def_id is
  passed first, the FluxDefId is wrong, and later lookups by assoc_id won't match.

  With plain cargo flux, the iteration order over defs happens to call this with sized_id first (or the Sized trait is looked up first). With --only-check,
  a different def triggers the first call, so def_id is some unrelated DefId, and the FluxDefId for size_of/align_of is permanently wrong.

  The fix is to replace def_id with sized_id on lines 50 and 54.
```

@nilehmann does this "fix" seem correct to you?

### DIFF
--- a/crates/flux-driver/src/callbacks.rs
+++ b/crates/flux-driver/src/callbacks.rs
@@ -331,13 +331,11 @@ fn trigger_queries(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult {
             genv.generics_of(def_id)?;
             genv.predicates_of(def_id)?;
             genv.refinement_generics_of(def_id)?;
-            genv.assoc_refinements_of(def_id)?;
         }
         DefKind::Impl { .. } => {
             genv.generics_of(def_id)?;
             genv.predicates_of(def_id)?;
             genv.refinement_generics_of(def_id)?;
-            genv.assoc_refinements_of(def_id)?;
         }
         DefKind::Fn | DefKind::AssocFn => {
             genv.generics_of(def_id)?;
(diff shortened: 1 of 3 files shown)


nilehmann (approved):
Oh funny bug

ranjitjhala:
This was an interesting experience. CC was first just floundering around with all sorts of BS suggestions and theories. Then I got fed up and decided to do it the old fashioned way with TRACE here and there, which led me to this builtin_assoc_reft file; at which point I went back to CC and asked it what was fishy there -- at which point it sussed out the bug instantly...

PR #1676, opened 2026-06-21
Core pointer spec part 3: offset + pointer comparisons

enjhnsn2:
This PR adds specifications for the remaining common pointer functions not covered by the core spec.
Specifically, this PR:
1. adds specs for `offset`, `byte_offset`, `offset_from` (comparison), `offset_from_unsigned`, `null`, `null_mut`, the method versions of read and write (`ptr.read(4)` rather than `read(ptr, 4)`), and `replace`.
2. Updated our `ptr_spec` macro to handle functions that exist only for mutable pointers.
3. Made the specs for common arithmetic functions stronger. Previously, we required only the output of `add`,`sub`, and related functions to be in bounds of their allocated buffer. However, after inspecting the safety specs closer, it is also immediate UB if these functions are called on oob pointers. I have therefore added preconditions to enforce this requirement. 
4. Added tests for the above specs

This is the last set of pointer specs that I plan to work on. After this, I will work on `NonNull<T>`

### DIFF
--- a/lib/flux-core/src/ptr/mod.rs
+++ b/lib/flux-core/src/ptr/mod.rs
@@ -83,6 +83,22 @@ macro_rules! ptr_specs {
                 -> *$mutable[p.base, p.addr + count, p.size - count] T
                     requires count <= p.size && p.addr + count >= p.base)]
             unsafe fn byte_offset(self, count: isize) -> Self;
+
+            // - Both `self` and `origin` must be derived from a pointer to the same allocation,
+            //   and the memory range between them must be in bounds of that allocation.
+            //   (`p.base == op.base` captures same-allocation provenance; `addr >= base` for
+            //   each captures the in-bounds requirement from below.)
+            // - The distance between the pointers in bytes must be an exact multiple of
+            //   `size_of::<T>()`, so that it corresponds to a whole number of elements.
+            // - `T` must not be a ZST; the function panics via `assert!(0 < size_of::<T>())`.
+            // See: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/ptr/const_ptr.rs#L614-L628
+            #[spec(fn (me: *$mutable[@p] T, origin: *const[@op] T) -> isize[(p.addr - op.addr) / T::size_of()]
+                requires p.base == op.base &&
+                         p.addr >= p.base && op.addr >= op.base &&
+                         (p.addr - op.addr) % T::size_of() == 0 &&
+                         T::size_of() > 0)]
+            unsafe fn offset_from(self, origin: *const T) -> isize
+            where T: Sized;
         }
     };
 }
--- a/lib/flux-core/src/ptr/mod.rs
+++ b/lib/flux-core/src/ptr/mod.rs
@@ -99,6 +99,18 @@ macro_rules! ptr_specs {
                          T::size_of() > 0)]
             unsafe fn offset_from(self, origin: *const T) -> isize
             where T: Sized;
+
+            // - All safety conditions of `offset_from` apply.
+            // - Additionally, `self` must be greater than or equal to `origin`.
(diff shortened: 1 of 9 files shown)


lib/flux-core/src/ptr/mod.rs
```diff
@@ -1,83 +1,161 @@
...
+    // do not overlap.
     fn nonoverlapping(p: ptr, q: ptr, num_bytes: int) -> bool {
         p.addr + num_bytes <= q.addr || q.addr + num_bytes <= p.addr
     }
+
+    // The pointer itself lies within its allocation: addr >= base and size >= 0.
+    // One-past-the-end pointers (size == 0) satisfy this — they are valid
+    // starting points for pointer arithmetic but cannot be dereferenced.
+    // Required as a precondition for all pointer arithmetic methods.
+    fn in_bounds(p: ptr) -> bool { dereferenceable(p, 0) }
```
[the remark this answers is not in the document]
enjhnsn2: @copilot `ptr::null::<T>().add(0)` is not ub, per https://doc.rust-lang.org/std/primitive.pointer.html "If the computed offset is non-zero, then self must be [derived from](https://doc.rust-lang.org/std/ptr/index.html#provenance) a pointer to some [allocation](https://doc.rust-lang.org/std/ptr/index.html#allocation), and the entire memory range between self and the result must be in bounds of that allocation.". Do you have reason to believe it is UB?

lib/flux-core/src/ptr/mod.rs
```diff
@@ -1,83 +1,161 @@
...
+        requires valid(p, T::size_of()) && aligned_to(p, T::align_of()))]
+    unsafe fn write(self, val: T)
+    where
+        T: Sized;
+    /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/ptr/mut_ptr.rs#L1490
+    #[spec(fn (me: *mut[@p] T, src: T) -> T
+        requires valid(p, T::size_of()) && aligned_to(p, T::align_of()))]
+    unsafe fn replace(self, src: T) -> T
+    where
+        T: Sized;
```
nilehmann, line 154: I'd put this outside the macro in a separate `impl *mut T` block

nilehmann (commented):
LGTM, but this is getting too gnarly for me to be confident we are not missing something. 

It is also unclear which parts we are not covering or approximating (e.g., dangling pointers, aliasing model, ...). It'd be good if we could have a somewhat rigorous claim, such as "we are covering everything except x, y, and z"

lib/flux-core/src/ptr/mod.rs
enjhnsn2: > I'd put this outside the macro in a separate `impl *mut T` block

That's what I tried originally, but it fails with:
```
   --> lib/flux-core/src/ptr/mod.rs:140:6
    |
140 | impl<T> *mut T {
    |      ^ extern spec for `core::ptr::mut_ptr::<impl *mut T>::T` redefined here
    |
note: previous extern spec for `core::ptr::mut_ptr::<impl *mut T>::T` defined here
   --> lib/flux-core/src/ptr/mod.rs:56:14
    |
 56 |         impl<T> *$mutable T {
    |              ^
...
137 | ptr_specs!(mut);
    | --------------- in this macro invocation
    = note: this error originates in the macro `ptr_specs` (in Nightly builds, run with -Z macro-backtrace for more info)
```

enjhnsn2:
> LGTM, but this is getting too gnarly for me to be confident we are not missing something.
> 
> It is also unclear which parts we are not covering or approximating (e.g., dangling pointers, aliasing model, ...). It'd be good if we could have a somewhat rigorous claim, such as "we are covering everything except x, y, and z"

I feel pretty confident (as confident as one can be without testing :) ) about what we are/are not proving. In fact, I just added a comment describing exactly these properties.
The short summary is:
1. spatial safety: yes
2. per method UB safety contracts (alignment, non-overlapping, no dangling pointers): yes
3. temporal safety: no

nilehmann (approved):

PR #1680, opened 2026-06-26
spec for windows and position

enjhnsn2:
This PR adds specs for `position` and `slice::windows`.
The spec for position specifies that the output (if it is `Some`) returns an idx < `iter.size`.
The spec for `windows` propagates size constraints across the iter. The lets us verify code like
```rust
pub fn test_windows_index_safe(xs: &[i32]) {
    let mut it = xs.windows(2);
    if let Some(w) = it.next() {
        let _diff = w[1] - w[0];
    }
```

These were both added in service of verifying the insulin pump.


ranjitjhala:
no comments! amazing!

nilehmann (approved):

PR #1684, opened 2026-06-30
Core ptr specs part 4: `NonNull<T>`

enjhnsn2:
This PR adds flux-core specs for `NonNull<T>`.
Specifically it:
1. Adds specs for `new`, `new_unchecked`, `as_ptr`, `add`, `sub`, `offset`, ` byte_add`, `byte_sub`, `byte_offset`, `read`, `write`, `offset_from`, and `slice_from_raw_parts`. For those most part, these specs are syntactically identical to those in ptr. 
2. Adds a definition and invariant for `NonNull<T>` 
```rust
#[refined_by(base: int, addr: int, size: int)]
#[invariant(addr != 0)]
struct NonNull<T>;
```
3. Adds tests for the above

Technical note: I could have indexed `NonNull<T>` by a `ptr` rather than by `base, addr, size`. I chose to use the 3 indices to 1) make the specs as syntactically identical to those in ptr as possible, and 2) allow us to do direct indexing when updating a single one of those fields. If you would rather have it indexed by a ptr, I can do that.

### DIFF
--- a/lib/flux-core/src/ptr/mod.rs
+++ b/lib/flux-core/src/ptr/mod.rs
@@ -52,6 +52,8 @@
 /// These specs do NOT prove temporal safety.
 /// To prove temporal safety, Flux would need to tie the lifetimes of allocations to the pointers derived from them,
 /// which is currently out of scope.
+mod non_null;
+
 use flux_attrs::*;
 
 macro_rules! ptr_specs {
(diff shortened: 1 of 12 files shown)


enjhnsn2:
The lean-demo is failing, but I'm pretty sure its not my fault :)

nilehmann (approved):

PR #1687, opened 2026-07-16
Add "spacer" as a back-end

ranjitjhala:
hooking into the `smt-horn` translation.

### DIFF
--- a/crates/flux-infer/src/fixpoint_encoding.rs
+++ b/crates/flux-infer/src/fixpoint_encoding.rs
@@ -31,10 +31,9 @@ use flux_middle::{
 };
 use itertools::Itertools;
 use liquid_fixpoint::{
-    FixpointStatus, KVarBind, SmtSolver, Stats, VerificationResult,
+    FixpointStatus, KVarBind, SmtSolver, VerificationResult,
     parser::{FromSexp, ParseError},
     sexp::Parser,
-    smt_horn::SmtFormatter,
 };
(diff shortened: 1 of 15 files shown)


crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -913,8 +917,17 @@ where
             return result.clone();
         }
         let result = metrics::time_it(TimingKind::FixpointQuery(def_id, kind), || {
-            task.run()
-                .unwrap_or_else(|err| tracked_span_bug!("failed to run fixpoint: {err}"))
+            match backend {
+                Backend::Spacer => {
+                    let path = dbg::item_dump_path(genv.tcx(), def_id, "horn.smt2");
```
ranjitjhala: this is a bit icky -- but because the current "dump" hardwires the fixpoint backend; I'll have to clean it up.

crates/flux-infer/src/infer.rs
```diff
@@ -241,7 +247,8 @@ impl<'genv, 'tcx> InferCtxtRoot<'genv, 'tcx> {
         fcx.generate_lean_files(def_id, task)
     }
 
-    pub fn execute_fixpoint_query(
+    // We use this for the `fixpoint` or `SMTLIB` queries
```
ranjitjhala, line 242:
```suggestion
    // We use this for the `fixpoint` or `SmtHorn` queries
```

PR #1692, opened 2026-07-21
Callee Specs

ranjitjhala:
Get CC to add "callee-spec-events" to the trace, to make it easier to see what contracts are used for callee.

So at each call site to `foo(...)` we emit an event for the SPEC for `foo` where the SPEC is the fn-sig but as a raw string.

The main hassle is for extern-specs for extern-crates (e.g. flux-alloc etc.) which is where this is most required, where the whole pipeline has to be tweaked to serialize the spec string.

Changes across 6 files:

1. crates/flux-middle/src/lib.rs — Added spec_attr_strings: UnordMap<DefId, String> to Specs, with set_spec_attr_string / get_spec_attr_string accessors.
2. crates/flux-middle/src/global_env.rs — Added GlobalEnv::spec_attr_string() that checks local specs, falls back through extern_id_to_local_id for same-crate extern specs, then falls back to cstore for cross-crate.
3. crates/flux-middle/src/cstore.rs — Added spec_attr_string to the CrateStore trait.
4. crates/flux-metadata/src/lib.rs — Added spec_attr_string field to Tables, serialized it during encoding for Fn/AssocFn items, implemented the CrateStore method on CStore, and added it to merge_extern_tables.
5. crates/flux-driver/src/collector/mod.rs — At the three sites where set_spec_attr_span is called, also extract the snippet and call set_spec_attr_string.
6. crates/flux-driver/src/collector/extern_specs.rs — Switched collect_extern_fn, collect_extern_impl_fn, and collect_extern_trait_fn from attrs.fn_sig() to attrs.fn_sig_with_attr_span() so that extern spec functions also capture their spec attribute span and string.
7. crates/flux-common/src/dbg.rs — Updated call! macro to use genv.spec_attr_string() instead of span-based snippet extraction.


crates/flux-middle/src/lib.rs
```diff
@@ -439,6 +439,7 @@ impl Specs {
         self.spec_attr_spans.get(&def_id).copied()
     }
 
+
```
ranjitjhala:
```suggestion
```

ranjitjhala:
@nilehmann ok I think this is ready for you to take a look at?

ranjitjhala:
@nilehmann actually wait, I want to add ONE more thing...

ranjitjhala:
Ok added -- @nilehmann this is ready for you to look at

nilehmann (commented):
LGTM

PR #1693, opened 2026-07-22
Dump callgraph

ranjitjhala:


ranjitjhala:
@nilehmann can we merge this?

crates/flux-middle/src/call_graph.rs
```diff
@@ -208,3 +211,94 @@ impl<'tcx> fmt::Display for CallGraph<'tcx> {
...
+
+    let json_nodes: Vec<JsonNode> = topo_order
+        .iter()
+        .enumerate()
+        .map(|(uid, &orig_idx)| {
+            let def_id = def_ids[orig_idx];
+            let dep_uids: Vec<usize> = adj[orig_idx].iter().map(|&j| uid_of[j]).collect();
+            JsonNode {
+                uid,
+                path: tcx.def_path_str(def_id),
```
nilehmann, line 251: As I mentioned in Zulip, this is not faithfully representing the data because it's losing the instantiated generic arguments.

crates/flux-driver/src/callbacks.rs
```diff
@@ -142,6 +142,17 @@ fn check_crate(genv: GlobalEnv) -> Result<(), ErrorGuaranteed> {
...
+        if config::dump_call_graph() {
+            let graph = genv.call_graph();
+            if let Err(e) = graph.dump_json(genv.tcx(), config::log_dir()) {
+                tracing::warn!("failed to dump call graph: {e}");
+            }
+            if let Err(e) = flux_middle::type_graph::dump_type_graph(genv.tcx(), config::log_dir())
+            {
+                tracing::warn!("failed to dump type graph: {e}");
+            }
+        }
```
nilehmann: The convention is to dump stuff at the point we compute them. That'd be `flux_opt::call_graph::build_call_graph` for the call graph

crates/flux-middle/src/call_graph.rs
ranjitjhala: Ok, but I don't think that matters (at this point) where we're just getting a _very_ rough dependency analysis -- mostly as a hint for CC -- and not trying to do anything precise about these dependencies anyways?

ranjitjhala: (remains to be seen if its at all useful there either for that matter, but its at least probably more precise than whatever claude cooks up with grep...)

nilehmann: Agree, but let's dump what it is instead of a weird thing.

ranjitjhala: ok I'm lost, what exactly do you want dumped here? For my use-case I *only* want the `DefId` and *not* the instantiated generic arguments, since I just want a coarse dependency of "check this DefId before that DefId etc..."?

ranjitjhala: (continuing discussion on Zulip)

PR #1697, opened 2026-07-24
Add test with enum-record-style variant

ranjitjhala:
Since I couldn't find one in the regtests!


ranjitjhala:
merging as this just adds a test.

PR #1702, opened 2026-07-29
Guarded Mut References

ranjitjhala:
CC summary:

1. ghost_statements.rs — new Point::AtReturn(Location) with its own statement map, plus dump support.
2. ghost_statements/fold_unfold.rs — Mode::ret emits the argument folds at Point::AtReturn instead of Point::BeforeLocation.
3. checker.rs — check_ret collects those folds and runs them between the return-value subtyping and check_ensures, all inside the existing evar scope. A separate fold_span keeps fold errors on the closing brace (this preserved the spans in neg/error_messages/localize04.rs and neg/surface/issue-299.rs).
4. type_env/place_ty.rs — new unblock_fields: when folding a struct/tuple whose field is Blocked(t) (a live mutable borrow into it), the field is folded using the borrow's guard t, unpacked to a fresh variable with its invariants assumed. The resulting aggregate stays blocked, since the borrow is still live. This is what makes the invariant check succeed for a strong guard and fail for a weak one.

Tests

- tests/tests/pos/surface/guarded_mut00.rs — 6 cases: simple field, write-then-return, nested field (s.inner.bar), unconstrained field, whole-argument borrow, tuple.
- tests/tests/neg/surface/guarded_mut00.rs — 3 cases: no guard, a guard that holds of the current bar but is still too weak (v + 1 >= me.foo), and nested.

**Scope note:** this makes the body check sound and precise. The caller side is unchanged — after let r = test_ok(&mut foo); flux does not block foo while r is live, so writing through r won't be reflected in foo's index. Your // want callers to "block s" comment suggests you already know this; it's a separate piece of work.

### DIFF
--- a/crates/flux-refineck/src/type_env/place_ty.rs
+++ b/crates/flux-refineck/src/type_env/place_ty.rs
@@ -867,6 +867,28 @@ fn downcast_enum(
     Ok(variant_sig.fields.to_vec())
 }
 
+/// A field of an aggregate can be *blocked* if there's a live mutable borrow into it, e.g., after
+/// `let r = &mut s.f`. To fold `s` back we use the *guard* of the borrow, i.e., the (upper) bound
+/// `r` is allowed to write back, as the type of the field. The aggregate itself remains blocked
+/// because the borrow is still live.
+///
+/// Returns whether any field was blocked together with the unblocked fields.
+fn unblock_fields(infcx: &mut InferCtxtAt, fields: &[Ty]) -> (bool, Vec<Ty>) {
+    let mut blocked = false;
+    let fields = fields
+        .iter()
+        .map(|ty| {
+            if let TyKind::Blocked(bound) = ty.kind() {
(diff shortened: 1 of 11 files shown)


crates/flux-refineck/src/type_env/place_ty.rs
```diff
@@ -867,6 +867,28 @@ fn downcast_enum(
...
+///
+/// Returns whether any field was blocked together with the unblocked fields.
+fn unblock_fields(infcx: &mut InferCtxtAt, fields: &[Ty]) -> (bool, Vec<Ty>) {
+    let mut blocked = false;
+    let fields = fields
+        .iter()
+        .map(|ty| {
+            if let TyKind::Blocked(bound) = ty.kind() {
+                blocked = true;
+                infcx.hoister(true).hoist(bound)
```
ranjitjhala: @nilehmann any idea what this is doing?

PR #1703, opened 2026-07-30
Make flux-core mods public

ranjitjhala:
So we can use the defs within.


ranjitjhala:
@nilehmann I'll merge if CI is ok?

PR #1704, opened 2026-08-01
Fix PtrToPtr cast to preserve pointer refinements

### LINKED ISSUE #1700: Lossy cast to `*mut T`
The following fails to verify as the ptr as *mut T is a PtrToPtr cast that erases all refinements? Is there any reason for this? Or can I update flux to preserve the refinements here?

```rust
extern crate flux_core;
use std::ptr::NonNull;

// --- casts ---

#[flux_rs::spec(fn (ptr: *const[@p] T) -> Option<NonNull<T>{nn: nn.base == p.base && nn.addr == p.addr && nn.size == p.size}>)]
fn new<T>(ptr: *const T) -> Option<NonNull<T>> {
    match NonNull::new(ptr as *mut T) {
        None => None,
        Some(nn) => Some(nn),
    }
}

Hamid0004:
Fixes #1700

Following @ranjitjhala's suggestion to look at how `IntToInt` casts preserve
refinements, I traced the issue to `check_cast` in
`crates/flux-refineck/src/checker.rs`.

`CastKind::PtrToPtr` was falling into the same catch-all as casts that
should legitimately erase refinements (e.g. `FloatToInt`), calling
`self.refine_default(to)?` and discarding the source pointer's index
(`base`, `addr`, `size`). This meant `ptr as *mut T` always lost any
refinement info tracked on `ptr`.

Following the existing pattern used for `MutToConstPointer`, I gave
`PtrToPtr` its own arm: when the pointee type is unchanged, it clones the
source pointer's index onto the destination pointer type instead of
falling back to `refine_default`. When the pointee type *does* change, it
still falls back to the default (safe) behavior.

### Changes
1. Added a dedicated `CastKind::PtrToPtr` arm in `checker.rs` that
   preserves the refinement index when the pointee type matches.
2. Added a regression test at `tests/with_deps/pos/ptr_cast_1700.rs`
   mirroring the exact `NonNull::new(ptr as *mut T)` repro from the issue.
3. Added a simpler unit test (`const_to_mut_ptr`) in
   `tests/pos/casts/pointers.rs`, mirroring the existing
   `mut_to_const_ptr` test.

### Testing
Ran the full local test suite — all passing, no regressions:

\`\`\`
test result: ok. 449 passed; 0 failed; 4 ignored   (pos)
test result: ok. 407 passed; 0 failed; 2 ignored   (neg)
test result: ok. 102 passed; 0 failed; 1 ignored   (with-deps pos)
test result: ok.  65 passed; 0 failed; 0 ignored   (with-deps neg)
\`\`\`

Also specifically verified the new regression test:
\`\`\`
test [ui] pos/ptr_cast_1700.rs ... ok
\`\`\`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ fixpoint
 
 .DS_Store
 fixpoint
+*.log
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -1770,10 +1770,22 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                     _ => self.refine_default(to)?,
                 }
             }
+            CastKind::PtrToPtr => {
+                match (from.kind(), to.kind()) {
+                    (
+                        TyKind::Indexed(BaseTy::RawPtr(from_inner_ty, _), idx),
+                        RustTy::RawPtr(to_inner_ty, to_mutbl),
+                    ) if from_inner_ty.to_rustc(self.genv.tcx())
+                        == to_inner_ty.to_rustc(self.genv.tcx()) =>
+                    {
+                        Ty::indexed(BaseTy::RawPtr(from_inner_ty.clone(), *to_mutbl), idx.clone())
+                    }
+                    _ => self.refine_default(to)?,
+                }
+            }
             CastKind::FloatToInt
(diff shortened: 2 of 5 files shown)


nilehmann (approved):
LGTM. Thanks @Hamid0004!

ranjitjhala:
Thanks @nilehmann and @Hamid0004 !

PR #1705, opened 2026-08-03
Implement `qualifier!` macro

### LINKED ISSUE #1389: Create a `qualifier!` macro
Spun out of #1386

@nilehmann you'll see that the bulk of the work here is really the 

```
invariant qualifier Auto(x1:s1,... ) { e }
```

I think we should probably also JUST have a convenient way to generate that with a macro, so I can write something like

```rust
#[spec(fn (n: usize) -> usize[n])]
pub fn test(n: usize) -> usize {
    let mut i = n;
    let mut res = 0;
    while i > 0 {
        qualifier!(my_plus(res, i) == n); // just generates the qualifier stuff
        assert(res + i == n) // reuses the existing assert
        i -= 1;
        res += 1;
    }
    res
}
```

Its not perfect, but at least you can 

1. Put the `assert` and the `qualifier` next to each other;
2. Make the qualifier "local";
3. Decouple the syntax-intersection issue, so you can use whatever you like in refinements.

So basically you get a "localized" way to drop qualifier hints wherever.

_Originally posted by @ranjitjhala in https://github.com/flux-rs/flux/issues/1386#issuecomment-3597862909_

kris-gaudel:
Resolves #1389, follow-up to #1386.

Code example:

```rust
defs! {
    fn my_plus(x: int, y: int) -> int { x + y }
}

#[spec(fn (n: usize) -> usize[n])]
pub fn test(n: usize) -> usize {
    let mut i = n;
    let mut res = 0;
    while i > 0 {
        qualifier!(my_plus(res, i) == n);
        assert(res + i == n);
        i -= 1;
        res += 1;
    }
    res
}
```

## Changes
- `lib/flux-rs/src/lib.rs` — the `qualifier!` macro (`tt`-captured body, plus an optional `params ; body` arm for sorts) and its re-export in `flux_rs::macros`
- `crates/flux-syntax/src/parser/mod.rs` — uniquify hint qualifier names with `next_node_id()`
- `crates/flux-syntax/src/surface.rs` — `free_vars` no longer collects call callees
- `tests/tests/with_deps/pos/surface/qualifier_macro00.rs` — new test
- `book/src/guide/specifications.md` — new "Qualifier Macro" section, `{{#include}}`ing the test

### DIFF
--- a/crates/flux-syntax/src/surface.rs
+++ b/crates/flux-syntax/src/surface.rs
@@ -933,14 +933,12 @@ impl Expr {
                         if let [segment] = path.segments.as_slice() {
                             self.vars.insert(segment.ident);
                         }
-                        visit::walk_expr(self, expr);
                     }
                     ExprKind::Call(callee, args) => {
                         // The callee of a call is a refinement function, not a variable, so
-                        // don't collect it. Anything more complex than a single-segment path
-                        // may contain free variables, so visit it normally.
-                        if !matches!(&callee.kind, ExprKind::Path(path) if path.segments.len() == 1)
-                        {
+                        // don't collect it. Anything that's not a path may contain free
+                        // variables, so visit it normally.
+                        if !matches!(&callee.kind, ExprKind::Path(_)) {
                             self.visit_expr(callee);
                         }
                         for arg in args {
--- a/crates/flux-syntax/src/surface.rs
+++ b/crates/flux-syntax/src/surface.rs
@@ -927,14 +927,29 @@ impl Expr {
 
         impl visit::Visitor for FreeVarsVisitor {
             fn visit_expr(&mut self, expr: &Expr) {
-                if let ExprKind::Path(path) = &expr.kind {
-                    // Only collect paths with a single segment
-                    if let [segment] = path.segments.as_slice() {
-                        self.vars.insert(segment.ident);
+                match &expr.kind {
+                    ExprKind::Path(path) => {
+                        // Only collect paths with a single segment
+                        if let [segment] = path.segments.as_slice() {
+                            self.vars.insert(segment.ident);
+                        }
+                        visit::walk_expr(self, expr);
                     }
+                    ExprKind::Call(callee, args) => {
+                        // The callee of a call is a refinement function, not a variable, so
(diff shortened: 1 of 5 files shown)


crates/flux-syntax/src/surface.rs
```diff
@@ -927,14 +927,29 @@ impl Expr {
...
+                        if let [segment] = path.segments.as_slice() {
+                            self.vars.insert(segment.ident);
+                        }
+                        visit::walk_expr(self, expr);
                     }
+                    ExprKind::Call(callee, args) => {
+                        // The callee of a call is a refinement function, not a variable, so
+                        // don't collect it. Anything more complex than a single-segment path
+                        // may contain free variables, so visit it normally.
+                        if !matches!(&callee.kind, ExprKind::Path(path) if path.segments.len() == 1)
```
nilehmann: nit: if it's a path, it's not necessary to recurse, regardless of the length. Also, I think we don't need the `visit::walk_expr` for the `ExprKind::Path` case

```suggestion
                        if !matches!(&callee.kind, ExprKind::Path(path))
```

lib/flux-rs/src/lib.rs
```diff
@@ -52,6 +52,34 @@ macro_rules! __private_invariant {
...
+/// # Example
+/// ```
+/// flux_rs::macros::qualifier!(my_plus(res, i) == n);
+/// flux_rs::macros::qualifier!(res: int, i: int ; res + i == n);
+/// ```
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __private_qualifier {
+    ($($param:ident : $ty:ty),+ ; $($body:tt)*) => {
+        #[flux::defs{
```
nilehmann: This will fail when compiling a crate without flux (plain `cargo build` on the crate). We should use `[cfg_attr(flux, ...)]` or `$crate::defs!`.

This is also wrong for `__private_invariant` and ` __private_detached_spec` btw.

nilehmann (commented):
Thanks @kris-gaudel! 

A couple of small comments.

nilehmann (approved):
Awesome!

PR #1706, opened 2026-08-04
CC restores brackets; (but why had they disappeared!)

ranjitjhala:
So that the log/checker has `usize[4]` instead of `usize4` ...

(I'd fixed this earlier, I wonder why this got reverted...)


ranjitjhala:
@nilehmann ok to merge?

nilehmann (approved):

PR #1707, opened 2026-08-04
Add snippet about importing flux def to specifications.md

ranjitjhala:


ranjitjhala:
@nilehmann merging this as its just a doc update (hopefully not wrong!)

PR #1710, opened 2026-08-05
Preserve pointer indices on cast

### LINKED ISSUE #1709: Remove source-ty / dst-ty check when casting?
@nilehmann and @enjhnsn2 -- in @Hamid0004 's PR #1704 we added this guard

https://github.com/flux-rs/flux/blob/78fadb29ef16f5870e042bd4118068d4893774a6/crates/flux-refineck/src/checker.rs#L1778-L1779

to only preserve the index if the inner types were unchanged.

However, since the index is only tracking base/addr/size -- can't we carry it along even if the types change?

This has a use case: in smallvec for example, we often cast from the outer container type to the inner element type, e.g.

https://github.com/flux-rs/smallvec/blob/flux-thawed/src/lib.rs#L667-L671 

?

ranjitjhala:
(but strip out the pointee refinements...) 

Fixes #1709

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -1771,14 +1771,18 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
                 }
             }
             CastKind::PtrToPtr => {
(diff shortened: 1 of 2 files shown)


nilehmann (approved):
LGTM

PR #1711, opened 2026-08-09
Implement wildcard literals

### LINKED ISSUE #1583: Add support for wildcard literals?
Its rather silly that flux complains about code like 

```rust
fn blah(arr: &[i32; 10]) -> i32 { 
   let mut tot = 0;
   for i in 0..10 {
      tot += arr[i];
   }
}
``` 

Even with the extern-spec for the iterator. The reason is that we need a qualifier like `i <= 10`, which is specific to the `10` here. We could solve this with the new `wildcard-literals` feature in fixpoint https://github.com/ucsd-progsys/liquid-fixpoint/pull/834 if we expose such qualifiers to flux.

kris-gaudel:
Resolves https://github.com/flux-rs/flux/issues/1583, using the wildcard-literals feature from ucsd-progsys/liquid-fixpoint#834.

Code example:
```rs
#![flux::defs {
    qualifier LeLit(x: int, a #: int) { x <= a }
}]

extern crate flux_core;

pub fn blah(arr: &[i32; 10]) -> i32 {
    let mut tot = 0;
    for i in 0..10 {
        tot += arr[i];
    }
    tot
}
```

## Summary
- A `is_wildcard` flag is propagated through the representation of every qualifier
- The flag rides `RefineParam` through surface and fhir, becomes a `List<bool>` mask alongside the Binder in `rty::Qualifier`
- Parser accepts `#` marker only in qualifier parameter position, the flag is eventually put back into a new `QualParam` struct in liquid-fixpoint

### DIFF
--- a/crates/flux-syntax/src/parser/mod.rs
+++ b/crates/flux-syntax/src/parser/mod.rs
@@ -1385,7 +1385,7 @@ fn parse_refine_param(cx: &mut ParseCtxt, require_sort: RequireSort) -> ParseRes
 }
 
 /// ```text
-/// ⟨qualifier_param⟩ := ⟨mode⟩? ⟨ident⟩ #? : ⟨sort⟩
+/// ⟨qualifier_param⟩ := #? ⟨mode⟩? ⟨ident⟩ : ⟨sort⟩
 /// ```
 fn parse_qualifier_param(cx: &mut ParseCtxt) -> ParseResult<RefineParam> {
     parse_refine_param_inner(cx, RequireSort::Yes, AllowWildcard::Yes)
@@ -1397,14 +1397,11 @@ fn parse_refine_param_inner(
     allow_wildcard: AllowWildcard,
 ) -> ParseResult<RefineParam> {
     let lo = cx.lo();
+    // `#a: int` rather than fixpoint's `a#: int` because rustc lexes the attribute first and
+    // rejects `a#` as a reserved prefix.
+    let is_wildcard = matches!(allow_wildcard, AllowWildcard::Yes) && cx.advance_if(token::Pound);
     let mode = parse_opt_param_mode(cx);
     let ident = parse_ident(cx)?;
-    // `a #: int` rather than fixpoint's `a#: int` because rustc lexes the attribute first and
-    // rejects `a#` as a reserved prefix.
-    let is_wildcard = matches!(allow_wildcard, AllowWildcard::Yes) && cx.peek(token::Pound);
-    if is_wildcard {
-        cx.advance();
-    }
     let sort = parse_sort_if_required(cx, require_sort)?;
     let hi = cx.hi();
     Ok(RefineParam {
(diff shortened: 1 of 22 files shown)


crates/flux-syntax/src/parser/mod.rs
```diff
@@ -1367,18 +1370,51 @@ fn parse_sort_if_required(cx: &mut ParseCtxt, require_sort: RequireSort) -> Pars
...
+) -> ParseResult<RefineParam> {
     let lo = cx.lo();
     let mode = parse_opt_param_mode(cx);
     let ident = parse_ident(cx)?;
+    // `a #: int` rather than fixpoint's `a#: int` because rustc lexes the attribute first and
+    // rejects `a#` as a reserved prefix.
+    let is_wildcard = matches!(allow_wildcard, AllowWildcard::Yes) && cx.peek(token::Pound);
+    if is_wildcard {
+        cx.advance();
+    }
```
nilehmann: I don't like that we are adding a field to `RefineParam` which is used for other stuff. How is `#a` lexed? Can we put the `#` before the identifier and avoid having to make it part of `RefineParam`?

ranjitjhala: @nilehmann do you mean you want to 

1. Just leave all this in the `Ident`,
2. Implicitly that start (or end?) with `#` as wildcard,
3. Ultimately you'd still have to do to the `List<bool>` mask alongside the `Binder<...>` though, to make the appropriate translation to fixpoint?

nilehmann: 1,2. I'm proposing `#ident`, which I think is lexed as two tokens (I would have to confirm), so it's like a mark you put on the identifier.
3. Yes, but I want to represent it like that earlier in the pipeline, so we don't have to add an extra field to `RefineParam`

kris-gaudel: > I don't like that we are adding a field to `RefineParam` which is used for other stuff. How is `#a` lexed? Can we put the `#` before the identifier and avoid having to make it part of `RefineParam`?

```rs
macro_rules! show {
    ($($t:tt)*) => {};
}

// `a #: int` — the current PR syntax. Lexes: Ident(a) Punct(#) Punct(:) Ident(int)
show!(qualifier LeLit(x: int, a #: int) { x <= a });

// `#a: int` — the marker moved in front. Lexes: Punct(#) Ident(a) Punct(:) Ident(int)
show!(qualifier LeLit(x: int, #a: int) { x <= a });

// `a#: int` — fixpoint's spelling
show!(qualifier LeLit(x: int, a#: int) { x <= a });
```

In the above rust program, running `rustc --edition 2024 --crate-type lib broken.rs` will result in  the third `show!` statement failing with the error, but `#a` is lexed as two tokens 😄 

```rs
error: prefix `a` is unknown
  --> broken.rs:23:31
   |
23 | show!(qualifier LeLit(x: int, a#: int) { x <= a });
   |                               ^ unknown prefix
   |
   = note: prefixed identifiers and literals are reserved since Rust 2021
```

So we could follow the `#a` convention, I'm going to try moving the `Vec<bool>` on the qualifier instead

ranjitjhala (approved):
This is awesome, thanks very much!!!

nilehmann (approved):
Looks great, thanks!

PR #1713, opened 2026-08-12
Check macro generated code

ranjitjhala:
Can mark `trusted` using `detached_spec!` if necessary.


book/src/guide/specifications.md
```diff
@@ -417,6 +421,44 @@ Here is an example
 {{#include ../../../tests/tests/neg/surface/ignore02.rs}}
 ```
 
+### Derive-generated code
+
+Flux checks the code generated by a `#[derive(..)]` just like any other code. Sometimes that
+code is something Flux cannot check — most commonly a derive on an [opaque
+struct](#opaque-structs), where the generated `Debug` or `Hash` reads the very fields that
+opacity hides. Marking it `#[flux_rs::trusted]` is not an option, because you cannot attach an
```
ranjitjhala:
```suggestion
opacity hides.
```

book/src/guide/specifications.md
```diff
@@ -417,6 +421,44 @@ Here is an example
...
+attribute to code you did not write.
+
+For these cases, the _type_ carries the annotation instead: `#[flux_rs::trusted_derive]` tells
```
ranjitjhala:
```suggestion
For these cases, `#[flux_rs::trusted_derive]` tells
```

crates/flux-middle/src/global_env.rs
```diff
@@ -770,6 +770,43 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
         annotation || self.matches_trusted_impl_pattern(MaybeExternId::Local(def_id))
     }
 
+    /// Whether code generated by a `#[derive(..)]` on this ADT should be trusted.
+    ///
+    /// A `#[flux::opaque]` type implicitly trusts its derives: the generated code reads the
```
ranjitjhala:
```suggestion
    /// A `#[flux::opaque]` type implicitly trusts its derives: 
```

crates/flux-middle/src/global_env.rs
```diff
@@ -770,6 +770,43 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
...
+    /// internal representation, which is exactly what opacity forbids, so it could never be
```
ranjitjhala:
```suggestion
```

crates/flux-middle/src/global_env.rs
```diff
@@ -770,6 +770,43 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
...
+    /// checked. An explicit `#[flux::trusted_derive(..)]` takes precedence, which lets an opaque
```
ranjitjhala:
```suggestion
    /// An explicit `#[flux::trusted_derive(..)]` takes precedence, which lets an opaque
```

crates/flux-middle/src/queries.rs
```diff
@@ -1329,8 +1329,17 @@ impl<'a> Diagnostic<'a> for QueryErrAt {
                         diag.span_label(cx_span, fluent::_subdiag::label);
                         if let ErrCtxt::FnCheck(_, fn_def_id) = self.cx {
                             let fn_span = tcx.def_span(fn_def_id);
-                            diag.arg("def_kind", tcx.def_descr(fn_def_id.to_def_id()));
-                            diag.span_label(fn_span, fluent::middle_query_opaque_struct_help);
+                            if fn_span.in_derive_expansion() {
+                                // Derive-generated code can't be annotated directly, so point at
```
ranjitjhala:
```suggestion
```

ranjitjhala:
@nilehmann did you want to take a look at this?

crates/flux-middle/src/global_env.rs
```diff
@@ -770,6 +770,42 @@ impl<'genv, 'tcx> GlobalEnv<'genv, 'tcx> {
...
+        // to fall back to "not opaque" here.
+        self.adt_def(def_id)
+            .is_ok_and(|adt_def| adt_def.is_opaque())
+    }
+
+    /// For a `def_id` generated by a `#[derive(..)]`, the ADT the derive was applied to, i.e. the
+    /// self type of the enclosing impl. Returns `None` if the self type is not a local ADT.
+    pub fn derive_self_ty(self, def_id: LocalDefId) -> Option<LocalDefId> {
+        let tcx = self.tcx();
+        let impl_id = tcx.local_parent(def_id);
```
nilehmann, line 801: A `#[derive(...)]` can generate arbitrary code, so there's no guarantee this will be an `impl_id`, nor that the the impl will be on the type you put the derive on 🤔

tests/tests/with_deps/pos/detached/detach_impl05.rs
```diff
@@ -9,6 +9,7 @@ pub mod wrapper {
     use flux_rs::specs;
 
     #[flux_rs::reflect]
+    #[flux_rs::trusted_derive(reason = "the detached spec below is the trusted contract for the derived `eq`")]
```
nilehmann, line 12: I'm curious about why Flux can't prove the derived `impl` against the detached spec?

crates/flux-middle/src/global_env.rs
ranjitjhala: "Genuinely a latent ICE, not just a style issue."

ranjitjhala: we could test if it IS an `impl`?

```rust
 if !matches!(tcx.def_kind(impl_id), DefKind::Impl { .. }) {
            return None;
 }
 ...       
```

tests/tests/with_deps/pos/detached/detach_impl05.rs
ranjitjhala: unclear, how does one actually see the derived code?

ranjitjhala: Apparently it looks like this

```rust
#[automatically_derived]
impl ::core::cmp::PartialEq for Role {
    #[inline]
    fn eq(&self, other: &Role) -> bool {
        let __self_discr = ::core::intrinsics::discriminant_value(self);
        let __arg1_discr = ::core::intrinsics::discriminant_value(other);
        __self_discr == __arg1_discr
    }
}
```

so presumably we'd need some sort of spec for `discriminant_value` ...

crates/flux-middle/src/global_env.rs
ranjitjhala: done.

tests/tests/with_deps/pos/detached/detach_impl05.rs
nilehmann: ah yikes

ranjitjhala:
ok to merge @nilehmann ?

nilehmann (approved):

PR #1723, opened 2026-08-14
Add specs for ControlFlow/? etc.

ranjitjhala:


tests/tests/with_deps/neg/extern_specs/flux_core_ptr_cast00.rs
```diff
@@ -0,0 +1,45 @@
+extern crate flux_core;
+
+// A ptr-to-ptr cast carries the `base`/`addr`/`size` index over to the casted
```
nilehmann, line 3: This file doesn't look related to this PR

nilehmann (approved):
LGTM, but in what situation do you know a Result is an error (or ok) and use `?` on it?

tests/tests/with_deps/neg/extern_specs/flux_core_ptr_cast00.rs
ranjitjhala: hmm yes, wonder how that crept in here... still may as well keep it?

ranjitjhala:
> LGTM, but in what situation do you know a Result is an error (or ok) and use `?` on it?

Indeed -- but with this thicket of break and continue its hard to tease apart what exactly is required.

Are you saying some of the tests are redundant?

ranjitjhala:
(CC added them, but they looked reasonable...)

nilehmann:
I'm wondering what prompted this. Is there some specific code that was failing that requires these specs?

ranjitjhala:
Yes, these:

- https://github.com/flux-rs/tinyvec/blob/flux-thawed/src/arrayvec.rs#L1766-L1773 
- https://github.com/flux-rs/tinyvec/blob/flux-thawed/src/arrayvec.rs#L1806-L1812

ranjitjhala:
Have to say its quite neat how all the bits snap together nicely to make this work. Shocking!

nilehmann:
Ah interesting. Perhaps worth adding something like that as a test

tests/tests/with_deps/pos/extern_specs/control_flow00.rs
```diff
@@ -0,0 +1,146 @@
...
+    let v = x?;
+    Ok(v)
+}
+
+#[spec(fn (n: usize) -> Option<usize[n-1]>[n > 0])]
+fn decr(n: usize) -> Option<usize> {
+    if n > 0 { Some(n - 1) } else { None }
+}
+
+#[spec(fn (n: usize) -> Option<usize[n-2]>[n > 1])]
```
ranjitjhala, line 141: Add test for dependency between Option and ?

PR #1728, opened 2026-08-17
Implement auto strong returned references

### LINKED ISSUE #1714: Auto strong returned references
When Flux checks a call whose _returned result_ is a &mut T, the returned reference lands in the caller's environment as a plain invariant Ref(re, T, Mut). This means if you have something like

```rust
#[spec(fn (x:&mut usize{v: v >= 10}) -> &mut usize{v: v >= 10})]
fn returns_a_mut(x: &mut usize) -> &mut usize {
  x 
}
```

and then you write a client

```rust
fn client(){
   let mut blah = 10;
   let tmp = returns_a_mut(&mut blah);
   let a = *tmp;
   let b = *tmp;
   assert(a == b) // fails! a, and b are unrelated -- they are both {v: v >= 10}
}
```

The assert fails ... the current "fix" is to rewrite the `client` as

```rust
fn helper(tmp: &mut usize) { 
   let a = *tmp;
   let b = *tmp;
   assert(a == b) // succeeds! a, and b are unrelated -- they are both {v: v >= 10}
}
 
fn client(){
   let mut blah = 10;
   let tmp = returns_a_mut(&mut blah);
   helper(tmp)
}
```

Because now the "auto-strong" heuristic kicks in, makes `tmp` a "strong reference", which lets FLUX understand that `a` and `b` are the same.

So the *goal* of this task is to enable this sort of "auto-strong" conversion, but without requiring the refactoring.

kris-gaudel:
Closes https://github.com/flux-rs/flux/issues/1714
```rs
#[flux::sig(fn(x: bool[true]))]
fn assert(_x: bool) {}

#[flux::sig(fn(x: &mut usize{v: v >= 10}) -> &mut usize{v: v >= 10})]
fn returns_a_mut(x: &mut usize) -> &mut usize {
    x
}

fn reads_agree() {
    let mut blah = 10;
    let tmp = returns_a_mut(&mut blah);
    let a = *tmp;
    let b = *tmp;
    assert(a == b) // now succeeds: before, `a` and `b` were unrelated values,
                   // each only known to satisfy `{v: v >= 10}`
}

fn read_after_write() {
    let mut blah = 10;
    let tmp = returns_a_mut(&mut blah);
    *tmp = 20; // a strong update, not a weak one
    assert(*tmp == 20)
}

#[flux::sig(fn(x: &mut usize{v: v >= 10}) -> (&mut usize{v: v >= 10}, usize[0]))]
fn tag(x: &mut usize) -> (&mut usize, usize) {
    (x, 0)
}

fn through_a_tuple() {
    let mut blah = 10;
    let (r, _n) = tag(&mut blah);
    *r = 20;
    assert(*r == 20) // references nested in a returned tuple or struct too
}
```

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -473,6 +473,20 @@ fn fold_local_ptrs(infcx: &mut InferCtxt, env: &mut TypeEnv, span: Span) -> Infe
     env.fold_local_ptrs(&mut at)
 }
 
+/// Auto-strong a `&mut T` returned by a call: instead of the invariant `&mut T`, which loses
+/// track of the value behind the reference, we hand back a pointer to a fresh local location
+/// holding `T`. Reads through the returned reference are then precise and writes are strong
+/// updates, which get checked against `T` when the pointer is folded back into a `&mut T` by
+/// [`fold_local_ptrs`] at the end of the block.
+fn unfold_returned_ref(infcx: &mut InferCtxt, env: &mut TypeEnv, ret: &Ty) -> InferResult<Ty> {
+    if let TyKind::Indexed(BaseTy::Ref(re, bound, Mutability::Mut), _) = ret.kind() {
+        let loc = env.unfold_local_ptr(infcx, bound)?;
+        Ok(Ty::ptr(PtrKind::Mut(*re), Path::from(loc)))
+    } else {
+        Ok(ret.clone())
+    }
+}
+
 fn promoted_fn_sig(ty: &Ty) -> PolyFnSig {
     let safety = rustc_hir::Safety::Safe;
     let abi = rustc_abi::ExternAbi::Rust;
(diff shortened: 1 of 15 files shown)


nilehmann:
This PR introduces (new) unsoundness. The following is correctly rejected in `main` but accepted by this PR

```rust
#[flux::sig(fn(bool[true]))]
fn assert(_: bool) {}

#[flux::sig(fn(&mut usize{v: v > 0}) -> &mut usize{v: v > 0})]
fn id(x: &mut usize) -> &mut usize {
    x
}

fn bad() {
    let mut w = 1;
    let r = id(&mut w);
    let _a = [&mut *r];
    *r = 0;           
    assert(w > 0);   
}
```

ranjitjhala:
(I haven't seen the code in the PR yet) but is the unsoundness because the effect of the `*r = 0` is not "back-propagated" to the `w` after the `r` expires? Is there a way to only do this auto-strong IF the returned `r` "expires" at the END of the caller? (Also, @nilehmann what is the  `let _a = [&mut *r];` doing??)

nilehmann:
I haven't reviewed the implementation, so I can't determine whether the issue is in the implementation or an existing unsoundness that's surfacing because there are more unfolds.

The `let _a = [&mut *r]` is triggering a `ptr_to_ref` on the unfolded location.

PR #1729, opened 2026-08-17
Add core specs for `alloc:Allocate`

kris-gaudel:
Addresses part of https://github.com/flux-rs/flux/issues/1546

Adds specifications for:
- `Allocator::allocate`
- `Allocator::deallocate`
- `Allocator::grow`
- `Allocator::shrink`


lib/flux-core/src/alloc/allocator.rs
```diff
@@ -0,0 +1,111 @@
...
+    // provenance (deallocating a stack local verifies), liveness (double free, use after
+    // free), and identity (the predicate never mentions `Self`). Closing those needs the
+    // block owned as a linear resource keyed by the allocator, which a `NonNull` index
+    // cannot carry, so those clauses are assumed rather than checked.
+    //
+    // Alignment is approximate for the same reason: the pointer does not carry the alignment
+    // it was allocated with, so all we check is that its address is a multiple of
+    // `layout.align()`. Deallocating an align-16 block with an align-8 layout is UB and
+    // verifies. Note the asymmetry — collapsing `size` only ever rejects legal code, this
+    // accepts some illegal code.
```
sweiglbosker: Can you clean up this comment?

kris-gaudel:
CC @enjhnsn2 @ranjitjhala

lib/flux-core/src/ptr/non_null.rs
```diff
@@ -135,6 +135,11 @@ impl<T> NonNull<T> {
 impl<T> NonNull<[T]> {
     /// Core impl: https://github.com/rust-lang/rust/blob/c871d09d1cc32a649f4c5177bb819646260ed120/library/core/src/ptr/non_null.rs#L1420
     /// Safety: https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety
+    /// The indices carry through unchanged: `size` is the extent of the *allocation*, not of
+    /// the slice, and `len` only has to fit inside it. Retracking as `len * size_of::<T>()`
+    /// would be unsound, since `layout_fits` reads `size` as the size a block was allocated
+    /// with — see `test_allocate_then_deallocate_narrowed_slice` in
+    /// `tests/with_deps/neg/extern_specs/flux_core_alloc01.rs`.
```
enjhnsn2, line 142: I don't really understand this comment.  "The indices carry through unchanged" -- from what?

lib/flux-core/src/alloc/allocator.rs
```diff
@@ -0,0 +1,95 @@
+#![cfg_attr(flux, flux::defs {
+    // A layout *fits* a block when the block is aligned to `layout.align()` and its size
```
enjhnsn2, line 2: does a block end need to be aligned to `layout.align`, or just the start?

lib/flux-core/src/alloc/allocator.rs
```diff
@@ -0,0 +1,95 @@
...
+    // `NonNull` carries one size index, so we collapse that interval and track a block as
+    // exactly the `layout.size()` it was allocated with; keeping only the upper bound would
+    // admit deallocating with a smaller layout than was allocated, which is UB.
+    // See: https://doc.rust-lang.org/std/alloc/trait.Allocator.html#memory-fitting
+    //
+    // Weaker than the "currently allocated via this allocator" the methods below require:
+    // provenance, liveness and allocator identity are not checked, and the alignment checked
+    // is the address's, not the one the block was allocated with. Those clauses are assumed.
+    fn layout_fits(p: NonNull, l: Layout) -> bool {
+        p.base == p.addr && l.size == p.size && p.addr % l.align == 0
```
enjhnsn2, line 13: why does p.base need to match p.addr?

lib/flux-core/src/alloc/allocator.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[spec(fn(&Self, layout: Layout[@l])
+        -> Result<NonNull<[u8]>{p: layout_fits(p, l)}, AllocError>)]
+    fn allocate_zeroed(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError>;
+
+    /// Core impl: https://github.com/rust-lang/rust/blob/dab8d9d1066c4c95008163c7babf275106ce3f32/library/core/src/alloc/mod.rs#L166
+    /// - `ptr` must denote a block of memory currently allocated via this allocator
+    ///   (assumed, not checked), and
+    /// - `layout` must fit that block of memory.
+    ///
+    /// Deallocating does not invalidate `ptr`, so a later use of it is not rejected.
```
enjhnsn2, line 46: "Deallocating does not invalidate pointer" -- do you mean that the rust compiler does not somehow dynamically invalidate the pointer, or do you mean that the pointer is indeed valid to use after deallocating?

enjhnsn2 (changes requested):
I found some of the comments somewhat confusing. I think the easy way to resolve this is provide a convincing reason as to why `layout_fits(...)` is correct.

lib/flux-core/src/alloc/allocator.rs
kris-gaudel: The pointer is dangling after deallocate, the compiler treats this as undefined behaviour (since deallocate spec has no post condition, we won't reject later use of the pointer)

lib/flux-core/src/alloc/allocator.rs
kris-gaudel: deallocate/grow/shrink would be unsound bc interior pointers would be accepted:

```rs
#[flux::spec(fn(&A, ptr: NonNull<u8>[@base, @addr, @size], layout: Layout[@lsize, @lalign])
    requires addr > base && lsize == size && addr % lalign == 0)]
pub unsafe fn test_deallocate_interior<A: Allocator>(a: &A, ptr: NonNull<u8>, layout: Layout) {
    a.deallocate(ptr, layout);
}
```
Without `p.base == p.addr`, `layout_fits holds`:
- `l.size == p.size` (`8 == 8`)
- `p.addr % l.align == 0` because A is a multiple of 8, so A + 8 is too

The behaviour will be undefined since allocator gets A + 8, which it never returned

lib/flux-core/src/alloc/allocator.rs
kris-gaudel: Only the start

lib/flux-core/src/ptr/non_null.rs
kris-gaudel: Indices here refers to the refinement indices of `NonNull` and `[base, addr, size]`. The output slice pointer will have the same indices as the `data` pointer passed in. I'll change the comment to be clearer

enjhnsn2 (approved):
I can't 100% wrap my head around it, but it seems right.
I suspect that we will find some subtle issues with the specs once we start using them, but I think we should merge.

kris-gaudel:
Thanks! I'll write some in-line comments on this PR that should hopefully clear things up!

PR #1730, opened 2026-08-18
Specs for min/max 

ranjitjhala:
Refactored using `common` definitions that are reused in iterators.

### DIFF
--- a/lib/flux-core/src/common.rs
+++ b/lib/flux-core/src/common.rs
@@ -1,8 +1,0 @@
-#![flux::defs {
-    fn min(a: int, b: int) -> int { if a < b { a } else { b } }
-    fn max(a: int, b: int) -> int { if a > b { a } else { b } }
-}]
-//! Small spec functions shared across `flux-core`.
-//!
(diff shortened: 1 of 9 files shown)


ranjitjhala:
@nilehmann or @enjhnsn2 did you want to take a look -- seems pretty straightforward ...

lib/flux-core/src/common.rs
```diff
@@ -0,0 +1,8 @@
+#![flux::defs {
+    fn min(a: int, b: int) -> int { if a < b { a } else { b } }
+    fn max(a: int, b: int) -> int { if a > b { a } else { b } }
```
nilehmann: There's the `num` module and we already have some num like definitions there. Maybe we can put these over there.

https://github.com/flux-rs/flux/blob/283ad737569f80951973747ef16f1a89d64d9728/lib/flux-core/src/num/mod.rs#L2-L16

nilehmann (approved):

lib/flux-core/src/common.rs
ranjitjhala: done!

PR #1731, opened 2026-08-18
Check macro generated definitions

ranjitjhala:
Removes the "hack" that explicitly skipped that check; adds a `trusted_derive` mechanism to skip the bodies if needed.


ranjitjhala:
@nilehmann I'm confused -- I thought I'd *already* pushed and merged this branch but apparently not?

nilehmann (approved):

PR #1732, opened 2026-08-18
misc specs for index/mut

ranjitjhala:


lib/flux-core/src/ops/index.rs
```diff
@@ -8,3 +8,13 @@ trait Index<Idx> {
...
+
+/// Without a trait-level spec, `index_mut`'s signature is the unrefined lifted one, so any impl
+/// carrying an `in_bounds` precondition fails the impl-vs-trait subtyping check. `IndexMut: Index`,
+/// so the two associated refinements are inherited rather than redeclared — the same way
+/// `flux-core`'s own `[T]` and `[T; N]` `IndexMut` impls are written.
+#[extern_spec(core::ops)]
+trait IndexMut<Idx>: Index<Idx> {
+    #[sig(fn(self: &mut Self[@v], index: Idx { <Self as Index<Idx>>::in_bounds(v, index) }) -> &mut Self::Output{out: <Self as Index<Idx>>::output_pred(v, index, out)})]
+    fn index_mut(&mut self, index: Idx) -> &mut Self::Output;
+}
```
nilehmann, line 23: This makes invalid all existing implementations of `IndexMut` that also have an implementation of `Index` with a non-trivial `in_bounds` and `output_pred`.

ranjitjhala: why? is this spec somehow inconsistent with those?

ranjitjhala: Specifically I thought this got added because we don't have a trait definition?

ranjitjhala: Oh you mean those _other_ implementations will have to be "backported" to use the "correct" in_bounds etc?

nilehmann: If an impl has an `Index` impl with a non-trivial `in_bounds`, in the trait you will get

`fn(self: &mut Self[@v], index: Idx { <Self as Index<Idx>>::in_bounds(v, index) }) -> &mut Self::Output{out: <Self as Index<Idx>>::output_pred(v, index, out)}`

But in the impl, you get the default signature for `index_mut`

`fn(self: &mut Self, index: Idx) -> &mut Self::Output`

So trait-impl subtyping doesn't hold

nilehmann: There are two things at play here

1. For extern specs, we never check impl-trait subtyping. So things like this silently introduce inconsistencies
2. When generating the default signature for a method in an impl, we never look at the trait

ranjitjhala: Also: which existing ones? There don't seem to be any in the CI?

nilehmann: the ones that have extern specs in `flux-core`

ranjitjhala:
Still puzzled.

Is the upshot that if we want this trait spec I should also go and suitably
copy/paste update the spec into existing IndexMut impl in flux-core?

- Ranjit.

lib/flux-core/src/ops/index.rs
nilehmann: @ranjitjhala Ah, sorry. I thought we didn't have extern specs for `IndexMut` impls, but we already have them. There was just the spec on the trait missing. So this should be fine. Still, if there's any `impl Index` without a corresponding `impl IndexMut` this will cause an inconsistency and we don't have any automatic (deterministic) way to know that.

ranjitjhala:
Ah ok that makes sense— maybe I’ll add a comment next to the Index trait
that any impl should ALSO impl IndexMut …


- Ranjit.

nilehmann (approved):

lib/flux-core/src/ops/index.rs
```diff
@@ -8,3 +8,13 @@ trait Index<Idx> {
     #[sig(fn(self: &Self[@v], index: Idx { <Self as Index<Idx>>::in_bounds(v, index) }) -> &Self::Output{out: <Self as Index<Idx>>::output_pred(v, index, out)})]
     fn index(&self, index: Idx) -> &Self::Output;
 }
+
+/// Without a trait-level spec, `index_mut`'s signature is the unrefined lifted one, so any impl
+/// carrying an `in_bounds` precondition fails the impl-vs-trait subtyping check. `IndexMut: Index`,
+/// so the two associated refinements are inherited rather than redeclared — the same way
+/// `flux-core`'s own `[T]` and `[T; N]` `IndexMut` impls are written.
```
nilehmann: I don't like this comment. I'm not sure how to explain it, but claude tends to comment the context at the point it implemented something rather than the thing it implemented.

ranjitjhala:
Uff my bad I thought I’d nuked all the comments (I find them too verbose!)
missed that one!


- Ranjit.

lib/flux-core/src/ops/index.rs
ranjitjhala: nuked it.

PR #1733, opened 2026-08-21
Add support for Associated Constants in Refinements

### LINKED ISSUE #1722: Associated Constants in Refinements
It would be nice to support associated constants in refinements, e.g. 

```rust
use flux_rs::attrs::*;

struct Person;

impl Person {
    const ID: i32 = 10;
}

#[spec(fn() -> i32[10])]
pub fn test0() -> i32 {
    let id = Person::ID;
    id
}

trait TraitWithConst {
    const TRAIT_ID: u32;

    #[spec(fn() -> u32{v: v < Self::TRAIT_ID})]
    fn silly_method() -> u32;
}

impl TraitWithConst for Person {
    const TRAIT_ID: u32 = 5;

    #[spec(fn() -> u32{v: v < Person::TRAIT_ID})]
    fn silly_method() -> u32 {
        3
    }
}

struct Straw;

impl TraitWithConst for Straw {
    const TRAIT_ID: u32 = 5;

    #[spec(fn() -> u32[15])]
    fn silly_method() -> u32 {
        15
    } // ~ ERROR: trait-subtyping error
}





#[spec(fn() -> u32[5])]
pub fn test1() -> u32 {
    let x = Person::TRAIT_ID;
    x
}
```

**EDITED** to account for @nilehmann remark below.

ranjitjhala:
In which CC fixes #1722


ranjitjhala:
And here are CC's notes summarizing the diff that I'm now going over

```
---

1. Surface — parsing and collecting the annotation

flux-syntax/src/surface.rs — The structural change everything else follows from. ImplItemFn { attrs, sig, node_id } becomes ImplItem { attrs, kind, node_id } with

pub enum ImplItemKind { Fn(Option<FnSig>), Const(ConstantInfo) }

The surface AST previously could not represent an annotated constant in an impl at all — only Fn. ConstantInfo is the existing surface type already used by free const items, so #[flux::constant(..)] parses identically in both positions.

flux-syntax/src/surface/visit.rs — walk_impl_item dispatches on the new kind; the Const arm calls vis.visit_constant(cst). This line is load-bearing beyond bookkeeping: the name resolver drives off this visitor, so without it the T::size_of() inside the annotation would never get resolved.

flux-driver/src/collector/mod.rs — The actual bug fix. collect_impl_item gated everything on ImplItemKind::Fn, which is why the annotation was silently ignored rather than rejected. New else if branch: on a const item with attrs.constant(), insert an ImplItemKind::Const.

collector/{detached_specs,extern_specs}.rs — Mechanical: three constructor sites rewritten to ImplItemKind::Fn(sig).

---

2. FHIR — desugaring

flux-middle/src/fhir.rs — ImplItemKind::Const → Const(Option<Expr>). None means "an ordinary constant"; Some carries the annotation body.

flux-middle/src/fhir/visit.rs — Visits the expression when present.

flux-desugar/src/desugar/lift.rs — One character of meaning: lifting from HIR (no flux annotation) produces Const(None).

flux-desugar/src/desugar.rs — desugar_impl_item splits. The Const arm desugars the annotation expression and calls lift_generics() — the impl's generics, which is what lets the body mention T.

flux-desugar/src/resolver.rs, resolver/refinement_resolver.rs, flux-middle/src/lib.rs — Type renames only, no behavior change.

---

3. rty — the representation, and conv into it

Two independent representation changes here; they're the heart of the branch.

flux-middle/src/rty/mod.rs — ConstantInfo goes from a two-variant enum to a struct:

pub struct ConstantInfo { sort: Option<Sort>, value: Option<EarlyBinder<Expr>> }

Two reasons. First, the old enum could not express "I know its sort but not its value", which is precisely a constant declared in a trait. Second, EarlyBinder — the annotation body mentions the impl's generics, so the value has to be bound over them and instantiated later. Fields are private behind unsupported/opaque/interpreted constructors to keep the invariant that a value implies a sort.

flux-middle/src/rty/expr.rs — ConstDefId(DefId) → ConstDefId(DefId, GenericArgs). This is the soundness fix. Without args, <A as Tr>::C and <B as Tr>::C collapse to the same symbol and flux proves false things — neg/surface/assoc_const02.rs is exactly that. For a trait constant the args are the trait ref's, matching what rustc puts in UnevaluatedConst; empty for free and inherent constants. Pretty-printing appends <args>.

flux-middle/src/rty/fold.rs — ConstDefId stops being a fold/visit leaf and folds its args. This is what rewrites the trait's [Self] into the impl's [Thingy<T>] during trait-to-impl instantiation; without it the two sides would be different symbols and every impl-vs-trait check would fail.

flux-middle/src/rty/binder.rs — EarlyBinder gains Eq/PartialEq/Hash, needed to put it inside ConstantInfo.

flux-middle/src/queries.rs — The extern/default provider. Previously gave up with Uninterpreted for anything non-integral; now falls back to sort_of_def_id so constants from other crates aren't second-class.

flux-fhir-analysis/src/conv/mod.rs — Two changes. conv_type_relative_const_path now takes the fhir::Ty qself (so res is available) and, when the primitive incoherent_impls probe finds nothing, delegates to conv_type_relative_path(AssocTag::Const, …) — the same bound-probing associated types and refinements already used. And conv_const splits: it still demands a value (preserving missing_constant_annot.rs), while the new conv_opaque_const needs only a sort and takes the args.

flux-fhir-analysis/src/lib.rs — constant_info gains an arm for annotated impl constants; trait constants return opaque(sort) instead of erroring; scalar evaluation extends from integers to bool/char and uses the computed sort rather than a hardcoded Sort::Int. Separately, generics_of learns DefKind::Const/AssocConst — without it you get an internal flux error, because a constant never previously had an expression that could mention a generic.

flux-fhir-analysis/src/wf/mod.rs — Pattern update.

---

4. Refineck, normalization, fixpoint

flux-refineck/src/checker.rs — The MIR side. check_uneval_const no longer gives up when const_eval_resolve fails; if tcx.trait_of_assoc(uneval.def).is_some(), the new check_trait_assoc_const builds the same symbol conv builds. It deliberately mirrors ty::TraitRef::refine: same refiner, and refine_generic_args(tcx.parent(uneval.def), …). That correspondence is why a spec saying Self::C and a body reading Self::C agree.

flux-infer/src/projections.rs — New normalize_assoc_const, hooked into both try_fold_expr and structurally_normalize_expr beside the existing AliasReft case. Selects the impl, then takes the value of the constant defined there, instantiated at the impl's args — note the translation: the key carries trait-ref args, the annotation is written against impl generics. Falls back to const_eval_resolve, which resolves through the impl itself and so also covers a constant inherited from a trait default. Impl lookup deliberately comes first, so an annotation wins over an evaluated value.

flux-infer/src/fixpoint_encoding.rs — New ConstKey::AssocConst(DefId, GenericArgsRef) and define_const_for_assoc_const, keyed on to_rustc args with regions erased, mirroring define_const_for_alias_reft. Encoded as a plain constant of the constant's own sort, not a 0-ary application — an associated constant takes no refinement arguments. assume_const_values explicitly does nothing for AssocConst: values arrive via normalization, not as assumed equations.

flux-infer/src/fixpoint_encoding/decoding.rs — Both reverse-mapping sites. An AssocConst never heads an application, so it's grouped with RustConst.
---
```

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1589,7 +1593,16 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
         }
         let (expr, sort) = match &candidates[..] {
             [candidate] => self.conv_const(fhir_expr.span, candidate.def_id)?,
-            [] => self.report_assoc_item_not_found(fhir_expr.span, AssocTag::Const)?,
+            // Otherwise, resolve it as a constant declared in a trait, e.g., `Self::C` or
```
ranjitjhala, line 1596: important

crates/flux-middle/src/rty/mod.rs
```diff
@@ -1284,12 +1284,41 @@ pub enum SortArg {
...
-    /// A non-integral constant whose value is specified by the user
-    Interpreted(Expr, Sort),
+pub struct ConstantInfo {
+    sort: Option<Sort>,
+    value: Option<EarlyBinder<Expr>>,
+}
+
+impl ConstantInfo {
+    /// A constant whose type has no sort, and so cannot appear in a refinement.
+    pub fn unsupported() -> Self {
```
ranjitjhala: Why is this needed?

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -215,27 +216,46 @@ fn constant_info(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::Con
...
                 .into_conv_ctxt()
                 .conv_constant_expr(expr)?;
-            Ok(rty::ConstantInfo::Interpreted(expr, sort))
+            Ok(Some(rty::ConstantInfo { sort, value: Some(rty::EarlyBinder(expr)) }))
+        }
+        fhir::Node::ImplItem(fhir::ImplItem {
+            kind: fhir::ImplItemKind::Const(Some(expr)),
+            ..
+        }) => {
+            // As above, but the value may mention the generics of the impl, so we bind it under
```
ranjitjhala: tricky

crates/flux-fhir-analysis/src/lib.rs
```diff
@@ -215,27 +216,46 @@ fn constant_info(genv: GlobalEnv, def_id: MaybeExternId) -> QueryResult<rty::Con
...
+            // them and let normalization instantiate it at the impl's arguments.
+            let owner = def_id.map(|def_id| rustc_hir::OwnerId { def_id });
```
ranjitjhala, line 227: tricky

crates/flux-infer/src/fixpoint_encoding.rs
```diff
@@ -2276,6 +2282,45 @@ impl<'genv, 'tcx> ExprEncodingCtxt<'genv, 'tcx> {
             .name
     }
 
+    /// Returns the constant used to represent one instantiation of an associated constant,
+    /// creating and adding it to the const_map if necessary. Unlike an alias refinement this is
+    /// not a function: an associated constant takes no refinement arguments, so it encodes as a
+    /// plain constant of the constant's own sort.
+    fn define_const_for_assoc_const(
```
ranjitjhala, line 2288: check

crates/flux-infer/src/projections.rs
```diff
@@ -768,12 +779,75 @@ pub fn structurally_normalize_expr<'tcx>(
...
+        }
+        _ => Ok(expr.clone()),
+    }
+}
+
+/// Normalizes a constant declared in a trait, mirroring [`normalize_alias_reft`]: we use the
+/// trait solver to find the impl and take the value of the constant defined there, if we know
+/// one. Falls back to const evaluation, which resolves through the impl itself and so also
+/// covers a constant the impl inherits from the trait's default.
+fn normalize_assoc_const<'tcx>(
```
ranjitjhala, line 799: hairy!

ranjitjhala (commented):
@nilehmann this is all CC with minor tweaks here and there; I went through all the diffs quite closely and they look reasonable, but have especially marked the bits I'm uncertain about (with a comment). Do you want to take a look? Thanks!

nilehmann:
@ranjitjhala  I'll take a look this afternoon!

crates/flux-middle/src/rty/binder.rs
```diff
@@ -19,7 +19,7 @@ use super::{
     subst::{self, BoundVarReplacer, FnMutDelegate},
 };
 
-#[derive(Clone, Debug, TyEncodable, TyDecodable)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, TyEncodable, TyDecodable)]
```
nilehmann: These derives are unnecessary. `ConstantInfo` has a `#[derive(Eq, PartialEq, Hash)]`, that's why claude added them here, but we can remove them from `ConstantInfo` as well.

crates/flux-infer/src/projections.rs
nilehmann: Look a bit into it. This is wrong in various subtle ways. I'm prompting my local claude to create a couple of examples that trigger wrong/weird behavior. Will come back!

ranjitjhala: are you referring to this specific function `fn normalize_assoc_const` or the PR in general? either way, lmk!

nilehmann: I found some issues in `Checker`, but they're mostly cosmetic. This function is more problematic. Here are some examples that misbehave

<details>
<summary>assoc_const_generics</summary>

```rust
// `<Arr<N> as Len>::LEN` is the const generic `N`. Const evaluation reports `TooGeneric`,
// so normalization falls back to the opaque symbol, while the body's `N` is refined to
// `Expr::const_generic`. The postcondition of `len` then relates two different symbols
// standing for the same number.
//
// `test` verifies today: at `Arr<3>` the arguments are concrete and const evaluation
// closes the gap. It is here to isolate genericity as the cause.

use flux_attrs::*;

pub trait Len {
    const LEN: usize;

    #[spec(fn() -> usize[Self::LEN])]
    fn len() -> usize;
}

pub struct Arr<const N: usize>;

impl<const N: usize> Len for Arr<N> {
    const LEN: usize = N;

    #[spec(fn() -> usize[Self::LEN])]
    fn len() -> usize {
        N // ERROR
    }
}

#[spec(fn() -> usize[3])]
pub fn test() -> usize {
    <Arr<3> as Len>::len() // OK
}
```

</details>

<details>
<summary>assoc_const_nested</summary>

```rust
// `<Wrap<T> as Tr>::C` is defined to be `<T as Tr>::C`, which is exactly the symbol
// `conv` builds for `T::C` written in a refinement -- no new machinery is needed to
// express it. But the only path from an impl item to a refinement runs through rustc's
// const evaluator, which reports `TooGeneric` here, so the two sides of `get`'s
// postcondition remain unrelated opaque symbols.
//
// Unlike the reflected-enum and const-generic cases, this one cannot be closed by
// widening the set of values const evaluation accepts: it needs a path that converts
// an impl constant's body to a refinement expression directly.
//
// `test` verifies today: at `Wrap<Base>` the arguments are concrete and const
// evaluation closes the gap. It is here to isolate genericity as the cause.

use flux_attrs::*;

pub trait Tr {
    const C: usize;

    #[spec(fn() -> usize[Self::C])]
    fn get() -> usize;
}

pub struct Base;

impl Tr for Base {
    const C: usize = 5;

    #[spec(fn() -> usize[Self::C])]
    fn get() -> usize {
        5
    }
}

pub struct Wrap<T>(T);

impl<T: Tr> Tr for Wrap<T> {
    const C: usize = T::C;

    #[spec(fn() -> usize[Self::C])]
    fn get() -> usize {
        T::get() // ERROR
    }
}

#[spec(fn() -> usize[5])]
pub fn test() -> usize {
    <Wrap<Base> as Tr>::get() // OK
}
```

</details>

<details>
<summary>assoc_const_reflected</summary>

```rust
// `Simple::INIT` is a literal variant of a reflected enum and `Self` is fully concrete,
// but normalization cannot recover its value: `constant_info` only evaluates constants
// of integer, bool and char type, and `normalize_assoc_const` re-applies the same guard
// before trying const evaluation. So `Self::INIT` stays an opaque symbol and neither
// function below verifies. Replacing `State` with `bool` makes both pass.

use flux_attrs::*;

#[reflect]
pub enum State {
    On,
    Off,
}

pub trait Machine {
    const INIT: State;

    #[spec(fn() -> State[Self::INIT])]
    fn start() -> State;
}

pub struct Simple;

impl Machine for Simple {
    const INIT: State = State::On;

    #[spec(fn() -> State[Self::INIT])]
    fn start() -> State {
        State::On // ERROR
    }
}

#[spec(fn() -> State[State::On])]
pub fn test() -> State {
    <Simple as Machine>::start() // ERROR
}
```

</details>

<details>
<summary>const_trait_default</summary>

```rust
// A `#[constant]` annotation on a trait constant's *default* is silently ignored.
//
// `Inherit<T>` does not override `C`, so `impl_item_implementor_ids` has no entry for it
// and step 1 of `normalize_assoc_const` never consults `constant_info` on anything that
// carries the annotation. Step 2 then resolves the instance to the trait's default, whose
// body reads `size_of::<Self>()` with `T` generic, so const evaluation reports
// `TooGeneric` and `Self::C` stays opaque.
//
// The annotation says `Self::size_of() == 0`, and the impl defines `size_of` to be `1`,
// so `Self::C` is `false` and returning `100` should satisfy `10 < v`. It does not.
//
// The root cause is upstream of normalization: `fhir::TraitItemKind::Const` carries no
// expression, so the `constant_info` provider hardcodes `value: None` for every trait
// constant. That is right for a bare declaration but wrong for an annotated default.
//
// Moving the same annotation onto an impl item works today -- see step 4 of
// `tests/tests/pos/surface/assoc_const01.rs`.

use flux_attrs::*;

#[assoc(fn size_of() -> int)]
trait Tr: Sized {
    #[constant(Self::size_of() == 0)]
    const C: bool = size_of::<Self>() == 0;

    #[spec(fn() -> u32{v: if Self::C { v == 0 } else { 10 < v }})]
    fn m() -> u32;
}

struct Inherit<T>(T);

#[assoc(fn size_of() -> int { 1 })]
impl<T> Tr for Inherit<T> {
    #[spec(fn() -> u32{v: if Self::C { v == 0 } else { 10 < v }})]
    fn m() -> u32 {
        100
    }
}
```

</details>

PR #1735, opened 2026-08-27
Add FLUXLOCALFLAGS for targeted packages

### LINKED ISSUE #1608: Add Local Flags
The current "-Finclude=<pattern>" cannot be used with the `FLUXFLAGS` because that pattern is passed to ALL the crates, including the extern crates like `flux-core`, which means we get errors like the below whenever any code uses those operations (i.e. everywhere!)

```
error[E0999]: use of struct `std::ops::Range` that was not included when checking external crate
  --> crypto-playground/src/utils.rs:9:9
   |
 9 |     dst[0..n].copy_from_slice(&src[0..n]);
   |         ^^^^
   |
help: when checking external crate, include the file or module where the excluded item is defined
  --> /Users/rjhala/.rustup/toolchains/nightly-2025-11-25-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/range.rs:82:12
   |
82 | pub struct Range<Idx> {
   |            ^^^^^

error[E0999]: use of trait `std::clone::Clone` that was not included when checking external crate
   --> crypto-playground/src/utils.rs:13:1
    |
 13 | pub fn ms_memset<T: std::clone::Clone>(dst: &mut [T], c: T, n: usize) {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

Following the discussion here [#dev > Strange "not included" errors](https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/Strange.20.22not.20included.22.20errors/with/595527889) we should have a `FLUXLOCALFLAGS` which are ONLY passed to all targets in the current invocation, so if we have multiple "local" crates or packages use the `FLUXLOCALFLAGS` ONLY when checking those local packages.

kimjune01:
Fixes #1608.

`FLUXFLAGS` applies to every Flux-enabled package, including dependencies. This adds `FLUXLOCALFLAGS`, which applies only to package IDs selected by the current `cargo flux` invocation.

This builds on the targeting logic introduced by #1650 rather than duplicating Cargo package selection as #1609 did. It also complements `--only-check`: local flags can carry any Flux option and do not override configured include patterns.

- Preserve existing `FLUXFLAGS` behavior.
- Remove `FLUXLOCALFLAGS` before spawning Cargo so dependencies cannot inherit it as ambient state.
- Cover single and multiple package selection, workspace exclusions, whitespace parsing, and subprocess environment containment.
- Document the new variable alongside `FLUXFLAGS`.

Testing: `cargo xtask test`


kimjune01:
I should have validated the original report against current `main` first; #1650 already functionally addresses the reported use case, so I’m closing this as unnecessary.

PR #1738, opened 2026-08-28
Render extern specs in rustdoc

### LINKED ISSUE #1669: Add rust docs for extern specs
We should add rust-docs for `flux-core` and `flux-alloc` to the website, to make the extern spec library (more) easily navigable / searchable.

kimjune01:
Makes the `flux-core` and `flux-alloc` extern-spec libraries navigable and searchable in the generated rustdoc.

## Design

Rustdoc cannot attach documentation to items owned by another crate. Like [Creusot’s resolution of the same limitation](https://github.com/creusot-rs/creusot/issues/1104), this emits separate, searchable documentation-only items. The approach is based on [Creusot PR #1851](https://github.com/creusot-rs/creusot/pull/1851).

Flux cannot emit Creusot-style surrogate functions with their original signatures: Flux refinement syntax and bodyless inherent impls do not always type-check as ordinary Rust. Instead:

- each extern item gets a marker page containing its complete original declaration and contracts;
- impl methods become nested documentation modules, keeping method names searchable without displaying fabricated signatures;
- stable, path-sensitive 64-bit suffixes prevent collisions between specs for the same type;
- every page explicitly says that its marker is not a real Rust item.

Following the concern raised in [Creusot PR #1963](https://github.com/creusot-rs/creusot/pull/1963), broken intra-doc links are denied in `flux-core` and `flux-alloc` so generated links cannot silently regress.

Spec-only modules are included under `cfg(doc)`, while module-level Flux definitions remain restricted to `cfg(flux)`.

## Validation

Tests cover free-function and impl expansion, path-sensitive naming, target summaries, and hash stability. `cargo xtask doc`, rustfmt, and the macro tests pass.

Fixes #1669.

### DIFF
--- a/lib/flux-core/src/slice/mod.rs
+++ b/lib/flux-core/src/slice/mod.rs
@@ -2,7 +2,7 @@
 //!
 //! # Soundness of `SliceIndex`-based annotations
 //!
-//! Several annotations rely on `SliceIndex` being a sealed trait: it
+//! Several annotations rely on [`SliceIndex`] being a sealed trait: it
 //! requires `private_slice_index::Sealed`, which has no public path outside of
 //! `core`, so its set of implementations is fixed and exhaustive. The sealed implementations
 //! were inspected to ensure these specs are sound.
--- a/lib/flux-core/src/slice/mod.rs
+++ b/lib/flux-core/src/slice/mod.rs
@@ -2,7 +2,7 @@
 //!
 //! # Soundness of `SliceIndex`-based annotations
 //!
-//! Several annotations rely on [`SliceIndex`] being a sealed trait: it
+//! Several annotations rely on `SliceIndex` being a sealed trait: it
 //! requires `private_slice_index::Sealed`, which has no public path outside of
 //! `core`, so its set of implementations is fixed and exhaustive. The sealed implementations
 //! were inspected to ensure these specs are sound.
--- a/lib/flux-alloc/src/lib.rs
+++ b/lib/flux-alloc/src/lib.rs
@@ -2,20 +2,20 @@
 #![cfg_attr(flux, feature(allocator_api))]
 #![cfg_attr(flux, flux::no_suggestions)]
 
-#[cfg(flux)]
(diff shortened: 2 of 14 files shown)


nilehmann:
Thanks for taking a stab at this @kimjune01 

I took a screenshot of how this looks when rendered

image

image


thoughts @ranjitjhala ?

ranjitjhala:
Haha, that's undoubtedly a little scary looking BUT I think it's an
improvement over right now (where we simply cannot use the rustdoc search
at all, and have to rattle around grepping in the source).

I think we should merge as a starting point and then maybe cut a new issue
to make this look nicer?

ranjitjhala:
That said - I wonder if there is some easy way right now to at least
preserve the original source level formatting? It looks like line breaks
are getting inserted in odd places?


- Ranjit.

kimjune01:
*edit* now looking into prettifying and/or formatting.

Should we use `prettyplease` dep? I'm thinking not for this PR.

It seems that formatting can be improved for custom docs but not generated ones.

`flux_core::slice::TSliceSpec_95eefcac29f5f8b2`
Screenshot 2026-08-28 at 11 17 38 AM

ranjitjhala:
Hi @kimjune01 is the above what it looks like now -- if so, that's pretty awesome, thanks!!!

lib/flux-core/src/slice/mod.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! # Soundness of `SliceIndex`-based annotations
 //!
-//! Several annotations rely on [`SliceIndex`] being a sealed trait: it
+//! Several annotations rely on `SliceIndex` being a sealed trait: it
```
ranjitjhala:
```suggestion
//! Several annotations rely on [`SliceIndex`] being a sealed trait: it
```

ranjitjhala (approved):

ranjitjhala:
Sorry i broke the thing! will merge if/when it goes green!

PR #1741, opened 2026-08-28
Fix slice iterator as_slice remaining-length spec

kimjune01:
## Summary

- specify `slice::Iter::as_slice` with the remaining length, `len - idx`
- add a positive regression for the length after consuming an element
- add a negative regression rejecting an index into an exhausted remainder

The current spec gives `as_slice` the iterator original length. After consuming the only element, Flux therefore accepts an access that Rust panics on:

```rust
let mut it = &[42][..].iter();
it.next();
let remaining = it.as_slice();
remaining[0];
```

The regression fails on `main` because the expected bounds error is missing, and passes on this branch.

## Context

- `slice::Iter` received adjacent iterator specifications in #1680.
- Exhausted iterator state is handled by #1740.

## Test

`PATH="$HOME/.cargo/bin:$PATH" cargo xtask test flux_core_slice00`

### DIFF
--- a/lib/flux-core/src/slice/iter.rs
+++ b/lib/flux-core/src/slice/iter.rs
@@ -7,7 +7,7 @@ struct Iter<'a, T>;
 #[extern_spec(core::slice)]
 impl<'a, T> Iter<'a, T> {
     #[no_panic]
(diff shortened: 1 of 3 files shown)


ranjitjhala (approved):
Thanks!

PR #1742, opened 2026-08-31
Add specs for Ptr/NonNull casts

sweiglbosker:
This pr adds specs that preserve pointer/NonNull indices through their `cast*` methods. This is already how we handle `as` casts (see https://github.com/flux-rs/flux/pull/1710).


enjhnsn2 (approved):
lgtm

PR #1743, opened 2026-08-31
Preserve exhausted slice windows state in specs

kimjune01:
## Summary

- preserve `slice::Windows` remaining state after the iterator is exhausted
- align the associated `Iterator::step` relation with the `next` postcondition
- cover initially exhausted, final-item, and repeated-exhaustion transitions
- add a negative regression preventing inconsistent exhausted state from creating false proof context

The current specification decrements `remaining` on every call to `next`, including calls after exhaustion. Rust leaves an exhausted `Windows` iterator unchanged. Once the ghost state becomes negative, it can introduce inconsistent assumptions and make invalid programs verify.

## Test

- `PATH="$HOME/.cargo/bin:$PATH" cargo xtask test flux_core_slice03`
- `PATH="$HOME/.cargo/bin:$PATH" cargo xtask test flux_core_slice04`

### DIFF
--- a/lib/flux-core/src/slice/iter.rs
+++ b/lib/flux-core/src/slice/iter.rs
@@ -53,12 +53,17 @@ struct Windows<'a, T>;
 #[assoc(
     fn size(x: Windows) -> int { if x.window_size > x.remaining { 0 } else { x.remaining - x.window_size + 1 } }
     fn done(x: Windows) -> bool { x.remaining < x.window_size }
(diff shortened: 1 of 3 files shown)


enjhnsn2 (approved):
This took a bit of a deep read because the windows specs are kind of obtuse, but I'm pretty confident this is right.

ranjitjhala:
Thanks @enjhnsn2 !!!

PR #1744, opened 2026-08-31
Preserve exhausted take iterator state

kimjune01:
## Summary

- preserve a `Take` iterator’s quota and inner state when `next` is called with no quota remaining
- add repeated-exhaustion coverage
- add a negative regression showing that a negative ghost quota creates false proof context

Rust returns `None` without modifying either field when `Take::next` is called with `n == 0`. The current Flux transition decrements `n` unconditionally, producing `-1`. Because the quota originates as a `usize`, that inconsistent state can make invalid assertions verify.

## Test

`PATH="$HOME/.cargo/bin:$PATH" cargo xtask test flux_core_iter00`


ranjitjhala (approved):
Thanks for catching this!!!

PR #1746, opened 2026-08-31
Preserve existential relationships in array elements

### LINKED ISSUE #1736: Loses existential relationships when constructing arrays
Flux loses the relationship between tuple fields when existentially refined values are collected into an array.

Here is a minimal example:

```rs
#[flux_rs::spec(fn() -> {a, b. (usize[a], usize[b]) | a < b})]
fn tuple() -> (usize, usize) {
    (10, 20)
}

#[flux_rs::spec(fn() -> [{a, b. (usize[a], usize[b]) | a < b}; _])]
fn tuples() -> [(usize, usize); 2] {
    let x0 = tuple();
    let x1 = tuple();
    [x0, x1]
}

fn test() {
    let xs = tuples();
    let (a, b) = xs[0];
    flux_rs::assert(a < b);
}
```

On current `main`, Flux reports that the return postcondition `a < b` cannot be proved. The later assertion fails for the same reason.

This came up while discussing the SmallVec/TinyVec work. The likely issue is that the existentials need to be pulled out before generating the array-element kvar.

Zulip context: https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/notes.20on.20fluxing.20SmallVec.20.2B.20TinyVec/near/617772601

kimjune01:
Fixes #1736.

Array construction filled refinement holes before existential indices nested inside tuple elements were hoisted. This produced independent kvars that could not express relationships between tuple fields.

Hoist the element indices first, then generate a single kvar over all of them. This allows relationships such as `a < b` to be inferred when existentially refined tuples are collected into an array.

Adds a regression test covering array construction and subsequent indexing.

### DIFF
--- a/crates/flux-refineck/src/checker.rs
+++ b/crates/flux-refineck/src/checker.rs
@@ -20,11 +20,12 @@ use flux_middle::{
     query_bug,
     rty::{
         self, AdtDef, AliasReft, BaseTy, BinOp, Binder, Bool, Clause, Constant,
-        CoroutineObligPredicate, EarlyBinder, Expr, FnOutput, FnSig, FnTraitPredicate, GenericArg,
-        GenericArgsExt as _, Int, IntTy, Mutability, Path, PolyFnSig, PtrKind, RefineArgs,
-        RefineArgsExt,
+        CoroutineObligPredicate, EarlyBinder, Expr, ExprKind, FnOutput, FnSig, FnTraitPredicate,
+        GenericArg, GenericArgsExt as _, Int, IntTy, Mutability, Path, PolyFnSig, PtrKind,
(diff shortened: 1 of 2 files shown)


ranjitjhala:
@nilehmann AFAICT this PR implements your suggestion at the zulip [#dev > notes on fluxing SmallVec + TinyVec @ 💬](https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/notes.20on.20fluxing.20SmallVec.20.2B.20TinyVec/near/617772601) -- did you want to take a quick look or shall I merge?

nilehmann:
Of course I want to take a look. This is all vibecoded.

PR #1750, opened 2026-09-02
Add specs for dangling pointer functions

ryanfanchiotti:
- PR adds specs for std::ptr::dangling / dangling_mut / NonNull::dangling, which create a dangling pointer


enjhnsn2 (approved):
lgtm

PR #1752, opened 2026-09-02
Update generic Iterator::next() specs

ranjitjhala:
to allow reasoning about size.


ranjitjhala:
@nilehmann or @enjhnsn2 does either of you want to take a quick look at this?

enjhnsn2 (approved):
It's gnarly, but it looks right to me.

PR #1753, opened 2026-09-04
Add specs for raw pointer comparisons

ryanfanchiotti:
- This PR adds specs for PartialEq pointer equality functions (for *mut T, *const T, and NonNull<T>), and adds logic to reduce comparisons of raw pointers to comparisons of their addresses
- Note that comparing raw pointers across allocations should be well defined, since they implement Ord
- Implementing PartialOrd comparisons is left to a future PR, since they return an Ordering

### DIFF
--- a/lib/flux-core/src/ptr/mod.rs
+++ b/lib/flux-core/src/ptr/mod.rs
@@ -139,11 +139,18 @@ macro_rules! ptr_specs {
             unsafe fn read(self) -> T
             where T: Sized;
 
-            // Core impl: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/ptr/const_ptr.rs#L48
+            /// Core impl: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/ptr/const_ptr.rs#L48
             #[spec(fn(me: *$mutable[@p] T) -> *$mutable[p.base, p.addr, p.size] U)]
             fn cast<U>(self) -> *$mutable U;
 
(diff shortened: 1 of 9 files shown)


lib/flux-core/src/ptr/mod.rs
```diff
@@ -139,12 +139,19 @@ macro_rules! ptr_specs {
...
             #[spec(fn(me: *$mutable[@p] T) -> *$mutable[p.base, p.addr, p.size] U)]
             fn cast<U>(self) -> *$mutable U;
 
             $($($extra)*)?
         }
+
+        #[extern_spec(core::ptr)]
+        impl<T> PartialEq for *$mutable T {
+            /// Core impl: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/ptr/const_ptr.rs#L1614
+            #[spec(fn (me: &*$mutable[@m] T, other: &*$mutable[@o] T) -> bool[m.addr == o.addr])]
```
sweiglbosker:
```suggestion
            #[spec(fn (me: &*$mutable[@m] T, other: &*$mutable[@o] T) -> bool[m == o])]
```

Does this work? that way we have one source of truth

sweiglbosker (approved):
LGTM, but can you add a positive test for `.eq()` and `==` where the  pointers have the same address but different base and/or size? As written, all the tests would pass if the specs were testing structural equality.

nilehmann (commented):
Internal changes to flux LGTM. I'll let @enjhnsn2 judge the extern spec part of this PR

enjhnsn2 (approved):
External specs look good. I trust nico that the internal parts are sound

PR #1755, opened 2026-09-09
Add specs for Ordering + more raw pointer comparisons

ryanfanchiotti:
- Followup to https://github.com/flux-rs/flux/pull/1753; gets more functions that compare raw pointers / NonNull pointers working
- Refines `Ordering` in the same way the standard library does (`-1 = Less, 0 = Equal, 1 = Greater`)

### DIFF
--- a/lib/flux-core/src/cmp.rs
+++ b/lib/flux-core/src/cmp.rs
@@ -36,6 +36,27 @@ trait Ord {
         Self: Sized;
 }
 
+#[extern_spec]
+#[assoc(
+    fn is_lt(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_le(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_gt(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_ge(x: Self, y: Rhs, res: bool) -> bool { true }
+)]
+trait PartialOrd<Rhs: PointeeSized = Self>: PointeeSized {
+    #[spec(fn(&Self[@x], &Rhs[@y]) -> bool{v: <Self as PartialOrd<Rhs>>::is_lt(x, y, v)})]
+    fn lt(&self, other: &Rhs) -> bool;
+
+    #[spec(fn(&Self[@x], &Rhs[@y]) -> bool{v: <Self as PartialOrd<Rhs>>::is_le(x, y, v)})]
+    fn le(&self, other: &Rhs) -> bool;
+
+    #[spec(fn(&Self[@x], &Rhs[@y]) -> bool{v: <Self as PartialOrd<Rhs>>::is_gt(x, y, v)})]
+    fn gt(&self, other: &Rhs) -> bool;
+
+    #[spec(fn(&Self[@x], &Rhs[@y]) -> bool{v: <Self as PartialOrd<Rhs>>::is_ge(x, y, v)})]
+    fn ge(&self, other: &Rhs) -> bool;
+}
+
 #[extern_spec(core::cmp)]
 #[assoc(
     fn min_res(a: int, b: int, res: int) -> bool { res == min(a, b) }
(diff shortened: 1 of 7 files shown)


lib/flux-core/src/cmp.rs
```diff
@@ -36,9 +36,51 @@ trait Ord {
...
+    #[variant(Ordering[0])]
+    Equal,
+    #[variant(Ordering[1])]
+    Greater,
+}
+
+#[extern_spec(core::cmp)]
+#[assoc(
+    fn is_eq(m: Ordering, o: Ordering, res: bool) -> bool { res <=> (m.res == o.res) }
+    fn is_ne(m: Ordering, o: Ordering, res: bool) -> bool { res <=> (m.res != o.res) }
```
enjhnsn2: I think the correct way to do this is `-> bool[m.res==o.res]`
In general, the "direct" `[...]`-style refinement is preferable to existential refinements `{...}`, as the existential refinements introduce an additional variable.

lib/flux-core/src/cmp.rs
```diff
@@ -36,9 +36,51 @@ trait Ord {
         Self: Sized;
 }
 
+#[extern_spec]
+#[assoc(
+    fn is_lt(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_le(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_gt(x: Self, y: Rhs, res: bool) -> bool { true }
+    fn is_ge(x: Self, y: Rhs, res: bool) -> bool { true }
+)]
```
enjhnsn2, line 45: Can we simplify this to a single associated refinement `partial_cmp` of type `(x: self, y:Rhs,) -> Option<int>`, effectively mirroring the trait impl? We can then derive the specs for lt/le/gt/ge from that one method (like how the PartialOrd trait derives these methods)

enjhnsn2 (changes requested):

lib/flux-core/src/cmp.rs
ryanfanchiotti: We could, but won't this add an additional `ite` into smt formulas? I don't think a couple lines of boilerplate is worth that but I could be wrong (I think `ite` is a pretty fast operation for the solvers?)

lib/flux-core/src/cmp.rs
ryanfanchiotti: I don't think this is an existential refinement? There aren't any arguments (usually it would be like `{v: v == x ...}`, and [] syntax yields a compile error. I did just copy from the other assoc implementations here though so not really sure why they're defined like this

enjhnsn2: Oh, maybe you're right that its not creating an evar...
@nilehmann Why does `[...]` refinement not work here?

nilehmann: We need to define the associated refinement as a predicate on the inputs and output for it to work with an implementation where we can't be exact. When used monomorphically, it goes through the spec for the method below, which uses the index syntax.

lib/flux-core/src/cmp.rs
```diff
@@ -36,9 +36,51 @@ trait Ord {
...
+
 #[extern_spec(core::cmp)]
 #[assoc(
     fn min_res(a: int, b: int, res: int) -> bool { res == min(a, b) }
     fn max_res(a: int, b: int, res: int) -> bool { res == max(a, b) }
 )]
 impl Ord for usize {}
+
+#[extern_spec(core::cmp)]
+#[refined_by(res: int)]
```
nilehmann: Did you try using `#[reflect]` for this?

https://flux-rs.github.io/flux/guide/specifications.html#reflecting-enums

I don't remember if it works for extern specs. Also, it probably won't work if we actually need to know the numeric value of the discriminant

ryanfanchiotti: Nope but it does seem to work. Reflect does seem like a more user friendly idea since we don't have to remember that -1 and 1 are Less and Greater respectively. I don't think we'll need to cast from Ordering to u8 much but it could come up somewhere. Up to you all which version is better

PR #1757, opened 2026-09-12
Don't block when materializing under shared borrow

### LINKED ISSUE #1754: ICE (unsolved evar) when type checking mut raw ptr and opaque type
Similar error to https://github.com/flux-rs/flux/issues/1720:

```
struct Token<T>(*mut T);

fn _spawn<F>(_future: impl FnOnce() -> F) -> Token<impl Sized> {
    Token(std::ptr::null_mut::<F>())
}
```
```
thread 'rustc' (253418) panicked at crates/flux-infer/src/infer.rs:415:10:
called `Result::unwrap()` on an `Err` value: UnsolvedEvar(?4e)
Error: process exited unsuccessfully: exit status: 101
```
fails while

```
struct Token<T>(*const T);

fn _spawn<F>(_future: impl FnOnce() -> F) -> Token<impl Sized> {
    Token(std::ptr::null_mut::<F>())
}
```

does not; this is a simplified version of [this function](https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/raw/mod.rs#L262).

I did a little debugging here and this seems to have something to do with [subtyping and variance](https://doc.rust-lang.org/nightly/nomicon/subtyping.html), where mutable raw pointers' invariance causes Flux to check that `Token<F>` is a subtype of `Token<impl Sized>` and vice versa, which it can only do in one direction.

I wrote a hacky "fix" that is probably wrong, hopefully it's helpful for fixing this: [diff](https://github.com/flux-rs/flux/compare/main...ryanfanchiotti:flux:silly-type-hacking)

ranjitjhala:
Fixes #1754

maybe? (This is CC's doing, but the cause and fix sound plausible to me?)

## Cause

The real error was **`incompatible types: †Self - Self[?1e]`** — a blocked type reaching subtyping.
Because `Self` is `?Sized` in a trait, `&mut Self` itself implemented `Buf`, so `self.remaining()` passed `&_1` — a shared reference wrapping the `ptr(mut, ℓ)` — which forced `flux` to materialize a real `&mut` inside it, and materializing a `&mut` always blocks its location. The block was then never released, because unblocks are only emitted when a _mutable_ borrow expires and the borrow responsible here was a shared one.

## Fix

Flux now tracks whether subtyping is descending underneath a shared reference, and when it is, 
materializing a `&mut` no longer blocks its location. That's safe because **you can never mutate through a `&&mut T`**, 
so the location can't change and there was nothing to block in the first place — which is also why no unblock 
was ever coming.

### DIFF
--- a/crates/flux-infer/src/infer.rs
+++ b/crates/flux-infer/src/infer.rs
@@ -740,6 +740,19 @@ impl fmt::Debug for TypeTrace {
     }
 }
 
+/// Whether converting a `ptr(mut, ℓ)` into a `&mut` should block `ℓ`.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Blocking {
+    /// Block `ℓ` for the duration of the borrow. This is the normal case: the resulting `&mut`
+    /// can be used to mutate the pointee, so `ℓ` must not be used until the borrow expires.
+    Block,
+    /// Leave `ℓ` alone. Used when the resulting `&mut` sits *underneath* a shared reference: a
+    /// `&&mut T` cannot be reborrowed mutably, so the pointee cannot change and there is nothing
+    /// to block. Blocking here would be unreleasable anyway, because no mutable borrow expires to
+    /// trigger the matching `Unblock` ghost statement.
+    NoBlock,
+}
+
 pub trait LocEnv {
     fn ptr_to_ref(
         &mut self,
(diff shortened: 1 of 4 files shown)


ranjitjhala:
@nilehmann can you please take a look at this? its the biggest blocker (ha!) in `bytes` so an important thing to fix.

The fix itself is entirely CC but it seems to make sense to me?

PR #1760, opened 2026-09-14
Use rustc upvar tys to create correct env for closure body

### LINKED ISSUE #1756: ICE: loc not found
Here's another blocking ICE -- that CC thinks is related to #1754 

```rust
//! ICE group 2 (~24 sites in `bytes`): `loc not found`
//!
//!     error: internal compiler error:
//!       crates/flux-refineck/src/type_env/place_ty.rs:355:32: loc not found a0
//!      --> group2_loc_not_found.rs:29:24
//!       |
//!    29 |         (|| { let n = self.remaining(); self.advance(2); n })()
//!       |                       ^^^^^^^^^^^^^^^
//!
//! Reproduces on editions 2015/2018/2021/2024.
//!
//! Ingredients (each necessary — drop any one and it checks cleanly):
//!   1. a default trait method taking `&mut self`,
//!   2. whose body calls BOTH a `&self` and a `&mut self` method on `self`
//!      *from inside a closure*,
//!   3. an impl whose `Self` is `&mut T` with `T: ?Sized`.
//!
//! Same three ingredients as group 1 (`group1_unsolved_evar.rs`); -- github issue #1754 
//! the only difference is that here the two reborrows happen inside a closure, which
//! moves the failure from `TypeEnv::ptr_to_ref` to the `get_loc` lookup. That
//! is good evidence the two groups share a root cause in the location/strong-
//! pointer machinery.

pub trait Buf {
    fn remaining(&self) -> usize;
    fn advance(&mut self, cnt: usize);

    fn get_u16(&mut self) -> usize {
        (|| {
            let n = self.remaining(); // shared reborrow, inside a closure
            self.advance(2); // mutable reborrow, inside the same closure
            n
        })()
    }
}

impl<T: Buf + ?Sized> Buf for &mut T {
    fn remaining(&self) -> usize {
        (**self).remaining()
    }
    fn advance(&mut self, cnt: usize) {
        (**self).advance(cnt)
    }
}
```

ranjitjhala:
**Bug**  (fixes #1756): `flux` represents `&mut x` internally as a pointer to a location name like `a0` that only means something inside the current function. When a closure captured one, `flux` was supposed to convert it back into an ordinary `&mut` first — but it only did that when the pointer was the entire -- top-level -- captured value, so a pointer inside a tuple, or behind another reference, slipped through. The closure body was then checked with a fresh environment that has never heard of `a0`, and looking it up blows up.

**Fix** Instead of that ad-hoc top-level conversion, each captured value is now related against the type `rustc` says it should have, refined with fresh holes. That target never mentions locations, so the existing subtyping walk converts every pointer at any depth — through tuples, arrays and references alike — exactly as it already does when checking a call against a function's parameter types. A two-line prerequisite was needed so `ptr_to_ref` relates types using the real environment rather than `DummyEnv`, since nested pointers now reach it.


crates/flux-refineck/src/checker.rs
```diff
@@ -1389,25 +1389,35 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
-                }
+                let ty =
+                    ty.replace_holes(|binders, kind| infcx.fresh_infer_var_for_hole(binders, kind));
+
+                // Refining wraps a unit-sorted type (a reference) in a top-level `Constr`, and
+                // `place_ty` cannot deref through that wrapper, so strip it. An indexed type is
+                // wrapped in `Exists` instead, with the `Constr` inside, so this is a no-op there.
+                // But we defensively do the check here, just in case (even though it should be a no-op).
+                let (ty, pred) = ty.unconstr();
+                infcx.at(stmt_span).check_pred(&pred, ConstrReason::Other);
```
ranjitjhala, line 1412: this check is a no-op but CC suggests keeping it just in case some day it becomes a non-no-op...

ranjitjhala:
ok @nilehmann this is the other group of errors, but this is the fix separate from #1754 ...

crates/flux-refineck/src/checker.rs
nilehmann: Does any test fail if we don't do this?

ranjitjhala: nope (just double checked).

ranjitjhala: but CC insisted I keep it just in case in the future the strip peeled away something useful

nilehmann: Then let's add a test that exercises this code

ranjitjhala: do you mean the `check_pred(...)` -- there is no test -- that fails -- because the only pred that's stripped is for references with trivial pred.

ranjitjhala: or do you mean even the `ty.unconstr()` part?

nilehmann: So, any of the tests already fail without the `ty.unconstr()`?

If that's the case, that's why I was confused. I thought the comment applied to the entire sequence of `ty.unconstr() + check_pred`

ranjitjhala: So:

1. If you leave out the `ty.unconstr()` then `tests/tests/pos/surface/ptr02.rs` fails, with
    
```
crates/flux-refineck/src/type_env/place_ty.rs:198:30: invalid deref `{ &'?0 mut { Vec<{ i32 | * }> | * } | * }`
```

which makes sense, I suppose as the test is below, and it doesn't expect a `Constr` at the top; 
    
``` rust
pub fn test() {
    let mut vec = Vec::new();
    let thing = Some(1);
    thing.map(|x| vec.push(x));
}
```

2. If you leave out the `.check_pred()` though, nothing fails, and AFAICT nothing *can* fail, because the only time we strip out is when the top-level thing was a reference, in which case the pred is inert

?

nilehmann: Ok, then keeping the `check_pred` is fine. I just think the comment is a bit confusing. I thought the `ty.unconstr()` was also a no-op

nilehmann (approved):
Looks good, modulo making clear the no-op is the check_pred part

PR #1762, opened 2026-09-15
Fix reference to pointer conversion unsoundness when pointed-to type is unsized

### LINKED ISSUE #1761: Internal flux error when converting slice to raw ptr
```
extern crate flux_core;

fn main() {
    let slice: &[u32] = &[1, 2, 3];
    let raw_slice: *const [u32] = slice as *const [u32];
    flux_rs::assert(!raw_slice.is_null());
}
```
errors with
```
error: internal flux error: crates/flux-infer/src/projections.rs:812:22
  |
  = note: error selecting Obligation(predicate=TraitPredicate(<[u32] as std::marker::Sized>, polarity:Positive), depth=0): Unimplemented

```
I think this is because we're assuming everything we cast to a raw pointer has `Sized`. `raw_ptr_with_size()` looks like it makes an attempt to check for this but some basic debugging shows we are hitting the `Ok(ty)` path.

ryanfanchiotti:
- Fixes https://github.com/flux-rs/flux/issues/1761 by doing a best effort check for whether the underlying type has `Sized`, and not specifying anything related to alignment or size if so
- Maybe in the future we can get this to work a little better for unsized types? We could say the size must be a multiple of the underlying type for slices, etc


crates/flux-refineck/src/checker.rs
```diff
@@ -2134,12 +2134,47 @@ impl<'ck, 'genv, 'tcx, M: Mode> Checker<'ck, 'genv, 'tcx, M> {
...
+    let addr = Expr::field_proj(&nu, rty::FieldProj::RawPtr { field: rty::RawPtrField::Addr });
+    let size = Expr::field_proj(nu, rty::FieldProj::RawPtr { field: rty::RawPtrField::Size });
+
+    // For slices and other fat pointer types, we don't know the alignment and size
+    // and so must drop those assertions.
+    if !has_sized {
+        let pred = Expr::and_from_iter([
+            Expr::eq(base, addr.clone()),
+            Expr::ne(addr.clone(), Expr::zero()),
+        ]);
```
```suggestion
        let pred = Expr::and(
            Expr::eq(base, &addr),
            Expr::ne(&addr, Expr::zero()),
        );
```

tests/tests/with_deps/pos/extern_specs/flux_core_ptr12.rs
```diff
@@ -0,0 +1,14 @@
+extern crate flux_core;
```
nilehmann, line 4: Can you put a comment with a link to the issue please?

nilehmann (commented):
This is certainly better than crashing, thanks @ryanfanchiotti!

Just a couple of comments. Ping me after fixing it so I can merge.

ryanfanchiotti:
Should be fixed now @nilehmann

nilehmann (approved):

PR #1766, opened 2026-09-16
Implement cargo flux flags

ryanfanchiotti:
- Adds functionality to `cargo flux`'s flags by converting them to the same format as environment variable flags
- The flags struct only supports parsing each flag with an include pattern once here (not sure if this is how the FLUXFLAGS version of this is typically used, it seems like multiple `-Finclude` etc are expected?)
- Flags still take two `-` instead of one; making clap handle one `-` args seems like a bit of a pain


enjhnsn2:
@nilehmann this seems reasonable to me, but I never touch Flux's CLI. Thoughts?

nilehmann:
I think this could be implemented a bit more elegantly using clap's metadata without having to change parsing inside the flux driver. Something like this:

<details>
<summary>Produced by claude. haven't checked if it works</summary>

```rust

#[derive(clap::Args)]
pub struct CompileOpts {
    // ...
    flux_flags: FluxFlags, 
}

impl Flags {
    /// Re-emit the flags the user actually passed, in the `-Fkey=value` syntax
    /// the driver and `FLUXFLAGS` already understand.
    pub fn rustflags(matches: &ArgMatches) -> Vec<String> {
        let spec = Flags::augment_args(Command::new("flux-flags"));
        let mut out = Vec::new();
        for arg in spec.get_arguments() {
            let id = arg.get_id().as_str();
            if matches.value_source(id) != Some(ValueSource::CommandLine) {
                continue;
            }
            // `long` is already `Flog-dir`, so one dash gives the driver form.
            let Some(long) = arg.get_long() else { continue };
            match matches.get_raw(id) {
                Some(values) => out.extend(
                    values.map(|v| format!("-{long}={}", v.to_string_lossy()))
                ),
                None => out.push(format!("-{long}")),
            }
        }
        out
    }
}

#[derive(Clone, Debug, Default)]
pub struct FluxFlags(pub Vec<String>);

impl FromArgMatches for FluxFlags {
    fn from_arg_matches(m: &ArgMatches) -> Result<Self, clap::Error> {
        Ok(FluxFlags(Flags::rustflags(m)))
    }
    fn update_from_arg_matches(&mut self, m: &ArgMatches) -> Result<(), clap::Error> {
        self.0 = Flags::rustflags(m);
        Ok(())
    }
}

impl Args for FluxFlags {
    fn augment_args(cmd: Command) -> Command { Flags::augment_args(cmd) }
    fn augment_args_for_update(cmd: Command) -> Command { Flags::augment_args_for_update(cmd) }
}
```

</details>

But more generally, I'm not a fan of adding flux-driver flags directly to cargo-flux. I think it makes the distinction between the build system and the compiler driver more confusing. Besides, what's wrong with `FLUXFLAGS="..." cargo flux`?

But if anyone feels strongly about adding this, I think it's fine.

nilehmann:
I suppose the current state is even more confusing because you see the flags when running `cargo flux --help`

enjhnsn2:
i have to ask claude every time I want to add flags anyways, so either way works for me :)

ryanfanchiotti:
From a new user's perspective, having flag documentation is very nice. FLUXFLAGS flags aren't really documented anywhere besides the Flags struct itself (and you don't get any info on them from passing flux -h either). Clap does validation a little more cleanly, and I think it's a better experience to be able to see flags from -h and immediately input them, instead of passing a very slightly different version into an environment var.

The metadata implementation is nice (and seems to work), was looking for something like that and missed it haha

ranjitjhala:
I generally agree re: new user pains (also old users like me -- I always have a devil of a time recalling what the flaggery is :-) )

crates/flux-config/src/flags.rs
```diff
@@ -17,273 +41,280 @@ macro_rules! flux_arg {
...
         long = flux_arg!("pointer-width"),
         value_name = "WIDTH",
-        default_value = "64",
-        value_parser = default_pointerwidth
+        value_parser = parse_pointer_width_value
     )]
-    pub pointer_width: PointerWidth,
-    /// If present switches on query caching and saves the cache in the provided path
-    #[arg(long = flux_arg!("cache"), value_name = "PATH", value_parser = panicking_parser)]
+    pub pointer_width: Option<PointerWidth>,
```
nilehmann: Do we need to change all these to `Option` with the new implementation?

ryanfanchiotti: Nope, good catch

crates/flux-config/src/flags.rs
```diff
@@ -185,63 +177,59 @@ pub struct Flags {
...
     /// Path to the Flux sysroot directory. If not set, the driver infers it from its own binary location.
-    #[arg(long = flux_arg!("sysroot"), value_name = "PATH", value_parser = panicking_parser)]
+    #[arg(long = flux_arg!("sysroot"), value_name = "PATH")]
     pub sysroot: Option<PathBuf>,
     /// If `true`, all code is trusted by default. You can selectively untrust items by marking them with `#[trusted(no)]`. The default value of this flag is `false`, i.e., all code is untrusted by default.
     #[arg(
         long = flux_arg!("trusted"),
         num_args = 0..=1,
-        default_missing_value = "false"
+        default_missing_value = "true"
```
nilehmann, line 186: These were wrong before, right?

ryanfanchiotti: Yeah, default_missing_value is what clap puts as ={} if not specified, I don't think there's ever a scenario where you'd want it to be false (--feature would expand to --feature=false)

nilehmann (approved):
Ok, LGTM!

PR #1768, opened 2026-09-17
Support default values for const generic parameters

### LINKED ISSUE #1767: Flux rejects code with default values for const generic parameters
It would be nice if flux could handle generic parameters with default values.
Right now, the following code is accepted by rustc but rejected by flux:

```rust
pub struct S<const N: usize = 1>;

pub fn use_default(_: S) {}
```

error:
```
> error[E0999]: this struct takes 1 generic argument but 0 generic arguments were supplied
>  --> src/lib.rs:3:23
>   |
> 3 | pub fn use_default(_: S) {}
>   |                       ^ expected 1 generic argument
```

sweiglbosker:
Fixes https://github.com/flux-rs/flux/issues/1767


enjhnsn2:
seems reasonable, I'm going to ask copilot what it thinks :)

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1932,16 +1932,35 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
                 param.kind,
                 rty::GenericParamDefKind::Type { has_default: true }
                     | rty::GenericParamDefKind::Base { has_default: true }
+                    | rty::GenericParamDefKind::Const { has_default: true }
```
nilehmann: This `debug_assert!` is now redundant

crates/flux-fhir-analysis/src/conv/mod.rs
```diff
@@ -1932,16 +1932,35 @@ impl<'genv, 'tcx: 'genv, P: ConvPhase<'genv, 'tcx>> ConvCtxt<P> {
...
+                        .type_of(param.def_id)?
+                        .instantiate(self.tcx(), into, &[])
+                        .to_ty();
+                    into.push(self.try_to_ty_or_base(param.kind, span, &ty)?.into());
+                }
+                rty::GenericParamDefKind::Const { .. } => {
+                    let tcx = self.tcx();
+                    let cst = tcx
+                        .const_param_default(param.def_id)
+                        .instantiate_identity()
```
nilehmann: This should be `skip_binder`

```suggestion
                        .skip_binder()
```

nilehmann (changes requested):
Some nits

nilehmann (approved):
Thanks @sweiglbosker !

PR #1770, opened 2026-09-19
Fix for opaque return type bug

ryanfanchiotti:
- Fix for https://github.com/flux-rs/flux/issues/1754, since https://github.com/flux-rs/flux/pull/1757 seems to not actually fix it? I think it fixes other things, but the ICE on the return impl trait function still happens on that branch as far as I can tell
- The main problem this fixes is that rustc leaves opaque types from returning impl trait signatures around when we hit `generic_args()`, even though there is no reason for them to be left opaque (as far as I can tell) since returning an existential type that implements a trait only ever causes type checks in rustc to fail
- `generic_args()` expects it's args' base types to be the same, so we just unify the base type of opaques to be the other arg's base type
- One other way to do this is to add another pattern match for subtyping in `btys()` that is the reverse of what we already do for `other subtype opaque`. This is probably unsound in general but maybe we only ever need to know that an opaque type is a subtype of another type when they are guaranteed to be equal?


nilehmann:
Sorry for the delay, but this was a bit higher bandwidth than other PRs.

I'm afraid this is not so simple. This PR accepts the following two unsound examples that were previously rejected (because they crashed)

<details>
<summary>New unsound examples</summary>

```rust
pub trait Tr {
    type Out;
    fn get(self) -> Self::Out;
}

pub struct Wrap<T>(T);

impl<T> Tr for Wrap<T> {
    type Out = T;
    fn get(self) -> T {
        self.0
    }
}

// Example 1

#[flux::sig(fn(Wrap<i32{v: v > 0}>))]
fn requires_pos(_x: Wrap<i32>) {}

pub fn f(b: bool) -> impl Tr {
    if b {
        requires_pos(f(false)); // f(false) returns Wrap(-1)
    }
    Wrap(-1)
}

// Example 2

fn id<T>(x: T) -> T {
    x
}

#[flux::spec(fn(n: i32) -> impl Tr<Out = i32{v: v >= n}>)]
pub fn f(n: i32) -> impl Tr<Out = i32> {
    if n > 0 {
        return id::<Wrap<i32>>(f(n - 10));
    }
    Wrap(n)
}

#[flux::sig(fn(i32{v: v >= 5}))]
fn requires_ge5(_x: i32) {}

pub fn client() {
    requires_ge5(f(5).get()); // f(5).get() == -5 at runtime
}

```

</details>


Granted, the implementation was already unsound and accepted this

<details>
<summary>Existing unsound example</summary>

```rust
#[flux::spec(fn(n: i32) -> impl Tr<Out = i32{v: v >= n}>)]
pub fn f(n: i32) -> impl Tr<Out = i32> {
    if n > 0 {
        return f(n - 10);
    }
    Wrap(n)
}

#[flux::sig(fn(i32{v: v >= 5}))]
fn requires_ge5(_x: i32) {}

pub fn client() {
    requires_ge5(f(5).get()); // f(5).get() == -5 at runtime
}
```

</details>

Opaque types were never implemented properly, and I think we need to rethink them if we want to support the feature more generally.

ryanfanchiotti:
Hmm, shouldn't those examples fail rustc type checking though? When I remove the specs and try to compile them I get errors from rustc for both functions, since returning an opaque type (what we get back from id) isn't allowed and we can't pass an impl Tr into requires_pos() since it takes a Wrap. I do see that this breaks reasoning about nested refinement types in these things though... definitely seems possible to make trait methods that need to reason about nested types here.

Is the way to fix this just by creating a proper conversion from opaque type to concrete + extra refinements from nested types?

nilehmann:
> Hmm, shouldn't those examples fail rustc type checking though? When I remove the specs and try to compile them I get errors from rustc for both functions

How are you trying it? All of them work just fine for me

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=818d84091874e2fe57c0ed8e20fecf31

> and we can't pass an impl Tr into requires_pos() since it takes a Wrap. 

Inside `f` (this particular) `impl Tr` *is equal* to `Wrap<i32>`. Opaque types are opaque to callers, but inside the function, rustc knows exactly what type they are equal to.

> Is the way to fix this just by creating a proper conversion from opaque type to concrete + extra refinements from nested types?

It's not that you convert opaque types to concrete types; you make all occurrences of an opaque *equal* to a concrete type. But in Flux, we can't use syntactic equality, and we have the added complexity that we need to check associated type projections that can also have refinements. So it requires some thought...

nilehmann:
Moving the discussion to https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/Opaque.20types.20rework/with/626406691

ryanfanchiotti:
Oh I see, the duplicate defs of f caused my copy paste to think the opaque type was outside (and created some convincing error messages)... for some reason I thought the actual type wasn't visible to recursive calls but it clearly is.

nilehmann:
@ryanfanchiotti, for some reason I thought you were on Zulip, but I can't find you there.

Do you still want to work on this? I think the strategy I laid out in https://flux-rs.zulipchat.com/#narrow/channel/486099-dev/topic/Opaque.20types.20rework/with/626449902 should work for your original example and make Flux more sound

ryanfanchiotti:
Yeah (was distracted yesterday, sorry). Will make a zulip account and respond there, the kvar idea seems pretty straightforward

PR #1771, opened 2026-09-20
Fall back to primitive types and sorts during name resolution

### LINKED ISSUE #1769: Modules can shadow built-in sorts
sorts are resolved in rust's type namespace, so the `ptr` module will shadow the `ptr` sort here

```rust
use core::ptr::{self, NonNull};

#[flux::refined_by(p: ptr)]
struct S {
    pointer: NonNull<u8>,
}
```

Resulting in the error `expected a sort, found module`.

This isn't necessarily a bug, but it would be cool if flux could use the fact that the shadowing binding cannot denote a sort to determine that the refinement is probably referring to the built in `ptr` sort and fall back to that. 
I haven't worked out how much this would complicate shadowing rules, or whether there is a simpler solution

sweiglbosker:
This PR implements more rust-like name resolution.

Namely, if a path fails to resolve or resolves to a module, we try to resolve its first segment as a primitive type, and if that fails, a primitive sort.

Fixes #1769


nilehmann (approved):
Awesome!

PR #1772, opened 2026-09-21
Specs for task::Waker

ryanfanchiotti:
- Note that this only models the data stored inside (for now, maybe tracking function ptr data as well soon?)


lib/flux-core/src/task/wake.rs
```diff
@@ -0,0 +1,35 @@
+use flux_attrs::*;
+
+#[extern_spec(core::task)]
+#[refined_by(d_base: int, d_addr: int, d_size: int)]
+struct RawWaker;
```
enjhnsn2, line 5: can we just refine it by a single pointer?

lib/flux-core/src/task/wake.rs
```diff
@@ -0,0 +1,35 @@
...
+#[extern_spec(core::task)]
+impl RawWaker {
+    // Core impl: https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/core/src/task/wake.rs#L56
+    #[spec(fn(data: *const[@p] (), vtable: &RawWakerVTable) -> RawWaker[p.base, p.addr, p.size])]
+    const fn new(data: *const (), vtable: &'static RawWakerVTable) -> RawWaker;
+}
+
+#[extern_spec(core::task)]
+#[refined_by(d_base: int, d_addr: int, d_size: int)]
+struct Waker;
```
enjhnsn2, line 16: also refine this by a pointer

enjhnsn2 (changes requested):

lib/flux-core/src/task/wake.rs
ryanfanchiotti: Fixed

enjhnsn2 (approved):
lgtm

PR #1773, opened 2026-09-21
Add bv128 theory helpers

sweiglbosker:
This PR adds support for the bv128/int conversions added to liquid-fixpoint in https://github.com/ucsd-progsys/liquid-fixpoint/pull/856.


nilehmann (approved):
:+1:

